PR #2, opened 2019-03-08
Big int

ulrikeyang:
Hi all,
I have pretty much finished the BigInt stuff. Please, take a look and give me any comments you might have.
Thanks
Ulrike


rfalgout:
Wow!  You put in some serious work here, Ulrike.  Looks great!  I am going to do some testing and let you know if I have the same issues with the regression tests that you mentioned.

One question.  I thought you said you were going to restore Euclid and ParaSails to their original form, since they are not yet working in the mixed int case.  Did you change your mind or am I mis-remembering.  I'm happy either way, just curious.

ulrikeyang:
Good point. I still need to do this.
Looks like I also broke the GPU stuff ☹. I can’t even compile this.

liruipeng:
> Good point. I still need to do this. Looks like I also broke the GPU stuff ☹. I can’t even compile this. 

oh,no.... :( What's the error?

ulrikeyang:
The problem is the _hypre_utilities.h file, but I have a version now that compiles.
Ulrike

rfalgout:
I think I've got a new version of the printf stuff that fixes the regression test problems.  The regression tests haven't all finished yet, but the openmp test is now working as are all of the others just prior to the first valgrind test.  So, I'll push it to the branch in a few minutes.

rfalgout:
Just to follow up on my previous comment... all of the tests completed without errors.  Nice job, Ulrike!

rfalgout (approved):
Looks great!

rfalgout (approved):

oseikuffuor1 (approved):

src/parcsr_mv/par_csr_matrix.c
```diff
@@ -227,8 +249,8 @@ hypre_ParCSRMatrixSetNumNonzeros( hypre_ParCSRMatrix *matrix )
    hypre_CSRMatrix *offd;
    HYPRE_Int *offd_i;
    HYPRE_Int local_num_rows;
-   HYPRE_Int total_num_nonzeros;
-   HYPRE_Int local_num_nonzeros;
+   HYPRE_BigInt total_num_nonzeros;
```
liruipeng, line 252: maybe total_num_nonzeros should be of type long int anyway. a matrix <2B can have  nnz >2B. but this is not a big prob

src/seq_mv/csr_matop.c
```diff
@@ -128,6 +128,114 @@ hypre_CSRMatrixAdd( hypre_CSRMatrix *A,
...
+/*--------------------------------------------------------------------------
+ * hypre_CSRMatrixBigAdd:
+ * adds two CSR Matrices A and B and returns a CSR Matrix C;
+ * Note: The routine does not check for 0-elements which might be generated
+ *       through cancellation of elements in A and B or already contained
+ in A and B. To remove those, use hypre_CSRMatrixDeleteZeros 
+ *--------------------------------------------------------------------------*/
+
+hypre_CSRMatrix *
+hypre_CSRMatrixBigAdd( hypre_CSRMatrix *A,
```
liruipeng, line 140: Why is Add is so special that we need a Big version of it separately? Just curious

src/seq_mv/csr_matrix.c
```diff
@@ -82,7 +83,10 @@ hypre_CSRMatrixDestroy( hypre_CSRMatrix *matrix )
       if ( hypre_CSRMatrixOwnsData(matrix) )
       {
          hypre_TFree(hypre_CSRMatrixData(matrix), HYPRE_MEMORY_SHARED);
-         hypre_TFree(hypre_CSRMatrixJ(matrix), HYPRE_MEMORY_SHARED);
+         if (hypre_CSRMatrixJ(matrix))
```
liruipeng, line 86: I removed some if-statements of this kind, since TFree checks: if (NULL) return; but this is not important

src/seq_mv/csr_matrix.c
```diff
@@ -82,7 +83,10 @@ hypre_CSRMatrixDestroy( hypre_CSRMatrix *matrix )
...
+            hypre_TFree(hypre_CSRMatrixJ(matrix), HYPRE_MEMORY_SHARED);
+         if (hypre_CSRMatrixBigJ(matrix))
```
liruipeng, line 88: same as above

liruipeng (commented):
Thank Ulrike for doing such a lot of work. I went through most of the files and left some comments.

liruipeng (approved):

src/seq_mv/seq_mv.h
```diff
@@ -46,9 +46,12 @@ typedef struct
 {
    HYPRE_Int     *i;
    HYPRE_Int     *j;
+   HYPRE_BigInt  *big_j;
    HYPRE_Int      num_rows;
    HYPRE_Int      num_cols;
```
liruipeng, line 51: @ulrikeyang For the same reason of having *big_j, do we need a _big_num_cols_?

ulrikeyang:
No, we don’t, since we should never have both j and big_j at the same time. Indices are either 32 bits or 64 bits. It doesn’t affect the number of columns in the matrix.


---
Sent from Workspace ONE Boxer<https://whatisworkspaceone.com/boxer>

On April 3, 2019 at 12:49:28 AM GMT+2, Ruipeng Li <notifications@github.com> wrote:

@liruipeng commented on this pull request.

In src/seq_mv/seq_mv.h<https://github.com/hypre-space/hypre/pull/2#discussion_r271524752>:

> @@ -46,9 +46,12 @@ typedef struct
 {
    HYPRE_Int     *i;
    HYPRE_Int     *j;
+   HYPRE_BigInt  *big_j;
    HYPRE_Int      num_rows;
    HYPRE_Int      num_cols;


@ulrikeyang<https://github.com/ulrikeyang> For the same reason of having *big_j, do we need a big_num_cols?

ulrikeyang:
Hi Ruipeng,
I misunderstood your question. It is actually more complicated. Yes, if we plan to actually use a CSRMatrix with large indices we might want to have a big_num_cols, although it wouldn’t be quite that simple to implement. We are getting away without it because we are mainly using big_j as an auxiliary array where we temporarily store global indices. We generally do not use CSRMatrixNumRows in connection with big_j or this would be a problem. There is at least one place where we create a matrix with a global num_rows, which could be a problem since the number could overflow, but since we don’t check for negative num_cols and don’t use the number, we get away with it, but it is something to look further into and at least make sure it’s well documented.
Generally, we want a CSRMatrix to be local and not so big to require a 64-bit num_cols.
Thanks
Ulrike

liruipeng:
Hi, @ulrikeyang ,

Thanks for the elaborate reply! I agree with what you said that "we don’t use num_cols" in most places. A  place where we use CSR to store a global matrix (in its column space) is extracting external rows from other processes, where the column indices are global. So the **correct** num_cols should also be BigInt. Since it is not actually used,  it does not matter anyway. But we should keep it in mind that this number will overflow. You actually pointed this out in your comment in _par_lr_restr.c_. But it is used more commonly in many other places as well

Thanks

-Ruipeng

ulrikeyang:
Hi Ruipeng
I am well aware of where in hypre this is used after going through the library for the last 4 months and painstakingly replacing ints with bigints :)
I thought I clearly explained what you just repeated down there. I guess it wasn’t, nevertheless we should generally try to use CSRMatrix mainly as a local matrix, where this is not an issue. We can further look into this when I am back.
Ulrike


---
Sent from Workspace ONE Boxer<https://whatisworkspaceone.com/boxer>

On April 3, 2019 at 6:45:48 PM GMT+2, Ruipeng Li <notifications@github.com> wrote:

Hi, Ulrike,

Thanks for the elaborate reply! I agree with what you said that "we don’t use num_cols" in most places. A place where we use CSR to store a global matrix (in its column space) is extracting external rows from other processes, where the column indices are global. So the correct num_cols should also be big_int. Since it is not actually used, so it does not matter anyway. But we should keep it in mind that this number will overflow. You actually pointed this out in your comment in par_lr_restr.c. But it is used more commonly in many other places as wellThanks

-Ruipeng

Hi Ruipeng, I misunderstood your question. It is actually more complicated. Yes, if we plan to actually use a CSRMatrix with large indices we might want to have a big_num_cols, although it wouldn’t be quite that simple to implement. We are getting away without it because we are mainly using big_j as an auxiliary array where we temporarily store global indices. We generally do not use CSRMatrixNumRows in connection with big_j or this would be a problem. There is at least one place where we create a matrix with a global num_rows, which could be a problem since the number could overflow, but since we don’t check for negative num_cols and don’t use the number, we get away with it, but it is something to look further into and at least make sure it’s well documented. Generally, we want a CSRMatrix to be local and not so big to require a 64-bit num_cols. Thanks Ulrike From: Ruipeng Li notifications@github.com<mailto:notifications@github.com> Sent: Tuesday, April 2, 2019 3:49 PM To: hypre-space/hypre hypre@noreply.github.com<mailto:hypre@noreply.github.com> Cc: Yang, Ulrike Meier yang11@llnl.gov<mailto:yang11@llnl.gov>; Mention mention@noreply.github.com<mailto:mention@noreply.github.com> Subject: Re: [hypre-space/hypre] Big int (#2<https://github.com/hypre-space/hypre/pull/2>) @liruipeng<https://github.com/liruipeng> commented on this pull request.
________________________________ In src/seq_mv/seq_mv.h<#2 (comment)<https://github.com/hypre-space/hypre/pull/2#discussion_r271524752>>:
@@ -46,9 +46,12 @@ typedef struct

PR #5, opened 2019-04-01
bug fix FCFJacobi

liruipeng:
I've run into an issue with hypre_BoomerAMGRelax_FCFJacobi (parcsr_ls/pre_relax_more.c, line 675). The if-condition tests if it is the last level, by checking **cf_marker == NULL** (the last level has no cf_marker, apparently). If so, do normal relax, o.w. do FCF. This makes sense in most cases. But it failed to address the following scenario...

_If there is one rank has local size 0 on some coarser levels for some reason (why?), so it has cf_marker == NULL but others have nonzero sizes and thus non-null cf_marker, that rank goes to the "if"-branch the others go to the "else". The processes go to different branches here. This will create problems: the else-branch has a loop of doing Relax 3 times but there is only 1 Relax in the if-branch, and consequently MPI crashed afterwards in the interpolation (par_cycle.c: line 687)._

The problem is that we cannot rely on **cf_marker == NULL** to say it is the last level. I propose this fix which moves the if-logic out of hypre_BoomerAMGRelax_FCFJacobi but inside hypre_BoomerAMGCycle (see the modified code) and tests if **level == num_levels-1** indeed. Fortunately, this is the only place where FCFJacobi is called.

### DIFF
--- a/src/parcsr_ls/par_cycle.c
+++ b/src/parcsr_ls/par_cycle.c
@@ -502,12 +502,20 @@ hypre_BoomerAMGCycle( void              *amg_vdata,
               }
               else if (relax_type ==17)
               {
-                 hypre_BoomerAMGRelax_FCFJacobi(A_array[level],
-                                              Aux_F,
-                                              CF_marker_array[level],
-                                              relax_weight[level],
-                                              Aux_U,
-                                              Vtemp);
+                 //printf("Proc %d: level %d, n %d, CF %p\n", my_id, level, local_size, CF_marker_array[level]);
+                 if (level == num_levels - 1)
+                 {
+                    /* if we are on the coarsest level ,the cf_marker will be null
+                       and we just do one sweep regular jacobi */
+                    hypre_assert(cycle_param == 3);
+                    hypre_BoomerAMGRelax(A_array[level], Aux_F, CF_marker_array[level], 0, 0, relax_weight[level],
+                                         0.0, NULL, Aux_U, Vtemp, NULL);
+                 }
+                 else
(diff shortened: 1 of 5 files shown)


rfalgout:
This makes sense to me @liruipeng .  @ulrikeyang should chime in, though, since she is much more familiar with what's going on in these files.

liruipeng:
A process that has zero rows of a ParCSR may have an offd-part CSR of size 0-by-num_col_offd with num_col_offd > 0 (think this ParCSR was computed by RAP and that process has no C-pts). When computing MATVEC, this process will still communicate with others to receive those offd entries. But this communication is totally unnecessary. For more general cases than the 0-sized case, an optimization is to prune the offd indices that are not involved in _col_map_offd_, squeeze them out of the offd CSR, and shrink num_col_offd after a ParCSR is formed.

rfalgout:
This also makes perfect sense to me.  Thanks, @liruipeng !

oseikuffuor1:
This makes sense to me too. Thanks @liruipeng

liruipeng:
clean run of regression tests and GPU regression tests

ulrikeyang:
Hi Ruipeng,
this is fine
Thanks
Ulrike

ulrikeyang (approved):

liruipeng:
@rfalgout @oseikuffuor1 Can you please approve this if you also think it is ok? Thanks :)

oseikuffuor1 (approved):

rfalgout (approved):
I'm happy

Issue #16, opened 2019-05-03
Access to the final residual norm


nncarlson:
This is the remaining part of an issue from the old github project https://github.com/LLNL/hypre/issues/39

The various ParCSR Krylov solvers employ a combination of the relative residual norm and absolute residual norm for their convergence criterion. They currently provide a method for getting the final relative residual norm (e.g., `HYPRE_ParCSRHybridGetFinalRelativeResidualNorm`) but not the final (absolute) residual norm. The latter is also needed.

PR #17, opened 2019-05-03
Add HYPRE_ParCSRHybridGetFinalResidualNorm

nncarlson:
Also adds HYPRE_{PCG,GMRES,BiCGSTAB}GetFinalResidualNorm and some others as a consequence.

This PR aims to address #16. It is likely incomplete (other solvers, documentation, ...), but if you point me in the right direction I'd be happy to try and address the deficiencies.

### DIFF
--- a/src/krylov/HYPRE_bicgstab.c
+++ b/src/krylov/HYPRE_bicgstab.c
@@ -205,6 +205,17 @@ HYPRE_BiCGSTABGetFinalRelativeResidualNorm( HYPRE_Solver  solver,
 }
 
 /*--------------------------------------------------------------------------
+ * HYPRE_BiCGSTABGetFinalResidualNorm
+ *--------------------------------------------------------------------------*/
+
+HYPRE_Int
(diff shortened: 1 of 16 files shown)


ulrikeyang:
Hello Neil,
Thank you for providing these changes to hypre. We are currently in the process of changing the license. Therefore we do not want to accept any pull requests until this process is completed to avoid having to get permission from additional institutions for the change. We don’t know how long this process will take. How urgently do you need these changes?
Thanks
Ulrike

nncarlson:
We're already using these changes on a fork, so there's no great urgency. I'd just like to get them merged at some point so that I don't need to maintain a patch as the official hypre evolves.

However I expect there are some gaps in this PR, so if someone has time to take a quick look at it now and point out what is missing/wrong, I'd be able to try addressing them in the meantime.

koning:
These changes would also benefit the mfem_hypre_interface replacement for FEI.

PR #26, opened 2019-06-20
Improve output from hypre_BoomerAMGSetupStats

victorapm:
In this PR, I'm adding a new column in "Interpolation Matrix Information" that gives the average number of Fine-Coarse connections per row of the prolongation operator for each level. Also, I'm fixing minor details such as the format of the tables and deletion of a few trailing whitespaces.

### DIFF
--- a/src/utilities/hypre_printf.c
+++ b/src/utilities/hypre_printf.c
@@ -116,6 +116,20 @@ free_format( char *newformat )
    return 0;
 }
 
+HYPRE_Int
+hypre_ndigits( HYPRE_BigInt number )
+{
+   HYPRE_Int     ndigits = 0;
+
+   while(number)
+   {
+      number /= 10;
+      ndigits++;
+   }
+
+   return ndigits;
+}
+
 /* printf functions */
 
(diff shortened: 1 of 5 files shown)


src/utilities/hypre_printf.c
```diff
@@ -116,6 +116,20 @@ free_format( char *newformat )
    return 0;
 }
 
+HYPRE_Int
+hypre_ndigits( HYPRE_BigInt number )
```
rfalgout, line 120: I'm not sure about the location of this function.  Everything in this file is related to standard C printf facilities.

rfalgout (approved):
I am good with these changes, except that we might want to put the ndigits function in a different place.

src/utilities/hypre_printf.c
victorapm: Thanks for the comment @rfalgout, I was not sure either about the location and ended up putting it there since its usage is meant to help the printf functions.

ulrikeyang (approved):
This is fine with me.

PR #36, opened 2019-08-08
Cmake updates dev

osborn9:
These are some updates to the cmake interface, mainly enabling SuperLU_Dist and updates to the test directory cmake.

(Notably not in these changes are any GPU/kokkos/raja additions. That's for a future PR.)


ulrikeyang (approved):
Looks fine to me.

src/test/CMakeLists.txt
```diff
@@ -1,83 +1,19 @@
-# Copyright 1998-2019 Lawrence Livermore National Security, LLC and other
```
rfalgout: Why was all of this removed (e.g., the license header, the sequential make option, etc.)?

src/test/ij.c
```diff
@@ -6236,7 +6236,7 @@ main( hypre_int argc,
          HYPRE_BoomerAMGSetMultAddTruncFactor(pcg_precond, add_trunc_factor);
          HYPRE_BoomerAMGSetRAP2(pcg_precond, rap2);
          HYPRE_BoomerAMGSetKeepTranspose(pcg_precond, keepTranspose);
-#ifdef HAVE_DSUPERLU
```
rfalgout, line 6239: Does the normal unix build system use this variable also?

rfalgout: Never mind.  I see that it does and the changes have also been made.  Thanks!

victorapm (approved):
Ok for me! There's only a minor detail that I noted while compiling the code on summit, i.e., the following warning message:

>CMake Warning (dev) at CMakeLists.txt:861 (find_package):
  Policy CMP0074 is not set: find_package uses <PackageName>_ROOT variables.
  Run "cmake --help-policy CMP0074" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.
  Environment variable MPI_ROOT is set to:
    /autofs/nccs-svm1_sw/summit/.swci/1-compute/opt/spack/20180914/linux-rhel7-ppc64le/xl-16.1.1-3/spectrum-mpi-10.3.0.1-20190611-aqjt3jo53mogrrhcrd2iufr435azcaha

  >For compatibility, CMake is ignoring the variable. This warning is for project developers.  Use -Wno-dev to suppress it.

To fix it, I propose adding the following piece of code at line 7 of hypre/src/CMakeLists.txt

```
if (POLICY CMP0074)
   cmake_policy(SET CMP0074 NEW) 
endif (POLICY CMP0074)
```

osborn9:
Great, thanks Victor!

src/test/CMakeLists.txt
osborn9: The license header getting removed was my mistake. Thanks for catching it! 

As for the remaining changes, I've changed the build structure so that it no longer builds a separate project for the tests, it will now build them in hypre_BINARY_DIR/test. My reasoning was to avoid any issues with using one set of configurations to build hypre, then a different set to build hypre-test. 

If creating a separate location hypre-test (not tied to the hypre_BINARY_DIR) is desired, I can keep the original hypre-test structure and pull the same configuration options that were used for hypre.

rfalgout: This makes perfect sense.  One question, though.  Are the test drivers always built or is an additional "target" needed to make them?  Users who are only interested in building/installing the library may not want to worry about the test drivers.

Thanks!

osborn9: It's a cmake option that specifies whether or not to build the test drivers (similarly for the example drivers), then as you mentioned users can specify which drivers (if any) are built. 

Both options (HYPRE_BUILD_TESTS and HYPRE_BUILD_EXAMPLES) are off by default.

rfalgout: Works for me.  Thanks!

rfalgout (approved):

oseikuffuor1:
Hi @osborn9, overall, this looks really good. I do have a few comments, however, and maybe the rest of the team can chime in. 
1. Is there a reason why all the source information from lower level directories are build in one CMakeLists file? In practice, it is a good idea to keep things modular in a sense. In other words, each subdirectory under src would have its own CMakeLists file. The top level CMakeLists file typically handles the compiler information, cmake options, and handling of TPLs (i.e., things that are not typically found in the lower level directories). This will make future edits easier, since the addition of any new code would just require editing the CMakeLists file in that directory (better code maintenance). 
2. Third party libraries FindSuperLU module could be moved to a separate cmake_modules directory (under src) so that the top-level CMakeLists file only needs to make Find<Package> calls instead inserting the code there. It looks like for now, it is only SuperLU, so this is not so much of a big deal. Ofcourse if the TPL already has a Find<package> module in cmake, then we can just call that (eg. blas and lapack).
3. It appears there's some inconsistency in the naming options for configure and cmake. I see you correctly edited most of the options for TPL to something like HYPRE_WITH_<library>. However, the hypre-specific features would typically use the _ENABLE_ keyword. I know we don't follow this for all the options, but I think there are a few in the configure.in script that could be made consistent with the cmake version. For example, SHARED, MIXEDINT, LONG_DOUBLE, SINGLE, COMPLEX.

Like as said before, I think it looks great. These changes are fairly minor, since you have done all the hard work already. I am happy to look through with you if you have any questions. Alternatively, if everyone is happy with the current status, I am happy to approve it so as not to delay it further.

osborn9:
Thanks @oseikuffuor1 for your helpful comments. Overall, I agree with your recommendations. Here are some replies:
1. This was the original structure of the CMakeLists.txt, but I agree with you that having them in each sub-directory is a better way to go and avoid having monolithic file that's hard to understand. I'll add this to have it a part of the PR. 
2. The thought here was to not rely on any cmake "magic" and mimic autoconf, where if you want to use SuperLU_Dist then you must specify the exact include dir/library location and no find logic will be invoked if those paths aren't correct. For now, I may leave this as-is.
3. Yes, you are right. I didn't match all of the cmake configure options with the configure.in script. I'll also fix those and add to this PR. 

Thanks!

oseikuffuor1 (approved):
Looks good to me. Thanks @osborn9.

PR #46, opened 2019-09-13
Fixes for Cray compiler

bavier:
These changes allow hypre to build out-of-the-box with Cray's C compiler.

### DIFF
--- a/src/struct_ls/red_black_gs.h
+++ b/src/struct_ls/red_black_gs.h
@@ -299,9 +299,9 @@ typedef struct
 #ifdef HYPRE_USING_OPENMP
 #define HYPRE_BOX_REDUCTION 
 #ifdef WIN32
-#define Pragma(x) __pragma(#x)
+#define Pragma(x) __pragma(HYPRE_XSTR(x))
 #else
-#define Pragma(x) _Pragma(#x)
+#define Pragma(x) _Pragma(HYPRE_XSTR(x))
 #endif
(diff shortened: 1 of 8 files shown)


liruipeng:
This looks good to me. One question for `__builtin_ffs`: I suppose it is a GNU built-in function. How portable is it? Apparently, it works for Cray. How about others like Intel. Thanks!

bavier:
The Cray compiler has a problem with `__builtin_ffs`, but Intel and Clang compilers support it, afaik.  The benefit of `ffs()` is that it is a POSIX-defined function.

liruipeng:
> The Cray compiler has a problem with `__builtin_ffs`, but Intel and Clang compilers support it, afaik. The benefit of `ffs()` is that it is a POSIX-defined function.

Yes, right. I misread. You changed `__builtin_ffs` to `ffs()` not the other way around. This makes more sense. Thanks!

liruipeng:
@bavier The regression tests on our side seem clean. Can you please run tests on your side again before we merge the PR? Thanks!

-Ruipeng

bavier:
I was able to successfully build and run `make  check` with the 2 commits you made.

liruipeng:
@rfalgout @ulrikeyang Can you please review this PR? It looks good for me. Thanks!

rfalgout (approved):
Looks good to me!

ulrikeyang (approved):

PR #51, opened 2019-10-01
add hypre_ParCSRMatrixCompleteClone() compatibility macro - this way …

balay:
…PETSc, mfem [and perhaps other codes] can continue to work with hypre-2.18.1 - without local changes


liruipeng (approved):
looks good to me. Thanks!

PR #61, opened 2019-11-12
Hypre debug

liruipeng:
HYPRE Debug mode
1. ./configure --with-debug enables the debug mode build of hypre. Macro `HYPRE_DEBUG` is defined. 
2. By default, `HYPRE_DEBUG` is undefined and the standard `NDEBUG` macro is defined
3. When `HYPRE_DEBUG` is not defined, `hypre_assert` becomes a non-op. Need to dealt with the "variable was set but never used" compiler warning, which was really a pain. See the new definition of `hypre_assert` in `hypre_error.h`


src/seq_mv/csr_spgemm_device_rowbound.c
```diff
@@ -151,7 +151,7 @@ void csr_spmm_symbolic(HYPRE_Int  M, /* HYPRE_Int K, HYPRE_Int N, */
 
    char failed = 0;
 
-#if DEBUG_MODE
+#ifdef HYPRE_DEBUG
    assert(blockDim.z              == NUM_WARPS_PER_BLOCK);
```
rfalgout, line 155: Should these all be hypre_assert?

src/utilities/_hypre_utilities.h
```diff
@@ -1050,14 +1050,19 @@ void hypre_error_handler(const char *filename, HYPRE_Int line, HYPRE_Int ierr, c
 #define hypre_error(IERR)  hypre_error_handler(__FILE__, __LINE__, IERR, NULL)
 #define hypre_error_w_msg(IERR, msg)  hypre_error_handler(__FILE__, __LINE__, IERR, msg)
 #define hypre_error_in_arg(IARG)  hypre_error(HYPRE_ERROR_ARG | IARG<<3)
-#ifdef NDEBUG
-#define hypre_assert(EX)
+
+#ifdef HYPRE_DEBUG
+#define hypre_assert(EX) do { if (!(EX)) {hypre_fprintf(stderr,"hypre_assert failed: %s\n", #EX); hypre_error(1);} } while (0)
```
rfalgout: @liruipeng Maybe you can explain this to me.  I know you are trying to avoid compiler complaints, but this seems a bit complicated.

rfalgout (changes requested):
Hi @liruipeng ,

This mostly looks really good.  I have a couple of comments/questions, one concerning the use of 'assert' instead of 'hypre_assert', and the other about the new macro for hypre_assert.  Let me know.  Thanks!

-Rob

src/utilities/_hypre_utilities.h
oseikuffuor1: @liruipeng would it suffice to just change the hypre_assert macro to ignore the argument like you have below in line 1060? In other words, replace #define hypre_assert(EX) in the previous version with #define hypre_assert(EX) do { (void) (EX); } while (0) (Not sure if we need the do {} while(0) though). From your initial comments on this issue, it looks like that should fix the compiler complaints. However, your changes (particularly in config.h and configure.in) suggest to me that HYPRE_DEBUG may have been undefined even when --enable-debug is used, which is contrary to part 1 of your initial comment, so I am a bit confused there. Otherwise, the other changes look really good and make sense to me, and I agree with @rfalgout that regular asserts should probably be replaced with hypre_assert.

Thanks

src/seq_mv/csr_spgemm_device_rowbound.c
liruipeng: Yes, they should be but `hypre_assert` is just a print and I want the program to abort at some cases

src/utilities/_hypre_utilities.h
liruipeng: @oseikuffuor1 with `--enable-debug`,  `HYPRE_DEBUG` is defined. see in `configure.in` ```AC_DEFINE([HYPRE_DEBUG], 1, [HYPRE in debug mode])```
I am not sure I understand about your "ust change the hypre_assert macro to ignore the argument like you have below in line 1060" comment.

For the use of `do {} while(0)`, see e.g., http://www.bruceblinn.com/linuxinfo/DoWhile.html.  and https://www.quora.com/What-is-the-purpose-of-using-do-while-0-in-macros. It is a good code practice but it is for multi-line macro, used in other places in hypre also, so it is not really necessary in ```#define hypre_assert(EX) do { (void) (EX); } while (0)```. But it does not hurt

rfalgout: Sounds good.  Thanks!

oseikuffuor1: @liruipeng,
1.  looking at the diffs of configure.in, it looks like change to add HYPRE_DEBUG in AC_DEFINE([HYPRE_DEBUG], 1, [HYPRE in debug mode]) wasn't there before, hence my confusion. 
2. Regarding my comment in this file about just changing the hypre_assert macro, again this came from the diffs. It looks like your changes reversed the #ifdef condition. I would assume that #ifdef NDEBUG is the 'else' part of your modification, so replacing #define hypre_assert(EX) (after #ifdef NDEBUG) with your new assert statement and making the corresponding change to the #else part (your #ifdef  HYPRE_DEBUG part) should yield the same result.
3. Thanks for the clarification about the do {} while(0).

liruipeng: @oseikuffuor1 Yes, you were right. It is just the matter of the ordering of #if and #else. We also want to use `HYPRE_DEBUG` macro here instead of `NDEBUG`

rfalgout (approved):
Hi @liruipeng ,

This all looks good to me.  Thanks for cleaning up the asserts as well.  It wasn't necessary to change the ones in the FEI directory, because that stuff is going to be removed at some point anyway.  But, I guess it's all done now.  You've done the work.  And it was a lot of work! :)  Thanks again!

-Rob

oseikuffuor1 (approved):
Thanks Ruipeng!

rfalgout (approved):
This looks great, @liruipeng !  Thanks!

rfalgout (approved):
Looks great, Ruipeng!  Thanks!

liruipeng:
@ulrikeyang Hi, Ulrike, could you please approve this if you think it is fine. Thanks! -Ruipeng

ulrikeyang (approved):

PR #67, opened 2019-11-28
Improve coarsening

victorapm:
This PR adds new functionalities to the coarsening phase of BoomerAMG and a few other details.

Most important updates are:
1. HYPRE_BoomerAMGSetCoarsenCutFactor: sets a relative threshold value for choosing points that have too many nonzero connections as special fine (isolated) in various coarsening routines.
2. HYPRE_BoomerAMGSetIsolatedFPoints: sets a list of points to be considered as special fine (SF_PT) in the finest level. Use global indices.
3. HYPRE_BoomerAMGSetFPoints: sets a list of points to be considered as fine (F_PT) in the finest level. Use global indices.
4. HYPRE_BoomerAMGSetCpointsToKeep: use global indices for consistency. @oseikuffuor1, I made some changes in the MGR side and wanted to make sure with you that I didn't break anything.

Minor changes:
1. Add support to the previous functions in the IJ driver.
2. Extend the conversion function of a CSRMatrix to a ParCSRMatrix for the case that global partitioning is not used (HYPRE_NO_GLOBAL_PARTITION).
3. The same as 2, but for vectors.
4. Add gitignore file.

Thank you,
Victor


src/parcsr_ls/HYPRE_parcsr_ls.h
```diff
@@ -223,21 +223,38 @@ HYPRE_Int HYPRE_BoomerAMGSetMinCoarseSize(HYPRE_Solver solver,
...
- * a large strength threshold, such as 0.9, is often better. The
- * strong threshold for R is strong connections used in building an
- * approximate ideal restriction, and the filter threshold for R a
- * threshold to eliminate small entries from R after building it.
+ * a large strength threshold, such as 0.9, is often better.
  **/
 HYPRE_Int HYPRE_BoomerAMGSetStrongThreshold(HYPRE_Solver solver,
                                             HYPRE_Real   strong_threshold);
 
+/**
```
rfalgout, line 245: @ulrikeyang should review which of these functions we want to promote to the reference manual.  We don't always want to advertise interface routines to the general user.

src/parcsr_ls/HYPRE_parcsr_ls.h
```diff
@@ -1228,10 +1247,34 @@ HYPRE_Int HYPRE_BoomerAMGSetDSLUThreshold (HYPRE_Solver solver,
...
+                                          HYPRE_Int   *cpt_coarse_index);
+
+/**
+ * (Optional) Set fine points in the first level.
+ *
+ * @param solver [IN] solver or preconditioner
+ * @param num_fpt [IN] number of fine points
+ * @param fpt_index [IN] global indices of fine points
+ **/
+HYPRE_Int HYPRE_BoomerAMGSetFPoints(HYPRE_Solver solver,
```
rfalgout: Should this routine have a similar name to HYPRE_BoomerAMGSetCpointsToKeep()?  I actually think that "ToKeep" is somewhat redundant, but if we remove it we may cause backward compatibility issues for some users (probably not many though).  I would be interested in hearing opinions here.

src/parcsr_ls/par_amg_setup.c
```diff
@@ -1307,33 +1341,51 @@ hypre_BoomerAMGSetup( void               *amg_vdata,
...
+         }
+
+         if (hypre_ParAMGDataCPointsKeepLevel(amg_data) > 0)
          {
             if (block_mode)
             {
-               hypre_printf("Keeping coarse nodes in block mode is not implemented\n");
+               if (my_id == 0)
+               {
+                  hypre_printf("Keeping coarse nodes in block mode is not implemented\n");
```
rfalgout: We really shouldn't have printf without guarding it print level.  This should probably throw an error code instead.

rfalgout (commented):
@victorapm These generally look okay to me.  I made a few small comments.  Someone like @ulrikeyang , @oseikuffuor1 , or @liruipeng should take a closer look at the larger logic changes just to make sure some algorithmic behavior wasn't accidentally changed.  I assume you ran the regression tests, though?

Thanks!

src/parcsr_ls/HYPRE_parcsr_ls.h
oseikuffuor1: @rfalgout, the interface is a bit different between the two, although the idea is similar. One sets F points *only* for the finest level, whereas the other keeps specific points as C points until some coarse level is reached. So the notion of "keep"-ing the points is not too bad (for MGR, I used MGRSetReservedCoarseNodes). That said, I don't think SetFPoints needs "ToKeep" to be descriptive. I also don't think many external users are using the SetCpointsToKeep() routine. I use it in MGR, but it should be an easy change if we switch the name.

victorapm: Hi all, I chose to drop "ToKeep" for the reason that Daniel pointed out.

src/parcsr_ls/par_amg_setup.c
victorapm: Hi Rob, I modified this to use a call to hypre_error_w_msg instead. See db237b821593a039917c3b91eae42428a0f2d8bc

victorapm:
> @victorapm These generally look okay to me. I made a few small comments. Someone like @ulrikeyang , @oseikuffuor1 , or @liruipeng should take a closer look at the larger logic changes just to make sure some algorithmic behavior wasn't accidentally changed. I assume you ran the regression tests, though?
> 
> Thanks!

Thanks for the review, Rob. I ran machine-tux.sh on my desktop and got no errors.

src/parcsr_ls/HYPRE_parcsr_ls.h
rfalgout: Thanks @oseikuffuor1 and @victorapm .  If we do anything here, I would vote for removing "ToKeep" from the SetCPoints function.  But, I don't have an issue with leaving it the way it is currently either.

ulrikeyang: We could provide both functions with and without ToKeep, but deprecate the ToKeep one and hide it.

src/parcsr_ls/par_amg_setup.c
```diff
@@ -1307,33 +1341,51 @@ hypre_BoomerAMGSetup( void               *amg_vdata,
...
+               }
+            }
+         }
+
+         if (hypre_ParAMGDataCPointsKeepLevel(amg_data) > 0)
          {
             if (block_mode)
             {
-               hypre_printf("Keeping coarse nodes in block mode is not implemented\n");
+               if (my_id == 0)
```
rfalgout: Most of the time we don't restrict these error messages to process 0.  I think we probably shouldn't ever do that, but I can see the potential downside of getting lots and lots of warnings printed when debugging code.  On the other hand, errors aren't universally going to be the same on all processes and we could, in theory, try to manage the message output in the error routines themselves somehow.  Thoughts?  Comments?

bensworth: I personally find it pretty annoying to get hundreds of error messages and have to parse through for anything else that may be important/don't see any downside to just having one processor print, that's just my two cents though

ulrikeyang:
How likely is it for this printout (which should be replaced by hypre_error_w_msg) to occur on the first level which should include proc 0?

liruipeng:
About .gitignore, people have their own ones; I am not sure if we should put it in the repo.

rfalgout:
> About .gitignore, people have their own ones; I am not sure if we should put it in the repo.

Agreed.  I had intended to make this comment as well, but forgot.  The '.gitignore' file should be removed from the pull request.

src/parcsr_ls/par_amg_setup.c
rfalgout: After a bit more thought, I think we should *not* restrict error messages (e.g., with hypre_error() or hypre_error_w_msg()) to a process number.  All of this output goes to stderr and there are a variety of ways to deal with stdio in the mpirun command:

  https://www.open-mpi.org/doc/v3.0/man1/mpirun.1.php

These error messages would only ever be turned on during debugging sessions ('configure --with-print-errors'), and it is better to get the full parallel picture of what's going on when debugging.  From what I can tell, there are only a few other places in the code where we wrap the error messages, and we should remove those if statements when we come across them.

victorapm: In this particular case, block_mode is a variable that assumes the same value across all processors. So, I think it makes sense to throw the error message only from proc 0 and not having redundant messages from other procs as Ben pointed out. Regarding the message output management, I thought about adding a flag to hypre_error_w_msg which decides if the message will be printed by all processors or just the master. Also, in the first case, we could add the processor rank at the beginning of the message, so we would know which rank threw the error when reading it. I saw this in PETSc, by the way.

victorapm:
> How likely is it for this printout (which should be replaced by hypre_error_w_msg) to occur on the first level which should include proc 0? 

It would occur only when the user chooses a block relaxation or interpolation method and tries to set a list of C points to keep up to a certain coarse level.

victorapm:
> > About .gitignore, people have their own ones; I am not sure if we should put it in the repo.
> 
> Agreed. I had intended to make this comment as well, but forgot. The '.gitignore' file should be removed from the pull request.

Ok! Done with 5923a6a

src/parcsr_ls/par_amg_setup.c
victorapm: > After a bit more thought, I think we should _not_ restrict error messages (e.g., with hypre_error() or hypre_error_w_msg()) to a process number. All of this output goes to stderr and there are a variety of ways to deal with stdio in the mpirun command:
> 
> https://www.open-mpi.org/doc/v3.0/man1/mpirun.1.php
> 
> These error messages would only ever be turned on during debugging sessions ('configure --with-print-errors'), and it is better to get the full parallel picture of what's going on when debugging. From what I can tell, there are only a few other places in the code where we wrap the error messages, and we should remove those if statements when we come across them.

Thanks for pointing this out, Rob. I was not aware of those options of mpirun. I deleted the "if my_id == 0" statement for that warning message and another one in par_amg_setup. See 5d0ffec

rfalgout (approved):

src/parcsr_ls/HYPRE_parcsr_amg.c
```diff
@@ -1859,11 +1859,24 @@ HYPRE_BoomerAMGSetDSLUThreshold (HYPRE_Solver solver,
 
 HYPRE_Int
 HYPRE_BoomerAMGSetCpointsToKeep(HYPRE_Solver solver,
-                                HYPRE_Int  cpt_coarse_level,
-                                HYPRE_Int  num_cpt_coarse,
-                                HYPRE_Int *cpt_coarse_index)
+                                HYPRE_Int    cpt_coarse_level,
+                                HYPRE_Int    num_cpt_coarse,
+                                HYPRE_Int   *cpt_coarse_index)
```
oseikuffuor1: @victorapm, should this be HYPRE_BigInt * instead since it is global? Similarly for HYPRE_BoomerAMGSetCPoints( ... ).

victorapm: Yes! Thanks for pointing this out, Daniel. Fixed with 82d208c

liruipeng (approved):

oseikuffuor1:
This PR looks good to me. I don't think I am a reviewer, so once Ulrike approves it, it can be merged. Thanks @victorapm .

ulrikeyang (approved):

liruipeng:
@victorapm I assume you have run the GPU regression tests ...

PR #78, opened 2020-01-10
Install CMake config file, use target-based CMake features

ZedThree:
This PR makes it easier to consume HYPRE in other CMake projects by installing a CMake config file. With this PR, finding and linking against HYPRE can be done in two lines:

```cmake
find_package(HYPRE)
target_link_libraries(my_project PUBLIC HYPRE::HYPRE)
```

This ensures that `my_project` is correctly linked against HYPRE and its dependencies (MPI, BLAS, etc), as well as transitively forwarding those dependencies to other targets linked against `my_project`.

To do this, I had to make some changes to the CMake files so that everything is forwarded correctly, which used some features from more recent versions of CMake that the current minimum supported version.

Slightly unrelated, I also changed the handling of `CMAKE_INSTALL_PREFIX` and `CMAKE_BUILD_TYPE`: previously, they were ignored in favour of the `HYPRE_*` versions. Now they take their values from the `HYPRE_*` versions if they are set, but they can also be set directly.

There's still a few bits left to sort out, notably FEI and SuperLU support.


osborn9 (approved):
This looks good to me. Thanks for the updates!

rfalgout (approved):

oseikuffuor1 (approved):
Looks good to me too.

PR #83, opened 2020-01-22
cmake: Change to use default install location CMAKE_INSTALL_PREFIX is default value.

osborn9:
This is a small change to the previous PR to install hypre in the default location (src/hypre) by default (and not /usr/local).


rfalgout (approved):
This looks good to me, Sarah.  Thanks!

PR #90, opened 2020-02-06
Gpu interp new

liruipeng:
1. Added additional data in SoC that is more helpful in building P on GPUs
2. Based on 1. changed implementation of Dir-Interp on GPUs
3. New interp type 15


rfalgout (approved):
Hi @liruipeng .  I'll leave it to others to look at this more closely.  It looked fine to me.  My only comment/question is, why do you need a new interptype (=15) here?  Thanks!

liruipeng:
> Hi @liruipeng . I'll leave it to others to look at this more closely. It looked fine to me. My only comment/question is, why do you need a new interptype (=15) here? Thanks!

The original interp 3 (dir) was kept the same mathematically with a different (better) implementation on GPUs. Interp 15 (which can be just a temporary number in this branch) is new, which modified the dir interp formula with some ideas as in "collapsing the stencil". Some preliminary tests showed "15" is better than "3". We can talk about them later when merging into the master. I have too many GPU branches to manage; just want to merge some of them here.

PR #93, opened 2020-02-28
Amg setup

liruipeng:


src/utilities/hypre_cub_allocator.h
```diff
@@ -0,0 +1,844 @@
+/******************************************************************************
+ * Copyright 1998-2019 Lawrence Livermore National Security, LLC and other
+ * HYPRE Project Developers. See the top-level COPYRIGHT file for details.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR MIT)
+ ******************************************************************************/
```
rfalgout: I'm not sure it's okay to put this copyright header in this file.

rfalgout (approved):
Hi @liruipeng .  Whew!  That's a lot of code changes. :)  It all looks good to me, though I do have a couple of comments/questions:

- I would like to have a discussion about the memory model changes you made at one of our upcoming meetings.  I am still a little bit concerned about supporting this memory location stuff at this level of granularity.
- I didn't understand why there are '_v2' routines of various kinds.
- With the hypre_Vector, you use the 'Complex' data type.  What about situations where we really want 'Real'?

That's it.  I'm good with this as is for now, with the exception of the CUB copyright statement (I want to make sure we do this correctly).  The above issues can be sorted out later.

Nice job!  Thanks!

-Rob

bjorn2 (approved):
Lot of changes, but it looks all right.

liruipeng:
@bjorn2 @victorapm Please do not push to this branch with any more new development but just the fixes only. I want to delete it after the merge is done. You may want to do `git branch -b new_dev` from this branch for your new development . Thanks! -Ruipeng

ulrikeyang (commented):
Ruipeng, why are you changing l1_norms to a vector?

osborn9 (approved):

liruipeng:
> Ruipeng, why are you changing l1_norms to a vector?

Since it can be either on device or host, which depends on different paths of code, it is nicer to put it into a container like hypre_vector that has `memory_location`, so we don't need to keep track where is it, for instance when freeing it. `hypre_SeqvectorDestroy` will do the free properly

oseikuffuor1:
This looks good to me. Thanks for all the work!

oseikuffuor1 (approved):

ulrikeyang (approved):
I approve this for now, but agree with Rob that we need to change the use of hypre_ParVector you introduced in the future, maybe with some hypre_RealArray data structure or so.

victorapm (approved):

liruipeng (commented):
Have rerun regression tests and reviewed all the files changed. Ready to be merged into master

Issue #96, opened 2020-03-15
Seg faults in coarsening


bensworth:
Our code began getting seg faults in the coarsening routines starting at this commit: https://github.com/hypre-space/hypre/commit/b2f6d96e7cbc7e93f13c4ab81de126b8038bdcb6

Our code constructs parallel matrices in MFEM and interfaces to hypre. Below is where we get seg faults for two different types of coarsening schemes. This does not happen prior to the above commit. 

Falgout:
```
[quartz8:mpi_rank_0][error_sighandler] Caught error: Floating point exception (signal 8)
==12819== 
==12819== Process terminating with default action of signal 8 (SIGFPE): dumping core
==12819==    at 0x838A4FB: raise (pt-raise.c:36)
==12819==    by 0x838A62F: ??? (in /usr/lib64/libpthread-2.17.so)
==12819==    by 0x87180F: hypre_BoomerAMGCoarsenRuge (par_coarsen.c:901)
==12819==    by 0x87492E: hypre_BoomerAMGCoarsenFalgout (par_coarsen.c:2005)
==12819==    by 0x85E72B: hypre_BoomerAMGSetup (par_amg_setup.c:1237)
==12819==    by 0x83FBA7: HYPRE_BoomerAMGSetup (HYPRE_parcsr_amg.c:47)
==12819==    by 0x836F5C: hypre_GMRESSetup (gmres.c:216)
==12819==    by 0x8351B6: HYPRE_GMRESSetup (HYPRE_gmres.c:37)
==12819==    by 0x841978: HYPRE_ParCSRGMRESSetup (HYPRE_parcsr_gmres.c:60)
==12819==    by 0x6640DC: mfem::HypreGMRES::Mult(mfem::HypreParVector const&, mfem::HypreParVector&) const (hypre.cpp:2496)
==12819==    by 0x66372E: mfem::HypreSolver::Mult(mfem::Vector const&, mfem::Vector&) const (hypre.cpp:2283)
==12819==    by 0x5F0DA1: TRTDSAOperator::NewtonStepProd(mfem::Vector&, mfem::Vector&, mfem::GridFunction&, double, mfem::Vector const&, mfem::Vector&) (TRTOperator.cpp:1433)

```
HMIS:
```
[quartz5:mpi_rank_0][error_sighandler] Caught error: Floating point exception (signal 8)
==30205== 
==30205== Process terminating with default action of signal 8 (SIGFPE): dumping core
[...]

PR #98, opened 2020-03-19
Fix issue96

victorapm:
When aggressive coarsening is used, the second stage strength matrix S2 might have zero size. In this case, computing the average number of nonzeros per row, as done by BoomerAMGCoarsenRuge, causes a SIGFPE (division by zero). With this PR, the computation of avg_nnzrow is done only when it makes sense. A new regression test is added to make sure the code works in this corner case. This fixes issue #96.


rfalgout (approved):
Looks good to me.  Thanks @victorapm !

ulrikeyang (approved):

liruipeng (approved):

PR #102, opened 2020-03-27
access global var _hypre_handle via hypre_handle(); and L1-norm

liruipeng:
This issue was raised by `MFEM` after we changed the `master` lately, where `HYPRE_Init` and `HYPRE_Finalize` are not called in MFEM and it is difficult to add these calls (see https://github.com/mfem/mfem/issues/1376#issuecomment-605086854), and so `hypre_handle` cannot be created. This PR is a workaround for non-GPU builds that can fix this issue.

Global variable is renamed:
```
hypre_Handle *_hypre_handle
```
as its name starting with `_` suggests, it should not be accessed directly, so an accessor is added
```
static inline hypre_Handle*
hypre_handle()
{
   if (!_hypre_handle)
   {
      _hypre_handle = hypre_HandleCreate();
   }

   return _hypre_handle;
}
```
which will create the handle if it has not been created. Most of changes in this PR were changing `hypre_handle` to `hyper_handle()`

Ideally, we still want the users to call `HYPRE_Init` and `HYPRE_Finalize`


rfalgout (approved):
Looks good to me, @liruipeng .  Thanks!  -Rob

src/struct_mv/struct_matrix.c
```diff
@@ -1770,7 +1770,7 @@ hypre_StructMatrixMigrate( hypre_StructMatrix *from_matrix,
    hypre_assert( constant_coefficient == hypre_StructMatrixConstantCoefficient( to_matrix ) );
 
    mat_num_values = hypre_StructMatrixNumValues(from_matrix);
-   hypre_assert( mat_num_values = hypre_StructMatrixNumValues(to_matrix) );
+   hypre_assert( mat_num_values == hypre_StructMatrixNumValues(to_matrix) );
```
liruipeng, line 1773: embarrassingly bad bug!

liruipeng:
More changes were added to resolve the incompatible interface issue regarding `L1-norm` of several `hypre_` functions that are used by MFEM, which was a bad decision I made to change it at the beginning. Changed function arguments `L1-norm` back to `HYPRE_Real *` but `L1-norm` itself is still kept as a `hypre_Vector` in `hypre_ParAMGData`

liruipeng:
@ulrikeyang @oseikuffuor1 Any comments on this PR? Thanks! -R

ulrikeyang (commented):
I still don't like to use the vector for the l1 norms. We still have the problem with complex. Why can't we introduce something like hypre_RealArray or so?

liruipeng:
> I still don't like to use the vector for the l1 norms. We still have the problem with complex. Why can't we introduce something like hypre_RealArray or so?

Hi, @ulrikeyang, @rfalgout @oseikuffuor1, Having something like `hypre_Array` for L1-norms makes sense to me, but I am not sure what's the best way to address this problem. Shall we have individual `hypre_RealArray`, `hypre_IntArray`, etc for each data type? Or, we want to have one new type for all (C++ template is a good thing but we cannot use it since we use C). Another approach is to use `void *`, like
```
typedef struct hypre_array
{
   HYPRE_Int size;
   HYPRE_Int elem_type; /* 0: Int; 1: double; 2: ... */
   void *data;
}
```
But `void *` is not preferred either. We should think about it a bit more.

Anyways, shall we address this issue in this PR, or we can do it later? The current master breaks MFEM, so I would like to let the fix happen asap

Thanks

-Ruipeng

rfalgout:
Just to add my 2 cents to the vector discussion, I think this can wait until after the current PR.  However, I agree with Ulrike that we should have IntArray, RealArray, ComplexArray instead.  I know it's not templated C++, but neither are our current Int, Real, and Complex types.  Also, the arrays you originally replaced with vectors had specific types.

-Rob

ulrikeyang (approved):
ok. I approve for now :)

oseikuffuor1 (approved):

PR #104, opened 2020-04-09
added new extended(interptype 16), extended+i(iinterptype 17) and i

ulrikeyang:
Hi all,
I have added the new interpolation routines to hypre. The regression tests on tux were clean (except cmake, since I have an old cmake). I also have not run the gpu regression tests.
There are basically 3 new interpolation routines: new extended, new extended+i and new 2stage extended. There is also a new option for 2stage extended+i (agg_interp 6), but it does not have all the new elements, just uses the new extended+i for one stage, but the old portion for the second stage.

### DIFF
--- a/src/parcsr_ls/CMakeLists.txt
+++ b/src/parcsr_ls/CMakeLists.txt
@@ -50,6 +50,7 @@ set(SRCS
   HYPRE_ams.c
   HYPRE_ads.c
   HYPRE_ame.c
+  par_2s_interp.c
   par_amg.c
   par_amg_setup.c
   par_amg_solve.c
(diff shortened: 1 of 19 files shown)


rfalgout (approved):
Looks great to me!  We should at least do the additional testing you mentioned.  Thanks!

-Rob

ulrikeyang:
Hi all,
I have fixed all cmake issues and GPU memory issues. The current versions run clean. Any comments from anybody besides Rob who approved it?
Can you please approve so this can be merged?
Thanks
Ulrike

oseikuffuor1 (approved):
Looks good to me. Thanks Ulrike.

liruipeng (approved):
Looks good to me. Thanks, Ulrike!

PR #106, opened 2020-04-16
SuperLU bug fixes

oseikuffuor1:
This PR fixes some issues with using superLU as a coarse grid solver within BoomerAMG:
1. Fixed a memory bug for cases when the superLU solver is destroyed without being created
2. Fixed mixedInt bug that resulted from inconsistent integer types between superLU and hyper
3. Updated hypre to be consistent with latest version of superLU.
NOTE: There are still some compiler warnings with building hypre with superLU, since superLU uses 32bit integer types for some of its function arguments even if xsdk_index = 64. Another warning that shows up is because hype's BigInt is of type long long, where as superLU uses long when xsdk_index = 64. I believe this warning is benign, but I am not sure how to turn it off.

### DIFF
--- a/src/parcsr_ls/HYPRE_parcsr_ilu.c
+++ b/src/parcsr_ls/HYPRE_parcsr_ilu.c
@@ -177,3 +177,11 @@ HYPRE_ILUGetFinalRelativeResidualNorm(  HYPRE_Solver solver, HYPRE_Real *res_nor
 {
    return hypre_ILUGetFinalRelativeResidualNorm(solver, res_norm);
 }
+/*--------------------------------------------------------------------------
+ * HYPRE_ILUSetLocalReordering
+ *--------------------------------------------------------------------------*/
+HYPRE_Int
+HYPRE_ILUSetLocalReordering(  HYPRE_Solver solver, HYPRE_Int ordering_type )
+{
+   return hypre_ILUSetLocalReordering(solver, ordering_type);
+}
--- a/src/parcsr_ls/HYPRE_parcsr_ls.h
+++ b/src/parcsr_ls/HYPRE_parcsr_ls.h
@@ -1195,6 +1195,12 @@ HYPRE_Int HYPRE_BoomerAMGGetGridHierarchy(HYPRE_Solver solver,
 #ifdef HYPRE_USING_DSUPERLU
 /**
  * HYPRE_BoomerAMGSetDSLUThreshold
+ *
+ * Usage:
+ *  Set slu_threshold >= max_coarse_size (from HYPRE_BoomerAMGSetMaxCoarseSize(...))
+ *  to turn on use of superLU for the coarse grid solve. SuperLU is used if the 
+ *  coarse grid size > max_coarse_size and the grid level is < (max_num_levels - 1) 
+ *  (set with HYPRE_BoomerAMGSetMaxLevels(...)).
  **/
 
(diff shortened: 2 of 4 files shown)


rfalgout (approved):
Looks good to me.  Thanks!  -Rob

ulrikeyang (approved):
Looks good to me

oseikuffuor1:
@osborn9, @victorapm, @liruipeng, can you take a look and approve when you get the chance?

Thanks.

osborn9 (approved):
Looks good and works for me. 

Back to Daniel/Rob's comments about documenting the version....it would probably be good in the Changelog of the next release with these change to note the version change of SuperLU_DIST.

oseikuffuor1:
Thanks @osborn9 . I was wondering where would be a good place to document this. The Changelog is a good idea.

victorapm:
Hi @oseikuffuor1,

The changes look good to me. Regarding the warnings, I think they are not a problem for 64-bit Linux users (LP64 data model) since "long long" and "long" have the same width in bits, however, this is not the case for Windows users (LLP64 data model) and this could be an issue (See [this link](https://en.cppreference.com/w/cpp/language/types)). We could insert a check in the code to warn for this particular situation or copy data that are "long long" (HYPRE_BigInt) to long before passing it to SuperLU_dist.

liruipeng (approved):

victorapm (approved):

oseikuffuor1:
Thanks for the info @victorapm. @ulrikeyang does the xsdk policy require 64bit integers across all architectures when XSDK_INDEX_SIZE=64 is set? The implementation of BigInt in hypre satisfies this, but as @victorapm mentioned, int_t in superLU will be a 32bit integer on windows systems. One option is to leave this as it is for now, as I suspect superLU will change this to conform with xsdk policy. The alternative will be to copy from BigInt arrays to int_t arrays before calling superLU. This will make the integration compatible across difference architectures. Ideally I would like to avoid the copying, but I am leaning towards the latter option to avoid possible issues later on. Any objections? thoughts??

liruipeng:
> Thanks for the info @victorapm. @ulrikeyang does the xsdk policy require 64bit integers across all architectures when XSDK_INDEX_SIZE=64 is set? The implementation of BigInt in hypre satisfies this, but as @victorapm mentioned, int_t in superLU will be a 32bit integer on windows systems. One option is to leave this as it is for now, as I suspect superLU will change this to conform with xsdk policy. The alternative will be to copy from BigInt arrays to int_t arrays before calling superLU. This will make the integration compatible across difference architectures. Ideally I would like to avoid the copying, but I am leaning towards the latter option to avoid possible issues later on. Any objections? thoughts??

I also think it is SUPERLU that should address its 64-bit integer interface compatibility per XSDK rule

victorapm:
I agree with Ruipeng. Meanwhile, we could implement the data conversion only when sizeof(long) != sizeof(long long). Then, the additional data copy would occur only for a small subset of users (I suppose most of them run hypre under a 64-bit Linux platform). If SuperLU updates "int_t" to "long long", this new data copy portion of hypre would become obsolete and we could delete it.

ulrikeyang:
I think we should check with Sherry on that, specifically the fact that this violates xSDK policies.

oseikuffuor1:
Thanks everyone for chiming in. I will do a check like Victor suggested and only do the copy as needed (this is only done in superlu.c anyway). In the mean time, I will email Sherry about this issue. When they make the change, I can go back and simplify the integration. 

Thanks!

Issue #108, opened 2020-04-28
Deadlock in `HYPRE_StructSMGSetup` for some specific domain decompositions


RemiLacroix-IDRIS:
Hello,

One of our users reported a deadlock with his application that I tracked it down to `HYPRE_StructSMGSetup` and more specifically `hypre_FinalizeCommunication`.

I was able to reproduce with this small sample code: https://gist.github.com/RemiLacroix-IDRIS/ec88a1ab976454a675f42084e97a668f. To compile, just use `mpif90 -g -o test.exe test.f90 -lHYPRE -Lpath/to/hypre/lib` and run with `mpirun -np 12 ./test.exe` (the test case is hard coded for 12 MPI processes).

The problem occurs only with some specific domain decompositions. For example, I can reproduce on a 256x256x384 domain with a 2x2x3 MPI topology but not with a 256x384x256 domain with a 2x3x2 MPI topology.

I added some `printf` to the functions `hypre_InitializeCommunication` and `hypre_FinalizeCommunication` in the `struct_mv/struct_communication.c` file to see exactly what is going on.

Here is what I get for processes 0 and 4 with a 256x256x384 domain and a 2x2x3 MPI topology:
 * process 0
```
req 0 - irecv --> receiving 16389 elements from 1
req 1 - irecv --> receiving 16389 elements from 2
req 2 - irecv --> receiving 261 elements from 4
req 3 - irecv --> receiving 261 elements from 5
req 4 - irecv --> receiving 133 elements from 7
req 5 - irecv --> receiving 133 elements from 8
req 6 - irecv --> receiving 7 elements from 10
req 7 - irecv --> receiving 7 elements from 11
1/8 - wait req 2 ok
2/8 - wait req 6 ok
3/8 - wait req 0 ok
4/8 - wait req 4 ok
```
 * process 4
```
req 0 - irecv --> receiving 261 elements from 0
req 1 - irecv --> receiving 261 elements from 2
req 2 - irecv --> receiving 16389 elements from 3
req 3 - irecv --> receiving 16389 elements from 5
req 4 - irecv --> receiving 7 elements from 6
req 5 - irecv --> receiving 7 elements from 8
req 6 - irecv --> receiving 133 elements from 9
req 7 - irecv --> receiving 133 elements from 11
```

As you can see both are waiting for each other to send  261 elements, hence the deadlock.

[...]

PR #110, opened 2020-05-11
Fix issue108

victorapm:
This PR fixes issue #108. 

When building the interpolation operator of SMG for a case with periodic boundary conditions and more than one processor in the coarsening direction, it may happen that a box has zero volume, meanwhile, its corresponding rbox has a non-zero volume and vice-versa. In this situation, the communication matrix is non-symmetric, resulting in a deadlock. With this PR, boxes are communicated (sent/received) only if their matching rboxes have non-zero volume. A new regression test was added to make sure the code works properly in this particular situation.

### DIFF
--- a/src/struct_ls/semi.c
+++ b/src/struct_ls/semi.c
@@ -81,7 +81,7 @@ hypre_StructInterpAssemble( hypre_StructMatrix  *A,
    hypre_CommInfoProjectSend(comm_info, index, stride);
    hypre_CommInfoProjectRecv(comm_info, index, stride);
 
-   for (s = 0; s < 3; s++)
+   for (s = 0; s < 4; s++)
    {
       switch(s)
       {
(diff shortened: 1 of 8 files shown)


rfalgout (approved):
Hi @victorapm .  This looks great to me!  Thanks for looking into it!  I was surprised when the report came in because SMG really hasn't changed in long time.  You would think this problem would have come up before, but maybe it's because periodic conditions are used all that often?  Anyway, nice job tracking it down and fixing it!

ulrikeyang (approved):

victorapm:
Hi Rob, thanks! It was an interesting task. I understood better how commPkg works in struct. I was surprised too! I believe it was for the reason you pointed out.

liruipeng:
Tuxbox regression tests break after this merge. gdb pointed to `struct_communication.c`

PR #111, opened 2020-05-12
Rl gpu

liruipeng:
1. GPU FFFC code. Wrote it using thrust, without writing any raw cuda kernels. Somewhat complicated but it will be directly portable to AMD GPUs. May consider more in this line of future development
2. Optimize CPU FFFC, so should produce the same matrices as by the GPU code
3. Make minor changes in the SoC matrix from the device function that generates 
4. Minor cosmetic format changes


rfalgout (approved):
There's no way for me to follow the details of this, but it otherwise looks fine.

ulrikeyang (approved):
Looks fine

oseikuffuor1 (approved):
Looks good to me.

ulrikeyang (approved):
Looks fine to me

PR #113, opened 2020-05-13
Fix regression tests after #110

victorapm:
This PR fixes the regression tests involving FAC after #110


rfalgout (approved):
Looks good.  Thanks!

ulrikeyang (approved):

PR #118, opened 2020-05-15
Updating autotest for SuperLU-dist version 6.3.1

rfalgout:
Needed to add -lstdc++ to link line in machine-tux-exlibs.sh and change version of superlu-dist.
Also changed 'superlu.c' to get rid of some warnings (and removed some whitespace).


oseikuffuor1 (approved):
Looks good and clean to me. Thanks @rfalgout . Are you testing the BigInt or MixedInt cases as well, or just the basic int case?

rfalgout:
Hi @oseikuffuor1 .  Just testing the regular case at the moment.  I am going to work on the spack idea we talked about first before trying to add other tests, because it may be easier to do it that way.  Thanks!

PR #119, opened 2020-05-16
Adding a Spack autotest script

rfalgout:
This adds a basic autotest capability for testing spack installs.
Currently, only a test of superlu-dist is done.
There is a commented out examples test that fails because of lobpcg differences.  I think the idea of completely separating the lobpcg tests should be done at some point.


AUTOTEST/machine-tux-spack.sh
```diff
@@ -0,0 +1,64 @@
...
+# spack load    $spackspec
+# spackdir=`spack location -i $spackspec`
+# test.sh basic.sh ../src -co: -mo: -spack $spackdir -eo: -spack $spackdir
+# ./renametest.sh basic $output_dir/basic-examples
+
+spackspec="hypre@develop~debug+superlu-dist"
+spack install $spackspec
+spack load    $spackspec
+spackdir=`spack location -i $spackspec`
+test.sh basic.sh ../src -co: -mo: -spack $spackdir -ro: -superlu
```
oseikuffuor1, line 57: Hi @rfalgout , is there a reason why you chose -superlu instead of -dsuperlu here? I assume we used the latter to indicate that we use super-dist (and not superlu).  Thanks.

rfalgout: Hi @oseikuffuor1 .  The '-superlu' indicates the tests in 'src/test/TEST_superlu', which are actually superlu-dist tests from what I can tell.  I am just repeating what we were already running in the 'machine-tux-exlibs.sh' autotest script.  Is this what you were asking?

oseikuffuor1: @rfalgout Yes, that's what I was asking. Thanks for the clarification.

oseikuffuor1 (approved):

osborn9 (approved):
Nice! Looks good and works for me.

liruipeng (approved):

ulrikeyang (approved):

PR #121, opened 2020-05-19
Gpudev bjorn

liruipeng:
GPU Ext/Ext+i interpolations


ulrikeyang (approved):
Looks good to me, although I trust that the CUDA changes are correct.

src/parcsr_ls/par_lr_interp.c
```diff
@@ -591,9 +576,7 @@ hypre_BoomerAMGBuildStdInterp(hypre_ParCSRMatrix  *A,
             {
                indx = ihat[i1];
                if (indx > -1)
-               {
                   ahat[indx] += A_diag_data[jj];
-               }
```
rfalgout: Why are all of the open/close squirly braces being removed?  Didn't you just go through a big effort to put them in?

rfalgout (approved):
Looks good, though I didn't understand why the '{' and '}' delimiters were removed in many places.  In general, the convention is to always put these in to avoid accidental logic issues.  We can wait to fix this later, though.

liruipeng:
> Looks good, though I didn't understand why the '{' and '}' delimiters were removed in many places. In general, the convention is to always put these in to avoid accidental logic issues. We can wait to fix this later, though.

I really don't know how it could happen. I had to manually add them back...

bjorn2 (approved):
Well, lot of changes.  Looks ok, assume you have tested that the code still gives correct results.

PR #122, opened 2020-05-21
Uninstall spack packages in hypre+superlu-dist test.

osborn9:


rfalgout (approved):
Hi @osborn9 .  This looks great!  I made a couple of minor changes to the script (the behavior is unchanged) to document what we are doing and to use quotes in a more bash-like way (at least in the way I've always done it :).  If you approve, go ahead and merge.  Thanks!

PR #128, opened 2020-06-02
new sparse mat-mat-dist, triple-mat-dist

liruipeng:
Re-implemented GPU R^TAP, R^TA and AP routines, which showed better performance than the previous ones.


rfalgout (approved):
Hi @liruipeng .  I can't follow the details on this, but I don't see anything to worry about.  It would be good to get an overview presentation of what you've done here versus what was there before.  Thanks!

-Rob

ulrikeyang (approved):
I haven't checked this in detail.

PR #134, opened 2020-06-09
bug fix (omp with large #threads)

liruipeng:
Fixed a problem in `hypre_ParCSRMatrixGenerateFFFC` with OMP, when `OMP_NUM_THREADS` is sufficiently large. This was accidentally found when doing tests on Lassen with `srun`, where `OMP_NUM_THREADS` was automatically set to be `160`!


src/parcsr_mv/gen_fffc.c
```diff
@@ -346,6 +346,12 @@ hypre_ParCSRMatrixGenerateFFFC( hypre_ParCSRMatrix  *A,
          {
             fpt = fpt_array[i];
             fpt2 = fpt_array[i-1];
+
```
liruipeng, line 349: A simple fix but hard to find. In the cases that `num_threads` is large enough, a thread can have no F-pt in its "partition", so we must skip the adds in the rest of the loop.

src/parcsr_mv/gen_fffc.c
```diff
@@ -346,6 +346,12 @@ hypre_ParCSRMatrixGenerateFFFC( hypre_ParCSRMatrix  *A,
...
+            if (fpt == fpt2)
```
liruipeng, line 350: A simple fix but hard to find. In the cases that num_threads is large enough, a thread can have no F-pt in its "partition", so we must skip the adds in the rest of the loop.

rfalgout (approved):
Looks good to me.

ulrikeyang (approved):
We probably should check if we have the same problem somewhere else.

Issue #126, opened 2020-05-27
Compilation error with 2.19


stefanozampini:
I got a compilation error when trying to build 2.19 with cuda support (same configuration without --with-cuda goes through fine, below is the error together with the configure line and compiler used)

```
Making pilut ...
make[2]: Entering directory '/home/zampins/hypre/src/distributed_ls/pilut'
nvcc -O2 -lineinfo -ccbin=mpicxx -gencode arch=compute_60,"code=sm_60" -expt-extended-lambda -dc -std=c++11 --x cu -Xcompiler "-fstack-protector -g3 -O2 -fPIC " -Xcompiler "-fPIC" -DHAVE_CONFIG_H -I../.. -I. -I./../.. -I./../../blas -I./../../utilities -I./../../distributed_matrix -I/opt/ecrc/cuda/10.2/include   -c debug.c 
debug.c(31): error: expected a member name

debug.c(43): error: expected a member name

debug.c(60): error: expected a member name

debug.c(82): error: expected a member name

debug.c(105): error: expected a member name

debug.c(129): error: expected a member name

debug.c(138): error: expected a member name

debug.c(178): error: expected a member name

debug.c(183): error: expected a member name

debug.c(199): error: expected a member name

debug.c(202): error: expected a member name

debug.c(217): error: expected a member name

debug.c(218): error: expected a member name

debug.c(219): error: expected a member name

14 errors detected in the compilation of "/tmp/tmpxft_00009be0_00000000-6_debug.cpp1.ii".
../../config/Makefile.config:42: recipe for target 'debug.o' failed
make[2]: *** [debug.o] Error 1
make[2]: Leaving directory '/home/zampins/hypre/src/distributed_ls/pilut'
Makefile:14: recipe for target 'all' failed
make[1]: *** [all] Error 1
make[1]: Leaving directory '/home/zampins/hypre/src/distributed_ls'
Makefile:86: recipe for target 'all' failed
make: *** [all] Error 1
```
configuration:
```
[...]

Issue #125, opened 2020-05-27
Cuda compute capability requirements?


stefanozampini:
I could not find any documentation about the current compute capability requirements of the 2.19 version. It seems sm_37 is not supported, by judging from the below compilation error. We are trying to finally integrate HYPRE GPU capabilities into PETSc. (A comment aside: it is not clear to me if BoomerAMG solve phase will still perform on the GPU when cuda-unified-memory is not used)
Thanks

Making seq_mv ...
make[1]: Entering directory '/home/zampins/hypre/src/seq_mv'
nvcc -O2 -lineinfo -ccbin=mpicxx -gencode arch=compute_37,"code=sm_37" -expt-extended-lambda -dc -std=c++11 --x cu -Xcompiler "-fstack-protector -g3 -O2 -fPIC " -Xcompiler "-fPIC" -DHAVE_CONFIG_H -I.. -I. -I./.. -I./../utilities -I/opt/ecrc/cuda/10.2/include   -c csr_spgemm_device_attempt.c 
csr_spgemm_device_attempt.c(52): error: no instance of overloaded function "atomicAdd" matches the argument list
            argument types are: (HYPRE_Complex *, HYPRE_Complex)
          detected during:
            instantiation of "HYPRE_Int csr_spmm_compute_row_attempt<attempt,HashType>(HYPRE_Int, HYPRE_Int, HYPRE_Int *, HYPRE_Int *, HYPRE_Complex *, HYPRE_Int *, HYPRE_Int *, HYPRE_Complex *, HYPRE_Int, volatile HYPRE_Int *, volatile HYPRE_Complex *, HYPRE_Int, HYPRE_Int *, HYPRE_Complex *, char &, volatile char *) [with attempt=1, HashType='L']" 
(251): here
            instantiation of "void csr_spmm_attempt<NUM_WARPS_PER_BLOCK,SHMEM_HASH_SIZE,attempt,HashType>(HYPRE_Int, HYPRE_Int *, HYPRE_Int *, HYPRE_Complex *, HYPRE_Int *, HYPRE_Int *, HYPRE_Complex *, HYPRE_Int *, HYPRE_Complex *, HYPRE_Int *, HYPRE_Int *, HYPRE_Complex *, HYPRE_Int *, HYPRE_Int *) [with NUM_WARPS_PER_BLOCK=20, SHMEM_HASH_SIZE=128, attempt=1, HashType='L']" 
(463): here

[...]

PR #135, opened 2020-06-09
Rl fix

liruipeng:
@rfalgout @ulrikeyang 
Several bug fixes on GPU build, PILUT compile errors with gcc, and interfacing with PETSc

1. It seems that GCC compilers (new versions, see #126) get confused by something in PILUT like 
```
typedef struct {
HYPRE_Int mype;
...
};
#define mype (global->mype)
```
where `mype` appears in both `#define` and the struct. Seems to have been fixed by changing the names in the struct.

2. `atomicAdd`, used in sparse matrix-matrix, was supported for `double` only after `CUDA_ARCH` 60. Add a workaround for older CUDA's (see https://github.com/hypre-space/hypre/issues/125)

3. In hypre 2.19, `hypre_AuxParCSRMatrixInitialize` was renamed to its `_v2` interface in with an additional `memory_location` argument. The original function is still needed by PETSc. Had to add it back with default `memory_location`. 

4. There was an issue [reported by Holger Jones, issue1772] from PETSc where `hypre_IJMatrixSetDiagOffdSizesParCSR` https://github.com/hypre-space/hypre/blob/f875c2a38b52d10264ecfa5a0a4114eb659510b0/src/IJ_mv/IJMatrix_parcsr.c#L197 was called, which set internal flag `hypre_AuxParCSRMatrixNeedAux(aux_matrix)` https://github.com/hypre-space/hypre/blob/f875c2a38b52d10264ecfa5a0a4114eb659510b0/src/IJ_mv/IJMatrix_parcsr.c#L237 to `0`, but this flag was manipulated in PETSc after (set back to `1`) for some reason. It was also related to our code refactor in 2.19 that broke PETSc.


src/distributed_ls/pilut/DistributedMatrixPilutSolver.h
```diff
@@ -33,23 +33,23 @@
 typedef struct
 {
 MPI_Comm hypre_MPI_communicator;
-HYPRE_Int mype, npes;
+HYPRE_Int _mype, _npes;
```
rfalgout, line 36: @liruipeng What's the reason for the underscores?

src/utilities/_hypre_utilities.h
```diff
@@ -1455,6 +1456,26 @@ hypre_int hypre_cuda_get_grid_warp_id()
           hypre_cuda_get_warp_id<bdim>();
 }
 
+#if defined(__CUDA_ARCH__) && __CUDA_ARCH__ < 600
```
rfalgout, line 1459: @liruipeng What's going on here?  Where is atomicAdd even called?

rfalgout (commented):
Hi @liruipeng ,

Could you provide a bit more description about what's going on with this pull request.  I'm having difficulty seeing what it is fixing, especially with respect to PETSc.  Thanks!

-Rob

src/distributed_ls/pilut/DistributedMatrixPilutSolver.h
```diff
@@ -78,31 +78,31 @@ HYPRE_Int Ul_timer;
 
 /* DEFINES for global variables */
 #define pilut_comm (globals->hypre_MPI_communicator)
-#define mype (globals->mype)
```
liruipeng, line 81: @rfalgout My guess is GCC compilers (new versions, see https://github.com/hypre-space/hypre/issues/126#issue-625738732) got confused from these lines (starting from 81). Must have something to do with name conflicts in, e.g., `#define mype (global->mype)` and `mype` in the struct. Seems working with the new changes.

rfalgout (approved):
Looks good, @liruipeng .  Thanks!

ulrikeyang (approved):
Looks good

PR #138, opened 2020-06-15
Headers

liruipeng:
The goal of this PR is to separate C++/CUDA headers with the traditional C headers. Namely, the C++ stuff in `_hypre_utilities.h` and `_hypre_struct_mv.h` has been moved out into C++ headers `_hypre_utilities.hpp` and `_hypre_struct_mv.hpp`, which are only included by the files that really need C++/CUDA. This was originated by the issues from PETSc who does not like the way we spread CUDA stuff everywhere in `_hypre_utilities.h`

1. There are a lot of changes, but most of them are simply due to the headers reorganizations.
2. In the `Makefiles`, we split the files into `CFILES` and `CUFILES`. The former ones are always compiled by `CC`, no matter with CUDA or not, but the latter ones are compiled with `CUCC` for GPU-build. In the GPU-build, `CUFILES` are compiled into `.obj` files instead of `.o`, while for regular CPU build they are still `.o`. This was how I dealt with the `makefile` compile rules. I don't know a cleaner way. If you do, please let me know.


src/IJ_mv/Makefile
```diff
@@ -40,12 +42,20 @@ FILES =\
...
- IJVector_parcsr.c\
+ IJVector_parcsr.c
+
+CUFILES =\
+ IJMatrix_parcsr_device.c\
  IJVector_parcsr_device.c
 
-OBJS = ${FILES:.c=.o}
+COBJS = ${FILES:.c=.o}
+ifeq ($(CUCC), )
```
rfalgout: Why not use a rule in Makefile.config.in instead of having this if statement?  The '.obj' rule can be the same as the '.o' rule when not using CUDA.

src/parcsr_ls/ams.c
```diff
@@ -8,6 +8,7 @@
 #include "_hypre_parcsr_ls.h"
 #include "float.h"
 #include "ams.h"
+#include "hypre_cuda_utils.h"
```
rfalgout: I suppose I can check myself, but I was wondering how many files that are not '*_device.c' files have this include?

src/parcsr_mv/par_csr_matvec.c
```diff
@@ -12,6 +12,8 @@
  *****************************************************************************/
 
 #include "_hypre_parcsr_mv.h"
+#include "hypre_cuda_utils.h" //RL: TODO par_csr_matvec_device.c, include cuda there
```
rfalgout: Looks like you wanted to move this into another file.

src/sstruct_ls/fac_zero_stencilcoef.c
```diff
@@ -6,6 +6,7 @@
  ******************************************************************************/
 
 #include "_hypre_sstruct_ls.h"
+#include "boxloop.h"
```
rfalgout: What's going on here?

src/test/ij.c
```diff
@@ -30,6 +30,8 @@
 #include "superlu_ddefs.h"
 #endif
 
+#include "../utilities/hypre_cuda_utils.h"
```
rfalgout: Users should not include files from the source tree, only installed files.  We try to make the test drivers somewhat independent of internal hypre stuff.  Why is this needed here?

src/test/struct.c
```diff
@@ -12,6 +12,7 @@
 #include "_hypre_utilities.h"
 #include "HYPRE_struct_ls.h"
 #include "HYPRE_krylov.h"
+#include "../struct_mv/boxloop.h"
```
rfalgout: Same question as for `../utilities/hypre_cuda_utils.h`.

src/utilities/hypre_memory.c
```diff
@@ -12,7 +12,13 @@
  *****************************************************************************/
 
 #include "_hypre_utilities.h"
+#include "hypre_cuda_utils.h"
```
rfalgout: Our header file stuff just got a lot more complicated.  Are all of these include files really needed?  I would like to maintain an overall strategy for header file use in the code (I realize there are inconsistencies already, but we should be trying to fix those).  Also, the directory-structure-dependent includes are something we should be avoiding.

rfalgout (commented):
Hi @liruipeng .  Thanks for taking a stab at cleaning up the CUDA compile issue!  I need to understand the overall strategy here before I can approve, so we'll need to iterate a bit first.

src/IJ_mv/Makefile
liruipeng: OK. The goal is to have a different rule to compile certain `.c` files to `.obj` when CUDA is enabled, and use the same rule from `.c` to `.o` without CUDA. I don't want to rename certain `.c` files to another extension such as `.cu`. I don't know if I understand what you were suggesting. It is not clear for me how to switch rules in `Makefile.config`.

src/parcsr_ls/ams.c
liruipeng: I don't like the fact that this file needs to be compiled by `CUCC` either. The problem is due to some junk such as this line https://github.com/hypre-space/hypre/blob/3f8ff325b4577e055d8f826eb2a88d52c0fe7b23/src/parcsr_ls/ams.c#L697, which needs CUDA and so the headers. It is an example of not isolating our functions properly. What we can do is to make a function for that particular operation, put it in some `_device` file, and hide the actual implementation there, so `ams.c` does not see it and should not care about it at all.

src/parcsr_mv/par_csr_matvec.c
liruipeng: Yep, there should be a `par_csr_matvec_device.c` which needs the cuda header instead

src/sstruct_ls/fac_zero_stencilcoef.c
liruipeng: The files in `struct_mv` and `struct_ls` are more problematic. I've moved boxloop headers that contain the implementations out of `_hypre_struct_mv.h` into a new file called `boxloop.h`. So, whenever a file has `BoxLoop` call, `boxloop.h` has to be included.

src/test/ij.c
liruipeng: There are a few places in this files directly calling CUDA routines. I included this file to simply include the needed headers.

src/test/struct.c
liruipeng: There are a few places in this files directly calling `BoxLoop`. I included this file to simply include the BoxLoop implementations.

src/utilities/hypre_memory.c
liruipeng: This file is very involving; it covers our memory model with all kinds of scenarios. Previously, these headers were hidden in `_hypre_utilities.h` but they had to be moved out now.

src/IJ_mv/Makefile
rfalgout: If you always did this
```
CUOBJS = ${CUFILES:.c=.obj}
```
Then, if the Makefile.config rule is as follows, these CUFILES files would be compiled with the `CUCC` compiler
```
.c.obj:
	$(CUCC) $(CUFLAGS) -c $< -o $@
```
On the other had, with this rule, they would be compiled with the `CC` compiler
```
.c.obj:
	$(CC) $(CFLAGS) -c $<
```
Isn't the `CUCC` thing determined at configure time?  If so, we should be able to write an appropriate rule for `.c.obj:`.  I think this makes sense and does not require renaming files, but maybe I'm confused.  I'm just trying to avoid having the same `ifeq` construct repeated in a bunch of makefiles.

Is there a special meaning attached to the `.obj` extension?

rfalgout:
Hi @liruipeng .  I'd like to take a step back and understand the basic goal here.  I thought it was a C/C++ thing, and that's why I originally suggested having, for example,  `_hypre_utilities.h` and `_hypre_utilities.hpp` files (and similarly elsewhere), with the latter mainly containing all of the CUDA stuff since that's what requires C++.  I may have misunderstood this, but I'd still like to come up with some overall plan so that we don't go too crazy with the header files.  Also, this type of approach would clearly reflect the overall goal.  Right now, things like `boxloop.h` seem arbitrary until you understand that it's because they are the only parts of `Struct` that sometimes requires C++.

Let's maybe have a face-to-face discussion?  That would probably be a bit more efficient. :)  Thanks again!

liruipeng:
> Hi @liruipeng . Thanks for taking a stab at cleaning up the CUDA compile issue! I need to understand the overall strategy here before I can approve, so we'll need to iterate a bit first.

@rfalgout Thank you for looking into these carefully! Absolutely fine and no rush. I actually want to hear everyone's opinion to have a second thought. I just did it once to find some way that seems working.

rfalgout (approved):
Hi @liruipeng .  I think this looks better.  Thanks for making it happen!

bjorn2 (approved):
It looks ok to me, but I am not expert on header files and build systems. In general I think it is a good idea to try to make the Cuda code more separate from the CPU code, as you have done here. That should also make it easier to replace building with Cuda by something else, such as HIP in the future. Is that correct ? Is that something you have thought about when you did these changes ? 
 It looks as if there are some device-OpenMP files grouped together with the Cuda files, I am assuming that you have tested that building with device-OpenMP works.

victorapm (approved):
Hi Ruipeng,

I had only a few comments regarding minor details. It was not very clear to me why some files using OpenMP 4.5 capabilities are listed as CUFILES, e.g. some of the fac_* files. They don't need to be compiled with nvcc, right? a C compiler such as gcc-6 would be enough. I was just wondering, probably there's a reason that I'm missing for your choice :)

Thank you,
Victor

oseikuffuor1 (approved):

osborn9 (approved):

src/IJ_mv/IJMatrix_parcsr_device.c
```diff
@@ -741,5 +742,21 @@ hypre_IJMatrixAssembleParCSRDevice(hypre_IJMatrix *matrix)
    return hypre_error_flag;
 }
 
+void
```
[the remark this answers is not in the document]
liruipeng: Fixed. Thanks!

liruipeng:
> Hi Ruipeng,
> 
> I had only a few comments regarding minor details. It was not very clear to me why some files using OpenMP 4.5 capabilities are listed as CUFILES, e.g. some of the fac_* files. They don't need to be compiled with nvcc, right? a C compiler such as gcc-6 would be enough. I was just wondering, probably there's a reason that I'm missing for your choice :)
> 
> Thank you,
> Victor

Hi, Victor,

The `fac_` files potentially need nvcc since they have `hypre_BoxLoop` that requires CUDA when built with CUDA. For openmp 4.5, we have not relied on modern `gcc` to take care of it as normal C files. But, instead, we use `xlc-gpu` on LC, so we still distinguish the files with OMP4.5 as `CUFILES`. Let me know if this makes sense.

liruipeng:
> It looks ok to me, but I am not expert on header files and build systems. In general I think it is a good idea to try to make the Cuda code more separate from the CPU code, as you have done here. That should also make it easier to replace building with Cuda by something else, such as HIP in the future. Is that correct ? Is that something you have thought about when you did these changes ?
> It looks as if there are some device-OpenMP files grouped together with the Cuda files, I am assuming that you have tested that building with device-OpenMP works.

Maybe, it can be helpful when we hipify. But I don't know much about it yet. device-OpenMP files were treated as CUDA files, since they require special compilers, like `xlc-gpu` on LC. Thanks, @bjorn2

victorapm:
> Hi, Victor,
> 
> The `fac_` files potentially need nvcc since they have `hypre_BoxLoop` that requires CUDA when built with CUDA. For openmp 4.5, we have not relied on modern `gcc` to take care of it as normal C files. But, instead, we use `xlc-gpu` on LC, so we still distinguish the files with OMP4.5 as `CUFILES`. Let me know if this makes sense.

Oh, yes! I understand now. Thanks, Ruipeng!

ulrikeyang (approved):

PR #141, opened 2020-06-19
GPU build doc

liruipeng:
Hi, All,

I wrote something in `chi-mis.rst` of the user manual on the GPU build. Please take a look. For the documentation of specific functions calls that support GPU, I think they should go to the ref-manual? Since it is generated from the headers, so I just need to add documentations to the `HYPRE_*.h`?

Thanks

-Ruipeng

### DIFF
--- a/src/docs/usr-manual/ch-misc.rst
+++ b/src/docs/usr-manual/ch-misc.rst
@@ -163,7 +163,7 @@ include:
 GPU build
 ------------------------------------------------------------------------------
 
-Hypre can support GPUs with CUDA and OpenMP (:math:`{\ge}` 4.5). The related ``configure`` options are listed as follows
+Hypre can support GPUs with CUDA and OpenMP (:math:`{\ge}` 4.5). The related ``configure`` options are
 
 .. code-block:: none
 
@@ -173,17 +173,7 @@ Hypre can support GPUs with CUDA and OpenMP (:math:`{\ge}` 4.5). The related ``c
   --with-device-openmp    Use OpenMP 4.5 Device Directives. This may affect
                           which compiler is chosen.
 
-  --enable-nvtx           Use NVTX (default is NO).
-
-  --enable-cusparse       Use cuSPARSE (default is YES).
-
-  --enable-cub            Use CUB caching allocator (default is NO).
-
-  --enable-cublas         Use cuBLAS (default is NO).
-
-  --enable-curand         Use cuRAND (default is YES).
-
-Environment variables related to CUDA
+The related environment variables
 
 .. code-block:: none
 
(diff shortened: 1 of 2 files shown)


rfalgout (approved):
Hi @liruipeng .  Thanks for writing this!  I am approving, but I do have a general suggestion that I think would be good to address before merging this.  Most of what you wrote is a listing of the configure options as they appear with `configure --help`.  There are a couple of problems with doing this:

1. It will be easy for the documentation to get out of sync with `configure`.
2. Listing the options again doesn't really provide any new information.

I realize we already have some configure options listed in this section, but we did try to minimize it as much as possible by listing the most commonly used, while pointing to `configure --help` for the rest.  The extra words you wrote around the options are what will help users most, and I would recommend adding more of that and eliminating some of the option listing.  Guidance about which combinations of configure options to use is the best thing to put here.

Just my 2 cents.  Thanks again!

liruipeng:
@rfalgout Thank you for the suggestion. I made changes accordingly. Can you take another look? Thanks!

rfalgout (approved):
This looks good to me.  @ulrikeyang has done more GPU compiling than me, so it will be good to get her feedback as well.  Thanks!

rfalgout:
Hi @liruipeng and @ulrikeyang .  The PETSc guys just asked about GPU documentation.  Is this pull request ready?

rfalgout:
Hi @liruipeng .  The info on GPU-supported AMG options should definitely go in the reference manual.  You might also want to say something in the user manual and point to the reference manual for details.  Your call.

liruipeng:
> Hi @liruipeng . The info on GPU-supported AMG options should definitely go in the reference manual. You might also want to say something in the user manual and point to the reference manual for details. Your call.

@rfalgout  OK. Thanks! I added description of the current AMG options for GPUs. Please take a look.

rfalgout:
Looks fine to me.  We need to remember to keep this up to date as we add new options.  That's the only downside to having it in the user manual.  Thanks!

bjorn2:
Hi,
I wanted to have a look at the documentation, but I am not sure how to build it. The below did not work, what am I doing wrong ?
/Bjorn

I went to: 

src/docs/usr-manual

and run: 

make latexpdf

with result:

sphinx-build -b latex -d _build/doctrees   . _build/latex
Running Sphinx v1.8.4

Extension error:
Could not import extension breathe (exception: No module named 'breathe')
make: *** [latexpdf] Error 2

rfalgout:
Hi @bjorn2 .  The easiest way to look at this is simply to open the file in github.  It will render the markup (with only a few minor glitches).  To build the documentation on your own machine, I think there is a README that describes how to get things installed.  It's not the easiest thing to do, unfortunately, but it's not necessary in this case.

ulrikeyang (approved):

PR #145, opened 2020-07-01
AMG-DD implementation

waynemitchell:
This includes implementation of the AMG-DD algorithm, which is built on top of BoomerAMG. 

AMG-DD may be used as a standalone solver or a preconditioner for Krylov methods (note AMG-DD is a non-symmetric preconditioner) similar to BoomerAMG using the following functions for setup and solve:
HYPRE_BoomerAMGDDSetup(...)
HYPRE_BoomerAMGDDSolve(...)

The following parameters for AMG-DD may be accessed through the following set/get functions:

Padding (recommended default 1): HYPRE_BoomerAMGSetAMGDDPadding(...)
Number of ghost layers (recommended default 1): HYPRE_BoomerAMGSetAMGDDNumGhostLayers(...)
Number of inner FAC cycles per AMG-DD iteration (default 2): HYPRE_BoomerAMGSetAMGDDFACNumCycles(...)
FAC cycle type: HYPRE_BoomerAMGSetAMGDDFACCycleType(...)
    1 = V-cycle (default)
    2 = W-cycle
    3 = F-cycle
Number of relaxations on each level during FAC cycle: HYPRE_BoomerAMGSetAMGDDFACNumRelax(...)
Type of local relaxation during FAC cycle: HYPRE_BoomerAMGSetAMGDDFACRelaxType(...)
    0 = Jacobi
    1 = Gauss-Seidel
    2 = ordered Gauss-Seidel
    3 = C/F L1-scaled Jacobi (default)


victorapm:
Hi @waynemitchell,

First of all, thanks for your contribution! I think that implementing all of this must have been really difficult. So, thanks for your effort! I will make a list of points that I consider important to address before merging and also would like to offer help in addressing these points.

- [x] Implement access to BoomerAMGDD from the IJ driver, including its setup options. This would provide a quick way of testing BoomerAMGDD and comparing it to other solver/preconditioners in hypre.
- [x] I see many functions needed internally to build BoomerAMGDD without the `hypre_` prefix. This can be a problem when hypre is linked to external applications that might implement functions with the same name, e.g., `ZeroInitialGuess`, `AddSolution`...
- [x] Move prototypes from the `par_amgdd` files to the `_hypre_parcsr_ls.h` header file, which must be generated with the `headers` script in the `parcsr_ls` folder.
- [x] The prototypes of `HYPRE_BoomerAMGDDSetup` and `HYPRE_BoomerAMGDDSolve` need to be added to the `parcsr_ls/HYPRE_parcsr_ls.h` header file. The same for other user APIs related to BoomerAMGDD such as `HYPRE_BoomerAMGSetAMGDDFACTol`, `HYPRE_BoomerAMGSetAMGDDPadding` ...
- [x] Do we really need to use `std::map` and  `std::set` in `par_amg_dd_helpers`? I think the same can be achieved without relying on C++ for this. The way it is now, the hypre library **cannot** be linked to a driver through a C compiler, even if the driver is `.c`, e.g., drivers in the `src/test` folders.
- [x] `par_amgdd_helpers.cxx` is included in `par_amgdd_setup.c`, but we normally don't include source files. The functionalities provided by `par_amgdd_helpers.cxx` should be provided through `_hypre_parcsr_ls.h`, i.e., its prototypes should be implemented in the later header file.
- [x] Should we implement `[HYPRE,hypre]_BoomerAMGDDSolveT`? It might be interesting if this could be easily done. Following the logic of BoomerAMG, for completeness.
- [x] Should the [HYPRE,hypre]_BoomerAMG[Set,Get]AMGDDFAC be renamed to [HYPRE,hypre]_BoomerAMGDD[Set,Get]FAC?
- [x] Use `hypre_error_w_msg` instead of `printf` for warning messages, e.g., https://github.com/hypre-space/hypre/blob/33f42321303bcbf9bf8445c6c04199a2991ee6cd/src/parcsr_ls/par_amg_setup.c#L323
- [x] In `parcsr_ls/par_amgdd_fac_cycle_device.c`, use hypreCUDAKernel_IVAXPY https://github.com/hypre-space/hypre/blob/33f42321303bcbf9bf8445c6c04199a2991ee6cd/src/utilities/hypre_cuda_utils.c#L591 instead of the VecScale function.
- [x] Use `HYPRE_CUDA_CALL`, `HYPRE_CUSPARSE_CALL` and `HYPRE_CUDA_LAUNCH` macros for CUDA calls, see e.g., https://github.com/hypre-space/hypre/blob/master/src/utilities/hypre_cuda_utils.c. Then, the definition of `cusparseHandle_t` and `cusparseMatDescr_t` objects in `par_amgdd_fac_cycle_device.c` won't be needed.
- [x] I think it makes sense moving the implementation of VecScaleMaskedKernel to `hypre_cuda_utils.c`. Since it can be potentially used by other parts of the code.
- [x] Following the logic of the previous item, implement hypre_CSRMatrixMaskedMult, which will call `cusparseDbsrxmv`  when the execution policy is set to device.
- [x] Use `hypre_Memset` instead of `memset`.
- [ ] Fix compiler warnings that appear when configuring hypre in debug mode (`--enable-debug`).
- [x] Remove trailing whitespaces and tabs in the `par_amgdd` files and break long lines. We use three whitespaces for each level of indentation.
- [ ] Refactor code style, e.g.,
```
if (...)
{
   ...
}
else
{
   ...
}
```
instead of
```
if (...) {...}
else {...}
```
Also:
```
for (...)
{
   ...
}
```
instead of
```
for (...) ...
```

I wanted to bring a point for discussion. Maybe @rfalgout and @ulrikeyang could give their opinions:

it looks like BoomerAMGDD should be a separate strategy (solver/preconditioner) from BoomerAMG even though it's built on top of it. At least, the way it's implemented, BoomerAMGDD has Setup and Solve methods, so I would expect it to have Create and Destroy and its own Set\*Options\* such as SetPadding. So we would need to implement `[hypre,HYPRE]_BoomerAMGDDCreate` `[hypre,HYPRE]_BoomerAMGDDDestroy` functions (following the logic of other preconditioners in hypre). Suggestion: add a new file called `par_amgdd.c` for `hypre_BoomerAMGDDCreate` and `hypre_BoomerAMGDDDestroy`; implement their uppercase counterparts in `HYPRE_parcsr_amgdd.c`.

waynemitchell:
Hi, @victorapm. Thanks for your thorough review! I am working through the changes now. I have a few comments and questions:
1. Implementation of a SolveT routine is actually non-trivial in general. If the connectivity of the row stencils for the matrices are different from the column stencils, then I think you would want to rebuild the AMG-DD composite grids based on the column stencils. If the stencils are at least symmetric (even if the values are not), then things get much easier. Either way, this is probably outside the scope of this pull requets.
2. Do y’all have a strong preference for excluding C++ code from hypre? I can rewrite the parts of my code that use C++ (and I’m willing to do so if that’s what is preferred), but I am curious whether there is an acceptable way of including some C++ functionality from your point of view.
3. If we want to make AMG-DD its own solver/preconditioner distinct from AMG with it’s own Create/Destroy functions, etc., I would propose something like the following. I can make a hypre_ParAMGDDData struct and move the AMG-DD related members/parameters here from the hypre_ParAMGData struct. Then hypre_ParAMGDDData can have a hypre_ParAMGData struct as a member. The user can then access the hypre_ParAMGData member in order to set parameters for the underlying AMG. So a typical usage would look something like:
```
// Create the AMG-DD solver
HYPRE_Sovler amgdd_solver;
HYPRE_BoomerAMGDDCreate(&amgdd_solver);

// Set AMG-DD parameters
HYPRE_BoomerAMGDDSet[…](amgdd_solver, …);
…

// Get the underlying AMG struct and set parameters
HYPRE_Sovler amg_solver;
HYPRE_BoomerAMGDDGetAMG(&amg_solver);
HYPRE_BoomerAMGSet[…](amg_solver, …);
…

// Setup, solve, destroy as usual
HYPRE_BoomerAMGDDSetup(…);
HYPRE_BoomerAMGDDSolve(…);
HYPRE_BoomerAMGDDDestroy(…);
```

victorapm:
Hi @waynemitchell,

Great! Commenting your points:

1. Perfect, now I understand this wouldn't be trivial, so we can skip the implementation of `BoomerAMGDDSolveT` for this PR.
2. We normally try to avoid using C++ features if the same work can be done via C. In `par_amgdd_helpers.cxx`, it looks like you are able to replace (my suggestion):
    1. `std::vector` by a C array. You allocate it with an estimate of the total size if not known beforehand via TAlloc and later call TReAlloc if needed.
    2. `std::set` by `hypre_UnorderedIntSet`.
    3. `std::map` by `hypre_UnorderedIntMap`. However, not everywhere, e.g., I see you have a `map< HYPRE_Int, map<HYPRE_Int, vector<pair<HYPRE_Int, HYPRE_Int>>>>` type. Do you think it's easy to refactor this?
- If it is not a big amount of work this refactoring, I think we should go for it. Otherwise, if we decide to keep using C++, we need to make sure users are able to link hypre to their C applications without needing to use a C++ linker.

3. Yes! That's exactly what I was thinking. This way we have a better separation between BoomerAMG and BoomerAMGDD helping code maintenance and improving usability as well. 

Lastly, I added one more point to the list I published previously (the one about memset). Let's make that the default checklist in case I come across additional things to address :)

Thanks!

victorapm:
Hi @waynemitchell,

1. Correct, this turns on HYPRE_CONCURRENT_HOPSCOTCH, which enables the implementation of certain functions related to Set and Map, e.g. `hypre_UnorderedIntSetPut`, `hypre_UnorderedIntSetContains` and others. However, if you prefer to avoid OpenMP, I can look at extending the implementation of these functions to work also when HYPRE_CONCURRENT_HOPSCOTCH is undefined (some of them already do).
2. inCapacity needs to be chosen so that `NearestPowerOfTwo(inCapacity+4096) + 4097 > size`, where `size` is the size of your set/map. Note that a resize operation is not implemented in these data structures, thus the code will break if trying to insert more entries than `NearestPowerOfTwo(inCapacity+4096) + 4097`. `concurrencyLevel` can be set to zero if not using OpenMP, otherwise, a good number seems to be `16*hypre_NumThreads()` (that's what is used in other parts of hypre, but I'm not expert in this, @rfalgout could confirm this info).
3. Right! Sounds a good idea! The same for using `hypre_UnorderedIntSetCopyToArray` :)

waynemitchell:
OK, I think this is about ready for another look. I've addressed most of the requests above. I have the following remaining questions on my side:
1. Since AMG-DD now requires hopscotch and openmp, what is your recommendation for the best way to macro out code and warn the user when insufficient config options are used?
3. I get compiler errors (undefined references to openmp related functions) when trying to compile the ij driver through "make test" with openmp and hopscotch turned on. Should there be a fix to a makefile somewhere? Compiling via cmake works fine. 
4. Regarding your request above to eliminate compiler warnings when hypre is configured with --enable-debug, I don't see any such compiler warnings on my machine. Any tips on reproducing this?

src/parcsr_ls/par_amgdd_fac_cycle.c
```diff
@@ -0,0 +1,564 @@
...
+ * Copyright 1998-2019 Lawrence Livermore National Security, LLC and other
+ * HYPRE Project Developers. See the top-level COPYRIGHT file for details.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR MIT)
+ ******************************************************************************/
+
+#include "_hypre_parcsr_ls.h"
+
+HYPRE_Int
+hypre_BoomerAMGDD_FAC_Cycle(void *amgdd_vdata, HYPRE_Int level, HYPRE_Int cycle_type, HYPRE_Int first_iteration);
```
victorapm: Move this and the following function prototypes to `protos.h` and re-generate `_hypre_parcsr_ls.h`

src/seq_mv/csr_matrix.c
```diff
@@ -444,6 +490,10 @@ hypre_CSRMatrixPrint( hypre_CSRMatrix *matrix,
    return ierr;
 }
 
+
+/*--------------------------------------------------------------------------
+ * hypre_CSRMatrixPrintCOO: print a CSRMatrix in coordinate format
```
victorapm: I think there was a typo with PrintCOO (it does not match the function below)

src/seq_mv/csr_matrix.c
```diff
@@ -145,6 +145,52 @@ hypre_CSRMatrixInitialize( hypre_CSRMatrix *matrix )
...
+HYPRE_Int 
+hypre_CSRMatrixResize( hypre_CSRMatrix *matrix, HYPRE_Int new_num_rows, HYPRE_Int new_num_cols, HYPRE_Int new_num_nonzeros )
+{
+   HYPRE_MemoryLocation memory_location = hypre_CSRMatrixMemoryLocation(matrix);
+   HYPRE_Int old_num_nonzeros = hypre_CSRMatrixNumNonzeros(matrix);
+   HYPRE_Int old_num_rows = hypre_CSRMatrixNumRows(matrix);
+  
+   if (!hypre_CSRMatrixOwnsData(matrix))
+   {
+      hypre_printf("Error: called hypre_CSRMatrixResize on a matrix that doesn't own the data\n");
```
victorapm: Use hypre_error_w_msg instead.

src/seq_mv/csr_matvec_device.c
```diff
@@ -110,5 +110,118 @@ hypre_CSRMatrixMatvecDeviceBIGINT( HYPRE_Complex    alpha,
...
+
+   if (x == y)
+   {
+      hypre_error_w_msg(HYPRE_ERROR_GENERIC,"ERROR::x and y are the same pointer in hypre_CSRMatrixMatvecMaskedDevice\n");
+   }
+
+   // TODO
+   if (offset != 0)
+   {
+      hypre_printf("WARNING:: Offset is not zero in hypre_CSRMatrixMatvecMaskedDevice :: \n");
```
victorapm: Use hypre_error_w_msg.

src/utilities/hypre_cuda_utils.c
```diff
@@ -617,6 +617,35 @@ hypreDevice_IVAXPY(HYPRE_Int n, HYPRE_Complex *a, HYPRE_Complex *x, HYPRE_Comple
...
+   /* trivial case */
+   if (n <= 0)
+   {
+      return hypre_error_flag;
+   }
+
+   dim3 bDim = hypre_GetDefaultCUDABlockDimension();
+   dim3 gDim = hypre_GetDefaultCUDAGridDimension(n, "thread", bDim);
+
+   HYPRE_CUDA_LAUNCH( hypreCUDAKernel_IVAXPY, gDim, bDim, n, a, x, y );
```
victorapm: Shouldn't it be `hypreCUDAKernel_MaskedIVAXPY`?

src/parcsr_ls/par_amgdd_helpers.c
```diff
@@ -0,0 +1,2866 @@
...
+         }
+      }
+
+      hypre_AMGDDCompGridNumNonOwnedNodes(compGrid[level]) += add_node_cnt;
+   }
+
+   return 0;
+}
+
+HYPRE_Int PackColInd(HYPRE_Int *send_flag,
```
victorapm: Add `hypre_` prefix

victorapm (changes requested):
Hi @waynemitchell,

thanks for all your work! This looks great to me. I'm commenting on a few minor issues that need to be addressed before merging. Regarding your previous post:

1. I fixed the dependency on OpenMP within hopscotch in a separate branch and once it's merged to master, you will be able to use amgdd without having to compile hypre with "--enable-hopscotch --with-openmp". So, the include guards on `par_amgdd_helpers` won't be needed.
2. I haven't been able to reproduce this, but it would be interesting to see the errors if you can share.
3. I got a few warnings with gcc-9, I copied them to this link: https://gist.github.com/victorapm/c1a59b2cfea44ddbf0284b0ae6e37b0a

Lastly, I tried to run `mpirun -np 2 ./ij -n 10 10 10 -solver 91` but got no output. If I use `-solver 90` instead, it works. Can you look at this? It would be nice also if we could select a few small runs like this one to test the capabilities of AMG-DD, then we could add them to the regression tests.

waynemitchell:
Hi, @victorapm. OK, I think I've cleaned up the code according to your remaining comments and fixed the -solver 91 option in the ij driver. Thanks for fixing up the hopscotch configuration options! I also added just a couple of lines to the solvers.jobs, solvers.sh, and solvers.saved in the test/TEST_ij directory. Let me know whether this is what you had in mind in terms of adding to the regression tests.

On another note, I was trying to test out the GPU version of the code, and I was having trouble with hypre_CudaCompileFlagCheck() called from HYPRE_Init() failing. It's not clear to me, however, how I should specify __CUDA_ARCH__ during configuration?

victorapm (approved):
Thanks, @waynemitchell! I'm approving the PR

There's only a small detail left from my side, can you fix these warnings?
```
par_amgdd_fac_cycle.c: In function ‘hypre_BoomerAMGDD_FAC_CFL1Jacobi’:
par_amgdd_fac_cycle.c:410:24: warning: unused variable ‘amgdd_data’ [-Wunused-variable]
  410 |    hypre_ParAMGDDData *amgdd_data = (hypre_ParAMGDDData*) amgdd_vdata;
      |                        ^~~~~~~~~~
mpicc -g -Wall  -DHAVE_CONFIG_H -I.. -I../distributed_ls/Euclid -I. -I./.. -I./../blas -I./../lapack -I./../multivector -I./../utilities -I./../krylov -I./../seq_mv -I./../parcsr_mv -I./../distributed_matrix -I./../matrix_matrix -I./../IJ_mv -I./../parcsr_block_mv    -c par_amgdd_helpers.c
par_amgdd_helpers.c: In function ‘hypre_BoomerAMGDD_AddToSendAndRequestDofs’:
par_amgdd_helpers.c:136:20: warning: unused variable ‘neighbor_global_index’ [-Wunused-variable]
  136 |          HYPRE_Int neighbor_global_index = hypre_ParCSRMatrixColMapOffd(A)[i];
      |                    ^~~~~~~~~~~~~~~~~~~~~
par_amgdd_helpers.c: In function ‘hypre_BoomerAMGDD_SetupNearestProcessorNeighbors’:
par_amgdd_helpers.c:595:28: warning: unused variable ‘num_nodes’ [-Wunused-variable]
  595 |    HYPRE_Int               num_nodes = hypre_ParCSRMatrixNumRows(A);
      |                            ^~~~~~~~~
```
Regarding your question about cuda architecture,  you can choose another value than the default by passing `HYPRE_CUDA_SM` via the configure script, i.e., for V100 you would do `HYPRE_CUDA_SM=70`.

Thanks for the work!

rfalgout (approved):
This looks good to me .  I would merge in the master one more time just to make sure everything is up to date on this branch.  Thanks @waynemitchell for the great work!  Thanks @victorapm for looking over this in detail and helping with the hopscotch stuff!  It will be great to have this solver option available in hypre!

waynemitchell:
OK, final compiler warnings should be fixed and I did another pull of master. Thanks again, @victorapm!

ulrikeyang (approved):
Overall this looks fine to me.  I trust that AMGDD works fine, but obviously can not check all that, but I see no issue for it causing any trouble for hypre in general.

liruipeng (approved):
This looks good to me. I looked at the GPU related stuff in particular. Thanks! @waynemitchell

ulrikeyang:
Since every requested reviewer approved, should we merge this now?

rfalgout:
Yes, but I wonder if we should use the suggested method in #188.  If we could wait just a tad longer to see if anybody has serious issues with this idea, then it would lead to a cleaner merge.

ulrikeyang:
Depending on what 'tad' means. Should we give them a deadline, or can we just use this PR as a first try in hypre to apply this method, sort of a proof how great this is :-) ?

rfalgout:
Hi @victorapm .  Since you took the lead on this pull request, would you mind doing the squash-merge?

Just a couple of notes about the commit message.  I like the idea of keeping the pull request number in the title, and I wouldn't include too much information on how to use the solver since that is documented in the header files and online API reference manual.  Definitely include a basic standalone description of what this solver is all about, though.  Thanks!

And thanks again to you, @waynemitchell !

victorapm:
It's done, @rfalgout. Thanks everyone :)

liruipeng:
Hi, @waynemitchell, in `hypre_CSRMatrixMatvecMaskedDevice`, a matvec is first done followed by a masked matvec, see https://github.com/hypre-space/hypre/blob/641f7a4e31037bb0e2d841a5d046d983548a7594/src/seq_mv/csr_matvec_device.c#L205

Is this what you actually want to do? Just asking. I am refactoring this part of the code.

Thanks!

-Ruipeng

waynemitchell:
Hi @liruipeng. I had actually noticed this and fixed it in my own code, but hadn't gotten around to pushing an update. You are correct: the first regular matvec is extraneous (a copy-paste error, I guess). Thanks for the catch!

PR #149, opened 2020-07-09
Compile hypre_nvtx.c with nvcc

pbauman:
Small fix for the case `--enable-nvtx`. It appears #138 missed moving hypre_nvtx.c to `CUFILES`. With this change, I'm now able to compile and run `test/ij` with HYPRE configure using `--with-cuda=yes --enable-nvtx`. Thank you.


liruipeng (approved):
You are absolutely right! Thank you, @pbauman

PR #155, opened 2020-07-20
Fix caliper

victorapm:
This PR fixes the usage of caliper in hypre according to its new API. It also adds caliper annotation to multigrid levels (supported by PFMG and BoomerAMG) and extends annotation to other solvers such as MGR and ILU.


rfalgout (approved):
Looks great, @victorapm !  Thanks!

ulrikeyang (approved):
Looks good.

Issue #70, opened 2019-12-06
Convergence tolerance defaults


rfalgout:
Hi @hypre-space/developers,

It looks like we have some inconsistencies with our convergence tolerance defaults.  In most cases, the default is 1.0e-6, but there are a couple that are 1.0e-7 (par_amg.c, par_mgr.c).  I think we should make them all 1.0e-6 and ensure that the header file comments also match.  Any concerns about making these changes?

Thanks!

-Rob

liruipeng:
@rfalgout Agreed on making all the same, like 1e-6

ulrikeyang:
I have no concerns
Ulrike

oseikuffuor1:
That's fine with me too. I have not checked this yet, but I'm wondering if this may cause some of the regression tests to fail? That is, if the test drivers are not explicitly setting the convergence tolerance (default). I think ij.c does, so it should not be an issue.

-Daniel

rfalgout:
Resolved with #206

PR #156, opened 2020-07-20
Fixed convergence on issue #70

SourangshuGhosh:
Respected @rfalgout ,
On your issue on Convergence tolerance defaults #70 you have asked to change the tolerance value from 1e-7 to 1e-6 on two parameters file par_amg.c and par_mgr.c. This PR fixes these tolerances on the above mentioned two files


rfalgout:
This is resolved in #206.  Thanks!

PR #157, opened 2020-07-23
A few GPU-specific fixes/updates

pbauman:
This PR includes some fixes and updates to help improve portability.

950a747 and 2f51e75 are updates to use `HYPRE_WARP_SIZE` instead of hardcoded 32, 31, or 16.  26a1a8a in an update for readability.

2e5ccb6 is an important fix to a sneaky issue. It turns out, `clang` and `nvcc` assign different precedence between `>>` and `-`. Indeed, the code before that commit elicited a warning from clang about the precedence. The code as it was would behave differently on clang than on nvcc. So those extra parentheses are critical. This commit merits close examination to make sure I understood the original intent of the code correctly.

26a1a8a is another fix for clang. This function would not compile without `result_type` being defined. (Alternatively, the functor could subclass `unary_functor`.)

I've been using the `ij` test with the following to exercise GPU support without managed memory:

`./ij -n 48 48 48 -pmis -keepT 1 -rlx 18 -rap 1 -mod_rap2 1 -mm_cusparse 1 -solver 1 -agg_nl 0 -exec_device`

and with managed memory:

`./ij -n 48 48 48 -pmis -keepT 1 -rlx 18 -rap 1 -mod_rap2 1 -mm_cusparse 1 -solver 1 -agg_nl 1 -exec_device`

Both of these tests are passing (in the sense that the solver converges) using CUDA 10.2 on V100.

These changes will be important for HIP support as the HIP compiler is based on clang. Thank you.

### DIFF
--- a/src/utilities/_hypre_utilities.hpp
+++ b/src/utilities/_hypre_utilities.hpp
@@ -422,7 +422,7 @@ T warp_prefix_sum(hypre_int lane_id, T in, T &all_sum)
    for (hypre_int d = 2; d <=HYPRE_WARP_SIZE; d <<= 1)
    {
       T t = __shfl_up_sync(HYPRE_WARP_FULL_MASK, in, d >> 1);
-      if ( (lane_id & (d - 1)) == d - 1 )
+      if ( (lane_id & (d - 1)) == (d - 1) )
       {
          in += t;
       }
@@ -440,7 +440,7 @@ T warp_prefix_sum(hypre_int lane_id, T in, T &all_sum)
    {
       T t = __shfl_xor_sync(HYPRE_WARP_FULL_MASK, in, d);
 
-      if ( (lane_id & (d - 1)) == d - 1)
+      if ( (lane_id & (d - 1)) == (d - 1))
       {
         if ( (lane_id & ((d << 1) - 1)) == ((d << 1) - 1) )
          {
--- a/src/utilities/_hypre_utilities.hpp
+++ b/src/utilities/_hypre_utilities.hpp
@@ -373,28 +373,28 @@ hypre_double atomicAdd(hypre_double* address, hypre_double val)
 
 template <typename T>
 static __device__ __forceinline__
-T __shfl_sync(unsigned mask, T val, hypre_int src_line, hypre_int width=32)
+T __shfl_sync(unsigned mask, T val, hypre_int src_line, hypre_int width=HYPRE_WARP_SIZE)
 {
    return __shfl(val, src_line, width);
 }
 
 template <typename T>
 static __device__ __forceinline__
-T __shfl_down_sync(unsigned mask, T val, unsigned delta, hypre_int width=32)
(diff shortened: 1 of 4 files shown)


liruipeng:
@pbauman 

Thank you for the changes. These all look good to me. In particular, I am glad you found the bug in `(d << 1 - 1)`. I have been careful about precedence but apparently I missed this one!
 
CUDA provides constant `warpSize`. Can it be an alternative to the macro `HYPRE_WARP_SIZE`? But it seems only existing in device code, so I am not sure. I am not asking to change but just want to hear what do you think. Thanks!

pbauman:
Hi @liruipeng 

> I have been careful about precedence but apparently I missed this one!

Yeah, this is the frustrating thing about there being no standard for kernel code - compilers are allowed to do different things.

> CUDA provides constant warpSize. Can it be an alternative to the macro HYPRE_WARP_SIZE?

In some cases, yes (like the loop bounds etc. in 950a747), but will require adding `__CUDA_ARCH__` guards since it's only defined in the device pass. In other cases, no: I actually tried that instead of `HYPRE_WARP_SIZE` in 2f51e75 (since that function was already guarded by `__CUDA_ARCH__`), but nvcc gave me a compiler error that `warpSize` cannot be used as a constant there.

In my opinion, I'm 6-in-one half-a-dozen-in-the-other about it. The former requires more `#if defined` guards, but can't be used everywhere so we can't entirely get rid of `HYPER_WARP_SIZE`. The latter leaves an otherwise redundant undesirable macro.

I'm happy to rework the PR as you prefer, just let me know. Thanks!

liruipeng:
> Hi @liruipeng
> 
> > I have been careful about precedence but apparently I missed this one!
> 
> Yeah, this is the frustrating thing about there being no standard for kernel code - compilers are allowed to do different things.
> 
> > CUDA provides constant warpSize. Can it be an alternative to the macro HYPRE_WARP_SIZE?
> 
> In some cases, yes (like the loop bounds etc. in [950a747](https://github.com/hypre-space/hypre/commit/950a747ae6bcee21fa238f4cfc6b034f2ea83f01)), but will require adding `__CUDA_ARCH__` guards since it's only defined in the device pass. In other cases, no: I actually tried that instead of `HYPRE_WARP_SIZE` in [2f51e75](https://github.com/hypre-space/hypre/commit/2f51e75013a0a14da470384af80f2d5aeeb47ac8) (since that function was already guarded by `__CUDA_ARCH__`), but nvcc gave me a compiler error that `warpSize` cannot be used as a constant there.
> 
> In my opinion, I'm 6-in-one half-a-dozen-in-the-other about it. The former requires more `#if defined` guards, but the latter can't be used everywhere so we can't entirely get rid of `HYPER_WARP_SIZE`.
> 
> I'm happy to rework the PR as you prefer, just let me know. Thanks!

Thank you for your opinion on this. It doesn't seem to be any cleaner to use `warpSize`. The PR is good enough. Thanks!

PR #163, opened 2020-07-29
Cuda 11 API

AerysL:
# ~~Draft, Do Not Merge, WIP~~
# Description

The Cuda 11 cuSPARSE API removes a number of sparse matrix functions which were previously present, replacing them with alternatives (called `old` and `new` respectively). As the functions are entirely removed, the new API must be used for
Cuda \(>=\) 11, with some of the new API being available as early as 10.1. This pull request introduces support for the new API, altering the sparse matrix-vector and sparse matrix-matrix calls.

# Major Changes

The two primary changes are updating the sparse-matrix vector product, and (when using cusparse) the sparse-matrix sparse-matrix product.

The new sparse-matrix vector product API is available in Cuda 10.1, which is enabled when the compiler detects such a cuda version, and disabled otherwise.

Notes:

  - The new Cusparse API includes additional error checking compared to typical BLAS interfaces. In particular, during Matrix-Vector products it ensures the size of the vectors x,y are appropriate given the rows/columns of the matrix. This
    is an issue as in some locations, the HYPRE\_vector.size is not updated after doing a VectorCopy where fewer than the original HYPRE\_vector.size are copied. As a work around I have added code which, if there is a size mismatch, will
    print a warning if HYPRE\_DEBUG is defined, a flag (‘HYPRE\_MV\_CHECK\_VECTOR\_SIZES‘) and override the size used.

If Cuda 11 is detected, the new sparse-matrix sparse-matrix API is used. Tests indicate that the sorting of the column indices in the CSR representations is no longer required. The new API appears to include performance enhancements, as
seen below.

Additionally this PR includes fixes from Ruipeng for the ij\_mm driver, as well as a fix for the OpenMP Device

# Minor Changes

  - A number of helper functions were added in src/seq\_mv/csr\_matrix\_cuda\_utils.{c,h}, including
    
      - A function to determine the cuda datatype corresponding with the currently enabled HYPRE\_Complex typedef
    
      - A function to determine the cuda datatype corresponding with the currently enabled HYPRE\_Int typedef (indices)
    
      - A function to sort the column indices for CSR matrices while using the old sparse-matrix product API
    
      - Functions to convert different matrix and vector hypre representations to those used by cuda
    
      - A near drop in replacement for Dcsrmv. Warning: this has less error detection than the full SpMV call

  - When transpose is required (oomp matvec, device matvec), if Cuda \(>=\) 10.1 is detected, use a newer csr2csc cuda call

  - Use a different CSR sorting when required for SpMV (Cuda \(<=\) 10.1)

  - A flag was added to ray and lassen regression tests to silence the deprecation errors with C++11

  - The ray and lassen regression tests ‘rocuda‘ now use a tolerance

# Argument for correctness

The numerical algorithms are provided from the Cuda implementation, The new code has been tested through a number of numerical experiments, all of which have achieved expected accuracy. The new code has been tested through the regression
tests, with cuda 10 and 11 for both Ray and Lassen. NOTE: The Cuda 11 code is hardcoded to error during compilation when used with the OpenMP device (see Remaining Issues)

# Performance description

The plots below correspond with the experiments seen in TESTS\_cuda\_ray, with some parameter changed. Each experiment was repeated 3 times (each shown as an ’x’), with the corresponding line being the mean and the shaded region the mean
\(\pm\) one standard deviation.

<a id="fig:api_plots">[fig:api_plots]</a>

ij_vector_setup:RHS and Initial Guess
init_times:Hypre init
mat_gen:Spatial Operator
setup_phase_times:PCG Setup
solve_phase_times:PCG Solve
total_time

We see in Figure [\[fig:api\_plots\]](#fig:api_plots) a comparison of the old and new implementations, using both the builtin hypre sparse-matrix sparse-matrix and the cusparse sparse-matrix sparse-matrix products. For example,

<a id="fig:trans_plots">[fig:trans_plots]</a>
ij_vector_setup:RHS and Initial Guess
init_times:Hypre init
mat_gen:Spatial Operator
setup_phase_times:PCG Setup
solve_phase_times:PCG Solve
total_time

Figure [\[fig:trans\_plots\]](#fig:trans_plots) uses the new implementation (this pull request, Cuda 11) and compares what happens if the transpose is explicitly calculated and kept.
# Remaining Issues

  - Due to the newness of Cuda 11, it is not fully supported on all platforms, which has resulted in a number of issues cropping up. It is likely these issues will be fixed by external parties
    
      - The CUB and Thrust libraries included with Cuda 11 issue deprecation warnings when used with C++ \< C++14. The IBM Compiler Version 16 does not currently support C++14, and so warnings displayed when compiled with the compiler. This
        is independent of this pull request. There exists a flag to silence these deprecation warnings, which has been enabled for the regression tests. Once the IBM Compiler by default compiles with C++14, or gets flags for C++14, the
        silencing flags can be removed.
    
      - For the OpenMP Device compilers have to support the Cuda version used. The default compiler does not include such support for Cuda 11 yet, but the ‘xl/2020.06.25-cuda-11.0.2‘ module has been added. In order to successfully link the
        executables, such a compiler must be used.

  - The Cuda 11 OpenMP Device code does not currently fully work (I have been unable to determine if this is an error on our end, NVIDIAs, or the toolstack) , so I have added a compile time error when the OpenMP Device code is compiled with
    Cuda 11

  - Hypre fails to compile with cuda 8. This is independent of this pull request.

# Recommendation

After discussion with @liruipeng a reasonable course of action seems to be to merge the pull request, to allow people who need Cuda 11 support to use it if required. Once Cuda 11 has higher adoption than Cuda 10, the regression tests can be
switched over.


liruipeng (approved):
Looks all good to me now.

liruipeng:
Hi, @AerysL I have made changes to this PR and it's ready to be merged to the master. If you have time, can you please take another look? Thanks a lot! -Ruipeng

rfalgout (approved):
Looks good to me.  I made one small change related to the reference manual documentation that I accidentally introduced in a previous pull request.

ulrikeyang (approved):
I will approve this, but I am a bit worried about the vector size being overwritten when it doesn't match. This could possibly break something in the additive code.  It's however a code that is not really used by anybody, but we should look into this and come up with a solution.

liruipeng:
> I will approve this, but I am a bit worried about the vector size being overwritten when it doesn't match. This could possibly break something in the additive code. It's however a code that is not really used by anybody, but we should look into this and come up with a solution.

@ulrikeyang I am not sure where the vector size gets overwritten. If you referred to this line https://github.com/hypre-space/hypre/blob/94b77225972496a3c50899e7c0cb2de969a0f210/src/seq_mv/csr_matrix_cuda_utils.c#L97, there is no problem, since only the temporary CUDA vector structure is changed while `hypre_Vector` is not. Maybe I missed something. Thanks! -Ruipeng

oseikuffuor1 (approved):
Looks good to me

AerysL (commented):
Hi, sorry for the late review. Took a quick look over the changes reported by Github on this review as well as offline diff of the latest merge of `master` into `cuda_11_api`. Two things I will note is that some of the new API was added in 10.x, and some in 11. I think as you did just keeping `CUSPARSE_NEWAPI_VERSION` as 11000 makes the most sense (I think the version added in 10.x was only on certain platforms and had various caveats). 

Secondly, I recall that depending on the combination of the cuda version and device (cuda, openmp, cpu) used could lead to compilation issues (and resulted in various #if statements), which seem to have been changed. If the codebase works in the different combinations than I think it should be fine, just something to be aware of.

Aside from those two things to be aware of (neither of which I think should require any changes) looks good to me!

liruipeng:
> Hi, sorry for the late review. Took a quick look over the changes reported by Github on this review as well as offline diff of the latest merge of `master` into `cuda_11_api`. Two things I will note is that some of the new API was added in 10.x, and some in 11. I think as you did just keeping `CUSPARSE_NEWAPI_VERSION` as 11000 makes the most sense (I think the version added in 10.x was only on certain platforms and had various caveats).
> 
> Secondly, I recall that depending on the combination of the cuda version and device (cuda, openmp, cpu) used could lead to compilation issues (and resulted in various #if statements), which seem to have been changed. If the codebase works in the different combinations than I think it should be fine, just something to be aware of.
> 
> Aside from those two things to be aware of (neither of which I think should require any changes) looks good to me!

@AerysL Thanks again for your work included in this PR and the reviews. -Ruipeng

PR #165, opened 2020-07-30
fixed some mixed-int issues.

ulrikeyang:
Hi all
I fixed a few bugs.


rfalgout (approved):
Looks good @ulrikeyang !  Thanks!

liruipeng (approved):
Looks good to me. Just happened to find a similar issue in the GPU code. There might be more in the GPU code. Will need to look through and test GPU build with mixed-int for large problems. I will be in charge of it. Thanks, Ulrike!

PR #176, opened 2020-08-10
Add a direct solver using SuperLU that has the same interface as other hypre solvers.

mquanbui:
BoomerAMG uses SuperLU as an option for the coarse-grid solve. This PR adds a similar option for MGR.
Since the coarse solver can be constructed outside of MGR, the new direct solver needs to have the same interface as other hypre solvers in order to use the same signatures for the function pointers to setup and solve routines.
The new direct solver is implemented separately inside MGR, and does not interfere with SuperLU in BoomerAMG.

For `hypre_MGRDirectSolverSolve`, I wrote a simple wrapper around `hypre_SLUDistSolve`. Since `hypre_SLUDistSetup` both creates and setups SuperLU, I split it into `hypre_MGRDirectSolverCreate` and `hypre_MGRDirectSolverSetup`. The `hypre_MGRDirectSolverSetup` routine duplicates almost all of the code in `hypre_SLUDistSetup`, since I couldn't find a clean way to avoid this using the new interface. Maybe these can be combined in the future.

I also remove a lot of unused code in test_mgr.c file that I usually use for testing. The diff is long so you can ignore it since it's only used by me.


src/parcsr_ls/par_mgr.h
```diff
@@ -7,8 +7,7 @@
 
 #ifndef hypre_ParMGR_DATA_HEADER
 #define hypre_ParMGR_DATA_HEADER
-/*--------------------------------------------------------------------------
- * hypre_ParMGRData
+/*-------------------------------------------------------------------------- * hypre_ParMGRData
```
rfalgout: Looks like an accidental deletion of an EOL

rfalgout (approved):
Looks great to me!  Thanks!

src/parcsr_ls/par_mgr.h
mquanbui: You're right. I've fixed this. Thank you!

oseikuffuor1 (approved):
Looks good. Passed regression tests. Thanks Quan.

PR #178, opened 2020-08-13
Fix hopscotch

victorapm:
This PR extends the implementation of a few functions in `hypre_hopscotch.h` to the case when `HYPRE_USING_OPENMP` is undefined (and consequently `HYPRE_CONCURRENT_HOPSCOTCH`). Now, one can use `Unordered[Int,BigInt][Set,Map]` capabilities without having to enable OpenMP.

### DIFF
--- a/src/utilities/hypre_hopscotch_hash.h
+++ b/src/utilities/hypre_hopscotch_hash.h
@@ -31,7 +31,7 @@
 //  Moran Tzafrir
 //  Tel-Aviv University
 //
-//  Date: July 15, 2008.  
+//  Date: July 15, 2008.
 //
 ////////////////////////////////////////////////////////////////////////////////
 // Programmer : Moran Tzafrir (MoranTza@gmail.com)
@@ -70,7 +70,8 @@ extern "C" {
  ******************************************************************************/
 
 #ifdef HYPRE_USING_ATOMIC
-static inline HYPRE_Int hypre_compare_and_swap(HYPRE_Int *ptr, HYPRE_Int oldval, HYPRE_Int newval)
+static inline HYPRE_Int
+hypre_compare_and_swap( HYPRE_Int *ptr, HYPRE_Int oldval, HYPRE_Int newval )
 {
 #if defined(__GNUC__) && defined(__GNUC_MINOR__) && defined(__GNUC_PATCHLEVEL__) && (__GNUC__ * 10000 + __GNUC_MINOR__ * 100 + __GNUC_PATCHLEVEL__) > 40100
   return __sync_val_compare_and_swap(ptr, oldval, newval);
@@ -84,7 +85,8 @@ static inline HYPRE_Int hypre_compare_and_swap(HYPRE_Int *ptr, HYPRE_Int oldval,
 #endif
 }
 
-static inline HYPRE_Int hypre_fetch_and_add(HYPRE_Int *ptr, HYPRE_Int value)
+static inline HYPRE_Int
+hypre_fetch_and_add( HYPRE_Int *ptr, HYPRE_Int value )
 {
 #if defined(__GNUC__) && defined(__GNUC_MINOR__) && defined(__GNUC_PATCHLEVEL__) && (__GNUC__ * 10000 + __GNUC_MINOR__ * 100 + __GNUC_PATCHLEVEL__) > 40100
   return __sync_fetch_and_add(ptr, value);
@@ -97,7 +99,8 @@ static inline HYPRE_Int hypre_fetch_and_add(HYPRE_Int *ptr, HYPRE_Int value)
 #endif
 }
 #else // !HYPRE_USING_ATOMIC
-static inline HYPRE_Int hypre_compare_and_swap(HYPRE_Int *ptr, HYPRE_Int oldval, HYPRE_Int newval)
+static inline HYPRE_Int
+hypre_compare_and_swap( HYPRE_Int *ptr, HYPRE_Int oldval, HYPRE_Int newval )
 {
    if (*ptr == oldval)
    {
@@ -107,7 +110,8 @@ static inline HYPRE_Int hypre_compare_and_swap(HYPRE_Int *ptr, HYPRE_Int oldval,
    else return *ptr;
 }
 
-static inline HYPRE_Int hypre_fetch_and_add(HYPRE_Int *ptr, HYPRE_Int value)
+static inline HYPRE_Int
+hypre_fetch_and_add( HYPRE_Int *ptr, HYPRE_Int value )
 {
    HYPRE_Int oldval = *ptr;
    *ptr += value;
@@ -125,12 +129,11 @@ static inline HYPRE_Int hypre_fetch_and_add(HYPRE_Int *ptr, HYPRE_Int value)
 #define HYPRE_HOPSCOTCH_HASH_BUSY  (1)
 
 // Small Utilities ..........................................................
-#ifdef HYPRE_CONCURRENT_HOPSCOTCH
-static inline HYPRE_Int first_lsb_bit_indx(hypre_uint x) 
+static inline HYPRE_Int
+first_lsb_bit_indx( hypre_uint x )
 {
   return ffs(x) - 1;
 }
-#endif
 /**
  * hypre_Hash is adapted from xxHash with the following license.
  */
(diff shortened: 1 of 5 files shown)


rfalgout (approved):
Hi @victorapm .  This all looks good to me, as long as it all passes the regression tests.  I wonder if we should ask the PETSc guys or someone else to give it a quick spin before merging.  @ulrikeyang What do you think?

liruipeng:
@victorapm I still need to understand the purpose of these changes. We can chat later on it. Thanks! -R

src/utilities/hypre_hopscotch_hash.h
```diff
@@ -581,9 +592,9 @@ void hypre_UnorderedIntMapDestroy( hypre_UnorderedIntMap *m );
 void hypre_UnorderedBigIntMapDestroy( hypre_UnorderedBigIntMap *m );
 
 // Query Operations .........................................................
-#ifdef HYPRE_CONCURRENT_HOPSCOTCH
-static inline HYPRE_Int hypre_UnorderedIntSetContains( hypre_UnorderedIntSet *s,
-                                                HYPRE_Int key )
+static inline HYPRE_Int
+hypre_UnorderedIntSetContains( hypre_UnorderedIntSet *s,
+                               HYPRE_Int              key )
```
liruipeng, line 597: @rfalgout @victorapm This function only existed when `HYPRE_CONCURRENT_HOPSCOTCH` was defined. You now want to make it available as well even when it is not defined. These functions are complicated and I have not understood how it works. But from a high level, how can we guarantee it works, especially if changing it by commenting out the lines involving `segment` (see lines below with the `#ifdef`)? There are more changes like this in this file. I have the same question for them as well.

Maybe you have dig into it so you understand how it works. If so, you can ignore my comments above.

src/utilities/hypre_hopscotch_hash.h
```diff
@@ -615,10 +630,12 @@ static inline HYPRE_Int hypre_UnorderedIntSetContains( hypre_UnorderedIntSet *s,
     if (hash == s->hash[currElm] && key == s->key[currElm])
       return 1;
     hopInfo &= ~(1U << i);
-  } 
+  }
 
+#ifdef HYPRE_CONCURRENT_HOPSCOTCH
   if (segment->timestamp == startTimestamp)
```
liruipeng, line 636: This is not related to this PR. Looking at the original code, it is mysterious for me. How can the if-statement on line 620 is not true if we set it equal in line 609. I don't understand the stuff in between can break the equality. Then, what's the point of having stuff after line 623? Maybe I missed a lot things here. Again, I haven't understood how HOPSCOTCH works yet.

liruipeng:
@victorapm @rfalgout I left some comments in the file. The changes make sense to me; as least I understand why you wanted to do it. But since I don't understand hopscotch yet, I could not make sure the changes can work without  `HYPRE_CONCURRENT_HOPSCOTCH`. If you can, just ignore my comments. I am totally fine with it.

Are the changes just for making it compiled fine without OpenMP? Have you tested it without `HYPRE_CONCURRENT_HOPSCOTCH`? I am just curious.

Thanks

-Ruipeng

src/utilities/hypre_hopscotch_hash.h
victorapm: Hi Ruipeng, I haven't dug too much into the details of the implementation either but the same approach (commenting lines involving `segments`) is used in other functions and I tried to replicate that. Moreover, these changes don't break the regression tests.

src/utilities/hypre_hopscotch_hash.h
victorapm: I agree with you, and the stuff after line 623 could be deleted. However, I decided to leave as it is since it doesn't affect the purpose of this PR.

victorapm:
Thanks for the review, @liruipeng! All the regression tests passed. These changes are basically to make it compile fine without OpenMP support and I tested some Unordered[Set, Map] capabilities with a simple driver both with and without `HYPRE_CONCURRENT_HOPSCOTCH` and I get the results expected.

src/utilities/hypre_hopscotch_hash.h
liruipeng: @victorapm 
OK. Maybe this is fine. It is actually the lines after 623 that bother me. Originally the function is guaranteed to `return 0` at line 621, since it is supposed to be with `#ifdef HYPRE_CONCURRENT_HOPSCOTCH` only. But now when without `#ifdef HYPRE_CONCURRENT_HOPSCOTCH`, it proceeds to the lines after. Just from a high level, the function ~~seems to have different behavior~~ goes to different paths.

liruipeng:
> Thanks for the review, @liruipeng! All the regression tests passed. These changes are basically to make it compile fine without OpenMP support and I tested some Unordered[Set, Map] capabilities with a simple driver both with and without `HYPRE_CONCURRENT_HOPSCOTCH` and I get the results expected.

@victorapm 
I was wondering if the regress tests actually test the code without `HYPRE_CONCURRENT_HOPSCOTCH` but with `HYPRE_HOPSCOTCH` defined, since the `#define` seems to be hard-coded, see https://github.com/hypre-space/hypre/blob/e42ab96f9797096c9a776fd1f92fc468cbfdffa7/src/utilities/protos.h#L126. But since you have tested it with a driver, I assume it is fine.

src/utilities/hypre_hopscotch_hash.h
victorapm: I understand what you are saying now. Previously those lines could never be reached but now this might be possible. Should I exclude them, or maybe put them inside an `#if 0 ... #endif ` region?

victorapm:
Good point, @liruipeng, I assumed it was being tested, however, I just looked at `machine-tux.sh` and there's no test that compiles Hypre with `--enable-hopscotch` (enabling HYPRE_HOPSCOTCH) or with `--with-openmp --enable-hopscotch` (which ultimately enables HYPRE_CONCURRENT_HOPSCOTCH). Should we add such a test? Moreover, the unordered set/map capabilities work regardless of the value of HYPRE_HOPSCOTCH and thus the library does not need to be configured with `--enable-hopscotch` (this was something that I had observed before). So does it make sense to have this config flag? Shouldn't it be renamed to `--enable-concurrent-hopscotch`?

rfalgout:
Hi @victorapm .  We definitely should add tux tests for hopscotch both with and without openmp.  Let me know if you need help with that.  Now that you've made these changes, do we need any configure options at all for hopscotch?

src/utilities/hypre_hopscotch_hash.h
liruipeng: @victorapm No, I think we may need it for the non-`HYPRE_CONCURRENT_HOPSCOTCH` case. It appears like a standard loop searching for hash values and the key. It is safer to leave it here. At least we can guarantee, with `HYPRE_CONCURRENT_HOPSCOTCH` the code is exactly as before. It doesn't seem this function is used anywhere in Hypre. Is this used in AMG-DD maybe? 

We need to remind ourselves to pay attention when using these functions without `HYPRE_CONCURRENT_HOPSCOTCH`. It doesn't seem we have tested them before. 

My 2 cents

liruipeng:
> Good point, @liruipeng, I assumed it was being tested, however, I just looked at `machine-tux.sh` and there's no test that compiles Hypre with `--enable-hopscotch` (enabling HYPRE_HOPSCOTCH) or with `--with-openmp --enable-hopscotch` (which ultimately enables HYPRE_CONCURRENT_HOPSCOTCH). Should we add such a test? Moreover, the unordered set/map capabilities work regardless of the value of HYPRE_HOPSCOTCH and thus the library does not need to be configured with `--enable-hopscotch` (this was something that I had observed before). So does it make sense to have this config flag? Shouldn't it be renamed to `--enable-concurrent-hopscotch`?

@victorapm 

~~Q: `HYPRE_CONCURRENT_HOPSCOTCH` is enabled when `HYPRE_USING_ATOMIC` is defined which is enabled when~~
```
if defined(__GNUC__) && defined(__GNUC_MINOR__) && defined(__GNUC_PATCHLEVEL__) && (__GNUC__ * 10000 + __GNUC_MINOR__ * 100 + __GNUC_PATCHLEVEL__) > 40100
```

~~Why does `--with-openmp` affect here?~~

Never mind. I was confused.

src/utilities/hypre_hopscotch_hash.h
victorapm: Thanks, @liruipeng! That makes sense. This function is not being used right now, however, its BigInt counterpart (which has the same logic) is being used in `aux_interp.c`

victorapm:
Hi @liruipeng and @rfalgout. Sorry for the delay in pushing these changes. I added a regression test in `machine-tux.sh` to test concurrent hopscotch capabilities. Note that I'm not comparing results to the `.saved` files (`HYPRE_NO_SAVED` flag is used) since this new test enables OpenMP, which produces slightly different convergence profiles in a few cases. There's no additional test for hopscotch capabilities (no concurrency) since they don't require any additional configure flag and thus are already accessible to the other tests in `machine-tux.sh`. All tests, including the newly added, passed in my tux box.

rfalgout:
Hi @victorapm .  This looks good.  Thanks!  One question.  It sounds like --enable-hopscotch is on by default *unless* openmp is enabled.  Is that correct?  When it is enabled, is it accessed in both cases by AMG, or only in the openmp case?

victorapm:
Hi, @rfalgout. The flag `--enable-hopscotch` has the only purpose of enabling the concurrent version of hopscotch (provided that openmp is on), this was the case before as well. On the other hand, the unordered set and map capabilities provided by the non-threaded hopscotch can be accessed by default, i.e., they don't require any additional configure flag. In BoomerAMG, since these capabilities are used inside regions guarded by `HYPRE_CONCURRENT_HOPSCOTCH`, this means that they are accessed only when `--enable-hopscotch --with-openmp` is used.

rfalgout:
Sounds good, @victorapm .  Thanks again!

liruipeng:
Looks good to me. Thanks, @victorapm

liruipeng (approved):

PR #179, opened 2020-08-17
Fix regtest error

oseikuffuor1:
This PR should fix recent regression test errors from the superlu build with spack, after a recent update to the MGR solver. The biggest change is the addition of a new file superlu.h to hold the hypre_DSLUData struct so that it is accessible outside the superlu.c file.


src/parcsr_ls/dsuperlu.c
```diff
@@ -9,11 +9,11 @@
 #include <math.h>
 
 #ifdef HYPRE_USING_DSUPERLU
-/*#include "superlu.h"*/
+#include "superlu.h"
```
rfalgout: The includes need to be changed to "dsuperlu.h" in multiple places

rfalgout (approved):
Looks great @oseikuffuor1 .  Thanks!

Issue #183, opened 2020-08-26
build error after hopscotch fix


suguanting:
Dear HYPRE developers,

I am setting up HYPRE on a new WIN10 machine using CMAKE, which return the following error:

```
Severity	Code	Description	Project	File	Line	Suppression State
Error (active)	E0020	identifier "ffs" is undefined	HYPRE	E:\hypre\hypre-master\src\utilities\hypre_hopscotch_hash.h	135	
```

It seems it concerns the new hopscotch fix, after which the code becomes:

```
// Small Utilities ..........................................................
static inline HYPRE_Int
first_lsb_bit_indx( hypre_uint x )
{
  return ffs(x) - 1;
}
```

whose original version that worked fine for me was

```
// Small Utilities ..........................................................
#ifdef HYPRE_CONCURRENT_HOPSCOTCH
static inline HYPRE_Int first_lsb_bit_indx(hypre_uint x) 
{
  return ffs(x) - 1;
}
#endif
```

So any suggestions? Thanks in advance for your help!

Guanting

liruipeng:
@suguanting @victorapm 

It seems that `ffs` is available for Linux only.

victorapm:
@suguanting, thanks for reporting this problem. We fixed it and merged the changes into master.

suguanting:
Hi @victorapm @liruipeng , it is working fine this time, thank you for the fix!

PR #184, opened 2020-08-26
Add implementation of ffs for MSVC

victorapm:
This fixes issue #183.


rfalgout (approved):
Works for me.  Thanks!

liruipeng:
Implementing own `ffs` might be the most portable solution. Although there seems to be more optimized but complicated implementation, if we don't need extreme efficiency of this function, I am fine with this one. It is simple and straightforward to understand. Thanks, Victor! @victorapm @rfalgout @ulrikeyang

ulrikeyang (approved):
Looks good.

liruipeng (approved):

PR #187, opened 2020-09-01
Epe gpu

ulrikeyang:
Hi all,
I wrote a CUDA implementation of the extended+e interpolation stealing extensively from previous implementations. So far it looks good, generates correct results, performs similarly to the CPU version, passes cuda-memcheck, performs well. I added a few regression tests for Ray.


ulrikeyang:
When I said it `performs' similarly to the CPU version, I meant this with regard to convergence not with regard to time.

rfalgout (approved):
Looks good to me, @ulrikeyang .  Thanks!

liruipeng (approved):
@ulrikeyang Thanks for implementing Ext+e on GPUs. It looks good to me. I made some minor changes.

bjorn2 (approved):

PR #189, opened 2020-09-02
Changing issue reporting to use github's issue tracker

rfalgout:
Hi @hypre-space/developers .  This pull request points users to our github issues page instead of our hypre-support email, which we decided to retire.  The roundup issue tracker will be kept around for our records and for handling remaining issues there, but we won't accept new issues with that tool.  I will also change our LLNL web page appropriately.


src/config/Makefile.config.saved
```diff
@@ -22,7 +22,7 @@ HYPRE_RELEASE_NAME    = hypre
 HYPRE_RELEASE_VERSION = 2.13.0
 HYPRE_RELEASE_DATE    = 2017/10/20
 HYPRE_RELEASE_TIME    = 00:00:00
-HYPRE_RELEASE_BUGS    = hypre-support@llnl.gov
+HYPRE_RELEASE_BUGS    = https://github.com/hypre-space/hypre/issues
 
 HYPRE_SRC_TOP_DIR = /g/g92/li50/workspace/hypre/src
```
victorapm: This is not the purpose of this PR, but do we need `Makefile.config.saved` to be under version control? I see there are some hardcoded paths like this one that might not be necessary. @liruipeng, can you check this?

victorapm (approved):
Looks good to me, Rob!

ulrikeyang (approved):
This looks good.

osborn9 (approved):

src/config/Makefile.config.saved
rfalgout: Hi @victorapm .  The purpose of this file was to provide users with an example `Makefile.config` that they could modify directly if they were having trouble with `configure`.  I'm not sure how important this is anymore, but if we keep it, we probably should update it to be more current.

liruipeng (approved):
Looks good to me. Nice to have one place to track issues. Btw, what happens if one sends email to hypre-support@llnl.gov in the future? @rfalgout Thanks!

src/config/Makefile.config.saved
liruipeng: @victorapm @rfalgout I actually did not know that this file was in the repo, which was introduced by me 3 years ago. It could be a mistake to push it. But if Rob think it is useful, I am fine with keeping it, but I would like to remove my username from it....Thanks!

rfalgout: I could either change the paths to generic ones or delete this file altogether.  I'm not sure it's all that useful anymore, since we have multiple ways of building hypre.  What's the consensus?

liruipeng: > consensus

I am for deleting it

victorapm: Thanks, @rfalgout! I agree with Ruipeng. In my view, if someone wants to change `Makefile.config` manually, this could be done easily since the file is not difficult to understand, so the `Makefile.config.saved` is not really necessary.

rfalgout: That's enough of a consensus for me.  I'll remove it.  Thanks!

rfalgout:
Hi @liruipeng .  I am going to direct people away from the hypre-support email, but we probably should also set it up to auto-reply to users to instead visit the issues page.

PR #195, opened 2020-09-18
aggressive coarsening and 2s interp on GPUs

liruipeng:
This PR contains the implementation of aggressive coarsening and 2-stage MM-extended-type interpolations on GPU. The main functions are 
1. Aggressive coarsening (hypre_BoomerAMGCreate2ndS)
2. Partial MM-ext Interpolation (hypre_BoomerAMGBuildModPartialExtInterpDevice)
3. Partial MM-ext+e Interpolation (hypre_BoomerAMGBuildModPartialExtPEInterpDevice)

### DIFF
--- a/src/parcsr_ls/par_lr_interp_device.c
+++ b/src/parcsr_ls/par_lr_interp_device.c
@@ -14,17 +14,17 @@
 #define MAX_C_CONNECTIONS 100
 #define HAVE_COMMON_C 1
 
-__global__ void compute_weak_rowsums( HYPRE_Int nr_of_rows, bool has_offd, HYPRE_Int *CF_marker, HYPRE_Int *A_diag_i, HYPRE_Complex *A_diag_a, HYPRE_Int *S_diag_j, HYPRE_Int *A_offd_i, HYPRE_Complex *A_offd_a, HYPRE_Int *S_offd_j, HYPRE_Real *rs );
+__global__ void hypreCUDAKernel_compute_weak_rowsums( HYPRE_Int nr_of_rows, bool has_offd, HYPRE_Int *CF_marker, HYPRE_Int *A_diag_i, HYPRE_Complex *A_diag_a, HYPRE_Int *S_diag_j, HYPRE_Int *A_offd_i, HYPRE_Complex *A_offd_a, HYPRE_Int *S_offd_j, HYPRE_Real *rs, HYPRE_Int flag );
 
 __global__ void compute_aff_afc( HYPRE_Int nr_of_rows, HYPRE_Int *AFF_diag_i, HYPRE_Int *AFF_diag_j, HYPRE_Complex *AFF_diag_data, HYPRE_Int *AFF_offd_i, HYPRE_Complex *AFF_offd_data, HYPRE_Int *AFC_diag_i, HYPRE_Complex *AFC_diag_data, HYPRE_Int *AFC_offd_i, HYPRE_Complex *AFC_offd_data, HYPRE_Complex *rsW, HYPRE_Complex *rsFC );
 
 void hypreDevice_extendWtoP( HYPRE_Int P_nr_of_rows, HYPRE_Int W_nr_of_rows, HYPRE_Int W_nr_of_cols, HYPRE_Int *CF_marker, HYPRE_Int W_diag_nnz, HYPRE_Int *W_diag_i, HYPRE_Int *W_diag_j, HYPRE_Complex *W_diag_data, HYPRE_Int *P_diag_i, HYPRE_Int *P_diag_j, HYPRE_Complex *P_diag_data, HYPRE_Int *W_offd_i, HYPRE_Int *P_offd_i );
 
-__global__ void compute_twiaff_w( HYPRE_Int nr_of_rows, HYPRE_Int first_index, HYPRE_Int *AFF_diag_i, HYPRE_Int *AFF_diag_j, HYPRE_Complex *AFF_diag_data, HYPRE_Complex *AFF_diag_data_old, HYPRE_Int *AFF_offd_i, HYPRE_Int *AFF_offd_j, HYPRE_Complex *AFF_offd_data, HYPRE_Int *AFF_ext_i, HYPRE_BigInt *AFF_ext_j, HYPRE_Complex *AFF_ext_data, HYPRE_Complex *rsW, HYPRE_Complex *rsFC, HYPRE_Complex *rsFC_offd );
+__global__ void compute_twiaff_w( HYPRE_Int nr_of_rows, HYPRE_BigInt first_index, HYPRE_Int *AFF_diag_i, HYPRE_Int *AFF_diag_j, HYPRE_Complex *AFF_diag_data, HYPRE_Complex *AFF_diag_data_old, HYPRE_Int *AFF_offd_i, HYPRE_Int *AFF_offd_j, HYPRE_Complex *AFF_offd_data, HYPRE_Int *AFF_ext_i, HYPRE_BigInt *AFF_ext_j, HYPRE_Complex *AFF_ext_data, HYPRE_Complex *rsW, HYPRE_Complex *rsFC, HYPRE_Complex *rsFC_offd );
 
 __global__ void compute_aff_afc_epe( HYPRE_Int nr_of_rows, HYPRE_Int *AFF_diag_i, HYPRE_Int *AFF_diag_j, HYPRE_Complex *AFF_diag_data, HYPRE_Int *AFF_offd_i, HYPRE_Int *AFF_offd_j, HYPRE_Complex *AFF_offd_data, HYPRE_Int *AFC_diag_i, HYPRE_Complex *AFC_diag_data, HYPRE_Int *AFC_offd_i, HYPRE_Complex *AFC_offd_data, HYPRE_Complex *rsW, HYPRE_Complex *dlam, HYPRE_Complex *d_tmp, HYPRE_Complex *dtmp_offd );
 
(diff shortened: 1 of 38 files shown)


src/utilities/hypre_cuda_utils.h
```diff
@@ -664,6 +664,19 @@ struct less_than : public thrust::unary_function<T,bool>
    }
 };
 
+template<typename T>
```
rfalgout, line 667: Should this be a '.hpp' file?

rfalgout (approved):
Looks good to me, @liruipeng .  Thanks!

src/parcsr_ls/par_lr_interp_device.c
```diff
@@ -1017,7 +1030,7 @@ hypreDevice_extendWtoP( HYPRE_Int      P_nr_of_rows,
 // For Ext+i Interp, scale AFF from the left and the right
 __global__
 void compute_twiaff_w( HYPRE_Int      nr_of_rows,
```
ulrikeyang: Should this also be a hypreCUDAKernel ?

ulrikeyang (commented):
Apart from Rob's and my comment this looks fine to me.
Also, shouldn't there be a par_2s_interp_device.c file?
I assume we also are waiting for Bjorn to check in the 2nd strength matrix.

bjorn2:
It looks ok to me. Do you want to merge this first, then I update my local branch before doing a pull request ?

liruipeng:
> It looks ok to me. Do you want to merge this first, then I update my local branch before doing a pull request ?
@bjorn2 
Would you like to push to this PR? I don't want to merge it to master, because it is still a draft. No regression tests have been done.

bjorn2:
ok, so should I merge your PR to my local branch first ? Or will you take care of testing that it all works together ?

liruipeng:
> ok, so should I merge your PR to my local branch first ? Or will you take care of testing that it all works together ?

@bjorn2 Yes, you may check out this branch, pull your branch into it, and push it back. Please let me know when the S2 routine is done. So I will test them all together. Thank you!

src/utilities/hypre_cuda_utils.h
liruipeng: @rfalgout In this `.h`, we put all kinds of cuda C++ related stuff, which will be dumped into `_hypre_utilities.hpp` along with 2 other C++ `.h`. Would you like to rename all with `.hpp`? Thanks!

src/parcsr_ls/par_lr_interp_device.c
liruipeng: @ulrikeyang Yes, it should be, and has been fixed. The idea is to prefix our kernels with `hypreCUDAKernel_`. There are still other places we missed, which I will change later. Thanks!

src/utilities/hypre_cuda_utils.h
rfalgout: If they are all getting dumped into `_hypre_utilities.hpp` and they contain C++ stuff, then I think it makes sense to rename all three of these files with `.hpp` extensions.  This is minor, though, and not crucial to getting this pull request and release done; we can worry about it later.

src/parcsr_ls/par_lr_interp_device.c
liruipeng: @ulrikeyang I missed `par_2s_interp_device.c`, which has been pushed.

src/utilities/hypre_cuda_utils.h
liruipeng: > If they are all getting dumped into `_hypre_utilities.hpp` and they contain C++ stuff, then I think it makes sense to rename all three of these files with `.hpp` extensions. This is minor, though, and not crucial to getting this pull request and release done; we can worry about it later.

Sure. No problem. Will do it later. Thanks!

bjorn2:
Ruipeng, I pushed the routine to aggcoarsen branch. I did some performance testing, and it seems to not be great, I might continue to test more.

liruipeng:
> Ruipeng, I pushed the routine to aggcoarsen branch. I did some performance testing, and it seems to not be great, I might continue to test more.

Great, Thanks! I will take a look.

src/parcsr_ls/par_strength2nd_device.c
```diff
@@ -0,0 +1,927 @@
...
+   hypre_ParCSRMatrix* SI = hypre_ParCSRMatrixClone( S, 1 );
+   add_diagonal( SI );
+
+   // 2. Add a data array to the strength matrix, needed in order 
+   //    to use matrix-matrix multiplication.
+   add_unit_data( SI );
+   hypre_NvtxPopRange();
+
+   // 2b. Scale diagonal by coeff   
+   if( coeff != 1 )
```
liruipeng, line 194: @bjorn2 @ulrikeyang Do we need to consider `coeff !=1`, which is fixed to be `1` here? My understanding is we need to compute `(S+2I)S` here. So, why `coeff` is 1?

bjorn2: My understanding was that it is not completely clear if it should be S+2*I or S+I. However, in the case of num_path=1, I think it makes no difference, so setting coeff=1 saves a few operations. The idea with introducing the variable was that we could easily choose one case or the other.

liruipeng:
Hi, @bjorn2 I tried to run aggressive coarsening on device but got seg-fault. I have identified some problems. I will try to fix them. Please do not push to this branch until I give the green light. Thanks!

ulrikeyang:
We discussed that S+2*I is needed when setting the number of paths greater than 1. For num_paths == 1 it doesn’t matter what the coefficient is, so in that case S+I is fine. If we can get better performance with S+I, which is equivalent to just the structure, no data values, we should implement that, but it might make sense to also add the option of num_paths > 1 at a later time. Right now, my main concern is to get aggressive coarsening with num_paths 1 working on GPUs.

liruipeng:
@bjorn2 I have moved your code around like taking C-C submatrix, removing diagonal of CSR, etc, so we can reuse them later. The original code was left there and commented out for now. Also, I made modification of computing CC submatrix of S^2 which hopefully can be more efficient. Thanks!
 
I have run some aggressive coarsening tests on GPUs, which seemed reasonable. The regression tests (CPU, GPU) were clean (aside from AMG-DD). Regression tests for aggressive coarsening have **not** been added, which will be done after @ulrikeyang run the tests for the report.

@ulrikeyang @bjorn2 @rfalgout This PR is ready for a final review.

Thank you!

rfalgout (approved):
This still looks good to me.  Thanks!

src/parcsr_ls/par_strength2nd_device.c
ulrikeyang: We discussed that S+2*I is needed when setting the number of paths greater than 1. For num_paths == 1 it doesn’t matter what the coefficient is, so in that case S+I is fine. If we can get better performance with S+I, which is equivalent to just the structure, no data values, we should implement that, but it might make sense to also add the option of num_paths > 1 at a later time. Right now, my main concern is to get aggressive coarsening with num_paths 1 working on GPUs.

ulrikeyang (approved):
This seems to be fine, although there is no way I could check all of this. If there are any issues I hope they will show up when I do the performance tests.

src/parcsr_ls/par_strength2nd_device.c
bjorn2: Ulrike, I still don't understand why it is S+2*I, and not S+I, perhaps you could explain ?
Note, the function creat2ndS works also for num_paths==2.
Bjorn

ulrikeyang: It is (S+I)^2-I. Are you saying that using S+I also leads to a correct S2 matrix for num_paths == 2?

bjorn2 (approved):
I made some runs, seems to work fine now!

liruipeng:
@bjorn2 @rfalgout @ulrikeyang @oseikuffuor1 @victorapm The PR should be good to merge. Below are the most recent changes and status.

1. I changed the order of some statements in PMISDevice, see https://github.com/hypre-space/hypre/blob/4abb4c54693f222eae91386b20d4085a091c5044/src/parcsr_ls/par_coarsen_device.c#L283, so the regression test results now are the same as before for non-aggressive coarsening. The GPU implementation should be more consistent with the behavior of the host one. For aggressive coarsening, the results can be slightly different from 1-2 days ago, up to 1 or 2 iterations. 
2. @ulrikeyang There was a typo `agg_mx` that was supposed to be `agg_Pmx` in `gpuboomer.jobs`. Also, you cannot generate `.saved` in `TEST_cuda_ray` on `rzansel` since they have different GPUs, but you can do that for  `TEST_cuda_lassen`.
3. Fixed the issue of running `hypre_BoomerAMGInterpTruncation` on the host (https://github.com/hypre-space/hypre/blob/4abb4c54693f222eae91386b20d4085a091c5044/src/parcsr_ls/par_amg_setup.c#L1657). @oseikuffuor1 Please review the changes and comments at https://github.com/hypre-space/hypre/blob/4abb4c54693f222eae91386b20d4085a091c5044/src/parcsr_ls/par_interp.c#L2717 and https://github.com/hypre-space/hypre/blob/4abb4c54693f222eae91386b20d4085a091c5044/src/parcsr_ls/par_interp_trunc_device.c#L91.
4. `machine-tux.sh`, `machine-lassen.sh` and `machine-ray.sh` are all clear now. @victorapm I disabled IJ with `device-openmp` in the GPU regression tests, see https://github.com/hypre-space/hypre/blob/4abb4c54693f222eae91386b20d4085a091c5044/AUTOTEST/machine-ray.sh#L64 and https://github.com/hypre-space/hypre/blob/4abb4c54693f222eae91386b20d4085a091c5044/AUTOTEST/machine-lassen.sh#L64. We may revisit the issues later.

rfalgout (approved):
This all looks good to me, @liruipeng !  Once you squash-merge this (remember to write an informative commit message; see the Wiki), I will start the release process and rerun all of these tests.  Thanks!

victorapm (approved):
Hi @liruipeng,

I just looked at the changes related to the regression tests. I agree on revisiting this later :)
Thanks,
Victor

oseikuffuor1 (approved):
Looks good to me.

PR #204, opened 2020-09-24
optimization in hypre_CSRMatrixBigJtoJ and JtoBigJ

liruipeng:
Code optimization as suggested in #132. @v-dobrev can you also take a look? Thanks!

### DIFF
--- a/src/seq_mv/csr_matrix.c
+++ b/src/seq_mv/csr_matrix.c
@@ -216,7 +216,6 @@ hypre_CSRMatrixBigJtoJ( hypre_CSRMatrix *matrix )
    HYPRE_Int     num_nonzeros = hypre_CSRMatrixNumNonzeros(matrix);
    HYPRE_BigInt *matrix_big_j = hypre_CSRMatrixBigJ(matrix);
    HYPRE_Int    *matrix_j = NULL;
-   HYPRE_Int     ierr = 0;
 
    if (num_nonzeros && matrix_big_j)
    {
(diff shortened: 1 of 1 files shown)


rfalgout (approved):
Looks good to me!  -Rob

src/seq_mv/csr_matrix.c
```diff
@@ -215,19 +216,23 @@ hypre_CSRMatrixBigJtoJ( hypre_CSRMatrix *matrix )
    HYPRE_Int     num_nonzeros = hypre_CSRMatrixNumNonzeros(matrix);
    HYPRE_BigInt *matrix_big_j = hypre_CSRMatrixBigJ(matrix);
    HYPRE_Int    *matrix_j = NULL;
-   HYPRE_Int     i;
+   HYPRE_Int     ierr = 0;
```
victorapm: This is outside the purpose of this PR, but should we do `return ierr` when the value of `ierr` can't be changed in a function? How about replacing it to `return hypre_error_flag`? My understanding is that the latter approach is preferred so I wanted to check this with everyone :)

victorapm (approved):
Looks good to me, thanks!

oseikuffuor1 (approved):
Looks good to me, thanks. I also agree with Victor that we should use hypre_error_flag instead of ierr, if ierr is not used. It might even throw a compiler warning if the unused-variable warning option is turned on, which would fail the regression test.

ulrikeyang (approved):

liruipeng:
Made changes per suggestions from @victorapm @oseikuffuor1. `machine-tux.sh` was clean.

PR #206, opened 2020-09-25
Set default convergence tolerance to 1.0e-6

rfalgout:
This sets the default convergence tolerance to 1.0e-6 uniformly in hypre.  Note that the 'test' drivers change this default in many cases, which is good because it invokes the SetTol() routines.  This addresses Issue #70 and pull request #156.


liruipeng (approved):
Looks good to me. Thanks!

oseikuffuor1 (approved):
Looks good to me. Thanks @rfalgout .

victorapm (approved):
Looks good to me, Rob!

ulrikeyang (approved):

Issue #210, opened 2020-09-29
Integer Overflow in HYPRE_EuclidSetup


ryagal:
Hello,

The preconditioner setup for Euclid occasionally crashes for matrices with about 1 to 5 million unknowns. For example, it will successfully setup the preconditioner for a matrix with 1, 2, 4, 8, 32, 64 processes but crash with 16. The problem occurs due to integer overflow in either Hash_i_dhLookup or Hash_i_dhInsert. The problem is that

    idx = (start + i*inc) % size;

can overflow when performing i*inc and produce a negative value for idx which is supposed to be a positive. Casting to 64 bit integers for these operations fixes the problem (when not building with BigInts), but I'm not sure how this should be handled in the case where HYPRE is built with BigInts.

liruipeng:
> Hello,
> 
> The preconditioner setup for Euclid occasionally crashes for matrices with about 1 to 5 million unknowns. For example, it will successfully setup the preconditioner for a matrix with 1, 2, 4, 8, 32, 64 processes but crash with 16. The problem occurs due to integer overflow in either Hash_i_dhLookup or Hash_i_dhInsert. The problem is that
> 
> ```
> idx = (start + i*inc) % size;
> ```
> 
> can overflow when performing i*inc and produce a negative value for idx which is supposed to be a positive. Casting to 64 bit integers for these operations fixes the problem (when not building with BigInts), but I'm not sure how this should be handled in the case where HYPRE is built with BigInts.

@ryagal Can you please checkout branch https://github.com/hypre-space/hypre/tree/Euclid-FIX, and let us know if it fixes the problem? Thanks!

-Ruipeng

ryagal:
@liruipeng Hi Ruipeng, sorry about the delayed response. I've verified that the Euclid-FIX branch fixed the problem for the matrices I was encountering this issue with. I see that you merged the changes today, and will close this issue. Thanks for your help!

PR #218, opened 2020-10-15
Euclid fix

liruipeng:
This PR tries to fix issue #210, which is an integer overflow problem when computing hash values in the form of
```
idx = (start + i*inc) % size;
```
where `i*inc` can be too big to be held by `HYPRE_Int`.

### DIFF
--- a/src/distributed_ls/Euclid/Hash_i_dh.c
+++ b/src/distributed_ls/Euclid/Hash_i_dh.c
@@ -63,8 +63,8 @@ void Hash_i_dhCreate(Hash_i_dh *h, HYPRE_Int sizeIN)
   struct _hash_i_dh* tmp;
 
   size = DEFAULT_TABLE_SIZE;
-  if (sizeIN == -1) { 
-    sizeIN = size = DEFAULT_TABLE_SIZE; 
+  if (sizeIN == -1) {
+    sizeIN = size = DEFAULT_TABLE_SIZE;
   }
   tmp = (struct _hash_i_dh*)MALLOC_DH( sizeof(struct _hash_i_dh)); CHECK_V_ERROR;
   *h = tmp;
@@ -135,7 +135,8 @@ HYPRE_Int Hash_i_dhLookup(Hash_i_dh h, HYPRE_Int key)
 */
 
   for (i=0; i<size; ++i) {
-    idx = (start + i*inc) % size;
+    /* idx = (start + i*inc) % size; */
+    idx = (start + hypre_multmod(i, inc, size)) % size;
 
 /* hypre_printf("   idx= %i\n", idx); */
 
@@ -146,7 +147,7 @@ HYPRE_Int Hash_i_dhLookup(Hash_i_dh h, HYPRE_Int key)
         retval = data[idx].data;
         break;
       }
-    } 
+    }
   }
   END_FUNC_VAL(retval)
 }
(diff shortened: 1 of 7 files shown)


oseikuffuor1 (approved):
@liruipeng This looks good to me, although I am not sure how the new hypre_multimod function is fixing the integer overflow issue. Is it just to limit the hash table size?

rfalgout (approved):
Looks good to me.  I assume there is no simple way to use the hash functions that @victorapm recently worked on for the AMG-DD code?  Don't go out of your way to make that happen.  I just thought I'd ask.

ulrikeyang (approved):

liruipeng:
@oseikuffuor1 @rfalgout This is a known problem, see https://www.geeksforgeeks.org/multiply-large-integers-under-large-modulo/. The new code is basically from there. To compute `(a*b)%m`, where `a*b` causes overflow, since `a*b = a + a + ... + a (b times)`, the simplest way to fix is to use the property of modulo, i.e, `(a + a + ... + a)%m=(((a + a) %m + a)%m + ... + a)%m`, which can avoid the overflow. But the problem is when `b` is large, it is not efficient, complexity O(`b`). So, we do it in a "log2" fashion, that divides `b` by 2 every time and double `a`.

src/distributed_ls/Euclid/Hash_i_dh.c
```diff
@@ -135,7 +135,8 @@ HYPRE_Int Hash_i_dhLookup(Hash_i_dh h, HYPRE_Int key)
 */
 
   for (i=0; i<size; ++i) {
-    idx = (start + i*inc) % size;
+    /* idx = (start + i*inc) % size; */
+    idx = (start + hypre_multmod(i, inc, size)) % size;
```
oseikuffuor1, line 139: @liruipeng is hypre_multmod replacing (i*inc) % size? If so should we be taking % size outside the parenthesis? Should it is idx = start % size + hypre_multmod(i, inc, size) ?

liruipeng: @oseikuffuor1 
```
idx = (start + i*inc) % size
    = (start + (i*inc) % size) % size
    = (start + hypre_multmod(i, inc, size)) % size
```

oseikuffuor1: Got it. Thanks for the earlier reference and explanation.

liruipeng:
> I assume there is no simple way to use the hash functions that @victorapm recently worked on for the AMG-DD code? 

@rfalgout @victorapm  Maybe we can use hopscotch hash. But I just did not want to "intrude" the code of Euclid too much. I actually don't know the algorithm, so only replace the particular operation with an equivalent.

victorapm (approved):
Thanks, @liruipeng! I believe hypre_hopscotch could be used here, but the changes would involve a reasonable amount of lines of code and I don't think it's worth the effort since the hash table implementation of Euclid does not show any problem (after your last fixes).

One last thing, I realized the operation `idx = (start + i*inc) % size` is performed also in other places, e.g.:

https://github.com/hypre-space/hypre/blob/636706acd70055d02fe5c85c07efef7083498303/src/distributed_ls/Euclid/Hash_dh.c#L107

and

https://github.com/hypre-space/hypre/blob/636706acd70055d02fe5c85c07efef7083498303/src/distributed_ls/Euclid/Hash_dh.c#L148

Could you fix those also?

Thanks!!

liruipeng:
> Thanks, @liruipeng! I believe hypre_hopscotch could be used here, but the changes would involve a reasonable amount of lines of code and I don't think it's worth the effort since the hash table implementation of Euclid does not show any problem (after your last fixes).
> 
> One last thing, I realized the operation `idx = (start + i*inc) % size` is performed also in other places, e.g.:
> 
> https://github.com/hypre-space/hypre/blob/636706acd70055d02fe5c85c07efef7083498303/src/distributed_ls/Euclid/Hash_dh.c#L107
> 
> and
> 
> https://github.com/hypre-space/hypre/blob/636706acd70055d02fe5c85c07efef7083498303/src/distributed_ls/Euclid/Hash_dh.c#L148
> 
> Could you fix those also?
> 
> Thanks!!

Fixed. Thank you, @victorapm

PR #225, opened 2020-10-23
Check compiler name using basename

frobnitzem:
When compiling with spack, I get link errors (can't find `omp_in_parallel()`) because spack uses `CC=/path/to/wrapped/gcc`, and hence the -fopenmp flag doesn't get added properly.


rfalgout (approved):
Looks good to me.

osborn9 (approved):

oseikuffuor1 (approved):
Looks good to me

PR #227, opened 2020-10-29
Improve portability for update-release script (for LINUX, UNIX and macOS).

oseikuffuor1:
This PR addresses portability issues in the update-release script, particularly for macOS. 
1. The use of the 'date' command has been modified to use GNUs 'date' command (if installed). 
2. In addition, single quotes for sed commands have been replaced by double quotes to allow the use of single quotes around internal variables. This appears to be more portable than the use of '\x27'. Note that this means shell meta-characters need to be escaped if they need to be treated as string literals.

### DIFF
--- a/src/docs/update-release.sh
+++ b/src/docs/update-release.sh
@@ -10,12 +10,16 @@ refconf="ref-manual/conf.doxygen"
 
 version=`../utilities/version -number`
 reldate=`../utilities/version -date`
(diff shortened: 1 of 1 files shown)


rfalgout (approved):
I made one small change for consistency.  These 'sed' lines are better than what I had before, so thanks for that.  I was originally worried that the '.*' would be expanded inside the double quotes (I think), but bash reference says only '$', '`', and '\\' are expanded.

PR #230, opened 2020-11-11
cmake: Optionally accept path to BLAS/LAPACK libraries.

osborn9:
This change allows the BLAS/LAPACK libraries to be specified within cmake configuration. Otherwise, if external BLAS/LAPACK is used it will use cmake's find_package to locate libraries. This makes the cmake more xsdk compliant as well.


rfalgout (approved):
Looks good to me.  Thanks!

oseikuffuor1 (approved):
Looks good. Thanks Sarah!

liruipeng (approved):

PR #231, opened 2020-11-11
Cuda triangular smoothers [draft]

liruipeng:
This PR was copied from #224 from a remote repo


PaulMullowney:
Any update on this PR?

src/seq_mv/vector_batched.c
```diff
@@ -1185,8 +1185,8 @@ HYPRE_Int hypre_SeqVectorMassDotpTwo( hypre_Vector *x, hypre_Vector *y,
    {
       for (j = 0; j < k; j++)
       {
-         res_x = result_x[j];
-         res_y = result_y[j];
+	 res_x = 0; //result_x[j];
```
rfalgout, line 1188: Why are the comments needed here?  Which RHS is correct here?

liruipeng: Good question. I am not sure.  it seems that @PaulMullowney made the changes. Thanks!

PaulMullowney:
As far as I can tell, result_x and result_y are not initialized properly. For instance, look at how the accumulation vars in
hypre_SeqVectorMassDotpTwo8
are initialized compared to
hypre_SeqVectorMassDotpTwo

I think setting res_x, res_y = 0 is correct but I would like someone else to confirm that.

-Paul

Issue #235, opened 2020-11-22
CF/FC Relax 18


liruipeng:
@rfalgout @ulrikeyang 

This line doesn't seem correct to me,
https://github.com/hypre-space/hypre/blob/2bc4228eca313f0eeb4fdb6bf1baa81ad8a7e317/src/parcsr_ls/par_cycle.c#L462

What we want here is CF/FC relaxations, CF down and FC up. So, we should have `if (cycle_param < 2)` instead of `if (cycle_type < 2)`? `cycle_param` controls the down/up/coarsest cycle, whereas `cycle_type` is for V-/W-cycles.

Any ideas?

Thanks

-Ruipeng

rfalgout:
@ulrikeyang is more familiar with this code than me, so I'll leave it to her to comment.

ulrikeyang:
@liruipeng @rfalgout 
Ruipeng is right. This should be cycle_param. Thanks for catching this.

PR #236, opened 2020-11-23
bug fix cycle_param

liruipeng:
This PR fixes the improper use of `relax_type` in relax-18 with CF relax, see #235

### DIFF
--- a/src/parcsr_ls/par_cycle.c
+++ b/src/parcsr_ls/par_cycle.c
@@ -459,7 +459,7 @@ hypre_BoomerAMGCycle( void              *amg_vdata,
                      /* need to do CF - so can't use the AMS one */
                      HYPRE_Int i;
                      HYPRE_Int loc_relax_points[2];
-                     if (cycle_type < 2)
+                     if (cycle_param < 2)
                      {
                         loc_relax_points[0] = 1;
                         loc_relax_points[1] = -1;
--- a/src/test/TEST_ij/smoother.saved
+++ b/src/test/TEST_ij/smoother.saved
@@ -47,8 +47,8 @@ BoomerAMG Iterations = 11
 Final Relative Residual Norm = 7.457693e-09
 
 # Output file: smoother.out.9
-BoomerAMG Iterations = 14
-Final Relative Residual Norm = 4.118037e-09
+BoomerAMG Iterations = 17
+Final Relative Residual Norm = 4.979125e-09
 
 # Output file: smoother.out.10
 BoomerAMG Iterations = 23


rfalgout (approved):
Looks good.  Thanks!

ulrikeyang (approved):
It's strange that convergence got worse.

rfalgout:
About Ulrike's convergence comment.  I think this makes sense from the following perspective.  For a 7-pt Laplacian, the first AMG coarse grid is basically a red/black grid, so this is reduction.  For reduction, CF/FC relaxation is like doing 1.5 sweeps vs. the 2 full sweeps that you would get from CF/CF or FC/FC.

ulrikeyang:
Ok. This makes sense. The test was AMG only vs AMG-PCG, where we might have noticed this bug earlier.

liruipeng:
> About Ulrike's convergence comment. I think this makes sense from the following perspective. For a 7-pt Laplacian, the first AMG coarse grid is basically a red/black grid, so this is reduction. For reduction, CF/FC relaxation is like doing 1.5 sweeps vs. the 2 full sweeps that you would get from CF/CF or FC/FC.

@rfalgout _"CF/FC relaxation is like doing 1.5 sweeps"_ Because of the 2nd "F" not doing relaxations?

rfalgout:
In pure reduction, the second F-relaxation is redundant, because ideal interpolation is equivalent to doing injection followed by F-relaxation.

rfalgout:
I just realized the situation is even worse than 1.5 sweeps.  If you think about the relaxations that occur between each coarse-grid correction, you have FCCF.  Both the first F and one of the C's is redundant, so you really only get CF.  That's closer to only doing one smoothing step over the course of several iterations (i.e., a V(1,0) cycle).

liruipeng:
> I just realized the situation is even worse than 1.5 sweeps. If you think about the relaxations that occur between each coarse-grid correction, you have FCCF. Both the first F and one of the C's is redundant, so you really only get CF. That's closer to only doing one smoothing step over the course of several iterations (i.e., a V(1,0) cycle).

@rfalgout This is only true with R-B coarsening where P is ideal and C^2=C, right?

rfalgout:
Yes, but in the non-reduction case, you would still get 1.5 sweeps per iteration because of the back-to-back C sweeps.  It might even be less than that because interpolation usually approximates ideal interpolation in some sense.

liruipeng:
> Yes, but in the non-reduction case, you would still get 1.5 sweeps per iteration because of the back-to-back C sweeps. It might even be less than that because interpolation usually approximates ideal interpolation in some sense.

Do two back-to-back C sweeps always have one redundant? I thought only in the case $A_{CC}$ is diagonal matrix it is true.

rfalgout:
Good point.  Not necessarily, depending on strength matrix, etc., but often C-points form an independent set.

liruipeng:
Thank you, @rfalgout. It is now clear to me.

PR #237, opened 2020-11-24
Reorg relax

liruipeng:
This PR refactors the relaxation routines on CPUs and modularize the various Jacobi and Gauss-Seidel (G-S) methods in two "core" kernels. Part of this work is for incorporating the triangular smoothers on GPUs in #231. 

Detailed changes:

1. `hypre_ParCSRRelax` used in AMS no longer has separate implementations of Jacobi or G-S. Instead, it calls the main `hypre_BoomerAMGRelax` routine,

2. In `par_cycle.c`, `relax_type == 18` no longer relies on the AMS relaxation routines.

3. Major code refactor in `par_relax.c`, where variants of Jacobi and G-S (weighted, L1, C/F, forward, backward, symmetric/SSOR, threaded/sequential) have been consolidated into `hypre_BoomerAMGRelaxWeightedJacobi_core` and `hypre_BoomerAMGRelaxHybridGaussSeidel_core` with some optimizations. These two kernels should be ported to GPUs later.

4. Move Kaczmarz in `hypre_BoomerAMGRelax`, which can be ported to GPUs as well.

5. Move `hypre_ParCSRRelax_L1_Jacobi` and `hypre_BoomerAMGRelax_FCFJacobi` into `par_relax_interface.c` and remove the redundant code. It calls the main `hypre_BoomerAMGRelax` routine. So, they are just "interfaces".

6. Minor changes on `SyncCudaComputeStream`.

7. ~~Add `hypre_Iterator` to emulate C++ iterator/reverse iterator for a nicer implementation of SSOR, without writing two identical for-loops that have reverse orders. Example code follows.~~

8. Move `hypre_gselim` into `src/utilities`


src/parcsr_ls/par_relax.c
```diff
@@ -16,4395 +16,1544 @@
...
-            requests= hypre_CTAlloc(hypre_MPI_Request,  num_recvs+num_sends, HYPRE_MEMORY_HOST);
+         for (sweep = 0; sweep < num_sweeps; sweep++)
+         {
+            hypre_Iterator iter;
+            hypre_IteratorFirst(&iter) = ns;
+            hypre_IteratorLast(&iter) = ne;
+            hypre_IteratorDirection(&iter) = num_sweeps == 1 ? gs_order : sweep == 0 ? 1 : -1;
 
-            if (num_cols_offd)
+            for ( i = hypre_IteratorBegin(&iter); i != hypre_IteratorEnd(&iter); i += hypre_IteratorStep(&iter) )
```
rfalgout: Hi @liruipeng .  Any idea about the overhead of this?  I see that you use 'static inline', which should have the potential to be quite fast, depending on the compiler.  I just don't have experience with it.

rfalgout (approved):
Hi @liruipeng .  It looks like you were able to eliminate quite a bit of code without losing functionality.  That's great!  I would be sure to do extensive testing and comparisons to the previous versions of these to make sure it's all still correct.  Otherwise, this looks good to me.  It's really @ulrikeyang who needs to chime in on this.  Thanks!

src/parcsr_ls/par_relax.c
liruipeng: `inline` is a hint to compilers to expand the code instead of making function calls. If the compiler takes the hint, there should be no cost. Even if it doesn't I assume it should not be too much problem, since sparse solvers generally are not CPU-bound.

liruipeng:
Thanks @rfalgout . Simplifying the CPU code is good by itself since for example we don't need to add some missing features such as the "persistent MPI" to all the routines; there is only one place now. However, the motivation was actually to reduce the complexity of porting them to GPUs.

src/parcsr_ls/par_relax.c
rfalgout: I would still do some testing/timing, just to be sure.  Compilers don't always do what we'd like them to do and loop overhead can't always be discounted, even in sparse linear algebra.

liruipeng: I realized the whole `hypre_Iterator` thing is not necessary, so I removed it.

rfalgout: Okay, looks good @liruipeng .  BTW, although removing the iterator looks like a better way to go in this case, note that I am not against the idea of iterators in general, nor the use of 'static inline' functions to define them.  I only wanted to be cautious about potential performance implications, which may have been nothing in the end, as you argued.  Thanks!

victorapm (commented):
Thank you for the reorganization, @liruipeng! I think it looks great. 

I was wondering if we could go further and move some contents of `ams.c` to other files, e.g., move `hypre_ParCSRMatrixSetDiagRows` to `par_csr_matrix.c`. There are a few functions in `ams.c` that operate exclusively on hypre_ParCSRMatrix and hypre_ParVector so I think it makes sense to have them closer to the data structures that they operate on. Let me know your opinion on this :)

Thank you!

liruipeng:
> Thank you for the reorganization, @liruipeng! I think it looks great.
> 
> I was wondering if we could go further and move some contents of `ams.c` to other files, e.g., move `hypre_ParCSRMatrixSetDiagRows` to `par_csr_matrix.c`. There are a few functions in `ams.c` that operate exclusively on hypre_ParCSRMatrix and hypre_ParVector so I think it makes sense to have them closer to the data structures that they operate on. Let me know your opinion on this :)
> 
> Thank you!

@victorapm Thanks! I feel these AMS routines may not be general enough to move to `par_csr_mv`. Still leave them there. Maybe we can revisit this later.

victorapm (approved):

src/parcsr_ls/par_relax.c
liruipeng: Thanks @rfalgout. There was actually performance implications, not in the `hypre_iterator` per se, but in the overall strategy of putting all cases together in one piece of code. So, I reverted some of my effort to re-make a bunch of separate `inline` functions for different cases put in a new file `par_relax.h`. The desired refactored `hypre_BoomerAMGRelaxHybridGaussSeidel_core` still keeps a nice shape. I've tested the correctness and the efficiency, which look fine. There was no noticeable performance loss, while in some cases it could be even slightly faster than before.

liruipeng:
Hi @ulrikeyang @oseikuffuor1 Please take a look at this PR when you have time. The correctness and performance of the tests after the changes look fine. Let me know if you want to chat about it. Thanks!

src/parcsr_ls/par_relax.c
```diff
@@ -14,4397 +14,1398 @@
...
+   {
+      // XXX GPU calls Relax7 XXX
+      return hypre_BoomerAMGRelax7Jacobi(A, f, relax_points, relax_weight, l1_norms, u, Vtemp);
+   }
+   else
 #endif
-            for (i = 0; i < n; i++)
-            {
+   {
+      /* in the case of non-CF, use relax-7 which is faster */
```
ulrikeyang, line 323: Are you sure about this? In my experience 0 is faster than 7 on the CPU.

liruipeng: @ulrikeyang This is to enforce consistency with the old code. Non-CF `relax-type == 18` used to call Jacobi in `hypre_ParCSRRelax` in `ams.c` that uses Matvec, which is the same as `relax-type == 7`. Also, 0 doesn't use L1-norm. Thanks!

ulrikeyang (approved):
Looks good to me.

oseikuffuor1 (approved):
Thanks @liruipeng! This looks good to me.

PR #243, opened 2020-12-10
Umpire

Ram3sh:
This PR contains the support of UMPIRE pooling allocators for host and GPU memory. Configure hypre with `--with-umpire`, device and uvm allocations and deallocations are done with umpire, whereas host pool is not enabled by default.

This PR also includes some other minor changes:
1. Adding `.gitignore` to the repo
2. Removing all `malloc/calloc/realloc/free` and regression testing on finding them
3. No longer compile `ij.c` with C++ compiler. It goes back to a C code now.
4. Introducing `HYPRE_USING_GPU`, which is equivalent to `HYPRE_USING_CUDA || HYPRE_USING_DEVICE_OPENMP`
5. Adding a few user-level interfaces: `HYPRE_SetMemoryLocation`, `HYPRE_SetExecutionPolicy`, `HYPRE_SetGPUMemoryPoolSize` and `HYPRE_CSRMatrixSetSpGemmUseCusparse`

## Configure hypre with the following options to enable GPU (device and UM) memory pool

```
./configure --with-cuda --enable-unified-memory --enable-debug --with-umpire --with-umpire-include=/g/g92/li50/workspace/umpire-5.0.0/install_xlC/include/ --with-umpire-lib-dirs=/g/g92/li50/workspace/umpire-5.0.0/install_xlC/lib --with-umpire-libs=umpire
```
after installing UMPIRE by something like
```
git clone https://github.com/LLNL/Umpire.git

cmake3 -DCMAKE_INSTALL_PREFIX=../install_xlC -DENABLE_CUDA=On -DENABLE_OPENMP=Off -DCMAKE_CXX_COMPILER=xlC -DCMAKE_C_COMPILER=xlc -DENABLE_CUDA=On -DCMAKE_CUDA_FLAGS="-arch sm_70" -DENABLE_C=On ../
```


.gitignore
```diff
@@ -0,0 +1,7 @@
+*.o
```
rfalgout, line 1: We have traditionally not included a '.gitignore' file in the repo.  It looks like many argue that it should be included.  Opinions?

Ram3sh: When you add new files, it is very easy to miss them when you do a git status and it is 2 pages long. With a good .gitignore only the files that matter will show up.

liruipeng: I generally don't like it in the repo, since I have my own in home directory that includes particular ignored files such as vi swap files. Having this one will override it. Just my preference.

rfalgout: I also have one in my home directory, but there is a decent argument for including one in the repo as well.  Especially for weird non-standard files that the library generates (we have our '.obj' files, for example), or files that you want to ensure people don't accidentally commit.

Ram3sh: Having both means a union of the local and global sets will be applied right ?

rfalgout: Not sure yet.  There are some precedence rules, but I need to read this again: https://git-scm.com/docs/gitignore

src/distributed_ls/ParaSails/DiagScale.c
```diff
@@ -192,7 +192,7 @@ DiagScale *DiagScaleCreate(Matrix *A, Numbering *numb)
         requests, replies_list);
 
     num_replies = FindNumReplies(A->comm, replies_list);
-    free(replies_list);
```
liruipeng, line 195: @Ram3sh I notice that you changed a bunch of `free` to `hypre_TFree`, which is correct. But there are still a number of places where `free` is used in conjunction with `hypre_TAlloc`. It is not a big problem with standard  `malloc/free` but with umpire it can be, correct? (allocating with umpire but deallocating with `free`... Is it a problem?).

@rfalgout does it make sense to add regression tests searching for all instances of using `malloc`/`calloc`/`realloc`/`free` and throw errors?

Ram3sh: I added this when testing host pools. Freeing an Umpire allocation using free is of course very bad. Umpire host pools are very expensive and nobody uses it AFAIK. The Umpire team is working on making it faster, so I have left the code in there for use in the future.

rfalgout: I was wondering about these changes myself.  Clearly we didn't change every 'free' in hypre, and I don't remember why (maybe there wasn't actually a reason).  I think it would be fine to change all of it and write a regression test to check that it stays that way.  I would be happy to write it.

liruipeng: "very bad" means it would crash, or would screw up umpire's memory pooling algorithms but can still proceed? But in either case, I consider using `free` with `hypre_TAlloc` is a mistake, and we should fix it. Thanks, @Ram3sh

Ram3sh: Segfault since you will be freeing an allocation somewhere in the middle.

liruipeng: > I was wondering about these changes myself. Clearly we didn't change every 'free' in hypre, and I don't remember why (maybe there wasn't actually a reason). I think it would be fine to change all of it and write a regression test to check that it stays that way. I would be happy to write it.

@rfalgout sounds good. Please do, and I will make the code changes. Thanks!

src/test/runall
```diff
@@ -0,0 +1,14 @@
+./runtest.sh TEST_ams/*.sh
```
rfalgout: This file should not be committed.

src/test/runsome
```diff
@@ -0,0 +1,6 @@
+./runtest.sh TEST_ams/*.sh
```
rfalgout: This file should not be committed.

.gitignore
rfalgout: Okay, I just did some more reading on this.  It appears that a union of gitignore patterns will happen in most cases, as speculated by @Ram3sh , but there is a prefix '!' that negates (un-ignores) a previous pattern, so having a precedence rule for the different gitignore files becomes important.  The local file takes precedence.

We could also add the following to this

    *.so
    config.log
    config.status

Given this, what do you think @liruipeng ?

liruipeng: @rfalgout Thank you for researching this. Given this and the fact that I don't have any negated patterns so far in my local one, I think it's safe to have a global hypre's `.gitignore` (at least for me). The addings you want to make look good to me. Thanks!

rfalgout: FYI, I just confirmed this.  We probably should also add the following, since we don't want them commited accidentally:

    Makefile.config
    HYPRE_config.h

rfalgout: I will make the change.  Thanks!

oseikuffuor1: @rfalgout there's also a autom4te.cache folder that is generated if the (autconf) build system is modified and a new configure file is generated. Should this make it into the .gitignore file as well?

rfalgout: Just added it.  I had it in my global gitignore file, but I couldn't remember what it was for.  Thanks!

src/distributed_ls/ParaSails/DiagScale.c
rfalgout: Just added a memory usage test that checks for 'malloc', 'calloc', and 'free'.  Let me know if you think I have missed something.  Thanks!

liruipeng:
@Ram3sh 

When running the following on 1 node of ray
```
mpirun -np 4 ./ij -n 256 256 128 -P 2 2 1       -pmis -keepT 1 -rlx 18        -exec_device -rap 1 -mod_rap2 1 -mm_cusparse 0 -interptype  3 -solver 1
```
I got seg-fault
```
terminate called after throwing an instance of 'umpire::util::Exception'
  what():  ! Umpire Exception [/g/g92/li50/workspace/umpire-5.0.0/src/umpire/alloc/CudaMallocAllocator.hpp:37]:  allocate cudaMalloc( bytes = 4294967312 ) failed with error: out of memory
```
Some rank failed to allocated 4GB GPU memory with Umpire. Note that hypre does `cudaSetDevice` based on the MPI-rank and the number of GPUs it sees, so each rank here got its own GPU. So, there should be enough memory for each of them. Interestingly, adding `mpibind` to the above command (or running with `lrun`) can fix the issue, where each rank only sees 1 GPU instead. 

Does it seem to be an Umpire issue with `cudaSetDevice`, i.e., all ranks attempted to allocated on the same GPU in the first case even if they set different devices?

Thanks!

-Ruipeng

Ram3sh:
> @Ram3sh
> 
> When running the following on 1 node of ray
> 
> ```
> mpirun -np 4 ./ij -n 256 256 128 -P 2 2 1       -pmis -keepT 1 -rlx 18        -exec_device -rap 1 -mod_rap2 1 -mm_cusparse 0 -interptype  3 -solver 1
> ```
> 
> I got seg-fault
> 
> ```
> terminate called after throwing an instance of 'umpire::util::Exception'
>   what():  ! Umpire Exception [/g/g92/li50/workspace/umpire-5.0.0/src/umpire/alloc/CudaMallocAllocator.hpp:37]:  allocate cudaMalloc( bytes = 4294967312 ) failed with error: out of memory
> ```
> 
> Some rank failed to allocated 4GB GPU memory with Umpire. Note that hypre does `cudaSetDevice` based on the MPI-rank and the number of GPUs it sees, so each rank here got its own GPU. So, there should be enough memory for each of them. Interestingly, adding `mpibind` to the above command (or running with `lrun`) can fix the issue, where each rank only sees 1 GPU instead.
> 
> Does it seem to be an Umpire issue with `cudaSetDevice`, i.e., all ranks attempted to allocated on the same GPU in the first case even if they set different devices?
> 
> Thanks!
> 
> -Ruipeng

Yes, Umpire assumes that you are using device 0. There is a mechanism to use more than 1 device but it is kludgy. Since the device setting will be already done by the time Hypre is called, I haven't done anything about it.

liruipeng:
@Ram3sh Thank you for confirming this. Just want to clarify: hypre doesn't use more than 1 GPUs per MPI-rank. But in the cases of multiple GPUs, each rank gets 1 GPU in a round-robin way, so it is not always GPU 0. In this case, what the users can do if they don't use `mpibind` or `lrun`?

Ram3sh:
> @Ram3sh Thank you for confirming this. Just want to clarify: hypre doesn't use more than 1 GPUs per MPI-rank. But in the cases of multiple GPUs, each rank gets 1 GPU in a round-robin way, so it is not always GPU 0. In this case, what the users can do if they don't use `mpibind` or `lrun`?

As long as calling code creates Umpire pools called HYPRE_UM_POOL and HYPRE_DEVICE_POOL hypre wil use those pools. If they don't exist, the pools are created in hypre_Init. I am doing this to make sure all the tests run. Whether Umpire should be used by Hypre if the host code is not using it is something we need to discuss and decide. On the one hand it is faster but the memory locked in the pool is no longer available to the host code.

liruipeng:
> > @Ram3sh Thank you for confirming this. Just want to clarify: hypre doesn't use more than 1 GPUs per MPI-rank. But in the cases of multiple GPUs, each rank gets 1 GPU in a round-robin way, so it is not always GPU 0. In this case, what the users can do if they don't use `mpibind` or `lrun`?
> 
> As long as calling code creates Umpire pools called HYPRE_UM_POOL and HYPRE_DEVICE_POOL hypre wil use those pools. If they don't exist, the pools are created in hypre_Init. I am doing this to make sure all the tests run. Whether Umpire should be used by Hypre if the host code is not using it is something we need to discuss and decide. On the one hand it is faster but the memory locked in the pool is no longer available to the host code.

@Ram3sh That's a good point! We should talk about it for sure. But I am not sure if it is clear for me to understand the "device 0" question asked earlier. For non-LLNL users who may not have `mpibind` or `lrun`, if hypre is run on a multi-GPU node, some rank has to use a device other than device 0. If UMPIRE cannot do this, how should they handle this case? Thanks!

Ram3sh:
If the host code sets CUDA_VISIBLE_DEVICES using a script , all is well. If the host code attaches to different GPUs based on rank in the MPI_SHARED_COMM, then they will need to initialize Umpire using a slightly different input parameter. But as long as they alias the pools using HYPRE_DEVICE_POOL and HYPRE_UM_POOL, Hypre will be use the Umpire pools without any problems.

Right now, if the host code **doesn't** create Umpire pools and Hypre is compiled with Umpire support, the pools are created in Hypre_Init. This pool creation assumes lrun or mpibind is being used. We can make it so that this emits an error message and exits.

liruipeng:
Hi, @Ram3sh 

I made some changes and refactorings to this PR. Main changes are:

1. In `hypre_handle`  (https://github.com/hypre-space/hypre/blob/511960a649df9bbc6086e456039bd62b83f276c6/src/utilities/hypre_handle.h#L29), we put `umpire_*_pool_name`, `umpire_*_pool_size`, and `umpire_resourcemanager` to have global access of these data, which are initialized in `hypre_UmpireInit` (https://github.com/hypre-space/hypre/blob/511960a649df9bbc6086e456039bd62b83f276c6/src/utilities/hypre_general.c#L243);
2. After `HYPRE_Init`, users can set the pool names and sizes (for using existing pools outside hypre) by API calls such as `HYPRE_SetUmpireDevicePoolName` and `HYPRE_SetUmpireDevicePoolSize`;
3. Pool initialization is removed from `HYPRE_Init` and postponed to the first allocation (see `hypre_memory.c`)
4. C++ `hypre_umpire_allocator` is hidden in (https://github.com/hypre-space/hypre/blob/511960a649df9bbc6086e456039bd62b83f276c6/src/utilities/hypre_cuda_utils.h#L95). The `HYPRE_THRUST_CALL` is changed so that we don't need to define `hypre_umpire_allocator ualloc;` before thrust calls.

Questions:
1. Do we need to finalize UMPIRE in `HYPRE_Finalize()`? It is currently commented out (https://github.com/hypre-space/hypre/blob/511960a649df9bbc6086e456039bd62b83f276c6/src/utilities/hypre_general.c#L207). Apparently, we don't want to release user-input pools. But should we do it for the pools created by hypre?
2. Is `umpire_resourcemanager` a unique global object in UMPIRE, or there can be many instances? Do users want to pass in their `umpire_resourcemanager`?

Please take a look and let me know if these make sense.

Thanks!

liruipeng:
Hi, @Ram3sh 

There seem to be seg-fault problems with the code. I rolled back the branch to your last commit 
```
commit 7eca6e14f353125a0f6ef7ad1e2009268f9131e8 (HEAD)
Author: Ramesh Pankajakshan <pankajakshan1@llnl.gov>
Date:   Tue Dec 8 16:13:31 2020 -0800

    Some cleanup
```

I am not able to run this on ray
```
mpirun -n 2  mpibind cuda-memcheck ./ij -n 32 32 1       -pmis -keepT 1 -rlx 18        -exec_device -rap 1 -mod_rap2 1 -mm_cusparse 0 -interptype  3 -solver 1
```

Ram3sh:
I think this is due today’s change in lrun. It runs fine with 4 ranks on Ansel.

Ramesh

Ram3sh:
To confirm , rerun the case after setting env variable LLNL_USE_OLD_CUDA_VISIBLE_DEVICES_FORMAT to 1.

Ramesh

liruipeng:
@Ram3sh 

> To confirm , rerun the case after setting env variable LLNL_USE_OLD_CUDA_VISIBLE_DEVICES_FORMAT to 1. Ramesh
> […](#)
> On Jan 13, 2021, at 5:40 PM, Ramesh Pankajakshan ***@***.***> wrote: ﻿ I think this is due today’s change in lrun. It runs fine with 4 ranks on Ansel. Ramesh >> On Jan 13, 2021, at 4:35 PM, Ruipeng Li ***@***.***> wrote: >> > ﻿ > Hi, @Ram3sh > > There seem to be seg-fault problems with the code. I rolled back the branch to your last commit > > commit [7eca6e1](https://github.com/hypre-space/hypre/commit/7eca6e14f353125a0f6ef7ad1e2009268f9131e8) (HEAD) > Author: Ramesh Pankajakshan ***@***.***> > Date: Tue Dec 8 16:13:31 2020 -0800 > > Some cleanup > I am not able to run this on ray > > mpirun -n 2 mpibind cuda-memcheck ./ij -n 32 32 1 -pmis -keepT 1 -rlx 18 -exec_device -rap 1 -mod_rap2 1 -mm_cusparse 0 -interptype 3 -solver 1 > — > You are receiving this because you were mentioned. > Reply to this email directly, view it on GitHub, or unsubscribe.

OK Thanks! The regression tests look all good now.

rfalgout (approved):
This all looks good to me.  Thanks @Ram3sh and @liruipeng !

Ram3sh:
> Hi, @Ram3sh
> 
> I made some changes and refactorings to this PR. Main changes are:
> 
>     1. In `hypre_handle`  (https://github.com/hypre-space/hypre/blob/511960a649df9bbc6086e456039bd62b83f276c6/src/utilities/hypre_handle.h#L29
>        ), we put `umpire_*_pool_name`, `umpire_*_pool_size`, and `umpire_resourcemanager` to have global access of these data, which are initialized in `hypre_UmpireInit` (https://github.com/hypre-space/hypre/blob/511960a649df9bbc6086e456039bd62b83f276c6/src/utilities/hypre_general.c#L243
>        );
> 
>     2. After `HYPRE_Init`, users can set the pool names and sizes (for using existing pools outside hypre) by API calls such as `HYPRE_SetUmpireDevicePoolName` and `HYPRE_SetUmpireDevicePoolSize`;
> 
>     3. Pool initialization is removed from `HYPRE_Init` and postponed to the first allocation (see `hypre_memory.c`)
> 
>     4. C++ `hypre_umpire_allocator` is hidden in (https://github.com/hypre-space/hypre/blob/511960a649df9bbc6086e456039bd62b83f276c6/src/utilities/hypre_cuda_utils.h#L95
>        ). The `HYPRE_THRUST_CALL` is changed so that we don't need to define `hypre_umpire_allocator ualloc;` before thrust calls.
> 
> 
> Questions:
> 
>     1. Do we need to finalize UMPIRE in `HYPRE_Finalize()`? It is currently commented out (https://github.com/hypre-space/hypre/blob/511960a649df9bbc6086e456039bd62b83f276c6/src/utilities/hypre_general.c#L207
>        ). Apparently, we don't want to release user-input pools. But should we do it for the pools created by hypre?
> 
>     2. Is `umpire_resourcemanager` a unique global object in UMPIRE, or there can be many instances? Do users want to pass in their `umpire_resourcemanager`?
> 
> 
> Please take a look and let me know if these make sense.
> 
> Thanks!

1. It needs to be called iff Hypre creates the Umpire pools. This brings up the bigger question of whether we should be creating Umpire pools if Hypre is compiled with Umpire support but the calling code is either not using them or not providing the right aliases. 
2. Yes the resource manager is a singleton.

The lazy pool initialization seems to add some overhead. PCG setup now takes 0.82s compared to the original 0.79s. Seems a bit on the high side. 
The fix for the Umpire problem above with 2 ranks-per-node is to use cudaGetDevice and change the umpire device to DEVICE::X where X is the device being use.

liruipeng:
@Ram3sh 

> 1. It needs to be called iff Hypre creates the Umpire pools. This brings up the bigger question of whether we should be creating Umpire pools if Hypre is compiled with Umpire support but the calling code is either not using them or not providing the right aliases.

The current design is that if hypre is configured with Umpire, we assume user wants to use Umpire pools for GPUs. If user doesn't provide pool name, hypre creates its default (HYPRE_DEVICE_POOL, HYPRE_UM_POOL). To use existing pools, user needs to provide the pool name through hypre's API. I will add some mechanism to release Umpire pools created by hypre.

> 2. Yes the resource manager is a singleton.
> 
> The lazy pool initialization seems to add some overhead. PCG setup now takes 0.82s compared to the original 0.79s. Seems a bit on the high side.

I won't worry about it if this is only a one-time cost and needs to be spent somewhere. The lazy initialization moved the cost from `HYPRE_Init` to the first setup call.

> The fix for the Umpire problem above with 2 ranks-per-node is to use cudaGetDevice and change the umpire device to DEVICE::X where X is the device being use.

I am not sure what this means. Do I need to change hypre's code? The  `LLNL_USE_OLD_CUDA_VISIBLE_DEVICES_FORMAT` works and I assume LC will fix it.

Ram3sh:
When Hypre is run with 2 ranks per node, it calls the round-robin code for device binding. Rank 1 grabs device 1 but Umpire always binds to device 0 and this causes the fault. The fix is to pass DEVICE::1 to Umpire instead of DEVICE. The env varibale is fine for now.

liruipeng:
Hi @davidbeckingsale Thanks for agreeing to review this PR that adds Umpire support to hypre. Please take a look at the following items:

1. The Umpire related data structures are stored in hypre's global object `hypre_handle`, which include the names, pool sizes, and ownership of host, device, um, and pinned pools, and also `umpire_resourcemanager` (is it pointer?) See, their declarations at https://github.com/hypre-space/hypre/blob/af9bf7363814ee5f3cc0cd4710d39eb9c54cf863/src/utilities/hypre_handle.h#L30.
2. The the above Umpire data are initialized to the default values in `hypre_UmpireInit` called by `HYPRE_Init` that is the first HYPRE function user should call. See https://github.com/hypre-space/hypre/blob/af9bf7363814ee5f3cc0cd4710d39eb9c54cf863/src/utilities/hypre_general.c#L241. The interface to (re)set the values are located in the same file. See the code that follows this line https://github.com/hypre-space/hypre/blob/af9bf7363814ee5f3cc0cd4710d39eb9c54cf863/src/utilities/hypre_general.c#L312
3. The creation of pooling allocator is done at the first allocation of host, device, um, or pinned memory, if it doesn't exist in `resourcemanager`, and in this case, hypre is assumed to **own** this allocator, which will be released by hypre. See, code from line
https://github.com/hypre-space/hypre/blob/af9bf7363814ee5f3cc0cd4710d39eb9c54cf863/src/utilities/hypre_memory.c#L1003 for all the `malloc`, `free` and `realloc` with Umpire. 
4. In `HYPRE_Finalize`, `hypre_UmpireFinalize` is called to release the pools that hypre owns. See, https://github.com/hypre-space/hypre/blob/af9bf7363814ee5f3cc0cd4710d39eb9c54cf863/src/utilities/hypre_general.c#L268
5. For C++ CUDA thrust, device pooling allocator class is defined in https://github.com/hypre-space/hypre/blob/umpire/src/utilities/hypre_umpire_allocator.h, which is given to thrust via `HYPRE_THRUST_CALL` macro. See https://github.com/hypre-space/hypre/blob/af9bf7363814ee5f3cc0cd4710d39eb9c54cf863/src/utilities/hypre_cuda_utils.h#L235

Please let us know if these look fine to you, or if you have any questions and comments. Thanks a lot!

-Ruipeng

liruipeng (approved):

davidbeckingsale:
Hey @liruipeng, I think this looks good.

> 1. The Umpire related data structures are stored in hypre's global object `hypre_handle`, which include the names, pool sizes, and ownership of host, device, um, and pinned pools, and also `umpire_resourcemanager` (is it pointer?) See, their declarations at https://github.com/hypre-space/hypre/blob/af9bf7363814ee5f3cc0cd4710d39eb9c54cf863/src/utilities/hypre_handle.h#L30

Perfect. The ResourceManager is actually a reference (since it's a singleton), but everything in the Umpire C API is wrapped inside structs. How you are storing and using it looks fine here.

> 2. The the above Umpire data are initialized to the default values in `hypre_UmpireInit` called by `HYPRE_Init` that is the first HYPRE function user should call. See https://github.com/hypre-space/hypre/blob/af9bf7363814ee5f3cc0cd4710d39eb9c54cf863/src/utilities/hypre_general.c#L241
> . The interface to (re)set the values are located in the same file. See the code that follows this line https://github.com/hypre-space/hypre/blob/af9bf7363814ee5f3cc0cd4710d39eb9c54cf863/src/utilities/hypre_general.c#L312

Looks good to me.

> 3. The creation of pooling allocator is done at the first allocation of host, device, um, or pinned memory, if it doesn't exist in `resourcemanager`, and in this case, hypre is assumed to **own** this allocator, which will be released by hypre. See, code from line
> https://github.com/hypre-space/hypre/blob/af9bf7363814ee5f3cc0cd4710d39eb9c54cf863/src/utilities/hypre_memory.c#L1003
> for all the `malloc`, `free` and `realloc` with Umpire.

I like this. Handling both creating hypre-owned allocators or allowing the use of an allocator with that name that already exists is useful, and this a similar pattern used in other applications and libraries (e.g. SAMRAI).

> 4. In `HYPRE_Finalize`, `hypre_UmpireFinalize` is called to release the pools that hypre owns. See, https://github.com/hypre-space/hypre/blob/af9bf7363814ee5f3cc0cd4710d39eb9c54cf863/src/utilities/hypre_general.c#L268

👍 

> 5. For C++ CUDA thrust, device pooling allocator class is defined in https://github.com/hypre-space/hypre/blob/umpire/src/utilities/hypre_umpire_allocator.h, which is given to thrust via `HYPRE_THRUST_CALL` macro. See https://github.com/hypre-space/hypre/blob/af9bf7363814ee5f3cc0cd4710d39eb9c54cf863/src/utilities/hypre_cuda_utils.h#L235

Awesome. @Ram3sh said this was a big performance win, I'm glad it's working.

> Please let us know if these look fine to you, or if you have any questions and comments. Thanks a lot!

Everything here looks fine. Please let me know if you have any more questions.

ulrikeyang (approved):
Looks good as far as I can tell. Thank you.

liruipeng:
@davidbeckingsale Thanks again for your review, which is very helpful! We will have Umpire support in hypre!

The regression tests looked fine. Need one more approval from @oseikuffuor1, and it should be good to go. @Ram3sh @rfalgout Please take a look at the recent changes, if needed. Thanks!

liruipeng:
> When Hypre is run with 2 ranks per node, it calls the round-robin code for device binding. Rank 1 grabs device 1 but Umpire always binds to device 0 and this causes the fault. The fix is to pass DEVICE::1 to Umpire instead of DEVICE. The env varibale is fine for now.

@Ram3sh I missed this comment from you for some reason. Sorry about that.

By `DEVICE`, did you mean the `"DEVICE"` string at this line, https://github.com/hypre-space/hypre/blob/bfe99dacfa341ecf2d2f9f1c8d9e02f8bb45ad7a/src/utilities/hypre_memory.c#L1072? I don't recognize the syntax of `DEVICE::1`. Is it C++, or a string? What is the general solution to set any device, `DEVICE::X`? I would like to fix the device number problem in hypre, with Umpire and without.

Thanks!

oseikuffuor1 (approved):
Looks good to me. Thanks Ruipeng and Ramesh.

Ram3sh:
> > When Hypre is run with 2 ranks per node, it calls the round-robin code for device binding. Rank 1 grabs device 1 but Umpire always binds to device 0 and this causes the fault. The fix is to pass DEVICE::1 to Umpire instead of DEVICE. The env varibale is fine for now.
> 
> @Ram3sh I missed this comment from you for some reason. Sorry about that.
> 
> By `DEVICE`, did you mean the `"DEVICE"` string at this line,
> 
> https://github.com/hypre-space/hypre/blob/bfe99dacfa341ecf2d2f9f1c8d9e02f8bb45ad7a/src/utilities/hypre_memory.c#L1072
> ? I don't recognize the syntax of `DEVICE::1`. Is it C++, or a string? What is the general solution to set any device, `DEVICE::X`? I would like to fix the device number problem in hypre, with Umpire and without.
> 
> Thanks!

It is a string. But the whole discussion is moot in light of our offline discussion about binding. HYPRE_Init should use the binding it inherits from the host code. Most codes are not doing any internal binding, so device 0 will work. So a cudaGetDevice check and a change to the DEVICE::X if it is not 0 should suffice.

liruipeng:
> moot

@Ram3sh Thanks again for explaining it. I will make another PR addressing the binding issue and make the Umpire changes in it.

PR #246, opened 2020-12-22
fix GPU SpMV for zero matrices

liruipeng:
This PR fixes the issue with zero-sized matrix SpMV on GPUs.


rfalgout (approved):
Looks good.  Thanks!

PR #250, opened 2021-01-04
fixed syntax error with --enable-gpu-aware-mpi

liruipeng:
This PR fixes a syntax error with `--enable-gpu-aware-mpi`


rfalgout (approved):
Looks good

ulrikeyang (approved):

victorapm (commented):
Thanks, Ruipeng! Should we erase `// TODO RL` in line 415 since it works now?

liruipeng:
> Thanks, Ruipeng! Should we erase `// TODO RL` in line 415 since it works now?

I left that comment there since it is still unclear whether we need to sync. CUDA stream before doing (GPU-)MPI. We may need to revisit it at some time later. Thanks @victorapm

victorapm (approved):
Thanks for clarifying, Ruipeng!

Issue #249, opened 2020-12-31
ParaSails setup hangs


slmcbane:
I am using ParaSails for a smoother inside BoomerAMG and found a reproducible, non-deterministic hang inside `PairaSailsSetupPattern`. The hang happens at an allreduce. I've put a representative backtrace below. [Here is a link](https://u.pcloud.link/publink/show?code=XZRPmQXZQOoequ8Y14H2su26YQE6Ey9L2KBk) to a reproducing example; it expects to be run as `mpirun -np 64 ./a.out NUM_ITERATIONS` and repeats the setup of a solver with the same preconditioning setup as I used `NUM_ITERATIONS` times. For me, this eventually hangs after a **non-deterministic** number of iterations - usually in the hundreds but sometimes quickly. The bug occurs more frequently with more processes, hence the use of 64 in a simple example. Unfortunately, I don't have a system handily available to test with a different MPI implementation.

Here is a backtrace:
```
#0  0x00007feff1414095 in ?? () from /usr/lib/x86_64-linux-gnu/libopen-pal.so.20
#1  0x00007feff13b89a9 in opal_progress () from /usr/lib/x86_64-linux-gnu/libopen-pal.so.20
#2  0x00007feff365f3f5 in ompi_request_default_wait_all () from /usr/lib/x86_64-linux-gnu/libmpi.so.20
#3  0x00007feff36afcb7 in ompi_coll_base_allreduce_intra_recursivedoubling ()
   from /usr/lib/x86_64-linux-gnu/libmpi.so.20
#4  0x00007feff366f47a in PMPI_Allreduce () from /usr/lib/x86_64-linux-gnu/libmpi.so.20
#5  0x00007feff3bd4414 in FindNumReplies ()
   from /workspace/sean/spack/opt/spack/linux-ubuntu18.04-zen/gcc-7.4.0/hypre-2.20.0-7mesmzmw25yvt2reaxs2zn4htdhht7b5/lib/libHYPRE-2.20.0.so
#6  0x00007feff3bd48d5 in ParaSailsSetupPattern ()
   from /workspace/sean/spack/opt/spack/linux-ubuntu18.04-zen/gcc-7.4.0/hypre-2.20.0-7mesmzmw25yvt2reaxs2zn4htdhht7b5/lib/libHYPRE-2.20.0.so
#7  0x00007feff3bd7e88 in hypre_ParaSailsSetup ()
[...]

PR #251, opened 2021-01-08
Modification to fix error code warning for ILU-NSH coarse level solver

oseikuffuor1:
This PR fixes error code warnings for the NSH Schur solver for hypre-ILU. This addresses an issue brought up in #249

### DIFF
--- a/src/parcsr_ls/par_ilu.c
+++ b/src/parcsr_ls/par_ilu.c
@@ -124,7 +124,7 @@ hypre_ILUCreate()
 
    /* -> SCHUR-NSH */
    hypre_ParILUDataSchurNSHSolveMaxIter(ilu_data) = 5;
-   hypre_ParILUDataSchurNSHSolveTol(ilu_data) = 1e-02;
+   hypre_ParILUDataSchurNSHSolveTol(ilu_data) = 0.0;
    hypre_ParILUDataSchurNSHDroptol(ilu_data) = NULL;/* this is not the default option, set it only when switched to */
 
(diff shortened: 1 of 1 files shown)


rfalgout (approved):
I don't really understand why changing a default should fix an error code issue, but if you think this is the right default value and it works, then I'm good with it.  Thanks!

oseikuffuor1:
@falgout it really just makes the solve on the Schur solve depend on only the max. number of iterations. The alternative is to avoid throwing the code when the maximum number of iterations is reached without reaching the tolerance. Think preconditioner for a Krylov method. We typically set the number of iterations to something small, with the tolerance set to zero. In the case of NSH as the Schur solver, we typically only do a few iterations so controlling convergence with max. iters should be fine.

rfalgout:
Hi @oseikuffuor1 .  My comment was mostly about this being a default value.  The default value for our solvers is not zero, it must be set to zero when used as a preconditioner.  Anyway, it sounds like these are the best defaults anyway, so I'm good with this.

oseikuffuor1:
@rfalgout yes you are correct. The default the for NSH solver tolerance is actually nonzero (See [here](https://github.com/hypre-space/hypre/blob/a6c852be523e8f0ff56d0a24f3d2de161384bd6d/src/parcsr_ls/par_ilu.c#L132)). Only when it is used as the Schur solver for ILU do we set it to zero through the default value of the ILU data structure.

rfalgout:
Perfect, thanks again!

liruipeng (approved):
Looks good to me. Thank you @oseikuffuor1

PR #256, opened 2021-01-13
triangular solve on GPUs; runcheck

liruipeng:
This PR fixes the following 3 issues

1. Incorrect memory location of `hypre_ParCSRMatrixRowindices(A)` and `hypre_ParCSRMatrixRowvalues(A)` after `hypre_ParCSRMatrixMigrate`

2. Inf/NaN problem of Gauss-Seidel (G-S) on GPUs using cuSPARSE, when having zero diagonal entries. For forward G-S,   is updated as:  i} A_{ij}x_j^{old}) / A_{ii}">. In the case that , we let . For the GPU implementation, we took a different formulation that can utilize sparse kernels, i.e., . L is stored separately from A. cuSPARSE triangular solve routine gives Inf/NaN when encountering zero pivots. In order to avoid it and have  with zero pivot i, the fix is to set  where  . The code is at https://github.com/hypre-space/hypre/blob/5e3c8876d17b218bb8a772aff6b3040f531209e1/src/seq_mv/csr_matop_device.c#L1299

3. `runcheck.sh` not being able to catch lines without numerical values

### DIFF
--- a/AUTOTEST/machine-lassen.sh
+++ b/AUTOTEST/machine-lassen.sh
@@ -38,7 +38,7 @@ shift
 # Basic build and run tests
 mo="-j test"
 eo=""
-roij="-ij -ams -rt -mpibind -rtol 1e-3 -atol 8e-3"
+roij="-ij -ams -rt -mpibind -rtol 1e-3 -atol 2e-2"
 ross="-struct -sstruct -rt -mpibind -rtol 1e-6 -atol 1e-6"
 rost="-struct -rt -mpibind -rtol 1e-8 -atol 1e-8"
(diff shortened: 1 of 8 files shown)


rfalgout (approved):
Looks good to me.  Thanks!

oseikuffuor1 (approved):

oseikuffuor1:
Looks good to me too. Thanks.

ulrikeyang (approved):

liruipeng:
I would like to wait for @oseikuffuor1 's new `runcheck.sh` before merging it.

PR #260, opened 2021-01-21
Fix MGR setup on GPU. 

oseikuffuor1:
This PR fixes an issue with the memory location for extracting submatrices to compute the non Galerkin Coarse Grid option for MGR when hypre is built with GPU options enabled.


rfalgout (approved):
Looks good to me.  Thanks!

src/parcsr_ls/par_mgr.c
```diff
@@ -2147,14 +2148,14 @@ hypre_MGRComputeNonGalerkinCoarseGrid(hypre_ParCSRMatrix    *A,
   //{
     if (ordering == 0) // interleaved ordering
     {
-      HYPRE_Int *A_h_correction_diag_i_new = hypre_CTAlloc(HYPRE_Int, n_local_cpoints+1, HYPRE_MEMORY_HOST);
-      HYPRE_Int *A_h_correction_diag_j_new = hypre_CTAlloc(HYPRE_Int, (bsize + max_elmts)*n_local_cpoints, HYPRE_MEMORY_HOST);
-      HYPRE_Complex *A_h_correction_diag_data_new = hypre_CTAlloc(HYPRE_Complex, (bsize + max_elmts)*n_local_cpoints, HYPRE_MEMORY_HOST);
+      HYPRE_Int *A_h_correction_diag_i_new = hypre_CTAlloc(HYPRE_Int, n_local_cpoints+1, memory_location);
```
liruipeng, line 2141: I assume your original matrix A is on GPU, so you want to allocate `A_h_correction_diag_*_new` on the same memory space. In what's following, you operate these arrays on host. This should work but may not efficiently. If you are not going to port this part to GPU, you may consider allocating CPU memory, doing the computation, and copying to GPU at the end.

liruipeng: You may also consider using `hypre_ParCSRMatrixRAPKT` and `hypre_ParCSRMatMat` instead of `hypre_ParMatmul` and `hypre_BoomerAMGBuildCoarseOperator` for GPUs, since they run on the host.

oseikuffuor1: @liruipeng does the memory_location variable not take care of this, since it takes its value from the memory location of A?

liruipeng: @oseikuffuor1 `memory_location == HYPRE_MEMORY_DEVICE` for GPU build, which is actually `hypre_MEMORY_UNIFIED` if you configure with UVM. This is how it can work I suppose.

oseikuffuor1: @liruipeng ok so are you suggesting explicitly using HYPRE_MEMORY_DEVICE for the GPU build to address both UVM and non UVM configure options?

liruipeng: No, I am not. I was saying allocating `A_h_correction_diag_*_new` on the host, after the for-loop allocating another array on GPU, copying to the GPU array and giving it to `A_h_correction_diag`. It may be better than directly operating on unified memory from the host, but I am not sure by how much. What you have here is also fine for now.

oseikuffuor1: Ok thanks. I will try to get these suggestions in.

mquanbui: Thanks @liruipeng for the suggestions! I was just putting in a quick fix to avoid crashing with GPU build for MGR with this non-Galerkin approach. This was not meant to be a full GPU implementation, and I'll keep your suggestions in mind when I start working on that.

liruipeng (approved):

oseikuffuor1:
@ulrikeyang and @mquanbui, can you take a look/ approve this PR when you get the chance? Thanks.

mquanbui (approved):

mquanbui:
Looks good to me. Thanks, @oseikuffuor1!

ulrikeyang (commented):
looks fine.

Issue #158, opened 2020-07-23
Hide hardcoded "warp size" in bit-shift operators


pbauman:
In several places in the code, specifically the GPU parts, there are statements like `HYPRE_Int wid  = threadIdx.x >> 5`. These are effectively hardcoding in warp size and will limit portability. I would like to hide the value behind a macro similar to `HYPRE_WARP_SIZE`. So there are two questions:

1. Do you agree in principle?
2. What is a good name for the macro? My kneejerk is `HYPRE_WARP_BIT_SHIFT`, but I'm not sure that's very clear.

I'm happy to cook up a PR to address this if there is agreement both on the strategy and the marco name. Thank you.

rfalgout:
Hi @pbauman .  Thanks for being mindful of portability!  Regarding your questions, I do agree in principle that the `>> 5` should not be hardcoded.  I also don't really like the macro name idea you throw out.  However, I think there is more to this than defining the number 5 with a macro.  This line effectively multiplies by 32, and the number 32 is important in related code, such as this code in `utilities/hypre_cuda_reducer.h`

      __shared__ T shared[32];        // Shared mem for 32 partial sums

So, we would either need another companion macro for the number 32 or use a different approach here.  Just my 2 cents.

pbauman:
Hi @rfalgout Agreed, I'm not super keen on the macro idea either. However, I was trying to follow precedence in the library. See #157 where I address exactly the the case you list, as well as a few others. I'm definitely open to suggestions.

rfalgout:
@pbauman .  Yep.  I wasn't paying attention.  Should have looked more carefully at your pull request first.

Anyway, why not change `>> 5` to `* HYPRE_WARP_SIZE`?  Is there a performance reason for the bit shift?

pbauman:
Hi @rfalgout, no worries, sorry for firing off all this at once.

My understanding is that `>> 5` would be equivalent to `/ HYPRE_WARP_SIZE`, not `* HYPRE_WARP_SIZE`?

[...]

PR #265, opened 2021-01-22
Add HYPRE_WARP_BITSHIFT Macro

pbauman:
This PR adds the macro `HYPRE_WARP_BITSHIFT` in the same vein as `HYPRE_WARP_SIZE`. The goal is the remove the magic number "5" from `>> 5` in several places in the code. This is forward-looking to adding ROCm/HIP support for which the bitshift will be `>> 6`. I tested this using the AMG Solve benchmark on CUDA 10.2 and CUDA 11.0 on V100. I'm happy to update the name if others have a better one/one they prefer. Thank you.

### DIFF
--- a/src/seq_mv/csr_spgemm_device.h
+++ b/src/seq_mv/csr_spgemm_device.h
@@ -32,7 +32,7 @@ hypre_int get_thread_id()
 static __device__ __forceinline__
 hypre_int get_warp_id()
 {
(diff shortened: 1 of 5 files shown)


pbauman:
This was discussed in #158.

rfalgout (approved):
This looks good to me.  Thanks @pbauman !

PR #268, opened 2021-01-29
GPU examples

liruipeng:
This PR adds GPU-support to hypre's examples. A new `Makefile_gpu` is provided in `src/examples` as an sample makefile, so one can compile the examples with `make -f Makefile_gpu` if hypre has been built with GPUs. We tried to not change the source code of `ex*.c` as much as we can. The main changes are in allocating memory. Static arrays on CPU like `double values[16]` are changed to  `values = malloc(...)`. The `malloc` has been redefined (as a macro) in `ex.h`. With `HYPRE_USING_CUDA`, `malloc` is redirected to `cudaMallocManaged` to allocate UVM on GPUs as follows

```
#define malloc(size) gpu_malloc(size)
static inline void*            
gpu_malloc(size_t size)      
{
   void *ptr = NULL; 
   cudaMallocManaged(&ptr, size, cudaMemAttachGlobal); 
   return ptr;
}
```

~~The problem of this approach is its portability, since it is allowed as GNU C extension (https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html)~~

These examples are meant to show the usage of hypre, instead of testing performance, in particular, for GPUs, due to the use of UVM and accessing UVM from both host and device So, these programs do not have the best performance as they can if device memory were used and everything had been ported to GPUs.

The following examples currently do not have GPU-support for different reasons.
- ex5f : no GPU support for fortran yet
- ex5big: no GPU support for `--enable-bigint`
- ex10 : FEI. Add support if needed
- ex12f : no GPU support for fortran yet
- ex15big : no GPU support for `--enable-bigint`
- ex17: no GPU support for `HYPRE_MAXDIM > 3`
- ex18: no GPU support for `HYPRE_MAXDIM > 3`
- ex18comp : no GPU support for `HYPRE_MAXDIM > 3`


src/examples/Makefile
```diff
@@ -3,38 +3,40 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+include ../config/Makefile.config
```
rfalgout: Hi @liruipeng .  We intentionally avoid coupling the examples to the hypre configuration stuff to make them as simple and instructive as possible.  Can you do this in a way that is more in the spirit of the original Makefile?

liruipeng: Hi, @rfalgout For GPU build, we need to change `CC` and more others. What is a more preferred way of doing it? Thanks!

rfalgout: Good question.  The main reason for the examples is to teach people how to use hypre, so we need to keep them simple.  Are you updating this so you can run the examples on GPUs as part of regression testing?  I'm not sure that would be a good reason, since we can test the code through our drivers.  If you want to illustrate how to use GPUs, then it might be better to introduce a "GPU example" that illustrates how the user code needs to change.  I would like to hear other opinions on this, though.

liruipeng: @rfalgout Thanks for the suggestions. I changed `Makefile` back, and added `Makefile_gpu`. So, one can compile the examples with `make -f Makefile_gpu` if hypre has been built with GPUs. Looks better now?

I actually want to have the sample examples as for CPUs with minor changes as you may have seen in the `malloc`. So, these examples should work on GPUs as well, not performing the best but should be good enough to show the usage. Some of them don't have GPU-support yet for different reasons.

Any thoughts? Thanks!

rfalgout: Hi @liruipeng .  Yes, this looks much better.  Thanks!

rfalgout:
This looks good to me.  Just a couple of comments:

- The redefine of `malloc` would only be a portability concern on GPU architectures, so I am okay with doing it this way.  I think it's better to use a term that is standard for C.
- Since the examples are supposed to help explain usage of hypre, it would be nice to point out the reason for some of the `malloc` usage, when fixed arrays are fine on the CPU.  Maybe adding a comment in front of all of them indicating that a static array would be okay on a CPU?

I would like to hear the opinions of you, @ulrikeyang, and @oseikuffuor1 on these points.

rfalgout (approved):
Looks good to me.  Thanks!

src/examples/ex1.c
```diff
@@ -170,6 +171,7 @@ int main (int argc, char *argv[])
                                                   defined above */
          int nentries = 5;
          int nvalues  = 30; /* 6 grid points, each with 5 stencil entries */
+         /* double values[30]; OK to use constant-length array for CPUs */
```
rfalgout: This looks good.

liruipeng:
> This looks good to me. Just a couple of comments:
> 
> * The redefine of `malloc` would only be a portability concern on GPU architectures, so I am okay with doing it this way.  I think it's better to use a term that is standard for C.

I rewrote them with inline functions. Please take another look.

> * Since the examples are supposed to help explain usage of hypre, it would be nice to point out the reason for some of the `malloc` usage, when fixed arrays are fine on the CPU.  Maybe adding a comment in front of all of them indicating that a static array would be okay on a CPU?

Good suggestion! Comments added.

> 
> I would like to hear the opinions of you, @ulrikeyang, and @oseikuffuor1 on these points.

rfalgout:
I thought the portability issue was with the redefinition of `malloc` itself.  Since this only affects the GPU, I don't really care how you do the redefine other than to suggest that you do something you would recommend for users and keep it as simple as possible.  Someone else can comment on the use of inline functions here.  Thanks!

ulrikeyang (approved):

oseikuffuor1 (approved):
Looks good. Thanks Ruipeng.

liruipeng:
> I thought the portability issue was with the redefinition of `malloc` itself. Since this only affects the GPU, I don't really care how you do the redefine other than to suggest that you do something you would recommend for users and keep it as simple as possible. Someone else can comment on the use of inline functions here. Thanks!

@rfalgout @ulrikeyang @oseikuffuor1 This is ready to merge. More thoughts on portability of `malloc`? Apparently, cannot use `hypre_TAlloc` here. Overriding `malloc` in the GPU examples keeps the code cleaner. Thanks!

oseikuffuor1:
> > I thought the portability issue was with the redefinition of `malloc` itself. Since this only affects the GPU, I don't really care how you do the redefine other than to suggest that you do something you would recommend for users and keep it as simple as possible. Someone else can comment on the use of inline functions here. Thanks!
> 
> @rfalgout @ulrikeyang @oseikuffuor1 This is ready to merge. More thoughts on portability of `malloc`? Apparently, cannot use `hypre_TAlloc` here. Overriding `malloc` in the GPU examples keeps the code cleaner. Thanks!

@liruipeng, this is fine with me. As @rfalgout noted, this only affects the GPU calls, otherwise it looks like just a standard malloc to me.

PR #269, opened 2021-01-31
Adding AMS support for 1D domains [ams-1d-dev]

mlstowell:
A CurlCurl preconditioner in 1D might seem unnecessary but this is needed for solving problems with anisotropic coefficients where the three possible polarizations of an HCurl field can mix, even in 1D.

This will also allow eigenmodes for 1D anisotropic problems to be obtained using HypreAME (which requires HypreAMS as a preconditioner).


tzanio (approved):
Neat stuff, thanks @mlstowell !

Issue #255, opened 2021-01-12
Add saved-file feature to runtest.sh script


rfalgout:
Having one set of saved output files for regression testing creates issues on different architectures that produce slightly different numerical results.  The `runcheck.sh` script allows some flexibility here, but sometimes requires the use of loose tolerances, which has become a problem on the GPU machines in particular.

We should extend the regression testing to allow for additional sets of "saved files".  To do this requires adding a saved-file-extension feature to the `runtest.sh` script (similar to -rtol and -atol), modifying the `TEST*/*.sh` scripts to use the extension, and creating the new saved files.

liruipeng:
@rfalgout is it easy to add additional arguments to all tests for GPUs, something like `./ij .... -exec_gpu`? Thanks!

rfalgout:
Hi @liruipeng .  Not currently.  It would not be difficult to add a feature that allows us to append something to every run string, if that's what you are asking.  So, if the normal run string is something like

    mpirun -np 8 struct -n 20 20 20 -solver 11 -rap 1 ...

it would be easy to update `runtest.sh` to add options at the very end of this.

PR #271, opened 2021-02-04
Add saved-file extension to runtest

rfalgout:
The main reason for this pull request is to add a `-save <ext>` feature to the `runtest.sh` script to allow testing against different saved files on different platforms such as GPU machines.  See Issue #255.  A few additional things were done:

- All of the checks against the saved files were moved out of the individual tests and into `runtest.sh`.  This just cleans up the tests a bit by reducing lots of replication.
- The output-file sanity checks that we had in many of the tests were modified so they no longer depend on the saved files.  In the process, some issues were also uncovered and fixed.

Saved files for other platforms have not yet been added.  I would like to get some feedback first on what they might be and how to name them, especially from @liruipeng .  For example, we could add files with extension `.saved.gpu` or `.saved.cuda` or whatever.  I don't want to name these in a machine-specific manner (e.g., ray).  Suggestions?


rfalgout:
Hi @liruipeng .  In looking at `machine-ray.sh`, it seems like having a set of saved files with extension `.saved.cuda` would be the most useful thing to have.  The `-rtol -atol` feature could still be used to handle small differences between platforms, and hopefully they would be much smaller than before.

I would like to add one set of extra files to this PR just to have an example of how this would be used.  From there, we can add and refine as needed.  So, maybe what I could do is use the output from the `# CUDA w.o UM` tests in `machine-ray.sh` and put them in files with the `.saved.cuda` extension as a starter.  You could add files with extension something like `.saved.cuda.um` later to handle the UM case.  How does that sound?

liruipeng (approved):
Looks good to me. Thank you @rfalgout

liruipeng:
> The main reason for this pull request is to add a `-save <ext>` feature to the `runtest.sh` script to allow testing against different saved files on different platforms such as GPU machines. See Issue #255. A few additional things were done:
> 
> * All of the checks against the saved files were moved out of the individual tests and into `runtest.sh`.  This just cleans up the tests a bit by reducing lots of replication.
> * The output-file sanity checks that we had in many of the tests were modified so they no longer depend on the saved files.  In the process, some issues were also uncovered and fixed.
> 
> Saved files for other platforms have not yet been added. I would like to get some feedback first on what they might be and how to name them, especially from @liruipeng . For example, we could add files with extension `.saved.gpu` or `.saved.cuda` or whatever. I don't want to name these in a machine-specific manner (e.g., ray). Suggestions?

I agree with @rfalgout. `.save.cuda` is good, and we can name for other platforms as `.save.hip`, `.save.iris`, etc.

liruipeng:
> Hi @liruipeng . In looking at `machine-ray.sh`, it seems like having a set of saved files with extension `.saved.cuda` would be the most useful thing to have. The `-rtol -atol` feature could still be used to handle small differences between platforms, and hopefully they would be much smaller than before.
> 
> I would like to add one set of extra files to this PR just to have an example of how this would be used. From there, we can add and refine as needed. So, maybe what I could do is use the output from the `# CUDA w.o UM` tests in `machine-ray.sh` and put them in files with the `.saved.cuda` extension as a starter. You could add files with extension something like `.saved.cuda.um` later to handle the UM case. How does that sound?

The `# CUDA w.o UM` tests are good ones to start, which include jobs in `TEST_CUDA_` and `TEST_struct`.

liruipeng:
@rfalgout There might not be need to have `saved.cuda.um`, since the results should not change with UVM or without, which is certainly true for `TEST_struct` and `TEST_cuda_`. For AMG, we may consider changing defaults for GPUs (such as coarsening, smoothers) that will change `.save.cuda`, so it makes sense we add them after.

rfalgout:
Okay @liruipeng .  I will move forward with the proposed plan.  Stay tuned and thanks!

rfalgout:
Okay, I've made the changes.  The only unfortunate thing I noticed is that the `.saved.cuda` files for these cases are identical to the `.saved` files.  I will run some final tests to make sure this is okay.  Otherwise, it's ready to go.

liruipeng:
> Okay, I've made the changes. The only unfortunate thing I noticed is that the `.saved.cuda` files for these cases are identical to the `.saved` files. I will run some final tests to make sure this is okay. Otherwise, it's ready to go.

@rfalgout Where were these `.saved.cuda` files generated? lassen? Thanks!

rfalgout:
On ray.  I got them from the `.out` files from the autotest runs on Feb 1.  I am running again on ray now to make sure this works.  I don't have lassen files from this month because autotest failed (I suppose I could check the January runs).

liruipeng:
> On ray. I got them from the `.out` files from the autotest runs on Feb 1. I am running again on ray now to make sure this works. I don't have lassen files from this month because autotest failed (I suppose I could check the January runs).

OK. It's good for now. I was thinking of only doing regression tests on lassen in the future so we don't have to worry about little discrepancies between ray and lassen. ray is mainly for development. Thoughts? Also, you mentioned they are identical to `.saved`...really? There should be some 10^{-15} errors...

rfalgout:
BTW, I just checked the Lassen files and they are the same also.  I guess `TEST_struct` is just well behaved.

liruipeng:
> BTW, I just checked the Lassen files and they are the same also. I guess `TEST_struct` is just well behaved.

Hmm...they are identical... I am puzzled why we had `rost="-struct -rt -mpibind -rtol 1e-8 -atol 1e-8"` in the original `machine-ray.sh`. The `rtol` and `atol` must have been added to address small errors. Something may have been changed.

rfalgout:
Agreed on the rtol/atol stuff.  BTW, I am also planning to update the lassen `rost` line since it appears that should work.  We can adjust all of this stuff more later.  At least now we will have more flexibility.

rfalgout:
The tux and ray tests run cleanly, so I just need @ulrikeyang and @oseikuffuor1 to approve, and this is ready to merge.  Thanks!

oseikuffuor1 (approved):
Looks good! Thanks Rob.

ulrikeyang (approved):

PR #275, opened 2021-02-11
fixed wrong pointer checking; error handler in debug mode

liruipeng:
This PR fixed a pointer comparison error in GPU matvec `y=Ax`, (if both pointers of `x` and `y` are `NULL`, should not throw errors). 

~~Found this by changing hypre's error handler behavior, printing out error messages in debug mode, so it may be useful in the future.~~


ulrikeyang (approved):
This looks good.

oseikuffuor1 (approved):

oseikuffuor1:
Good catch. Thanks Ruipeng

rfalgout (approved):
Looks good.  Thanks!

liruipeng:
Printing out error messages in debug mode messes up regression tests output, so I removed it.

PR #279, opened 2021-02-15
Remove 'memory.h' includes since it is not ISO standard

rfalgout:
The 'memory.h' header file is not standard ISO C, but it is standard C++.  Apparently, some C compilers may accidentally find a C++ version of this file and cause compile problems.  See issue #274 .  Removing it from hypre seems to compile just fine for me.


liruipeng (approved):
Looks fine to me. Thanks @rfalgout !

ulrikeyang (approved):

oseikuffuor1 (approved):
Thanks @rfalgout.

PR #280, opened 2021-02-16
Test jobs for enable-mixedint

liruipeng:
This PR separates the jobs that break with `--enable-mixedint` (CGC, ParaSail, Euclid) from regular job scripts, so regression tests can selectively run jobs with `--enable-mixedint`.

Status:
- [x] `machine-tux.sh`


rfalgout:
Hi @liruipeng .  After thinking for a bit, this is what I came up with.  Basically, I added a bunch of files named `runtests-option` that create an `-option` for the `run.sh` script (as well as for the `-ro:` lines in the higher level scripts).  The new files simply contain a list of tests for `runtest.sh` and support simple wildcard features to avoid having to list every runtest test individually.  Two good examples to look at are:

    AUTOTEST/runtests-sstruct
    AUTOTEST/runtests-sstruct-mixed

This should give us flexibility to create suites of tests that contain any sub-collection of runtest tests we want.  Combined with the `-save` feature, we should be able to do much more with our tests.

A few notes (none of this needs to happen, but I wanted to mention it):
- We should be able to get rid of some of the `TEST_*` directories such as `TEST_cuda_lassen` and even `TEST_longdouble` and `TEST_single`.  The tests in those directories could be moved into other directories now (or removed in favor of using existing tests) and called via an appropriate `runtest-option` file together with an appropriate `-save <extension>` on the saved files.
- Speaking of the save feature, the `HYPRE_NO_SAVED` could be replaced with option `-save off` or `-save none` or `-save nothankyou` :) , as long as these extensions don't exist, because saved files are only checked if they exist.
- We should revisit our use of the `HYPRE_NO_SAVED` because we could actually create saved files to compare with now in many cases.

I worry a little bit about this being confusing, so please let me know what you think.  I would like to hear from @ulrikeyang and @oseikuffuor1 as well.  Thanks!  And thanks for making the mixed-int testing possible!  I've run everything successfully on tux.

AUTOTEST/runtests-sstruct-mixed
```diff
@@ -0,0 +1,4 @@
+TEST_sstruct/[^n]*.sh
+TEST_sstruct/neighborpart.sh
+TEST_sstruct/neumann.sh
+
```
liruipeng, line 4: I was looking for a slightly cleaner way than this to exclude file names that start with `non`. No luck so far, not as easy as I thought. But this looks fine to me.

liruipeng:
> Hi @liruipeng . After thinking for a bit, this is what I came up with. Basically, I added a bunch of files named `runtests-option` that create an `-option` for the `run.sh` script (as well as for the `-ro:` lines in the higher level scripts). The new files simply contain a list of tests for `runtest.sh` and support simple wildcard features to avoid having to list every runtest test individually. Two good examples to look at are:
> 
> ```
> AUTOTEST/runtests-sstruct
> AUTOTEST/runtests-sstruct-mixed
> ```
> 
> This should give us flexibility to create suites of tests that contain any sub-collection of runtest tests we want. Combined with the `-save` feature, we should be able to do much more with our tests.
> 
> A few notes (none of this needs to happen, but I wanted to mention it):
> 
> * We should be able to get rid of some of the `TEST_*` directories such as `TEST_cuda_lassen` and even `TEST_longdouble` and `TEST_single`.  The tests in those directories could be moved into other directories now (or removed in favor of using existing tests) and called via an appropriate `runtest-option` file together with an appropriate `-save <extension>` on the saved files.
> * Speaking of the save feature, the `HYPRE_NO_SAVED` could be replaced with option `-save off` or `-save none` or `-save nothankyou` :) , as long as these extensions don't exist, because saved files are only checked if they exist.
> * We should revisit our use of the `HYPRE_NO_SAVED` because we could actually create saved files to compare with now in many cases.
> 
> I worry a little bit about this being confusing, so please let me know what you think. I would like to hear from @ulrikeyang and @oseikuffuor1 as well. Thanks! And thanks for making the mixed-int testing possible! I've run everything successfully on tux.

This looks good to me. I will update the GPU tests with this approach. Thanks, @rfalgout !

AUTOTEST/runtests-sstruct-mixed
rfalgout: Agreed.  Without using things like awk or sed, I don't know how to do it.  The wildcard stuff in linux is somewhat limited in what can be expressed.  If I come up with a clever idea, I'll let you know.  I am happy (I think) with the use of the individual files to specify the runtest tests.  The format of what goes in them could be improved.

liruipeng:
Just realized we need to test this on GPUs in another branch which supports `mixed-int` on  GPUs. So I leave `machine-ray.sh` as is here and will make the changes in a following PR. This should be good to go. Thanks!

rfalgout (approved):
Looks good to me.  Thanks!

oseikuffuor1:
> > Hi @liruipeng . After thinking for a bit, this is what I came up with. Basically, I added a bunch of files named `runtests-option` that create an `-option` for the `run.sh` script (as well as for the `-ro:` lines in the higher level scripts). The new files simply contain a list of tests for `runtest.sh` and support simple wildcard features to avoid having to list every runtest test individually. Two good examples to look at are:
> > ```
> > AUTOTEST/runtests-sstruct
> > AUTOTEST/runtests-sstruct-mixed
> > ```
> > 
> > 
> > This should give us flexibility to create suites of tests that contain any sub-collection of runtest tests we want. Combined with the `-save` feature, we should be able to do much more with our tests.
> > A few notes (none of this needs to happen, but I wanted to mention it):
> > 
> > * We should be able to get rid of some of the `TEST_*` directories such as `TEST_cuda_lassen` and even `TEST_longdouble` and `TEST_single`.  The tests in those directories could be moved into other directories now (or removed in favor of using existing tests) and called via an appropriate `runtest-option` file together with an appropriate `-save <extension>` on the saved files.
> > * Speaking of the save feature, the `HYPRE_NO_SAVED` could be replaced with option `-save off` or `-save none` or `-save nothankyou` :) , as long as these extensions don't exist, because saved files are only checked if they exist.
> > * We should revisit our use of the `HYPRE_NO_SAVED` because we could actually create saved files to compare with now in many cases.
> > 
> > I worry a little bit about this being confusing, so please let me know what you think. I would like to hear from @ulrikeyang and @oseikuffuor1 as well. Thanks! And thanks for making the mixed-int testing possible! I've run everything successfully on tux.
> 
> This looks good to me. I will update the GPU tests with this approach. Thanks, @rfalgout !

This sounds good to me @rfalgout. Thanks

oseikuffuor1 (approved):
Thanks @liruipeng. This looks good.

ulrikeyang (approved):
This looks good. Thanks for adding those.

PR #285, opened 2021-02-18
Bring recmat branch up to date with master

rfalgout:
This PR synchronizes master with the recmat branch by merging `recmat-merge` into `recmat`.  There are a number of things still to do and discuss before merging, so no further development should occur on the recmat branch until this PR is complete.

Because there were so many changes (due mainly to GPU work), there are too many differences to review in the normal manner for pull requests.  Instead, it's better to compare directly to master with the following double-dot diff (the `?w=1` at the end of the URL ignores whitespace changes):

  https://github.com/hypre-space/hypre/compare/master..recmat-merge?w=1

There are more than 300 files that differ from master.  Most are in the struct/sstruct code as they should be.  However, there are some changes to `IJ_mv`, `parcsr_mv`, `seq_mv`, and `utilities` to note.  Changes outside of the struct/sstruct code should really happen on master, so some discussion is needed about what to do here (integrate into master, modify code strategies, ...?).

Comparing the changes in `struct_mv`, `struct_ls`, `sstruct_mv`, and `sstruct_ls` to the recmat branch is more difficult with github since there are so many changes overall in this PR.  However, I tried it, and it's not as bad as I thought it might be.  Just scroll down to the struct/sstruct changes (have some patience while github updates its listing), then work through the files one at a time using the "Viewed" button to hide each file when done.  Another way to do this is through a tool like 'meld' (I used this extensively to help with the merge).  The diffs mainly involve GPU-related updates such as the new memory model.  However, although I tried to incorporate all updates made on `recmat`, some were reverted either because they broke regression tests or because I didn't understand why they were needed (more info on this below).  One thing I found useful when looking through these diffs is the following, which shows changes made on `recmat` since its last merge with master on May 2 2016 (commit 70767e0):

  https://github.com/hypre-space/hypre/compare/70767e0..06ad3a4?w=1

The basic strategy I used to synchronize with master and resolve the many merge conflicts (132 files) was:

- Start with code from master for everything outside of struct/sstruct, and merge in necessary changes made on recmat.
- Start with code from recmat for everything in struct/sstruct, and merge in (or reimplement) anything from master that was easy to do or necessary (e.g., all of the memory model stuff, but not all of the GPU implementation).

Once this was done, I looked through the diffs again, worked on getting the code to compile, then worked on getting the regression tests to run.

Currently, all of the regression tests run as they did in commit 4293971 on Jul 3 2017, with the exception of `TEST_sstruct/periodic`.  Here is a summary:

- 2 struct tests fail (emptyproc, pfmgconstcoeffmp).
- 10 sstruct tests fail (addtovalues, amr2d, default2d, default3d, emptyProc, miller, neighborpart, nonmixedint, periodic, solvers).

I couldn't figure out the problem with `periodic` because I don't know enough about the IJ code.  If you run the second job in the periodic test, the code fails because the `diag_j` pointer on line 2778 in file 'IJMatrix_parcsr.c' (function `hypre_IJMatrixAssembleParCSR`) is null and being accessed.  Maybe this is related to the new nnz optimizations?

I did not know how to run SSAMG, so that will need to be tested.  Because there were so many changes, it will likely require some debugging.

Here are some notes about what was reverted:

- These are the main files that had changes outside of struct/sstruct that I kept:

    src/IJ_mv/HYPRE_IJMatrix.c
    src/IJ_mv/HYPRE_IJ_mv.h
    src/IJ_mv/IJMatrix_parcsr.c
    src/IJ_mv/aux_parcsr_matrix.c
    src/IJ_mv/aux_parcsr_matrix.h
    src/parcsr_mv/par_csr_matrix.c
    src/parcsr_mv/par_csr_matrix.h

- There were major changes in the following that I didn't understand and didn't include:

    src/parcsr_mv/par_csr_bool_matop.c
    src/parcsr_mv/par_csr_matop.c
    src/parcsr_mv/par_csr_matvec.c
    src/seq_mv/csr_matop.c
    src/seq_mv/csr_matrix.c
    src/seq_mv/csr_matvec.c

- I reverted the changes to `StructMatmult` having to do with `Mgrid_in` and "exiting prematurely" because it broke the constant-coefficient case.

- I changed defaults in `test/struct.c` and `test/sstruct.c` to pass regression tests and match master.

- PFMG wasn't giving the right number of iterations and the new convergence test code does lots of extra matvecs, so I reverted to the previous way of doing things (that also corresponds to what is done on master).

- I needed to put the "data+size++ trick" back to get parallel runs working.

- I commented out the SquareMatvec routine because it was causing the cyclic reduction residual computation to fail in the sstruct driver.  I put a comment in the `struct_matvec.c` file about how to go about debugging this.


rfalgout:
Hi @victorapm and @ulrikeyang .  I just pushed some changes to recmat-merge that fix the remaining problems in the struct regression tests.  All of those tests and also the structmat tests run cleanly through valgrind.  Here is a summary of the changes:

- I had to make a few adjustments to the PFMG code to get the constant-coefficient stuff to line up with before, but I was not able to make it line up perfectly due to the new RAP Matmult routine, so I changed a couple of the saved files.
- I changed the point_relax code to basically just implement Jacobi by building on the matvec routine as we discussed in our meeting (this code should simply replace the current Jacobi code).
- There were a few memory leak problems that I cleaned up.

One thing I realized in the process of fixing this is that the Matmult routine doesn't really return an accurate result when constant coefficients are involved.  As long as a matrix-product coefficient is marked to be a variable coefficient, then the result will be correct regardless of the combination of constant and variable coefficients involved in the computation.  The problem is that the determination of whether a coefficient of the product matrix is constant or variable depends on the grid as well.  To see this, you only need to consider a simple all-constant matrix-matrix product to see that the resulting stencil is not necessarily the same along the boundaries as in the interior.  I'm not sure how we should handle this, yet.

We should turn the point_relax code into the Jacobi code and make a few optimizations:
- Change matvec to be optimized for residual calculations
- Fix Victor's unrolled square-matrix code and put it back in
- I will think about changing matvec to do the reverse communication thing we discussed so that we can use the unrolled kernels for all of matvecs, whether they are rectangular or square.

I am going to remove the sparse-msg code.  It's been hanging around for many years and doesn't have a future.

rfalgout:
BTW, I looked further into why I needed to change some of the saved files.  At first I thought it was due to the constant-coefficient point I made earlier about the new Matmult routine (that is still an issue to think about).  However, the test runs that needed to be changed all have fully variable coarse-grid matrices, even though the finest grid matrix has constant off-diagonals with variable diagonal.  So, the recmat-merge branch should have generated the same numbers as master.

I'm pretty confident that the reason they differ is because there is a small bug somewhere in one of the many RAP routines we have on the master branch.  There happened to be a `#define` flag I could flip the switch on to get the master code to use a different RAP routine than what it uses by default.  That routine produces matching results to recmat-merge.  Also, if I run recmat-merge with a constant off-diagonal matrix and compare that to a fully variable case (where I also force interpolation to match the constant case, so the only difference is that RAP has some constant values to deal with in one of the runs), the hierarchy and all of the iterations are identical.

rfalgout:
FYI, I just removed the SparseMSG solver

.gitignore
```diff
@@ -0,0 +1,28 @@
...
+*.err
+*.out
+config.log
+config.status
+Makefile.config
+HYPRE_config.h
+autom4te.cache
+*.TVD.v4breakpoints*
+
+# Drivers #
```
rfalgout, line 17: The gitignore file should be set up to match master, so we need to decide whether we should have these entries in master also.  There was some discussion about gitignore in one of the other issues or PRs.

rfalgout: FYI, the discussion about gitignore was in PR #243 .  As a strategy, we probably do want to add anything that might potentially be committed by accident by developers, so adding the driver executables probably makes sense.  On the other hand, this could get out of hand if you consider things like the AUTOTEST directory.  I think I will create an issue to have a short discussion and settle on a convention.

rfalgout:
Hi @victorapm and @ulrikeyang .  I just added more structmat regression tests and uncovered another bug in StructMatvec for the x==y case.  That's is now being tested along with different values for alpha and beta.

.gitignore
victorapm: Thanks for pointing this out, Rob! I have the drivers executables in my local gitignore and thought it could be useful to have it in the repo as well. The same for the TVD file which is created by totalview.

rfalgout:
Hi @victorapm .  The data_size problem should be fixed now.  The main change was in the `Resize()` routine, but I also did a little reorganizing and cleaning up of a few things.  The `structmat` and `struct` regression tests all run cleanly with valgrind.  Some additional `sstruct` tests run now as well (e.g., `miller` and parts of a few others).  Let me know if you have questions or issues with this in SSAMG.  Thanks!

victorapm:
Thank you @rfalgout! I see there are a couple of OutOfMemory and segfaults in TESTS_sstruct/solvers.err. Do you think it makes sense to solve this now?

rfalgout:
Hi @victorapm .  Some of the problems in `TEST_sstruct` might make sense to solve now, but I would have to look at them on a case by case basis.  I'm not sure I ever got the `TEST_sstruct/solvers` test running cleanly on the recmat branch.  Relative to commit 4293971 on Jul 3 2017, it looks like the same `TEST_sstruct` tests fail now as then with two exceptions: `miller` now runs but `periodic` does not (see the top description on this PR for the list from 2017).

It probably doesn't hurt to take a closer look at why these are failing, but I wouldn't spend too much time on it for this PR (some things are probably expected to fail until we complete additional implementation work).  We should definitely return to this on the `recmat` branch and try to get all of the regression tests running cleanly.

victorapm:
Yes, @rfalgout. I agree with this. So, I'll just take a quick look on these couple of tests failing due to segfault and see if It's a quick fix.

victorapm:
Thanks, Rob! Yes, I think we are very close to finishing this merge. We just need to complete #300 and possibly #341 (this can wait). SSAMG is working and I'm adding some tests now. There's still a problem with TESTS_sstruct/amr2D.sh related to my changes that I'm looking at.

src/sstruct_mv/sstruct_matmult.c
```diff
@@ -66,7 +66,7 @@ hypre_SStructMatmult( HYPRE_Int             nmatrices,
    HYPRE_Int                m, s;
    HYPRE_Int                part;
 
-#if DEBUG_MATCONV
+#if defined(HYPRE_DEBUG) && defined(DEBUG_MATCONV)
```
rfalgout, line 69: Why should this be changed in this way?

victorapm: Hi Rob, I was going back and forth between release and debug modes and wanted to run these statements only in debug mode. So, before, every time I compiled the code in release mode, I needed to go back to the code and undefine `DEBUG_MATCONV`. Now, I can keep such variable always defined and the statements bellow this line will run only when hypre is configured in debug mode. It was basically a smart way to have these debugging statements without needing to change code every time I changed compile mode.

src/struct_ls/pfmg_setup.c
```diff
@@ -377,17 +378,12 @@ hypre_PFMGSetup( void               *pfmg_vdata,
 
    if ((pfmg_data -> logging) > 0)
    {
-      max_iter = (pfmg_data -> max_iter);
       (pfmg_data -> norms)     = hypre_TAlloc(HYPRE_Real, max_iter+1, HYPRE_MEMORY_HOST);
       (pfmg_data -> rel_norms) = hypre_TAlloc(HYPRE_Real, max_iter+1, HYPRE_MEMORY_HOST);
    }
 
-#if DEBUG
+   if (print_level > 1)
```
rfalgout: I'm not sure we should use `print_level` to print to file.  We defined the meaning of this at one point, but maybe it's not documented well.  I will need to look this up.

victorapm: Hi Rob, I found the meaning of `print_level` is documented for some `ParCSR` solvers but not for `Struct` or `SStruct`. My reasoning was:
```
0: No printing.
1: Print convergence data.
2: Print MG matrices and vectors to file.
```
With the option 2, I eliminate the need of redefining the DEBUG variable every time I want to see that information. But having to do this is not a big deal either, so let me know if you prefer the way it was before and I'm happy to change.

rfalgout:
Hi @victorapm .  Once master is merged to this branch again, we can change the PR from draft to "ready for review" and I we can all take another look.  You might want to wait until the other PRs are done for the release before merging, but that's your call.  Thanks!

src/test/sstruct.c
```diff
@@ -1368,6 +1368,20 @@ DistributeData( ProblemData   global_data,
                                         pdata.graph_iuppers[box],
                                         ilower, iupper,
                                         int_ilower, int_iupper);
+
+                  /* Correct intersected box extents */
```
rfalgout, line 1374: Why is this different from the code on master?  Is there something wrong on master as well?

victorapm:
Hi @rfalgout and @ulrikeyang,

I'm marking this PR as ready for review. Almost all of the tux regression tests are passing (valgrind included). The ones that are not passing are:
```
FAILED : TEST_sstruct/addtovalues.err  (718)
FAILED : TEST_sstruct/emptyProc.err  (1228)
FAILED : TEST_sstruct/periodic.err  (7293)
FAILED : TEST_sstruct/solvers.err  (490)
```
I started looking at this and some of these errors originate from changes in `struct_communication.c`. For example, valgrind points to some invalid accesses in this file when running the case 31 from `TEST_sstruct/periodic.sh`. I think we can get back to these issues later.

Also, I put the changes that I made outside the `[s]struct_[mv,ls]` folders in #392.

Thanks!

victorapm:
Hi all,

these recent changes are mainly for enabling L1-Jacobi for SSAMG. It can be accessed with `-relax 2`, if you want to try out. 

Thanks!

rfalgout:
Hi @victorapm .  I have looked through the struct/sstruct code and it looks good to me.  I would like to take one final look after the merge of #392 .  Thanks!

victorapm (approved):

victorapm:
Hi all, In 7ef9ea8 I simplified the code for relaxation in SSAMG. This removed the `skip_diag` option, something that @rfalgout had required in one of our previous conversations. Thanks!

victorapm:
Hi all, I wanted to mention that PR #392 was merged to master and master was merged into this branch. Thus, this is ready for a final look. Thanks!

rfalgout:
Hi @victorapm .  This looks good, and I think we can merge.  I do have a few questions/comments that I would like to follow up on at some point:

- `.gitignore` - It looks like I never created an issue to discuss this.
- There are a few differences with `master` in the `IJ`, `parcsr`, and `utilities` directories.
- There are some differences with `master` in `test/sstruct.c` having to do with problem setup that I think we may have talked about, but I want to revisit.

Just for documentation sake, it looks like the following tests pass (with valgrind on): `structmat`, `struct`, and `IJ`.  There are four tests that currently fail in `sstruct`:

    addtovalues
    emptyProc
    periodic
    solvers

Nice job on this!

src/seq_mv/csr_matrix.c
```diff
@@ -33,79 +28,275 @@ hypre_CSRMatrixCreate( HYPRE_Int num_rows,
...
+
+#if defined(HYPRE_USING_CUSPARSE) || defined(HYPRE_USING_ROCSPARSE)
+      hypre_GpuMatDataDestroy(hypre_CSRMatrixGPUMatData(matrix));
+#endif
+
+      hypre_TFree(matrix, HYPRE_MEMORY_HOST);
    }
 
-   return hypre_error_flag;
+   return ierr;
```
ulrikeyang, line 91: Why doe we have 'ierr' here, when we used to have 'hypre_error_flag'? Why is ierr seet to 0?

victorapm: @ulrikeyang, I agree it should be `hypre_error_flag` as in the `recmat` branch. This change comes from master and there are other occurences like this in other functions. Should we change `return ierr` to `return hypre_error_flag` in `master` first and then update `recmat-merge` with the new `master`?

rfalgout: I say yes, change it in master first.  However, I don't think we need to wait for the change on master to finalize this PR.  We will continue to update `recmat` by regularly pulling in changes from master anyway.  My 2 cents.

ulrikeyang (approved):
I have looked through a lot of the 300+ files Rob mentions, but not in detail. I also looked more into IJ_mv, parcsr and seq_mv. I don't want to hold this up any longer. I have not seen anything of concern, but that is of course no guarantee. This entails a huge amount of work. Quite impressive We need to find a way to document this to make this more accessible to people who need to work on this in the future to make sure this is sustainable.

rfalgout:
Thanks @ulrikeyang !  These PR's stay around in GitHub, so the history is there.  What did you have in mind wrt documenting (we can discuss this face to face if you prefer)?

Hi, @victorapm .  I was just thinking that, instead of waiting for the recmat-mm PR #448 merge to happen, we could just merge this one and change the merge branch on the other.  Then, continue to keep the two branches up to date with master.  What do you think?  Thanks!

victorapm:
Thanks @ulrikeyang! 

Rob, I agree with your idea. We can change the destination branch to `recmat` in PR #448.

Are you going to do squash and merge for this one? Writing the commit message for this PR may take some time. Let me know if you want me to do it :) 

Thanks!
Victor

rfalgout:
Hmmmm...  That's a good question @victorapm!  Normally I would just do a merge when going branch to branch, but there were a fair number of changes here after the main merge from master, so maybe it would be good to summarize them in a single merge commit.  What do you think?

victorapm:
I think we can summarize them in a single merge commit now. This will help us in the future when we merge `recmat` into `master`. I can prepare the merge message if it helps :)

rfalgout:
Hi @victorapm .  Okay, sounds good.  I know that github suggests a message to begin with.  I am curious what that will look like.  Can you maybe email that to me before you start rewriting?  Also, I think we can keep the message somewhat high level, since the details can always be retrieved if needed.  Thanks again!

victorapm:
Yes, Rob! The original commit message generated by github contains many references of features that were added to master in the meantime (since recmat started diverging). I think we can exclude those, then I would go through the remaining ones and try to summarize them. Does that sound good?

rfalgout:
Sounds good.  Thanks!

PR #291, opened 2021-02-28
Changed the rhs in several TEST_sstruct/periodic runs

rfalgout:
The previous `-rhsone` option on tests 80-95 resulted in CG solving the problem exactly in 4 or 5 iterations.  This required the code to be bitwise accurate to pass the tests, which is not reasonable here.  The new cosine-based rhs produces more typical convergence for CG on a Laplacian problem.


liruipeng (approved):
Looks good to me. Thanks, @rfalgout

victorapm (approved):
Thank you, Rob! I see the tests are more robust now.

PR #292, opened 2021-03-02
Use HYPRE_USING_GPU Macro

pbauman:
This PR takes several instances of `#if defined(HYPRE_USING_CUDA) || defined(HYPRE_USING_DEVICE_OPENMP)` and changes them to `#if defined(HYPRE_USING_GPU)`. This is not totally exhaustive: I only chose ones that were (seemingly...) very obvious. This is a precursor to adding HIP support. I tested this PR on V100 with CUDA 10.2 by compiling HYPRE with CUDA support and building the `ij` test and running it.

### DIFF
--- a/src/distributed_matrix/distributed_matrix_parcsr.c
+++ b/src/distributed_matrix/distributed_matrix_parcsr.c
@@ -101,7 +101,7 @@ hypre_DistributedMatrixGetRowParCSR( hypre_DistributedMatrix *matrix,
    ierr = HYPRE_ParCSRMatrixGetRow( Parcsr_matrix, row, size, col_ind, values);
 
    // RL: if HYPRE_ParCSRMatrixGetRow was on device, need the next line to guarantee it's done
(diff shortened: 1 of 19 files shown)


rfalgout (approved):
Hi @liruipeng and @ulrikeyang .  If this is the approach we are moving towards, then I am good with this PR.  I think it would be good to do some grepping to make the changes consistent throughout hypre as well.

Thanks @pbauman !

pbauman:
No problem. If I may ask, I would hesitate to change any others at the moment as they will collide with the HIP work that I'm doing ATM. I did these because they shouldn't collide.

rfalgout:
That makes sense @pbauman .  Thanks again!

pbauman:
OK, I've rebased on master since there was a conflict with a recent change. I've also removed the changes in `HYPRE_struct_int.c`. In a separate branch, I've found that it's currently not compiling correctly with the HIP compiler so we'll avoid that path for now. (My primary goal is I want to get BoomerAMG working on AMD GPUs and get it upstreamed, then I can help diagnose.)

PR #296, opened 2021-03-08
removing device bindings in HYPRE_Init

liruipeng:
This PR removes GPU device binding in `HYPRE_Init()`.

PR #297, opened 2021-03-08
Add Basic HIP Support

pbauman:
This PR is the next step in HIP support, arguably the primary step. 

This adds the HIP path for many of the APIs in HYPRE. It does *not* add support for libraries other than Thrust (since it's header-only). The strategy was to mimic what was done for CUDA, namely use `HYPRE_USING_HIP` judiciously and then add supporting HIP APIs adjacent to the CUDA counterparts. 

I've tried to keep each of the commits small in order to facilitate ease of reading of each individual commit. Of course, this leads to many commits in the PR. Towards the end of this PR, starting roughly at 0b54652aa, is just turning on the compiler passes to HIP, i.e. adding `|| defined(HYPRE_USING_HIP)` alongside the `#if defined(HYPRE_USING_CUDA)` and similar. Everything before that commit hash is adding actual calls to the HIP APIs, so those probably warrant more careful examination.

This compiles with ROCm4.0/AMD GPUs using `./configure --with-hip=yes --without-MPI --enable-rocrand=no --enable-rocsparse=no --enable-rocblas=no`. I've not made a compiling pass with MPI yet nor has this PR been tested on AMD GPUs (I don't know what tests I could run that didn't require SpMV), but I wanted to get this going for review and possibly merge. I also checked that I didn't break compilation with CUDA 10.2 on V100. I built the `ij` test and ran it on V100, using `./ij -n 96 96 96 -pmis -keepT 1 -rlx 18 -exec_device -rap 1 -mod_rap2 1 -mm_cusparse 1 -interptype 6 -solver 1 -agg_nl 0` without unified memory enable and `./ij -n 96 96 96 -pmis -keepT 1 -rlx 18 -exec_device -rap 1 -mod_rap2 1 -mm_cusparse 1 -interptype 6 -solver 1 -agg_nl 1` with unified memory enabled, but nothing more than that.

My plan is to next add rocSPARSE support (rocSPARSE is the AMD sparse matrix library, analogous to cuSPARSE) and then test with MPI and then test using the AMG benchmark. rocRAND and rocBLAS support will be easy after that. rocTX support (analog with nvtx) also needs to be added, but will require some discussion.

Note that this will almost certainly conflict #296.

### DIFF
--- a/src/IJ_mv/HYPRE_IJMatrix.c
+++ b/src/IJ_mv/HYPRE_IJMatrix.c
@@ -433,7 +433,7 @@ HYPRE_IJMatrixSetValues2( HYPRE_IJMatrix       matrix,
       return hypre_error_flag;
    }
 
-#if defined(HYPRE_USING_CUDA)
+#if defined(HYPRE_USING_CUDA) || defined(HYPRE_USING_HIP)
    HYPRE_ExecutionPolicy exec = hypre_GetExecPolicy1( hypre_IJMatrixMemoryLocation(matrix) );
 
    if (exec == HYPRE_EXEC_DEVICE)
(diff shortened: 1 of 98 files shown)


pbauman:
I should also note that there's currently quite a few warnings spat out (`clang` is more strict than `nvcc`). I plan on helping silence the ones that are able to be silenced in a subsequent PR.

pbauman:
One more note: I just confirmed I was able to compile with MPICH on my local machine using: `./configure --with-hip=yes --with-MPI-include=/usr/include/x86_64-linux-gnu/mpich` so modulo the warnings and ROCm libraries support, this is about as complete as I can make it without being able to do full app tests.

PR #298, opened 2021-03-08
Hip support

liruipeng:
This PR (by @pbauman #297) is the next step in HIP support, arguably the primary step. 

This adds the HIP path for many of the APIs in HYPRE. It does *not* add support for libraries other than Thrust (since it's header-only). The strategy was to mimic what was done for CUDA, namely use `HYPRE_USING_HIP` judiciously and then add supporting HIP APIs adjacent to the CUDA counterparts. 

I've tried to keep each of the commits small in order to facilitate ease of reading of each individual commit. Of course, this leads to many commits in the PR. Towards the end of this PR, starting roughly at 0b54652aa, is just turning on the compiler passes to HIP, i.e. adding `|| defined(HYPRE_USING_HIP)` alongside the `#if defined(HYPRE_USING_CUDA)` and similar. Everything before that commit hash is adding actual calls to the HIP APIs, so those probably warrant more careful examination.

This compiles with ROCm4.0/AMD GPUs using `./configure --with-hip=yes --without-MPI --enable-rocrand=no --enable-rocsparse=no --enable-rocblas=no`. I've not made a compiling pass with MPI yet nor has this PR been tested on AMD GPUs (I don't know what tests I could run that didn't require SpMV), but I wanted to get this going for review and possibly merge. I also checked that I didn't break compilation with CUDA 10.2 on V100. I built the `ij` test and ran it on V100, using `./ij -n 96 96 96 -pmis -keepT 1 -rlx 18 -exec_device -rap 1 -mod_rap2 1 -mm_cusparse 1 -interptype 6 -solver 1 -agg_nl 0` without unified memory enable and `./ij -n 96 96 96 -pmis -keepT 1 -rlx 18 -exec_device -rap 1 -mod_rap2 1 -mm_cusparse 1 -interptype 6 -solver 1 -agg_nl 1` with unified memory enabled, but nothing more than that.

My plan is to next add rocSPARSE support (rocSPARSE is the AMD sparse matrix library, analogous to cuSPARSE) and then test with MPI and then test using the AMG benchmark. rocRAND and rocBLAS support will be easy after that. rocTX support (analog with nvtx) also needs to be added, but will require some discussion.

Note that this will almost certainly conflict #296.

### DIFF
--- a/src/seq_mv/csr_matvec_device.c
+++ b/src/seq_mv/csr_matvec_device.c
@@ -68,7 +68,7 @@ hypre_CSRMatrixMatvecDevice( HYPRE_Int        trans,
                              hypre_Vector    *y,
                              HYPRE_Int        offset )
 {
-#if defined(HYPRE_USING_CUDA)
+#if defined(HYPRE_USING_CUDA) || defined(HYPRE_USING_HIP)
    hypre_NvtxPushRange("CSRMatrixMatvec");
 #endif
 
@@ -114,7 +114,7 @@ hypre_CSRMatrixMatvecDevice( HYPRE_Int        trans,
 
    hypre_SyncCudaComputeStream(hypre_handle());
 
-#if defined(HYPRE_USING_CUDA)
+#if defined(HYPRE_USING_CUDA) || defined(HYPRE_USING_HIP)
    hypre_NvtxPopRange();
 #endif
 
--- a/src/seq_mv/csr_matvec_device.c
+++ b/src/seq_mv/csr_matvec_device.c
@@ -48,6 +48,8 @@ hypre_CSRMatrixMatvecDevice2( HYPRE_Int        trans,
 #endif
 #elif defined(HYPRE_USING_DEVICE_OPENMP)
    hypre_CSRMatrixMatvecOMPOffload(trans, alpha, A, x, beta, y, offset);
+#elif defined(HYPRE_USING_HIP)
+#warning SPMV NOT IMPLEMENTED FOR HIP YET
 #else // #ifdef HYPRE_USING_CUSPARSE
 #error HYPRE SPMV TODO
 #endif
--- a/src/utilities/_hypre_utilities.hpp
+++ b/src/utilities/_hypre_utilities.hpp
@@ -157,11 +157,23 @@ typedef struct hypre_cub_CachingDeviceAllocator hypre_cub_CachingDeviceAllocator
 
 struct hypre_CudaData
 {
+#if defined(HYPRE_USING_CURAND)
    curandGenerator_t                 curand_generator;
+#endif
+
+#if defined(HYPRE_USING_CUBLAS)
    cublasHandle_t                    cublas_handle;
+#endif
+
+#if defined(HYPRE_USING_CUSPARSE)
    cusparseHandle_t                  cusparse_handle;
    cusparseMatDescr_t                cusparse_mat_descr;
+#endif
+
+#if defined(HYPRE_USING_CUDA) || defined(HYPRE_USING_DEVICE_OPENMP)
    cudaStream_t                      cuda_streams[HYPRE_MAX_NUM_STREAMS];
+#endif
+
 #ifdef HYPRE_USING_CUB_ALLOCATOR
    hypre_uint                        cub_bin_growth;
    hypre_uint                        cub_min_bin;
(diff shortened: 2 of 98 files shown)


src/seq_mv/csr_matvec_device.c
```diff
@@ -48,6 +48,8 @@ hypre_CSRMatrixMatvecDevice2( HYPRE_Int        trans,
 #endif
 #elif defined(HYPRE_USING_DEVICE_OPENMP)
    hypre_CSRMatrixMatvecOMPOffload(trans, alpha, A, x, beta, y, offset);
+#elif defined(HYPRE_USING_HIP)
+#warning SPMV NOT IMPLEMENTED FOR HIP YET
```
rfalgout, line 52: Is this ISO standard C?  If so, which standard?  If not, we should handle this differently.

pbauman: I do not believe it is officially standard, but it is supported by GCC, clang. I'm pretty sure it's supported by Intel as well. Regardless, this is a stop-gap.

On the one hand, I wanted the code in the PR to compile with the HIP compiler, but I did not want to mix in rocSPARSE support with this PR. On the other hand, HYPRE will not compile GPU support without a supported SpMV implementation. I thought this was a reasonable stop-gap to allow compilation, but still spit out something to alert the developer+user that they should not expect their program to run successfully on AMD GPUs yet.

I'm open to other suggestions.

rfalgout (approved):
This all looks good to me, modulo doing appropriate testing, etc.  Thanks @pbauman and @liruipeng !

src/seq_mv/csr_matvec_device.c
rfalgout: If this is just a stop-gap, we can leave it as is.  Thanks!

src/utilities/_hypre_utilities.hpp
```diff
@@ -493,8 +588,20 @@ void __syncwarp()
...
 }
 
+#endif // #if defined(HYPRE_USING_HIP) || (CUDA_VERSION < 9000)
+
+
+// __any was technically deprecated in CUDA 7 so we don't bother
+// with this overload for CUDA, just for HIP.
+#if defined(HYPRE_USING_HIP)
+static __device__ __forceinline__
+int __any_sync(unsigned mask, int predicate)
```
liruipeng: changed `int` to `hypre_int`

src/seq_mv/csr_matvec_device.c
liruipeng: We have `hypre_CSRMatrixSpMVDevice` implementation of SpMV. If SpMV is the only thing that stops us running on AMD GPUs, we can try it (we should add it anyway as it's marked as "TODO").

liruipeng:
@pbauman This looks good to me! I made a few comments and changes. Also, not really specific to this PR, there are file/function/variable names in hypre that are tied to cuda, CU, nv, or NVIDIA, which you probably don't like :), we can change them to more generic names. Thanks a lot for all these changes!

src/seq_mv/csr_matvec_device.c
liruipeng: > Is this ISO standard C? If so, which standard? If not, we should handle this differently.

I assume you two are talking about `#warning`. I didn't know it's not part of C. How about `#error`? Thanks.

pbauman: SpMV was only stopping the compiling. I believe SpGEMM is also required to actually run a test case? Either way, my own preference would be just get rocSPARSE in ASAP.

liruipeng: > SpMV was only stopping the compiling. I believe SpGEMM is also required to actually run a test case? Either way, my own preference would be just get rocSPARSE in ASAP.

Yes, SpGEMM is needed for running AMG. We also have our own implementation for it (I am not sure if it has been compiled with HIP). I agree, getting rocSPARSE in hypre is probably easier to test first. Thanks!

src/utilities/_hypre_utilities.hpp
pbauman: Good catch, thanks.

pbauman (approved):
@liruipeng changes on top of my original PR look good.

ulrikeyang (approved):
Looks good. Thank you.

liruipeng:
Hi, @pbauman @ulrikeyang @rfalgout I found a glitch in `hypre_memory.c` with `HYPRE_USING_DEVICE_OPENMP`. Instead of simply fixing it, I redid the `#if` logic which is hopefully clearer with the newly introduced `Umpire` and `Hip` stuff.  Thanks!

pbauman:
OK, thanks @liruipeng.  Generally, I attempted to try to leave in place the logic with OMP+CUDA and not try to mix in HIP+OMP at this point. Mostly because I'm not clear yet on the expectations for that functionality. So I'm glad you caught that.

pbauman:
I should say, i also agree that the new processing logic looks clearer. Thank you.

liruipeng:
@pbauman Do we need to do some testings to make sure if hipified version still works fine after my changes? Otherwise regression tests are clean on our end. It is ready to go. Thanks!

pbauman:
Hi @liruipeng In fact I did pull the branch this morning and confirmed I can still compile HYPRE with HIP+MPICH, so I think this is good to go. Thanks!

PR #302, opened 2021-03-11
More hip support

liruipeng:
By @pbauman #301 

There were a couple of places that I missed in the first HIP pass in #297. Also I missed adding some pieces of the build system in #288. This PR takes care of those. 

Perhaps the more important thing here is the guarding of asserts in device kernels. Unfortunately, at present, asserts in device kernels behave sub-optimally and may either cause the compiler to fail (an ICE) or may produce code that behaves incorrectly.

The library itself will still compile in static mode with `--with-hip=yes`. However, building in shared mode or trying to compile the `ij` test (for example) will fail as there are a couple of missing functions. Those will be taken care of the in the (very shortly) forthcoming rocSPARSE support PR. I wanted to keep these changes separate from that.

### DIFF
--- a/src/seq_mv/csr_spgemm_device_util.c
+++ b/src/seq_mv/csr_spgemm_device_util.c
@@ -8,18 +8,19 @@
 #include "seq_mv.h"
 #include "csr_spgemm_device.h"
 
-#if defined(HYPRE_USING_CUDA)
+#if defined(HYPRE_USING_CUDA) || defined(HYPRE_USING_HIP)
 
 /* assume d_i is of length (m+1) and contains the "sizes" in d_i[1], ..., d_i[m]
    the value of d_i[0] is not assumed
  */
 void csr_spmm_create_ija(HYPRE_Int m, HYPRE_Int *d_i, HYPRE_Int **d_j, HYPRE_Complex **d_a, HYPRE_Int *nnz)
 {
-   cudaMemset(d_i, 0, sizeof(HYPRE_Int));
+   hypre_Memset(d_i, 0, sizeof(HYPRE_Int), HYPRE_MEMORY_DEVICE);
+
    /* make ghash pointers by prefix scan */
    HYPRE_THRUST_CALL(inclusive_scan, d_i, d_i + m + 1, d_i);
    /* total size */
-   cudaMemcpy(nnz, d_i + m, sizeof(HYPRE_Int), cudaMemcpyDeviceToHost);
+   hypre_Memcpy(nnz, d_i + m, sizeof(HYPRE_Int), HYPRE_MEMORY_HOST, HYPRE_MEMORY_DEVICE);
    if (d_j)
    {
       *d_j = hypre_TAlloc(HYPRE_Int, *nnz, HYPRE_MEMORY_DEVICE);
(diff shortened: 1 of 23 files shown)


rfalgout:
Hi @pbauman and @liruipeng .  Would it be better to use `hypre_assert` instead of `assert` and put all of the `#ifdef` stuff where it is defined in `utilities/_hypre_utillities.h`?  It would certainly clean the code up a bit, but I don't know if there are other considerations here.

liruipeng:
> Hi @pbauman and @liruipeng . Would it be better to use `hypre_assert` instead of `assert` and put all of the `#ifdef` stuff where it is defined in `utilities/_hypre_utillities.h`? It would certainly clean the code up a bit, but I don't know if there are other considerations here.

Nice suggestion! We should do that in some way. It's a bit trickier since this is a GPU assert; our `hypre_assert` cannot do it in its current form. We may be able to encapsulate more `#if` to have a uniform `hypre_assert` (CPU/GPU). An easier way is to add `hypre_device_assert`. Let's see.

pbauman:
This thought had occurred to me, but I didn't want to do down that path unless HYPRE devs wanted it. I can cook up something and push to this branch tomorrow for `hypre_device_assert`, that won't be hard.

pbauman:
OK, @liruipeng @rfalgout I've pushed two commits which adds a `hypre_device_assert` and then uses it. Let me know what you think. I tested both with HIP and with CUDA10.2

rfalgout (approved):
This all looks good to me, and I like the new assert macro.  If @liruipeng is okay with it and everything still runs cleanly, I'm happy with this PR.  Thanks!

ulrikeyang (approved):
This looks good to me. Thank you.
I have to say (and I believe Ruipeng mentioned this before elsewhere) using names like CUFLAGS etc is just plain weird right now, but changing this is obviously not a high priority right now. BTW would anybody ever imagine a machine that would use Nvidia and AMD GPUs at the same time,, such as a super-heterogeneous machine? I guess that would only make sense if each of them had some important properties the other one wouldn't have.

pbauman:
> I have to say (and I believe Ruipeng mentioned this before elsewhere) using names like CUFLAGS etc is just plain weird right now,

I agree @ulrikeyang it's awkward, but I thought it better to get support in and then discussion on how to edit the build system etc. rather than me injecting my own sensibilities into HYPRE.

> BTW would anybody ever imagine a machine that would use Nvidia and AMD GPUs at the same time,, such as a super-heterogeneous machine?

This is tough to imagine. Certainly we have workstations with both, but I don't think anyone executes on them simultaneously. I imagine software trying to use both simultaneously could be tough to maintain, but that's just my opinion.

pbauman (approved):

liruipeng:
@rfalgout @pbauman Thank you for adding `hypre_device_assert`, which looks good to me. I found there was misunderstanding of `assert`, which was my fault, (`hypre_assert` is supposed to wrap `assert` and have basically the same behavior), that is if not in debug mode (i.e., `NDEBUG` is defined), `assert(EX)` should be no-op, so `EX` should not be even executed, the current definition
```
#define hypre_assert(EX) do { (void) (EX); } while (0)
```
actually runs `EX`. So, it should be just simply 
~~`#define hypre_assert(EX)`~~
?

It doesn't matter for `assert(a>0)` but does for something like `assert(++a)` (if anyone actually writes in this way) 

Any thoughts? I want to correct it in this PR, if you don't mind.

Update: I remembered why I did this `#define hypre_assert(EX) do { (void) (EX); } while (0)` trick, to silence compiler warnings "unused variables in EX". Now I do it as the following which is even uglier 
```
#define hypre_assert(EX) do { if (0) { (void) (EX);  } } while (0)
```

pbauman:
Fine with me

rfalgout:
Fine with me, too.  I've never used `assert`, so it doesn't really matter to me how it works. :)  The nice thing about this now is we can change the behavior of assert by touching only one file.  Thanks!

src/seq_mv/csr_spgemm_device_util.c
```diff
@@ -34,11 +35,14 @@ void csr_spmm_create_ija(HYPRE_Int m, HYPRE_Int *d_i, HYPRE_Int **d_j, HYPRE_Com
 void csr_spmm_create_ija(HYPRE_Int m, HYPRE_Int *d_c, HYPRE_Int **d_i, HYPRE_Int **d_j, HYPRE_Complex **d_a, HYPRE_Int *nnz)
 {
    *d_i = hypre_TAlloc(HYPRE_Int, m+1, HYPRE_MEMORY_DEVICE);
-   cudaMemset(*d_i, 0, sizeof(HYPRE_Int));
+   hypre_Memset(d_i, 0, sizeof(HYPRE_Int), HYPRE_MEMORY_DEVICE);
```
liruipeng: Fixed this bug. It should be `*d_i`

PR #308, opened 2021-03-16
2 stage gs

liruipeng:
This PR (by @PaulMullowney #282) includes updates/changes to the 2 stage GS preconditioner. The algorithm has the general formulation.
r = relax_weight * (b - Ax)
z' = inv(D) (r - Lz)
z'' = inv(D) (r - Lz')
z''' = inv(D) (r- Lz'')
...
x = x + z'''

relax_type 11 stops at the first term, z'.
relax_type 12 stops at the second term z''. This replaces the old version of 12 which was not an effective preconditioner.

Experiments with more correction terms have been done. Going beyond 2 has not shown much benefit. Because we're reinterpreting the meaning of relax_type=12, we didn't need to make any API changes. In the future, one could add an integer parameter to specify how many terms to use however this induces significant API changes through the code. CPU and GPU give identical results.

One more important point. When either of these algorithms was used on its own as a preconditioner (i.e. outside of AMG), Line 329 of par_cycle.c was forcing the number of iterations to 1. This seems wrong so I've changed it to what's read in from the user num_sweeps parameter. Having a user defined number of sweeps as a critical run time parameter and should not be forced to 1, imo.

### DIFF
--- a/src/parcsr_ls/par_cycle.c
+++ b/src/parcsr_ls/par_cycle.c
@@ -326,7 +326,7 @@ hypre_BoomerAMGCycle( void              *amg_vdata,
          /* If no coarsening occurred, apply a simple smoother once */
          Aux_U = U_array[level];
          Aux_F = F_array[level];
-         num_sweep = 1;
+         num_sweep = num_grid_sweeps[0];
          /* TK: Use the user relax type (instead of 0) to allow for setting a
            convergent smoother (e.g. in the solution of singular problems). */
          relax_type = hypre_ParAMGDataUserRelaxType(amg_data);
--- a/src/parcsr_ls/_hypre_parcsr_ls.h
+++ b/src/parcsr_ls/_hypre_parcsr_ls.h
@@ -1963,6 +1963,8 @@ HYPRE_Int hypre_BoomerAMGRelaxKaczmarz( hypre_ParCSRMatrix *A, hypre_ParVector *
 
 HYPRE_Int hypre_BoomerAMGRelaxTwoStageGaussSeidelDevice ( hypre_ParCSRMatrix *A, hypre_ParVector *f, HYPRE_Real relax_weight, HYPRE_Real omega, hypre_ParVector *u, hypre_ParVector *r, hypre_ParVector *z, HYPRE_Int choice);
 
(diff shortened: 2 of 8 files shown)


src/parcsr_ls/par_cycle.c
```diff
@@ -326,7 +326,7 @@ hypre_BoomerAMGCycle( void              *amg_vdata,
          /* If no coarsening occurred, apply a simple smoother once */
          Aux_U = U_array[level];
          Aux_F = F_array[level];
-         num_sweep = 1;
```
liruipeng, line 329: @ulrikeyang Can you please comment on this change? This is a special 1-level case. Do we want to apply smoothers more than once here? @PaulMullowney commented on the use case of it as separate preconditioner outside AMG. But do you still use `hypre_BoomerAMGSolve` which calls `hypre_BoomerAMGCycle` to apply?

PaulMullowney: @ulrikeyang Yes, we want to apply the Preconditioner (1 level 2stage GS algorithm) more than once per Krylov iteration. Line 329 was forcing the number of sweeps to 1 every time which was reducing its effectiveness. 

I was still using BoomerAMG but with max_levels=1. Is there another way to do this problem so that this change isn't needed?

ulrikeyang: Yes, I think it's a good idea to allow this, and num_grid_sweeps[0] is not used anywhere., we only use k=1,2,3. We just need to make sure that the default for this continues to be 1, but users will then be able to set it with HYPRE_BoomerAMGSetCycleNumSweeps(solver,ns,0);
So, no extra interface changes needed.

liruipeng: @PaulMullowney This is also fine to me. Using preconditioner as smoother of AMG that has only 1 level can fulfill the requirement here, albeit a little tricky. If you want to do it without involving AMG, you need to implement an interface of preconditioner setup and application, and give it to Krylov solvers, which should be very simple. You can follow how it's done for `HYPRE_ParaSailsSetup` and `HYPRE_ParaSailsSolve` for example.

liruipeng:
All regression tests are passed. This PR is good to go. Thanks!

PaulMullowney (changes requested):
On line 1478 of par_relax.c, it looks like I wrapped the error checking code in 
#if 0
#endif
We should remove this protection before merging, or remove the block entirely.

Otherwise, looks good. Thanks

liruipeng:
> On line 1478 of par_relax.c, it looks like I wrapped the error checking code in
> #if 0
> #endif
> We should remove this protection before merging, or remove the block entirely.
> 
> Otherwise, looks good. Thanks

@PaulMullowney It has been fixed. Thank you!

PaulMullowney (approved):

PR #312, opened 2021-03-22
Fix s

ulrikeyang:
This pull request fixes the hopscotch problem. 
The generation of the strength matrix for nodal coarsening combined with unknown-based approach has been rewritten.
A new option HYPRE_BoomerAMGSetKeepSameSign has been added, which allows different generation of this strength matrix.
The default is 0, i.e. turned off, since this is necessary for good convergence for extended+i interpolation, but it hurts
classical interpolation and MM-extended interpolation (the latter needs further investigation). In the future I would like to get rid of this option, but for right now it might be useful for certain systems problems.
I have also included the unknown-based approach into MM-extended, MM-extended+i , MM-extended+e, and 2-stage MM-extended+e interpolation. My inclusion into 2-stage MM-extended interpolation doesn't work well, so I took it out (also needs further investigation).
I removed the use of col_offd_S_to_A and the functionality of HYPRE_BoomerAMGSetSCommPkgSwitch. The function is still there, to not break users' code in case somebody actually uses it, but it doesn't do anything.


src/test/TEST_ij/agg_interp.saved
```diff
@@ -74,7 +74,7 @@ Final Relative Residual Norm = 5.870410e-09
 
 
 Iterations = 16
-Final Relative Residual Norm = 5.087936e-09
+Final Relative Residual Norm = 5.163837e-09
```
rfalgout, line 77: Is it clear why these residuals are different now?  Just being cautious.

rfalgout (approved):
Hi @ulrikeyang .  Nice work on all these changes!  It looks good to me, though I did not look through the code logic at all, since others are more qualified for that.  The only question I have is on the regression testing and the differences in saved file numbers.  It's always good to be extra cautious when doing this, because even small changes can sometimes indicate underlying issues.  If you are confident about why they differ, then that's good enough for me.

Thanks again!

src/test/TEST_ij/agg_interp.saved
ulrikeyang: Yes, this was a systems problem, as are the ones below, so I would have expected some change based on the changes I made.

rfalgout: Perfect!

src/parcsr_ls/par_2s_interp.c
```diff
@@ -281,27 +314,83 @@ hypre_BoomerAMGBuildModPartialExtInterpHost( hypre_ParCSRMatrix  *A,
       {
          if (CF_marker[i] == -2)
          {
-            for (j=A_diag_i[i]; j < A_diag_i[i+1]; j++)
+            /*if (num_functions > 1)
```
liruipeng, line 317: This is commented out but not in the next one ext+e, which is fine. Just want to make sure it is what you want. Thanks!

src/parcsr_ls/par_nodal_systems.c
```diff
@@ -798,74 +798,64 @@ hypre_BoomerAMGCreateNodalA(hypre_ParCSRMatrix    *A,
...
-   HYPRE_Int           SN_num_nonzeros_offd;
+   HYPRE_Int           num_cols_offd_A = hypre_CSRMatrixNumCols(A_offd);
+   HYPRE_Int           A_num_nonzeros_diag;
+   HYPRE_Int           A_num_nonzeros_offd;
    HYPRE_Int           S_num_nonzeros_diag;
    HYPRE_Int           S_num_nonzeros_offd;
    HYPRE_BigInt        global_num_vars;
    HYPRE_BigInt        global_num_cols;
    HYPRE_BigInt        global_num_nodes;
+   HYPRE_Int           nnz, S_cnt, in;
```
liruipeng, line 858: This is not a new issue in this PR. Line 871 `if (nodal < 0) {}`: 1. not sure what's the meaning of `nodal < 0`, probably not documented. 2. `dof_func` is allocated as size of `num_coarse_nodes*num_functions` only for C points, which does not match the comment above this function, "scalar version of dof_array"

liruipeng (approved):
Looks good to me. Thanks!

src/parcsr_ls/aux_interp.c
```diff
@@ -354,7 +354,15 @@ static HYPRE_Int hypre_new_offd_nodes(HYPRE_BigInt **found, HYPRE_Int num_cols_A
         {
           if (!hypre_UnorderedBigIntSetContains(&set, big_i1))
           {
-            Sop_j[j] = -hypre_UnorderedBigIntMapGet(&col_map_offd_inverse, big_i1) - 1;
+            HYPRE_Int k = hypre_UnorderedBigIntMapGet(&col_map_offd_inverse, big_i1);
+            if (-1 == k)
+            {
+               hypre_UnorderedBigIntSetPut(&set, big_i1);
```
liruipeng, line 360: Is it still possible to happen after your changes that make sure S is a subset of A? Just curious. Thanks!

ulrikeyang:
It should not happen unless we set HYPRE_BoomerAMGSetKeepSameSign to 1. In that case, we could still have a problem, but it should not crash, it would be a potential convergence issue, but convergence is tricky in general for systems problems. Sometimes it might even help.
Note that hopscotch will now act exactly like the non-hopscotch version for this case.

ulrikeyang:
Yes , I commented this out on purpose, since the systems version seem to make things worse for this case on all the regression problems. It’s something that needs to be investigated, but I didn’t have time to do this, so I thought it’s better to leave it out for now.

oseikuffuor1 (approved):
Looks good to me. Thanks @ulrikeyang 
.

PR #304, opened 2021-03-15
Add rocSPARSE Support

pbauman:
This PR adds the first pass of rocSPARSE support. It is not complete: you will notice that there are a couple of "stubs" where I throw a runtime error message if that code path is taken and I have not yet included optimizations, e.g. the analysis API for dscrmv. Nevertheless, I am able to run BoomerAMG on AMD GPUs, specifically the `ij` test using aggressive coarsening with unified memory `./ij -n 96 96 96 -pmis -keepT 1 -rlx 18 -exec_device -rap 1 -mod_rap2 1 -mm_cusparse 1 -interptype 6 -solver 1 -agg_nl 1` and no aggressive coarsening without unified memory `./ij -n 96 96 96 -pmis -keepT 1 -rlx 18 -exec_device -rap 1 -mod_rap2 1 -mm_cusparse 1 -interptype 6 -solver 1 -agg_nl 0`.

That said, I'm currently chasing what seems to be a compiler bug (though I'm not entirely sure of that yet). The cases above do converge, but with optimization level `-O2`, the coarsening behaves differently than with `-O1` and `-O0`. Nevertheless, I wanted you good folks to have a look at this and with any luck I can try and pin down the issue by the time this is ready to merge. If not, then y'all can decide if you want to wait to merge or not.

PR #316, opened 2021-03-25
Roc sparse

liruipeng:
This PR (by @pbauman #304)  adds the first pass of rocSPARSE support. It is not complete: you will notice that there are a couple of "stubs" where I throw a runtime error message if that code path is taken and I have not yet included optimizations, e.g. the analysis API for dscrmv. Nevertheless, I am able to run BoomerAMG on AMD GPUs, specifically the `ij` test using aggressive coarsening with unified memory `./ij -n 96 96 96 -pmis -keepT 1 -rlx 18 -exec_device -rap 1 -mod_rap2 1 -mm_cusparse 1 -interptype 6 -solver 1 -agg_nl 1` and no aggressive coarsening without unified memory `./ij -n 96 96 96 -pmis -keepT 1 -rlx 18 -exec_device -rap 1 -mod_rap2 1 -mm_cusparse 1 -interptype 6 -solver 1 -agg_nl 0`.

That said, I'm currently chasing what seems to be a compiler bug (though I'm not entirely sure of that yet). The cases above do converge, but with optimization level `-O2`, the coarsening behaves differently than with `-O1` and `-O0`. Nevertheless, I wanted you good folks to have a look at this and with any luck I can try and pin down the issue by the time this is ready to merge. If not, then y'all can decide if you want to wait to merge or not.

### DIFF
--- a/src/seq_mv/csr_spgemm_device_rocsparse.c
+++ b/src/seq_mv/csr_spgemm_device_rocsparse.c
@@ -1,0 +1,179 @@
+/******************************************************************************
+ * Copyright 1998-2019 Lawrence Livermore National Security, LLC and other
+ * HYPRE Project Developers. See the top-level COPYRIGHT file for details.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR MIT)
+ ******************************************************************************/
+
+#include "seq_mv.h"
+#include "_hypre_utilities.hpp"
+
+#if defined(HYPRE_USING_HIP) && defined(HYPRE_USING_ROCSPARSE)
+
+HYPRE_Int
+hypreDevice_CSRSpGemmRocsparse(HYPRE_Int       m,
+                               HYPRE_Int       k,
+                               HYPRE_Int       n,
+                               HYPRE_Int       nnzA,
+                               HYPRE_Int      *d_ia,
+                               HYPRE_Int      *d_ja,
+                               HYPRE_Complex  *d_a,
+                               HYPRE_Int       nnzB,
+                               HYPRE_Int      *d_ib,
+                               HYPRE_Int      *d_jb,
+                               HYPRE_Complex  *d_b,
+                               HYPRE_Int      *nnzC_out,
+                               HYPRE_Int     **d_ic_out,
+                               HYPRE_Int     **d_jc_out,
+                               HYPRE_Complex **d_c_out)
+{
+  HYPRE_Int  *d_ic, *d_jc, baseC, nnzC;
+  HYPRE_Int  *d_ja_sorted, *d_jb_sorted;
+  HYPRE_Complex *d_c, *d_a_sorted, *d_b_sorted;
+
+  d_a_sorted  = hypre_TAlloc(HYPRE_Complex, nnzA, HYPRE_MEMORY_DEVICE);
+  d_b_sorted  = hypre_TAlloc(HYPRE_Complex, nnzB, HYPRE_MEMORY_DEVICE);
+  d_ja_sorted = hypre_TAlloc(HYPRE_Int,     nnzA, HYPRE_MEMORY_DEVICE);
+  d_jb_sorted = hypre_TAlloc(HYPRE_Int,     nnzB, HYPRE_MEMORY_DEVICE);
+
+  rocsparse_handle handle = hypre_HandleCusparseHandle(hypre_handle());
+
+  // FIXME: This is an abuse. Really, each matrix should have its own
+  //        rocsparse_mat_descr and rocsparse_mat_info and these should
+  //        not be global variables.
+  rocsparse_mat_descr descrA = hypre_HandleCusparseMatDescr(hypre_handle());
+  rocsparse_mat_descr descrB = hypre_HandleCusparseMatDescr(hypre_handle());
+  rocsparse_mat_descr descrC = hypre_HandleCusparseMatDescr(hypre_handle());
+  rocsparse_mat_info infoC = hypre_HandleRocsparseMatInfo(hypre_handle());
+
+  rocsparse_operation transA = rocsparse_operation_none;
+  rocsparse_operation transB = rocsparse_operation_none;
+
+  HYPRE_Int isDoublePrecision = sizeof(HYPRE_Complex) == sizeof(hypre_double);
+  HYPRE_Int isSinglePrecision = sizeof(HYPRE_Complex) == sizeof(hypre_double) / 2;
+
+  hypre_assert(isDoublePrecision || isSinglePrecision);
+
+  /* Copy the unsorted over as the initial "sorted" */
+  hypre_TMemcpy(d_ja_sorted, d_ja, HYPRE_Int,     nnzA, HYPRE_MEMORY_DEVICE, HYPRE_MEMORY_DEVICE);
+  hypre_TMemcpy(d_a_sorted,  d_a,  HYPRE_Complex, nnzA, HYPRE_MEMORY_DEVICE, HYPRE_MEMORY_DEVICE);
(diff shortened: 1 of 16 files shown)


src/seq_mv/csr_spgemm_device_rocsparse.c
```diff
@@ -0,0 +1,180 @@
...
+                                                           m, n, k,
+                                                           (float *) &alpha, // \alpha = 1
+                                                           descrA, nnzA, d_ia, d_ja_sorted,
+                                                           descrB, nnzB, d_ib, d_jb_sorted,
+                                                           NULL, // \beta = 0
+                                                           NULL,   0,    NULL, NULL,
+                                                           infoC, &rs_buffer_size) );
+   }
+
+   HYPRE_HIP_CALL(hipMalloc(&rs_buffer, rs_buffer_size));
```
liruipeng: @pbauman  Why not `hypre_TAlloc`?

src/seq_mv/csr_spgemm_device_rocsparse.c
```diff
@@ -0,0 +1,180 @@
...
+                                               descrB, nnzB, (float *) d_b_sorted, d_ib, d_jb_sorted,
+                                               NULL, // beta = 0
+                                               NULL,   0,    NULL,       NULL, NULL, // D is nothing
+                                               descrC,       (float *) d_c, d_ic, d_jc,
+                                               infoC, rs_buffer) );
+   }
+
+   // Free up the memory needed by rocsparse
+   HYPRE_HIP_CALL(hipFree(rs_buffer));
+
```
liruipeng, line 165: @pbauman  Why not `hypre_TFree`?

src/seq_mv/csr_sptrans_device.c
```diff
@@ -96,6 +86,66 @@ hypreDevice_CSRSpTransCusparse(HYPRE_Int   m,        HYPRE_Int   n,        HYPRE
...
+
+   size_t buffer_size = 0;
+   HYPRE_ROCSPARSE_CALL( rocsparse_csr2csc_buffer_size(handle,
+                                                       m, n, nnzA,
+                                                       csc_i, csc_j,
+                                                       action,
+                                                       &buffer_size) );
+
+   void * buffer;
+   HYPRE_HIP_CALL( hipMalloc(&buffer, buffer_size) );
```
liruipeng: @pbauman `hypre_TAlloc`?

src/seq_mv/csr_sptrans_device.c
```diff
@@ -96,6 +86,66 @@ hypreDevice_CSRSpTransCusparse(HYPRE_Int   m,        HYPRE_Int   n,        HYPRE
...
+   HYPRE_ROCSPARSE_CALL( rocsparse_dcsr2csc(handle,
+                                            m, n, nnzA,
+                                            d_aa, d_ia, d_ja,
+                                            csc_a, csc_j, csc_i,
+                                            action,
+                                            rocsparse_index_base_zero,
+                                            buffer) );
+   HYPRE_HIP_CALL( hipFree(buffer) );
+
```
liruipeng, line 134: @pbauman `hypre_TFree`?

src/seq_mv/csr_spgemm_device_rocsparse.c
pbauman: In this case, it was because this allocation didn't match the pattern of hypre_TAlloc (a count and a type). I wasn't sure if the preference would be to invoke hypre_MAlloc or just use the "native" API. Please let me know your preference and I'll update accordingly.

rfalgout: I would prefer to wrap this in a hypre call, and `hypre_MAlloc` seems appropriate.  The "memory tracker" stuff would not be available for these calls, but that looks like a debugging aid, so maybe not so important.

@liruipeng .  What do you think?  Also, why do we have `hypre_MAlloc` and `_hypre_MAlloc`?

rfalgout (approved):
Looks good to me.  Thanks!

ulrikeyang (approved):
This looks good, assuming the malloc/free questions are resolved.
Thank you.

src/seq_mv/csr_spgemm_device_rocsparse.c
liruipeng: > In this case, it was because this allocation didn't match the pattern of hypre_TAlloc (a count and a type). I wasn't sure if the preference would be to invoke hypre_MAlloc or just use the "native" API. Please let me know your preference and I'll update accordingly.

Not sure if I understand. Does this work, `rs_buffer = hypre_TAlloc(char, rs_buffer_size, HYPRE_MEMORY_DEVICE);`? I think we use this in many places where don't necessarily have a type but just want to allocate certain number of bytes. Thoughts?

rfalgout: This is fine with me.  Does the compiler complain about type mismatches?

liruipeng: > I would prefer to wrap this in a hypre call, and `hypre_MAlloc` seems appropriate. The "memory tracker" stuff would not be available for these calls, but that looks like a debugging aid, so maybe not so important.
> 
> @liruipeng . What do you think? Also, why do we have `hypre_MAlloc` and `_hypre_MAlloc`?

We prefer having all allocations via `hypre_TAlloc` so it goes through our memory model and can use memory pool potentially. `hypre_MAlloc` will also do, but as @rfalgout pointed out, it will not tracked by the "tracker", which is for debugging only. But I don't see where `hypre_MAlloc` works but `hypre_TAlloc` doesn't. 

`hypre_MAlloc` and `_hypre_MAlloc` were designed for different levels of memory space, conceptual for the former one and physical for the latter. In a few cases, we want to bypass the conceptual memory location but allocate on actual memory location, UVM vs device memory related, where we call `_hypre_MAlloc`. We also have the macro version of it, `_hypre_TAlloc`.

```
hypre_MAlloc(size_t size, HYPRE_MemoryLocation location)
_hypre_MAlloc(size_t size, hypre_MemoryLocation location)
```

liruipeng: > This is fine with me. Does the compiler complain about type mismatches?

Good point. It may not since `rs_buffer` is `void *`. If it does, a pointer casting should work I think.

rfalgout: Okay, thanks for the explanation @liruipeng .

pbauman: Hi @liruipeng you were correct, `rs_buffer = hypre_TAlloc(char, rs_buffer_size, HYPRE_MEMORY_DEVICE);` worked just fine. I have pushed that change in 6169f41. I've also grepped and made sure `hipMalloc` and `hipFree` only appear in `hypre_memory.c`. Sorry for the confusion on my part.

src/seq_mv/csr_spgemm_device_rocsparse.c
pbauman: Fixed in 6169f41

src/seq_mv/csr_sptrans_device.c
pbauman: Fixed in 6169f41

src/seq_mv/csr_sptrans_device.c
pbauman: Fixed in 6169f41

pbauman (approved):
I've pushed the malloc/free changes as suggested by @liruipeng. I tested with `--enable-debug` (which turns on `-Wall` on the HIP compiler as seen in this PR) and no warnings are emitted. I have no objection to the commits @liruipeng pushed on top of my PR (sorry about the white space, I'm trying to keep my emacs setting out of this, but I missed a few clearly). Thank you!

liruipeng:
@pbauman Thank you so much! I am now merging this PR.

PR #319, opened 2021-03-25
Roctx

liruipeng:
This PR (by @pbauman  #317) adds rocTX support, the ROCm analog to NVTX. There are basically three parts to this PR:

1. Change the configure option `--enable-nvtx` to `--enable-gpu-profiling` so that we hide the GPU profiling mode from the user and tie it to `--with-cuda` or `--with-hip`. And add `HYPRE_USING_ROCTX` which this is triggered and the user is using HIP.
2. Add the rocTX APIs alongside the NVTX APIs. Note that there is no analog `Color` API functionality in rocTX. This is because we visualize the timeline using Google Chrome (`chrome://tracing`). So I `#ifdef` around all the color management and just use what is necessary for the rocTX APIs in that case.
3. Change the API naming from `hypre_Nvtx` to `hypre_GpuProfiling` to help with the clarity in the library.

I tested this PR on both ROCm and CUDA 10.2 and confirmed the GPU profiling regions are captured.

This should be able to be applied independent of #304/#316 so it doesn't matter which gets merged first from my point-of-view. Thank you.

### DIFF
--- a/src/config/HYPRE_config.h.in
+++ b/src/config/HYPRE_config.h.in
@@ -223,6 +223,9 @@
 /* rocSPARSE being used */
 #undef HYPRE_USING_ROCSPARSE
 
+/* Define to 1 if using AMD rocTX profiling */
+#undef HYPRE_USING_ROCTX
+
 /* Define to 1 if using UMPIRE */
 #undef HYPRE_USING_UMPIRE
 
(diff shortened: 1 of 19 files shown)


liruipeng:
@pbauman Looks good to me. Thanks!

liruipeng:
Hi @rfalgout @ulrikeyang This one is good to go. Thanks!

rfalgout (approved):
Looks good.  Thanks!

ulrikeyang (approved):
This looks good. Thanks

PR #328, opened 2021-04-13
Fix colmap

victorapm:
When building the `col_map` array for prolongation operators that contain a high number of nonzero coefficients living in `P_offd`, see [hypre_build_interp_colmap](https://github.com/hypre-space/hypre/blob/0a0ddd0723fdbc1874e30553d6494408aac36174/src/parcsr_ls/aux_interp.c#L808), the code will make `P_offd_size * num_cols_P_offd` comparisons, and this can be very high depending on the test case (in one test case, I observed this quantity in the order of 10^11). This problem does not show up when enabling concurrent hopscotch since the code takes a different path and employs unordered sets to achieve the same objective. 

In this PR, I eliminate the usual code path that might be problematic and make the hopscotch one the new default. Note that this will work even when concurrent hopscotch is not enabled since the internal functions from hopscotch can handle the case when openmp support is turned off (this has been added with the amgdd branch). Tux tests are running fine.

For the test problem that I was looking at, these changes reduce setup time from 800 [s] to 20 [s].

Thanks!


src/utilities/hypre_merge_sort.c
```diff
@@ -613,7 +613,7 @@ void hypre_sort_and_create_inverse_map(HYPRE_Int *in, HYPRE_Int len, HYPRE_Int *
    hypre_merge_sort(in, temp, len, out);
    hypre_UnorderedIntMapCreate(inverse_map, 2*len, 16*hypre_NumThreads());
    HYPRE_Int i;
-#ifdef HYPRE_USING_OPENMP
```
liruipeng, line 616: @victorapm do you find if this function is used anywhere?

src/utilities/hypre_merge_sort.c
```diff
@@ -762,7 +762,7 @@ void hypre_big_sort_and_create_inverse_map(HYPRE_BigInt *in, HYPRE_Int len, HYPR
    hypre_big_merge_sort(in, temp, len, out);
    hypre_UnorderedBigIntMapCreate(inverse_map, 2*len, 16*hypre_NumThreads());
    HYPRE_Int i;
-#ifdef HYPRE_USING_OPENMP
```
liruipeng, line 765: Why do we want to restrict `omp parallelism` of this function with only hopscotch? It may not matter with the current code since this function is only called with `HYPRE_CONCURRENT_HOPSCOTCH`.

src/utilities/hypre_merge_sort.c
victorapm: Hi Ruipeng! Thanks for looking at this PR. I didn't understand your question. Can you elaborate?

liruipeng: Oh. Sorry @victorapm, I didn't make it clear. I just grep `hypre_sort_and_create_inverse_map` but did not find it used anywhere (maybe?). Just wanted to ask if you know where it is called. Thanks!

src/utilities/hypre_merge_sort.c
victorapm: I thought about keeping omp parallelism on this loop, however `hypre_UnorderedBigIntMapPutIfAbsent` is called inside this loop and it turns out that we need concurrent hopscotch turned on to have this loop omp parallelized in order to prevent the possibility of two threads trying to add the same key at the same time.

src/utilities/hypre_merge_sort.c
victorapm: Ah ok! Yes, you're correct. Only its `HYPRE_BigInt` counterpart (`hypre_big_sort_and_create_inverse_map`) is used in hypre. Maybe `hypre_sort_and_create_inverse_map` was being used before the addition of BigInt and it was not deleted after the changes?

src/utilities/hypre_merge_sort.c
victorapm: Moreover, this is a separate topic but I think we could make `HYPRE_CONCURRENT_HOPSCOTCH=1`  by default when using openmp in Hypre. Then the option `--enable-hopscotch` would become obsolete. Also, all the functions provided by hopscotch are ready to work when openmp is turned off, so hopscotch is already enabled by default in hypre.

liruipeng: Thanks for the clarification. I also looked at `hypre_UnorderedBigIntMapPutIfAbsent`. My impression was `CONCURRENT_HOPSCOTCH` is just an option of the implementation which doesn't affect outer loops that call it. If the calling loop can be parallelized only with concurrent hopscotch inside, we need to add some comments about it

src/utilities/hypre_merge_sort.c
liruipeng: Maybe, but it is still fine for me to leave it as is now

victorapm: Ok! Yes, it makes sense to leave it

src/utilities/hypre_merge_sort.c
victorapm: Yes, that's the case for enabling threaded parallelism on the calling loop. I agree that a comment about this should be added. Maybe at `hypre_hopscotch_hash.h` (where most of the hopscotch functions that can use threaded parallelism are defined)?

victorapm:
Hi all,

Since I'm changing the default code path for building `colmap` to address a corner problem, I wanted to make sure it doesn't hurt performance on usual problems. I made a few laplacian tests and concluded the change is not harmful, here are some results for a 3D 27-pt laplace problem, 1M DOFs/process, default input parameters for BoomerAMG except for `-Pmx=0` (`-Pmx=4` follows the same trend). On the left, results for master; on the right, results for fix-colmap.

|Nproc| Niter | T_prec      | T_solve   |
|-----|-------|-------------|-----------|
| 64  | 8 (8) | 39.2 (37.9) | 3.6 (3.5) |
| 128 | 8 (8) | 40.8 (40.7) | 3.4 (3.4) |
| 256 | 8 (8) | 40.7 (39.1) | 3.5 (3.5) |

Thanks!

rfalgout (approved):
Looks good to me.  Thanks!

liruipeng (approved):
looks good to me. Thanks!

ulrikeyang (approved):
Looks good.

Issue #309, opened 2021-03-18
Make GPU Matrix Descriptors Tied to Matrix


pbauman:
I alluded to this in some code comments when adding rocSPARSE support in #304, but currently the matrix descriptors, i.e. `rocsparse_mat_descr` and `rocsparse_mat_info` for rocSPARSE and `cusparseMatDescr_t` for cuSPARSE, are treated as global variables and not tied to any specific matrix instance. This will need to change in order to add some performance optimizations for rocSPARSE. But I think it's also the more correct thing to do. I would propose moving those objects to be stored in `hypre_CSRMatrix` and appropriate accessor APIs added there. I can cook up a PR for this, but wanted to open up discussion here before doing so to make sure everyone is on the same page and to make sure there isn't something I'm missing.

liruipeng:
@pbauman Thank you for checking with us. This is a correct thing to do; we actually wanted to do it too. If you don't mind, please make the change, and let us know if you need help on this. Thanks!

liruipeng:
We have addressed this issue in #331. So, I close it.

PR #331, opened 2021-04-14
Mat descr

liruipeng:
This PR (@pbauman #329) addresses #309. I anticipate there will be requested changes since I made some choices that the HYPRE developers may disagree with, but I figured it was easier/faster to see and fix diffs rather than discuss it without code. I've attempted to make each commit in this PR very small so it may be easier to review by each commit rather than looking at the final comparison.

This PR moves the `cusparseMatDescr_t`, `rocsparse_mat_descr`, and `rocsparse_mat_info` from being globals in the `hypre_CudaData` data structure to being part of the `hypre_CSRMatrix` class. This is done by putting them in a new `hypre_GpuMatData` struct (that is in `hypre_cuda_utils.h`, following the pattern used for `hypre_CsrsvData`) and then the `hypre_GpuMatData` is the thing that's added to `hypre_CSRMatrix`.

Then, the interfaces for `SpGEMM`, `MatVec`, and `CSRSort` are updated to pass `hypre_CSRMatrix` to the higher-level APIs, the MatDescr objects extracted, and then lower level APIs are updated to accept the MatDescr objects.

The choice that I anticipate may be cause for discussion is that, in order to avoid polluting the `seq_mv.h` header with the raw MatDescr types (since we're passing them to the new lower-level APIs), I created a new `seq_mv.hpp` header that is used in the right places. Of course there may be other objections as well.

I'm happy to update the PR as needed to address comments. Ultimately, this PR is needed to facilitate using some CSR analysis APIs that can substantially improve SpMV performance. Those changes will come in a subsequent PR once this PR is sorted out and merged. Thank you.


liruipeng:
@rfalgout @ulrikeyang @pbauman Passed all regression tests. It's good for me modulo some comments I made.

src/seq_mv/csr_matop_device.c
```diff
@@ -1322,12 +1343,9 @@ hypre_CSRMatrixTriLowerUpperSolveCusparse(char             uplo,
...
 #endif
 
-      hypre_SortCSRCusparse(nrow, ncol, nnzA, A_i, A_sj, A_sa);
+      hypre_SortCSRCusparse(nrow, ncol, nnzA, descr, A_i, A_sj, A_sa);
    }
 
-   cusparseHandle_t handle = hypre_HandleCusparseHandle(hypre_handle());
-   cusparseMatDescr_t descr = hypre_HandleCusparseMatDescr(hypre_handle());
-
    HYPRE_CUSPARSE_CALL( cusparseSetMatDiagType(descr, CUSPARSE_DIAG_TYPE_NON_UNIT) );
```
liruipeng, line 1349: `descr` of A is changed here (FILL_MODE, UNIT DIAG). But it seems that cusparse spmv doesn't pick them up. Not sure if rocsparse will do.

src/parcsr_ls/par_ilu_setup.c
```diff
@@ -7,6 +7,7 @@
 #include "_hypre_parcsr_ls.h"
 #include "_hypre_utilities.hpp"
 #include "par_ilu.h"
+#include "seq_mv.hpp"
```
rfalgout, line 10: I would like the `.hpp` header file inclusion to parallel the `.h` header inclusion, so ideally we should have an include of a single `_hypre_parcsr_ls.hpp` file here.  The hypre team can sort this out later, though, in a separate PR.

rfalgout (approved):
Looks good to me.  Thanks!

src/parcsr_ls/par_ilu_setup.c
liruipeng: Hi @rfalgout , `seq_mv.h` is included in `_hypre_parcsr_mv.h` which is included in `_hypre_parcsr_ls.h`. Since we have `seq_mv.hpp` now, shall we have `_hypre_parcsr_mv.hpp` and then `_hypre_parcsr_ls.hpp`?

rfalgout: Yes, that's the idea, but it's not essential that we do that in this PR.  Your call.  Thanks!

liruipeng: @rfalgout Thank you for confirming it.

ulrikeyang (approved):

src/seq_mv/csr_matop_device.c
pbauman: The APIs exist for doing this in rocSPARSE. But I'm unsure exactly what they affect.

pbauman (approved):
@liruipeng's changes on top of mine look good. In particular, I appreciate the update on the `hypreDevice_CSRSpGemm` API, I think it's an overall improvement. Thanks!

pbauman:
I failed to mention that I pulled this branch down and tested the `ij` test on my AMD GPU as
`./ij -n 96 96 96 -pmis -keepT 1 -rlx 18 -exec_device -rap 1 -mod_rap2 1 -interptype 6 -solver 1 -agg_nl 0` and the test behaved as expected. I did not try CUDA as it sounded like @liruipeng had already tried it. Thanks!

PR #336, opened 2021-04-16
Gpu default

liruipeng:
This PR changes AMG defaults regarding GPUs at various places, adds regression tests on GPUs, simplifies CUDA boxloop implementations. The details follow.

1. It removes `machine-ray.sh` and only keeps `machine-lassen.sh` that can be run on ray and lassen. No error threshold is needed on lassen any more but certain thresholds are needed on ray. I am not sure if we still want to run regression tests on both of the two machines. If so, we need to find a way to add the thresholds in the `.sh` (only when running on ray).
2. It removes `TEST_cuda_ray/*` and `TEST_cuda_lassen/*`, and adds two `benchmark.sh` in `TEST_ij` and `TEST_struct`. These benchmarks test large size problems, which are currently for measuring the GPU performance, used by `machine-lassen.sh`. When preparing future releases, we should watch the performance of these runs between releases.
3. It adds additional autoconf features. `--with-cuda-home` allows passing in CUDA home DIR. `--with-cuda-arch` is for CUDA SM number, which can be more than one number, such as `--with-cuda-arch="60 70"`. Changes default CUDA SM to 70 (V100, lassen). Changed `--enable-cub` to `--enable-device-memory-pool` for potentially resolving users' concern about having conflicts in their CUB and hypre's (which are none actually)
4. It silences printings from Pilut which can mess up regression tests. (I don't remember why it wasn't an issue before)
5. Adds a function that generate SoC needed by the GPU code from S that is created on the host
6. Changes AMG defaults for GPUs: `rap2, keepT, modu_rap, coarsen_type, rlx_type,agg_interp_type`. `ExecutionPolicy` now is `EXEC_DEVICE`.
7. Changes various places for `memory_location` to allocate on device memory
8. Changed default RNG on GPUs to `curand`. Adds "sequential rand" in `par_indepset_device.c`. @pbauman we need to add `rocrand` also.
9. Disabled all `SetDeviceLevel` in structured solvers (PFMG, SMG) 
10. Re-implemented some boxloop CUDA kernels (for simplicity)
11. Adds `.saved.cuda` files
12. Removes `amgdd` from GPU regression tests. Results are incorrect with CUDA-10. Seg-fault with CUDA-11. 
13. Corresponding changes in the drivers in `src/test/`
14. Various format changes

@pbauman We would like to set up some GPU regression tests (AMG only for now) using AMD GPUs on redwood, probably only the jobs in `TEST_ij/benchmark.sh`. Does it sound good to you?


src/utilities/hypre_cuda_utils.c
```diff
@@ -894,6 +925,63 @@ hypre_CudaDataCurandGenerator(hypre_CudaData *data)
...
+template <typename T>
+HYPRE_Int
+hypre_CurandUniform_core( HYPRE_Int          n,
+                          T                 *urand,
+                          HYPRE_Int          set_seed,
+                          hypre_ulonglongint seed,
+                          HYPRE_Int          set_offset,
+                          hypre_ulonglongint offset)
+{
+   curandGenerator_t gen = hypre_HandleCurandGenerator(hypre_handle());
```
liruipeng, line 939: @pbauman We need to add `rocRand`

src/struct_ls/_hypre_struct_ls.h
```diff
@@ -605,6 +603,13 @@ hypre_StructMatrix *hypre_SparseMSG2CreateRAPOp ( hypre_StructMatrix *R , hypre_
 HYPRE_Int hypre_SparseMSG2BuildRAPSym ( hypre_StructMatrix *A , hypre_StructMatrix *P , hypre_StructMatrix *R , HYPRE_Int cdir , hypre_Index cindex , hypre_Index cstride , hypre_Index stridePR , hypre_StructMatrix *RAP );
 HYPRE_Int hypre_SparseMSG2BuildRAPNoSym ( hypre_StructMatrix *A , hypre_StructMatrix *P , hypre_StructMatrix *R , HYPRE_Int cdir , hypre_Index cindex , hypre_Index cstride , hypre_Index stridePR , hypre_StructMatrix *RAP );
 
+/******************************************************************************
+ * Copyright 1998-2019 Lawrence Livermore National Security, LLC and other
+ * HYPRE Project Developers. See the top-level COPYRIGHT file for details.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR MIT)
+ ******************************************************************************/
+
```
ulrikeyang, line 409: Why are the headers in this file?

src/parcsr_mv/par_csr_matvec.c
```diff
@@ -146,8 +146,11 @@ hypre_ParCSRMatrixMatvecOutOfPlace( HYPRE_Complex       alpha,
    {
       if (!hypre_ParCSRCommPkgTmpData(comm_pkg))
       {
-         /* hypre_ParCSRCommPkgTmpData(comm_pkg) = hypre_TAlloc(HYPRE_Complex, num_cols_offd, HYPRE_MEMORY_DEVICE); */
+#if 1
+         hypre_ParCSRCommPkgTmpData(comm_pkg) = hypre_TAlloc(HYPRE_Complex, num_cols_offd, HYPRE_MEMORY_DEVICE);
+#else
          hypre_ParCSRCommPkgTmpData(comm_pkg) = _hypre_TAlloc(HYPRE_Complex, num_cols_offd, hypre_MEMORY_DEVICE);
+#endif
```
ulrikeyang, line 153: What does _hypre_TAlloc do compared to hypre_TAlloc?

src/struct_ls/_hypre_struct_ls.h
liruipeng: Are you asking about the "Copyright" disclaimer in this file? This file is now generated by `headers`

src/parcsr_mv/par_csr_matvec.c
liruipeng: `hypre_TAlloc` is the macro for normal
```
void * hypre_MAlloc(size_t size, HYPRE_MemoryLocation location);
```
Similarly, `_hypre_TAlloc` is for
```
void * _hypre_MAlloc(size_t size, hypre_MemoryLocation location);
```
The former one is for "conceptual" memory spaces (HYPRE_MemoryLocation) and the latter one is for physical spaces (hypre_MemoryLocation). Both are supposed for developers. In the case of hypre with UVM, the latter one can be used for allocating device memory since the former one will return UVM. Since we have removed dependencies of UVM, I changed them back to `hypre_TAlloc`. The name starting with `-` suggests that it is really an internal function. One needs to know what they are doing with it.

ulrikeyang:
Yes.

src/struct_ls/_hypre_struct_ls.h
rfalgout: Hi @liruipeng .  If you are introducing `headers` and `protos.h`, then you can remove the extra license header in `protos.h` and also all of the repeated prototypes for the capital `HYPRE_*` files that appear in `HYPRE_struct_ls.h`.  We can also do this later since this has been lying around for awhile already.  Thanks!

liruipeng: @ulrikeyang @rfalgout I made the changes, removing extra license header and `HYPRE_*` prototypes in `protos.h`, also  adding some of the `HYPRE_*` prototypes into `HYPRE_struct_ls.h`. Please take another look. Thanks!

rfalgout: Thanks!  -Rob

src/struct_ls/HYPRE_struct_pfmg.c
```diff
@@ -314,10 +314,10 @@ HYPRE_StructPFMGGetFinalRelativeResidualNorm( HYPRE_StructSolver  solver,
    return( hypre_PFMGGetFinalRelativeResidualNorm( (void *) solver, norm ) );
 }
 
-#if defined(HYPRE_USING_CUDA) || defined(HYPRE_USING_HIP)
+#if 0 //defined(HYPRE_USING_CUDA) || defined(HYPRE_USING_HIP)
```
rfalgout, line 317: Is this correct?

rfalgout: Never mind.  I guess this will be introduced later.

src/struct_ls/cyclic_reduction.c
```diff
@@ -501,7 +501,7 @@ hypre_CyclicReductionSetup( void               *cyc_red_vdata,
    HYPRE_Int               flop_divisor;
 
    HYPRE_Int               x_num_ghost[] = {0, 0, 0, 0, 0, 0};
-#if defined(HYPRE_USING_CUDA) || defined(HYPRE_USING_HIP)
+#if 0 //defined(HYPRE_USING_CUDA) || defined(HYPRE_USING_HIP)
```
rfalgout, line 504: Okay, so now I am confused again.  Why is this in master but being commented out here?

AUTOTEST/runtests-ij
```diff
@@ -1,2 +1,2 @@
-TEST_ij/*.sh
+TEST_ij/[^b]*.sh
```
rfalgout: The "standard" test directories like `TEST_ij`, `TEST_struct`, `TEST_sstruct` should only contain tests that represent "default" tests that are always run on the Linux boxes when testing new code.  I regularly do `runtest.sh TEST_ij/*.sh` to do a quick test of the IJ-related solvers when I am modifying code.  Same for the other two directories.  Once I get these basic tests working, then I usually move on to running the full suite of tests in the `AUTOTEST` directory.

So, I think these new benchmark tests should probably go in a new directory, say `TEST_bench` or something.

src/struct_ls/cyclic_reduction.c
liruipeng: @rfalgout @ulrikeyang This, and the other "#if 0" in the PR, were intended to completely disable an (incomplete) feature of having "CPU-GPU" switching levels in SMG and PFMG (the `data_location` is for that). We introduced it long time ago but it seems not always working (have not investigated to figure out why). We commented it out its usage in the driver and not suggested it. But some smart user figured out its meaning, tried to use it, and had seg-fault issue. I just want to #ifdef it out for now. If you don't like it, we change it back. Thanks!

AUTOTEST/runtests-ij
liruipeng: @rfalgout This makes more sense. Sorry about that. I will change it back. That `.sh` was in `TEST_cuda` and I remembered we talked about directories like it or `TEST_single, TEST_longdouble` could be combined with `TEST_ij` after we had `-save` option in `runtest.sh`. That was why I moved it. But now I can see why I should not have done that. Thanks!

rfalgout:
Hi @liruipeng .  Everything looks good so far to me.  I have a few high level questions/comments:

- How do you expect these changes will affect our users, especially since some defaults are being modified?
- Did you compare the new cuda saved files to the Linux ones, just to make sure they seem reasonable?
- You will need to merge master into this, but you knew that already.
- I'll need to remember to remove `ray` from my crontab.
- It would be nice to have a way to automatically check the benchmark performance numbers in AUTOTEST.  We have talked about this sort of thing in the past, but never completely followed through.  Maybe Daniel's `runcheck` scripting work could help for comparing timings?

Thanks!

src/struct_ls/cyclic_reduction.c
rfalgout: Okay, I understand.  Basically, this is a feature in waiting that will eventually either be added back in or abandoned.  This is fine.  Thanks!

liruipeng:
> Hi @liruipeng . Everything looks good so far to me. I have a few high level questions/comments:
> 
> * How do you expect these changes will affect our users, especially since some defaults are being modified?

These are all for GPUs. If the users are using the recommended (non-default) parameters, they should see basically the same performance. If they are not, they should see better performance in the next release with default settings.

> * Did you compare the new cuda saved files to the Linux ones, just to make sure they seem reasonable?

I remember they look OK when I created them. We found bugs together in the code with comparing with CPUs. I will look through again when I do the regression tests.

> * You will need to merge master into this, but you knew that already.

Will do.

> * I'll need to remember to remove `ray` from my crontab.

Yes, thanks for reminding.
> * It would be nice to have a way to automatically check the benchmark performance numbers in AUTOTEST.  We have talked about this sort of thing in the past, but never completely followed through.  Maybe Daniel's `runcheck` scripting work could help for comparing timings?

This is absolutely nice to have. I will think about how to do it. Thanks!
> 
> Thanks!

rfalgout (approved):
Looks good.  Thanks!

ulrikeyang (approved):
This looks fine.

rfalgout:
Hi @liruipeng .  These benchmark tests look great!  I assume you will change `machine-lassen.sh` to run them separately with a tolerance in the diff?  I think these benchmark tests could be a really good thing, and if it works out, we may want to test performance elsewhere as well (not just on GPUs).  Thanks!

liruipeng:
> Hi @liruipeng . These benchmark tests look great! I assume you will change `machine-lassen.sh` to run them separately with a tolerance in the diff? I think these benchmark tests could be a really good thing, and if it works out, we may want to test performance elsewhere as well (not just on GPUs). Thanks!

Hi, @rfalgout I wanted to add `./runcheck.sh TEST_bench/benchmark_struct.perf TEST_bench/benchmark_struct.perf.lassen rtol atol` somewhere. I am not sure if I should add it in `runtest.sh` or `TEST_bench/benchmark_struct.sh` and `TEST_bench/benchmark_ij.sh`, or use some other way to do it. You know, I want to have zero tolerance for `.out` and `.saved` but some nonzero tolerance for `.perf` and `perf.lassen`. We need to figure out a proper tolerance too. Any thoughts? Thanks!

rfalgout:
Hi @liruipeng .  Why not name the files `*.saved.perf.lassen` and use the rtol/atol feature already in `runtest.sh`?  You could put the actual tests in the existing `machine-lassen.sh` file.  Maybe I am missing something, though?

I was thinking we should check for anything greater than, say, 10-20% differences?  What do you and others think?  We might need to gain some experience with this to settle on a reasonable number.

Thanks!

rfalgout:
Hi @liruipeng .  Just to add a bit more to my previous comment.  I see that you are checking the correctness of the runs in the benchmark tests like we normally do in the other tests.  I don't think that's really necessary, since we already do quite a bit of correctness testing.  If you make these tests strictly about performance, then I think you can do what I suggested.  Make sense?

liruipeng:
Hi @rfalgout I think I may still want to test the correctness of the jobs in `TEST_bench`: 1. These are with recommended AMG parameters for GPUs. I am not sure if all the combinations are tested at other places like `TEST_ij` 2. These are bigger problems. Sometimes we need to have some big problems to uncover some GPU issues (inconsistent results from run to run, race condition, etc). Thanks!

rfalgout:
Okay.  Then I think the best place to add this right now is in `benchmark_struct.sh` etc.  If we find that we want to test the performance of most of the tests in hypre, then it may make sense to add another general feature to `runtest.sh`, but I don't think that will be the case.

If you want to use the same benchmark tests to do timings on different machines, you could add a case statement to `benchmark_struct.sh` (for example) that branches based on `hostname` or on an environment variable that you pass into the test via the `runtest.sh -D` option.

Just my 2 cents.  Thanks again!

liruipeng:
> Okay. Then I think the best place to add this right now is in `benchmark_struct.sh` etc. If we find that we want to test the performance of most of the tests in hypre, then it may make sense to add another general feature to `runtest.sh`, but I don't think that will be the case.
> 
> If you want to use the same benchmark tests to do timings on different machines, you could add a case statement to `benchmark_struct.sh` (for example) that branches based on `hostname` or on an environment variable that you pass into the test via the `runtest.sh -D` option.
> 
> Just my 2 cents. Thanks again!

Thanks, @rfalgout . I came up with something to compare performance in `.perf.out` and `.perf.saved.lassen`. Please take a look and let me know if this can work.

rfalgout:
Hi @liruipeng .  This looks good, except that I don't see the need for the `prtol` feature.  The tolerance can just be hardcoded into the `lassen` case in `benchmark_*.sh` since it is lassen specific.  Even if there is a good argument to pass the tolerance through `runtest.sh`, I also don't see why you can't use the existing `rtol` feature.  Maybe I am missing something, though.

My feeling is that performance runs and tolerances will be much more correlated and tailored to the specific machines, so it doesn't hurt to put the tolerances in the tests themselves.

My 2 cents again. :)  Thanks!

liruipeng:
> Hi @liruipeng . This looks good, except that I don't see the need for the `prtol` feature. The tolerance can just be hardcoded into the `lassen` case in `benchmark_*.sh` since it is lassen specific. 

I agree. Put a hard-coded number in `benchmark_*.sh` makes sense.

>Even if there is a good argument to pass the tolerance through `runtest.sh`, I also don't see why you can't use the existing `rtol` feature. Maybe I am missing something, though.

I thought about that but I could be wrong. If from `machine-lassen.sh` I pass in `-rtol x`, `x` will be used in `runtest.sh`  to test correctness         
```
(../runcheck.sh $TestName.out $SaveName $RTOL $ATOL >> $TestName.err 2>&1)
 ```
But I also want to have zero tolerance for correctness. This system is complicated to me, so I might be lost somewhere.

rfalgout:
I agree.  The system is a bit complicated.  It's been a constant struggle between keeping it as simple as possible and providing flexibility.  This was part of the reason I did not want to add another feature to runtest.  Regarding `rtol`, you are right.  I realized it shortly after I sent my last comment.  I still don't think we should add a `prtol` feature, though.

I will think more about how we might make the whole system a bit simpler without losing our current flexibility.  I am not totally happy with the organization of the recent saved-file stuff I added, for example.  If there are particular things you (or others) find confusing about the system, let me know.  Thanks again!

liruipeng:
- [x] Pass `machine.tux.sh` 
- [x] Pass `machine.lassen.sh`
- [x] Run `machine.ray.sh`
- [x] Compile on `redwood` and run `ij`

PR #341, opened 2021-04-20
Add OMP support to Mat/Mat add functions

victorapm:
This PR adds OpenMP support to `CSRMatrixAddHost` and `ParCSRMatrixAdd`. It was motivated by some performance tests showing that these two functions can become a bottleneck when running BoomerAMG with OpenMP and specific input options, such as `-rap 1 -mod_rap2 1`. The new `SStructMatmult` from `recmat` and other preconditioners, such as `hypre_ILU`, will benefit from this also. This PR depends on #300. Tux tests are running fine. Minor changes are:
1) Changed name `ParcsrAdd` to `ParCSRMatrixAdd`
2) Add `hypre_CSRMatrixAddFirstPass` and `hypre_CSRMatrixAddSecondPass` to reduce code duplication
3) Update `rownnz` support in `CSRMatrixAddHost`, `ParCSRMatrixAdd` and `hypre_ILUParCSRInverseNSH`.


rfalgout:
I will wait until #300 is finished before looking at this, since many of the changes are the same.

rfalgout:
Hi @victorapm .  Once you merge master into this branch, I'll take a look at the PR.  Thanks!

victorapm:
Thanks, Rob! All, this is up to date with master and ready for review

src/IJ_mv/IJMatrix_parcsr.c
```diff
@@ -2664,7 +2664,7 @@ hypre_IJMatrixAssembleParCSR(hypre_IJMatrix *matrix)
                   }
                   else if (j != diag_pos[i])
                   {
-                     diag_j[i_diag] = (HYPRE_Int)(local_j[j] - col_0);
+                     diag_j[i_diag] = local_j[j] - col_0;
```
liruipeng: Integer cast missing

victorapm: Thanks for catching! Fixed with c63928c

src/parcsr_mv/par_csr_matop.c
```diff
@@ -5048,79 +5053,53 @@ hypre_ParcsrAdd( HYPRE_Complex        alpha,
...
-
-   HYPRE_Int             num_procs, my_id;
-   HYPRE_Int             i, ii, j;
+   HYPRE_Int             num_rownnz_diag_C;
+   HYPRE_Int             num_rownnz_offd_C;
+   HYPRE_Int             num_rows_diag_C = num_rows_diag_A;
+   HYPRE_Int             num_cols_diag_C = num_cols_diag_A;
+   HYPRE_Int             num_rows_offd_C = num_rows_offd_A;
+   HYPRE_Int             num_cols_offd_C = num_cols_offd_A + num_cols_offd_B;
+   HYPRE_Int             twspace[hypre_NumThreads()];
```
liruipeng: may use `malloc` rather than variable length array

rfalgout (approved):
This all looks good to me.  Thanks!

src/parcsr_mv/par_csr_matop.c
victorapm: The use of `malloc` is preferred since VLA is not in C++ standard, right (just checking)? This is used here as well: https://github.com/hypre-space/hypre/blob/3f12d47651ecf18644d95bc9e975c572fe44ccf3/src/parcsr_ls/aux_interp.c#L683

I'll change these to malloc

liruipeng: Not familiar with VLA in C/C++. I just found out it is indeed in C99. (I thought this sort of allocation is not allowed in C...). So, I don't have strong opinions on it. Maybe other people can comment on it.

victorapm: Ok! I decided to change to malloc anyway. If we ever decide to use a C++ compiler in these files, it will not create problems. And this is a very small allocation happening very few times, so there aren't benefits of using stack over heap memory

rfalgout: I agree with changing to malloc anyway.  Thanks!

victorapm: Good! Thanks, Rob!

src/seq_mv/csr_matop.c
```diff
@@ -14,6 +14,325 @@
...
+                ( map_B2C && (nnzs_A == 0)));
+
+   /* Initialize marker vector */
+   for (ia = 0; ia < ncols_C; ia++)
+   {
+      marker[ia] = -1;
+   }
+
+   pos = C_i[rownnz_C ? rownnz_C[firstrow] : firstrow];
+   if ((map_A2C && map_B2C) || ( map_A2C && (nnzs_B == 0)) || ( map_B2C && (nnzs_A == 0)))
```
liruipeng, line 263: Can we combine the if-branch with the else-, having `jcol = map_A2C? map_A2C[A_j[ia]] : A_j[ia]` and similar for B? Just to simplify.

src/parcsr_mv/par_csr_matop.c
```diff
@@ -322,6 +322,14 @@ hypre_ParMatmul_RowSizes( HYPRE_MemoryLocation memory_location,
    *C_diag_size = (*C_diag_i)[num_rows_diag_A];
    *C_offd_size = (*C_offd_i)[num_rows_diag_A];
 
+#ifdef HYPRE_DEBUG
+   for (HYPRE_Int i = 0; i < num_rows_diag_A; i++)
```
liruipeng: Should avoid C++ style for-loop declaration for `i`.

src/parcsr_mv/par_csr_matop.c
```diff
@@ -5265,125 +5155,108 @@ hypre_ParcsrAdd( HYPRE_Complex        alpha,
...
+
+   /* Free memory */
+   hypre_TFree(twspace, HYPRE_MEMORY_HOST);
+   hypre_TFree(A2C_offd, HYPRE_MEMORY_HOST);
+   hypre_TFree(B2C_offd, HYPRE_MEMORY_HOST);
+
+   /* Create ParCSRMatrix C */
    row_starts_C = hypre_TAlloc(HYPRE_BigInt, 2, HYPRE_MEMORY_HOST);
    col_starts_C = hypre_TAlloc(HYPRE_BigInt, 2, HYPRE_MEMORY_HOST);
    memcpy(row_starts_C, hypre_ParCSRMatrixRowStarts(A), 2*sizeof(HYPRE_BigInt));
```
liruipeng: Maybe better use the one in hypre's memory model: `hypre_TMemcpy` Host To Host. It has the same effect as `memcpy` for now.

src/seq_mv/csr_matop.c
```diff
@@ -224,75 +524,140 @@ hypre_CSRMatrixBigAdd( hypre_CSRMatrix *A,
...
-         pos++;
+         C_i[nrows_A] = 0;
+         for (ic = 0; ic < num_threads; ic++)
+         {
+            C_i[nrows_A] += twspace[ic];
+         }
+
+         C = hypre_CSRMatrixCreate(nrows_A, ncols_A, C_i[nrows_A]);
+         hypre_CSRMatrixI(C) = C_i;
+         hypre_CSRMatrixInitialize_v2(C, 0, memory_location_C);
```
liruipeng: Why `0`?

src/parcsr_mv/par_csr_matop.c
victorapm: Hi Ruipeng, This is [C99](https://en.cppreference.com/w/c/language/for), but I can definitely change in order to be aligned to the style we use in hypre.

src/seq_mv/csr_matop.c
victorapm: I thought about doing this and decided to do the other way. The problem with `jcol = map_A2C? map_A2C[A_j[ia]] : A_j[ia]` is that the conditional statement is evaluated `num_nonzero_A` times and this can be quite big. In the way I'm currently doing, the conditional statements on `map_A2C` and `map_B2C` are evaluated only once. Do you think the more compact form should be used? I'm happy to change it if so

src/parcsr_mv/par_csr_matop.c
victorapm: Yes, agreed! I realized memcpy is used in other parts of the code as well. Maybe we need to update those

src/seq_mv/csr_matop.c
victorapm: This is wrong, thanks for catching!

src/seq_mv/csr_matop.c
liruipeng: OK. Thanks for explaining. I think we should keep the current form. The cost of  the conditional statements could be non-negligible. Thanks!

src/parcsr_mv/par_csr_matop.c
liruipeng: I don't think we use this style in hypre. Maybe just define `HYPRE_Int i` outside?

victorapm: Yes, will do it :)

liruipeng (approved):
Looks good to me. Thanks!

victorapm:
Hi @oseikuffuor1, I updated the use of memcpy in `hypre_ILU` to be in line with Ruipeng's comment:
> Maybe better use the one in hypre's memory model: `hypre_TMemcpy` Host To Host. It has the same effect as `memcpy` for now.

Can you take a look just to be sure I didn't create any problems? Tux tests are passing after the last changes.

Thanks!

oseikuffuor1 (approved):
Looks good to me. Thanks @victorapm

ulrikeyang (approved):
This looks fine

PR #349, opened 2021-04-27
Cmake cuda updates

osborn9:
These changes add (some) CUDA options for a CMake configure / build. 

Please take a look and let know what other CUDA/GPU features that should be added, etc, etc. I also can provide a sample cmake invocation script that I was using on Lassen for testing. 

Other minor changes include using ${PROJECT_NAME} when possible instead of HYPRE/hypre, and having the CMake project name be HYPRE instead of hypre (open for debate).

Some TODOs
- [x] Add a CUDA cmake test (for lassen)? 
- [x] Double check HYPRE_config.h produced by CMake is comparable to configure (some inconsistencies with #define foo 1 vs #define foo)


src/CMakeLists.txt
```diff
@@ -78,214 +89,309 @@ option(TPL_LAPACK_LIBRARIES          "Optional list of absolute paths to LAPACK
...
+    if (NOT CMAKE_CUDA_STANDARD OR CMAKE_CUDA_STANDARD EQUAL 98)
+      set(CMAKE_CUDA_STANDARD 11)
+    endif ()
+
+    set(CMAKE_CUDA_STANDARD_REQUIRED TRUE)
+
+    set(HYPRE_USING_CUDA ON CACHE BOOL "" FORCE)
+    set(HYPRE_USING_GPU ON CACHE BOOL "" FORCE)
+
+    if (HYPRE_ENABLE_UNIFIED_MEMORY)
```
osborn9, line 213: @liruipeng Is this the correct logic for the `*_MEMORY` variable?

victorapm (approved):
Thank you, @osborn9! I successfully built hypre on lassen with GPU support.

src/CMakeLists.txt
liruipeng: This looks good .

ulrikeyang (approved):
It looks good as far as I can tell, but I am not that familiar.

rfalgout (approved):

liruipeng (approved):
Looks good to me (modulo the conflicting `HYPRE_config.h` issue @osborn9 and I talked) . Thanks!

osborn9:
> Looks good to me (modulo the conflicting `HYPRE_config.h` issue @osborn9 and I talked) . Thanks!

@liruipeng I was unable to replicate that behavior, so I'm not sure what was going on there. I double checked how the cmake is treating the configure file and it looks ok to me. 

FYI, The issue that happened to Ruipeng was the cmake build used the `HYPRE_config.h` from src/ (generated via autoconf from a previous configure), instead of the cmake generated one in src/cmbuild.

liruipeng:
> > Looks good to me (modulo the conflicting `HYPRE_config.h` issue @osborn9 and I talked) . Thanks!
> 
> @liruipeng I was unable to replicate that behavior, so I'm not sure what was going on there. I double checked how the cmake is treating the configure file and it looks ok to me.
> 
> FYI, The issue that happened to Ruipeng was the cmake build used the `HYPRE_config.h` from src/ (generated via autoconf from a previous configure), instead of the cmake generated one in src/cmbuild.

Sounds good. I am not sure either. It might be just some cmake files having been messed up in that folder. It is not a big issue I think. Thank you!

oseikuffuor1 (approved):
Looks very good to me Sarah. Thanks.

PR #352, opened 2021-05-03
Thrust 1.10 Fixes

pbauman:
This PR addresses Thrust 1.10 breaking changes. Thrust 1.10 landed in CUDA 11.2 and will land in ROCm 4.2.

As discussed in NVIDIA/thrust#1379 (that internally references NVIDIA/thrust#1176), the behavior of `exclusive_scan` and `inclusive_scan` changed in the case where the input types and output types were not the same. There is no deprecation warning or error thrown by the compiler. Indeed, with Thrust 1.10, before this PR, the `exclusive_scan` calls that had `make_transform_iterator` used in the input types (silently!) generated incorrect results. *That means HYPRE is broken on GPUs today with CUDA 11.2 without this PR.* There are a couple of way to fix, but what I did for the fix for `exclusive_scan` was to just use the API where one specifies the initial value and that was enough to fix the issue. It did not appear to me that there were any `inclusive_scan` calls affected in HYPRE. And my tests with this PR with a ROCm 4.2 release candidate pass.

In addition, Thrust 1.10 deprecated the use of C++ before C++14 so I've added `-std=c++14` to the `HIPCXXFLAGS` argument in the `configure.in` (and bootstrapped). 

Thrust 1.12 introduces similar breakages for the `scan_by_key` cousins, see NVIDIA/thrust#1376. The fixes are similar and I dropped in explicit casts to `HYPRE_BigInt` in the (already existing) initial value for `exclusive_scan_by_key` (commit 9a3bb66). I've not tried to address `inclusive_scan_by_key` cases, but I do believe they will be broken. I strongly recommend adding unit tests for those calls and adding them to the test suite. Thrust 1.12 is supposedly going to land in CUDA 11.4 (at least according to that thrust release page). I do not know when it will land in a ROCm release.

Thank you.

### DIFF
--- a/src/IJ_mv/IJMatrix_parcsr_device.c
+++ b/src/IJ_mv/IJMatrix_parcsr_device.c
@@ -228,7 +228,7 @@ hypre_IJMatrixAssembleSortAndReduce1(HYPRE_Int  N0, HYPRE_BigInt  *I0, HYPRE_Big
          make_reverse_iterator(thrust::make_zip_iterator(thrust::make_tuple(I0,    J0))),
          make_reverse_iterator(thrust::device_pointer_cast<char>(X0)+N0),
          make_reverse_iterator(thrust::device_pointer_cast<char>(X) +N0),
-         0,
+         HYPRE_BigInt(0),
          thrust::equal_to< thrust::tuple<HYPRE_BigInt, HYPRE_BigInt> >(),
          thrust::maximum<char>() );
 
--- a/src/IJ_mv/IJVector_parcsr_device.c
+++ b/src/IJ_mv/IJVector_parcsr_device.c
@@ -276,7 +276,7 @@ hypre_IJVectorAssembleSortAndReduce1(HYPRE_Int  N0, HYPRE_BigInt  *I0, char  *X0
          make_reverse_iterator(thrust::device_pointer_cast<HYPRE_BigInt>(I0)),     /* key end */
          make_reverse_iterator(thrust::device_pointer_cast<char>(X0)+N0),          /* input value begin */
          make_reverse_iterator(thrust::device_pointer_cast<char>(X) +N0),          /* output value begin */
-         0,                                                                        /* init */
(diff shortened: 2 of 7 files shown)


liruipeng:
@pbauman Thank you for fixing this issue, which would be very hard to find without knowing the behavior change of thrust. For future development, do you know where we should keep looking at for such kind of changes? We have been using https://thrust.github.io/doc/index.html for main reference of thrust, which has been apparently not up to date. 

About C++14, we are aware of the fact C++11 has been deprecated by thrust >= 11.0. Currently we just ignore the warnings (https://github.com/hypre-space/hypre/blob/3f12d47651ecf18644d95bc9e975c572fe44ccf3/src/utilities/hypre_cuda_utils.h#L29). Probably we should change hypre's required C++ version to C++14. Any ideas on this? 

Thanks!

pbauman:
> For future development, do you know where we should keep looking at for such kind of changes?

Honestly, I don't have a good answer. It took me some time to narrow down the issue and then some additional digging by me and a member of our libraries team to ultimately find the GitHub issue I linked. In particular, the first change wasn't even documented in the release notes of Thrust, so there was no way to know without essentially watching every commit to Thrust. It's very frustrating and unnerving, to say the least. Unfortunately, I don't have any good ideas.

> Probably we should change hypre's required C++ version to C++14. Any ideas on this?

Ultimately, that's up to the HYPRE developers. But, as far as from the ROCm/HIP side, the compiler supports C++17 I believe, so requiring C++14 there is not a problem; furthermore, ROCm 4.1 is the minimum version that's really supported here in HYPRE because there was a compiler bug prior to that release that generated incorrect code in one of the kernels (when using `-O2` or higher). As far as I understand, C++ only infiltrates HYPRE in the GPU code paths, so I guess the question to ask would be how far back in time do you want to support CUDA and then, once that's determined, what C++ standard is supported by that compiler? Just my two cents.

Issue #332, opened 2021-04-14
batch solution of N small linear systems under the same MPI communicator?


p-costa:
Dear all,

I am solving a Poisson equation of this kind, for variable $\Phi$ with $f$ the right-hand-side vector using the `Struct` interface, `PFMG`:

![eq1](https://user-images.githubusercontent.com/29615530/114690719-82843880-9d06-11eb-863b-3aeb671ed600.png)
 
However, to speedup the solution, I use a FFT-based expansion to decouple the equations in one homogeneous direction (here the direction pertaining to index `j`), resulting in the following Helmholtz equation:

![eq2](https://user-images.githubusercontent.com/29615530/114690768-8ca63700-9d06-11eb-8e08-e371854a688a.png)

Where I need to have the tasks distributed in a pencil-like decomposition so that the FFTs in one direction can be performed (figure below, left). Then I solve $N$ independent 2D linear systems along the homogeneous direction where I perform this FFT-based reduction (using the `Struct` interface, `PFMG`).
 
I get the expected major speedup for a small number of cores. However, each of these small 2D systems has to be solved over `MPI_COMM_WORLD`, and tasks have to sync between each solve, which compromise the parallel scalability. My solution right now, which works like a charm, is to redistribute the tasks in a slab-like domain decomposition as illustrated in the following figure, and solve these $N$ 2D systems with the parallelization of the right, over a new MPI communicator encapsulating the tasks responsible for each of the small systems. This way the solution of the independent 2D systems is distributed.

![all-to-all](https://user-images.githubusercontent.com/29615530/114690808-962f9f00-9d06-11eb-81ee-b279c8d0dd52.png)
 
This redistribution, however, involves an *all-to-all* collective operation (actually two, since I need to perform the backward FFT-based transform to get back the solution), which is something that I feel could be avoided if a batch solution of $N$ systems would be supported by *HYPRE*. (Also, I have a hard limit of the number of tasks I can decompose among *slabs* dictated by the number of grid cells in that direction.)
 
So after this long explanation, which I hope was clear, my short question is -- do you see it feasible/likely that *HYPRE* will support a batch solution of this kind in the future for use cases like this? Or perhaps there is a workaround that can be explored using the current API?
 
Thank you in advance!
Pedro

rfalgout:
[...]

PR #355, opened 2021-05-06
Fix a corner-case divide-by-zero in PFMG (see github issue #332)

rfalgout:
- The 'mean' computed in PFMG can be zero, so don't divide by it.
- Added comments to better explain this part of the code.
- Added regression tests to check 3D problems that have all of the coefficients
  zeroed out in one dimension (so they are basically 2D).


victorapm (approved):
Thank you, Rob! I'll add tests for SSAMG as well when we merge this to recmat

ulrikeyang (approved):
This looks good.

PR #364, opened 2021-05-13
Fix so this compiles with hip

PaulMullowney:


pbauman:
@liruipeng Does that API get used else where in HYPRE? Perhaps it's useful to add a shim for this `GpuDeviceReset`, for example? Sorry for the trouble @PaulMullowney, I've only been using the `ij` test.

liruipeng:
> @liruipeng Does that API get used else where in HYPRE? Perhaps it's useful to add a shim for this `GpuDeviceReset`, for example? Sorry for the trouble @PaulMullowney, I've only been using the `ij` test.

@pbauman No, it is only used in the (two) drivers. The reason we put it was to correctly run `nvprof` with the driver. So, I think this fix is fine since this function is not part of the library.

pbauman (approved):
Thanks!

PR #366, opened 2021-05-14
Adding a new HYPRE_RELEASE_NUMBER macro

rfalgout:
The HYPRE_RELEASE_NUMBER macro in `HYPRE_config.h` is an integer representation of the release that can be compared numerically at compile time against other releases in user build systems.  Other changes include:

- Added `config/version.sh` that should be modified to create new releases
- Renamed the `config/bootstrap` script to `config/update.sh`


liruipeng (approved):
Looks good to me. Thanks, @rfalgout

victorapm (approved):
Looks good. Thanks, Rob!

liruipeng:
@rfalgout From now on, shall we run `./config/update.sh` after changing `configure.in` (as `bootstrap` before)?

rfalgout:
Yes.  I will update the Wiki after the merge is done.

ulrikeyang (approved):
This looks good. Thanks

oseikuffuor1 (approved):
Looks great. I like the change from bootstrap to update.

Thanks.

rfalgout:
FYI, when you run `config/update.sh` in the future, you will now always get a message about the release number being either less than, greater than, or the same as the current version.  This should help to prevent mistakes like setting a new release number that is smaller than a previous one, or forgetting to update the release number altogether for a release.

I think it would be smart to have a regression test as well that checks the value of HYPRE_RELEASE_NUMBER relative to the previous release, but I'm not sure how best to implement that.  Let me know if you have any great ideas.

osborn9 (approved):

PR #369, opened 2021-05-15
Fix hypre_CSRMatrixAddHost

liruipeng:
Fix problems in #341

### DIFF
--- a/src/seq_mv/csr_matop.c
+++ b/src/seq_mv/csr_matop.c
@@ -520,25 +520,13 @@ hypre_CSRMatrixBigAdd( hypre_CSRMatrix *A,
       HYPRE_Int     ia, ib, ic, num_nonzeros;
       HYPRE_Int     ns, ne, pos;
       HYPRE_BigInt  jcol;
-      HYPRE_Int     ii, size, rest, num_threads;
+      HYPRE_Int     ii, num_threads;
       HYPRE_Int     jj;
       HYPRE_Int    *marker = NULL;
 
       ii = hypre_GetThreadNum();
       num_threads = hypre_NumActiveThreads();
-
-      size = nrows_A/num_threads;
-      rest = nrows_A - size*num_threads;
-      if (ii < rest)
-      {
-         ns = ii*size+ii;
-         ne = (ii+1)*size+ii+1;
-      }
-      else
-      {
-         ns = ii*size+rest;
-         ne = (ii+1)*size+rest;
-      }
+      hypre_partition1D(nrows_A, num_threads, ii, &ns, &ne);
 
       marker = hypre_CTAlloc(HYPRE_Int, ncols_A, HYPRE_MEMORY_HOST);
       for (ia = 0; ia < ncols_A; ia++)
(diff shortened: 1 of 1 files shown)


src/seq_mv/csr_matop.c
```diff
@@ -412,24 +412,9 @@ hypre_CSRMatrixAddHost ( hypre_CSRMatrix *A,
 #endif
    {
       HYPRE_Int   ns, ne;
-      HYPRE_Int   ii, size, rest, num_threads;
       HYPRE_Int  *marker = NULL;
 
-      ii = hypre_GetThreadNum();
-      num_threads = hypre_NumActiveThreads();
-
-      size = nnzrows_A/num_threads;
```
liruipeng, line 421: @victorapm Note that you had `nnzrows_A` which should be `nnzrows_C` I believe. I also use the existing function `hypre_partition1D` to simplify the code.

victorapm: Yes, right! This looks much better and solves the issue. Should we use `hypre_partition1D` at https://github.com/hypre-space/hypre/blob/cd7cd75786646c289ab15cf86337c17249e4422d/src/seq_mv/csr_matop.c#L527 ; https://github.com/hypre-space/hypre/blob/cd7cd75786646c289ab15cf86337c17249e4422d/src/seq_mv/csr_matop.c#L737 ;
https://github.com/hypre-space/hypre/blob/cd7cd75786646c289ab15cf86337c17249e4422d/src/seq_mv/csr_matop.c#L1143 ;
https://github.com/hypre-space/hypre/blob/cd7cd75786646c289ab15cf86337c17249e4422d/src/seq_mv/csr_matop.c#L1372 ;

as well? Thanks!

victorapm (approved):
Thanks for this, Ruipeng!

src/seq_mv/csr_matop.c
liruipeng: Yes, I prefer using that function, but what's more important is we don't have similar bugs at other places. Can you please double check all of them? Thanks!

victorapm: Yes! I just double checked and they are correct. Thanks!

liruipeng:
Btw I found this problem by running `mpirun -np 2  ./maxwell_unscaled -in TEST_sstruct/sstruct.in.maxwell3d_unscaled -solver 1 -P 2 1 1` for which you also need to have 
```
   hypre_ParAMGDataRAP2(amg_data)              = 1;
   hypre_ParAMGDataModularizedMatMat(amg_data) = 1;
```
which are the default with `HYPRE_USING_GPU`. It also has the same problem running on CPUs

victorapm:
Interesting, should we add `mpirun -np 2 ./maxwell_unscaled -in TEST_sstruct/sstruct.in.maxwell3d_unscaled -solver 1 -P 2 1 1 -rap 1 -mod_rap2 1` to the CPU tests?

src/seq_mv/csr_matop.c
liruipeng: > Yes! I just double checked and they are correct. Thanks!

Great! Do you also want to make more changes using `hypre_partition1D`? I am fine with it or without. Thanks!

liruipeng:
> -rap 1 -mod_rap2 1

This driver doesn't take `-rap 1 -mod_rap2 1` as cmd-line args...It would be nice to test

src/seq_mv/csr_matop.c
victorapm: I think it's a good idea. Then we reduce code duplication. Can i do this and commit to this branch?

liruipeng: > I think it's a good idea. Then we reduce code duplication. Can i do this and commit to this branch?

Of course you can :)

victorapm:
> > -rap 1 -mod_rap2 1
> 
> This driver doesn't take `-rap 1 -mod_rap2 1` as cmd-line args...It would be nice to test

Oh, that's true. At least we can still rely on the GPU tests :)

liruipeng:
> > > -rap 1 -mod_rap2 1
> > 
> > 
> > This driver doesn't take `-rap 1 -mod_rap2 1` as cmd-line args...It would be nice to test
> 
> Oh, that's true. At least we can still rely on the GPU tests :)

Yes, for now. But eventually maxwell solvers will be ported to GPUs, then `hypre_CSRMatrixAddHost` will not be used anymore with GPU runs.

PR #370, opened 2021-05-17
fix reordering option for diagonal of square matrix transpose

oseikuffuor1:
Make optional the decision to move diagonal entries of transposed square matrices to the first entry. Remove reordering option from transpose routine and call reorder only when needed. This reverts some saved file changes made in PR#260.


src/parcsr_ls/par_mgr.c
```diff
@@ -3205,6 +3205,8 @@ hypre_MGRBuildRestrict(hypre_ParCSRMatrix     *A,
   }
   else if (restrict_type == 3)
   {
+    /* move diagonal to first entry */
+    hypre_CSRMatrixReorder(hypre_ParCSRMatrixDiag(AT));
```
liruipeng, line 3209: Hi, @oseikuffuor1 Do we need `DiagFirst` on GPUs? If so, I can make some changes to `hypre_CSRMatrixReorder` to have a runtime switch from CPU to GPU like others. Thanks!

liruipeng:
Looks good to me, modulo a comment on GPU cases.

src/parcsr_ls/par_mgr.c
oseikuffuor1: @liruipeng yes the same idea is needed for when this option is called on the GPU. I assume we just call the GPU version or reorder (hypre_CSRMatrixMoveDiagFirst) guarded by if hypre_using_cuda || hypre_using_hip to achieve this. Would it be cleaner to do it through the runtime switch in hypre_CSRMatrixReorder like you suggest?

liruipeng (approved):
Looks good to me. Thank you @oseikuffuor1

PR #372, opened 2021-05-18
Change HYPRE_RELEASE_NUMBER to int in CMake's generated HYPRE_Config.h file

osborn9:
This will resolve the warning (cast from pointer to integer of different size) that is causing the cmake regression tests to fail.


rfalgout (approved):
Yes.  I did that correctly on the Linux side but not here.  Thanks, Sarah!

PR #375, opened 2021-05-20
Gpu examples

liruipeng:
This PR adds `TEST_examples` to GPU regression tests.

### DIFF
--- a/AUTOTEST/machine-lassen.sh
+++ b/AUTOTEST/machine-lassen.sh
@@ -46,7 +46,8 @@ eo=""
 # CUDA with UM in debug mode [ij, ams, struct, sstruct]
 co="--with-cuda --enable-unified-memory --enable-persistent --enable-debug --with-cuda-arch=\\'60 70\\' --with-extra-CFLAGS=\\'-qmaxmem=-1 -qsuppress=1500-029\\' --with-extra-CXXFLAGS=\\'-qmaxmem=-1 -qsuppress=1500-029\\'"
 ro="-ij-gpu -ams -struct -sstruct -rt -mpibind -save cuda"
-./test.sh basic.sh $src_dir -co: $co -mo: $mo -ro: $ro -eo: -gpu
+eo="-gpu -rt -mpibind -save cuda"
+./test.sh basic.sh $src_dir -co: $co -mo: $mo -ro: $ro -eo: $eo
 ./renametest.sh basic $output_dir/basic-cuda-um
 
 # CUDA with UM with shared library [no run]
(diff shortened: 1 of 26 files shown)


src/examples/ex12.c
```diff
@@ -71,7 +71,7 @@ int main (int argc, char *argv[])
    HYPRE_Init();
 
    /* Print GPU info */
-   HYPRE_PrintDeviceInfo();
+   /* HYPRE_PrintDeviceInfo(); */
```
rfalgout, line 74: Why comment this out vs removing it altogether?

rfalgout (approved):
Looks good, Ruipeng.  Thanks!

src/examples/ex12.c
liruipeng: Wanted to remind users they can print out info of GPU that is used by that. But turning it on pollutes the output, which fails regression tests.

src/test/TEST_examples/gpu.saved.cuda
```diff
@@ -0,0 +1,72 @@
+# Output file: gpu.out.1
```
liruipeng, line 1: @rfalgout Do I name this file correctly (.saved.cuda) for `machine-lassen`? Not sure if the `save` mechanism of this test (with `-eo`) is treated the same as the others (with `-ro`). Thanks!

src/examples/ex12.c
rfalgout: Works for me.  Thanks!

src/test/TEST_examples/gpu.saved.cuda
rfalgout: The mechanism is the same for '-ro:' and '-eo:'.  You need to use the '-rt' flag to pass additional options through to the `runtest.sh` script.  So, it looks like you need `-eo: -gpu -rt -mpibind -save cuda`.

AUTOTEST/machine-lassen.sh
```diff
@@ -46,7 +46,7 @@ eo=""
 # CUDA with UM in debug mode [ij, ams, struct, sstruct]
 co="--with-cuda --enable-unified-memory --enable-persistent --enable-debug --with-cuda-arch=\\'60 70\\' --with-extra-CFLAGS=\\'-qmaxmem=-1 -qsuppress=1500-029\\' --with-extra-CXXFLAGS=\\'-qmaxmem=-1 -qsuppress=1500-029\\'"
 ro="-ij-gpu -ams -struct -sstruct -rt -mpibind -save cuda"
-./test.sh basic.sh $src_dir -co: $co -mo: $mo -ro: $ro
+./test.sh basic.sh $src_dir -co: $co -mo: $mo -ro: $ro -eo: -gpu
```
liruipeng: @rfalgout Just to confirm I need to have `-rt -mpibind -save cuda` in `-eo` even if I have them in `-ro`? Thanks again!

rfalgout: Yes.  The `-co:`, `-mo:`, `-ro:`, and `-eo:` options for `basic.sh` are completely separate from each other.  The information to put in each of them is passed directly to the `configure.sh`, `make.sh`, `run.sh`, and `examples.sh` tests.

liruipeng: Great! Thanks for explaining. I will make the change.

PR #376, opened 2021-05-21
fix shared library install

### LINKED ISSUE #208: Cannot build/install shared libraries
In HYPRE 2.20.0, the CMake project fails to build shared libraries (and install them, on Windows).

**Build**

In `src/CMakeLists.txt`, variable `BUILD_SHARED_LIBS` must be set before calling `add_library(HYPRE)`, otherwise it will not have any effect. This could be worked around by specifying both `-DBUILD_SHARED_LIBS=ON` and `-DHYPRE_ENABLED_SHARED=ON` from command-line to CMake.

**Install**

Even if shared libraries are built, the `install(TARGETS)` command does not install the DLLs on Windows, which are identified by the [`RUNTIME`](https://cmake.org/cmake/help/latest/command/install.html#targets) keyword (missing). The following would need added to the `install(TARGETS)` command:

`RUNTIME DESTINATION "${CMAKE_INSTALL_LIBDIR}"`
or (to place them in the `bin` subdirectory):
`RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"`

matthiasdiener:
Fixes #208


osborn9 (approved):
Thanks @matthiasdiener for the fix!

PR #387, opened 2021-05-27
Fix make clean

rfalgout:
This PR replaces the `rm -f *.o*` lines in the makefiles to `rm -f *.o *.obj`.  The previous trailing approach can delete user or developer files that are not intended to be removed.  The change is explicit about removing only object files.


liruipeng (approved):

victorapm (approved):
Thanks, Rob and Ruipeng!

PR #392, opened 2021-06-03
Update IJ interface with changes in recmat-merge

victorapm:
This PR brings changes on the IJ interface and utilities folder that were added in the recmat-merge branch. Some of the functions added here are not being used in master, but are needed in the recmat-merge branch. A detailed list of the changes is given bellow:
1. Add `HYPRE_IJMatrixTranspose` to compute the transpose of an IJMatrix.
1. Add `HYPRE_IJMatrixAdd` to compute mat/mat add for IJMatrices.
1. Add `HYPRE_IJMatrixNorm` as a generic function to compute the norm of an IJMatrix, the only option available currently is the infinity norm but we could add others when needed.
1. Add `hypre_ParCSRMatrixInfNorm` to compute the infinity norm of a ParCSRMatrix.
1. Add `hypre_ParCSRMatrixReorder` to reorder a ParCSRMatrix such that the first entry of each row of the diag `CSRMatrix` is the diagonal component.
1. Add `hypre_CSRMatrixInvDiagAxpy` to compute `y = alpha*inv(D)*x + beta*y` for a `CSRMatrix`
2. Add `hypre_ParCSRMatrixInvDiagAxpy` to compute `y = alpha*inv(D)*x + beta*y` for a `ParCSRMatrix`
3. Add more caliper annotation to ParCSR functions and BoomerAMG to get more detailed timings.
4. Move `hypre_Log2` from `sstruct_ls` to `utilities`.
5. Fix typo in `hypre_LocateAssummedPartition` (should be assumed without the double "m")
5. Remove `ParCSRMatrixOwns[Row,Col]Starts` and `ParCSRMatrixSet[Row,Col]StartsOwner` functions. Default behavior now is a ParCSRMatrix always own `row_starts` and `col_starts` (Since we don't have the global partition anymore, this is not harmful for performance)

The last item touched many files in hypre. This was motivated by a memory leak that I was trying to solve after changing the default row/col ownership behavior of the resulting matrix of a Mat/MatT multiplication. It turned out to be easier for me to remove the ownership control for solving the leak.

Thanks!


rfalgout (approved):
This all looks really good to me, @victorapm .  I especially like the removal of the `Owns[Row|Col]Starts` stuff and the additional cleaning up of global partition code, because this should make maintenance, debugging, and implementation easier in the future.  I have some questions about some of the matrix/vector kernels, but that is a larger discussion item for the whole team.  This looks good.

Finally, I did not run any of the regression tests, but we should make sure they all run (on the GPU machines as well).

Thanks!

liruipeng:
Hi all, If you don't mind, I would like to merge this in after upcoming `ams-gpu` PR, and I will review this after. Thanks!

victorapm:
Thanks everyone! Tux tests are passing, but I haven't run `machine-lassen`. Will do it now. 

Sure, @liruipeng! You can look at this later. There's something I would like to bring for discussion: I see there are two implementations for ParCSRMatrix/Matrix multiplication (`hypre_ParCSRMatMat` and `hypre_ParMatmul`) and Matrix Transpose/Matrix multiplication (`hypre_ParCSRTMatMatKT` and `hypre_ParTMatmul`). These are located in different files. Should we create a unified wrapper for each of these operations? i.e., have a generic `hypre_ParCSRMatrixMult` which calls each implementation for example. Also, I wanted to understand better when to use one or another implementation. I see only the first choice has a GPU branch in it, is there any other reason to use one over the other (specifically when thinking about CPU)?

Thanks!

liruipeng:
> Thanks everyone! Tux tests are passing, but I haven't run `machine-lassen`. Will do it now.
> 
> Sure, @liruipeng! You can look at this later. There's something I would like to bring for discussion: I see there are two implementations for ParCSRMatrix/Matrix multiplication (`hypre_ParCSRMatMat` and `hypre_ParMatmul`) and Matrix Transpose/Matrix multiplication (`hypre_ParCSRTMatMatKT` and `hypre_ParTMatmul`). These are located in different files. Should we create a unified wrapper for each of these operations? i.e., have a generic `hypre_ParCSRMatrixMult` which calls each implementation for example. Also, I wanted to understand better when to use one or another implementation. I see only the first choice has a GPU branch in it, is there any other reason to use one over the other (specifically when thinking about CPU)?
> 
> Thanks!

`hypre_ParMatmul` and `hypre_ParTMatmul` are older routines, which are the default for CPUs, with no GPU support. The other two you mentioned are so-called "modularized" Parcsr kernels since they call corresponding CSR modules. Mathematically, they are the same. For backward compatibility (regression tests, performance maybe, etc), we still use the older ones on CPUs. Maybe @ulrikeyang can chime in. Thanks!

ulrikeyang (approved):
This looks good. Thanks

src/parcsr_mv/par_csr_matrix.c
```diff
@@ -592,19 +550,14 @@ hypre_ParCSRMatrixRead( MPI_Comm    comm,
    hypre_ParCSRMatrixLastRowIndex(matrix) = row_e - 1;
    hypre_ParCSRMatrixLastColDiag(matrix) = col_e - 1;
 
-   hypre_ParCSRMatrixRowStarts(matrix) = row_starts;
-   hypre_ParCSRMatrixColStarts(matrix) = col_starts;
+   hypre_TMemcpy(hypre_ParCSRMatrixRowStarts(matrix), row_starts,
```
liruipeng: I don't think `hypre_ParCSRMatrixRowStarts` and `hypre_ParCSRMatrixColStarts` have been allocated here (before you do the copy). Btw, have you thought about having them on stack, i.e., `HYPRE_BigInt  row_starts[2];` in `hypre_ParCSRMatrix`, so we don't need to worry about allocating them any more?

src/parcsr_mv/par_csr_matrix.c
```diff
@@ -2246,13 +2197,13 @@ hypre_ParCSRMatrix * hypre_ParCSRMatrixUnion( hypre_ParCSRMatrix * A,
...
    hypre_assert( hypre_ParCSRMatrixFirstRowIndex( B )
                  == hypre_ParCSRMatrixFirstRowIndex( A ) );
-   hypre_ParCSRMatrixRowStarts( C ) = hypre_ParCSRMatrixRowStarts( A );
-   hypre_ParCSRMatrixOwnsRowStarts( C ) = 0;
-   hypre_ParCSRMatrixColStarts( C ) = hypre_ParCSRMatrixColStarts( A );
-   hypre_ParCSRMatrixOwnsColStarts( C ) = 0;
-   for ( p=0; p<=num_procs; ++p )
-      hypre_assert( hypre_ParCSRMatrixColStarts(A)
-                    == hypre_ParCSRMatrixColStarts(B) );
+   hypre_TMemcpy(hypre_ParCSRMatrixRowStarts(C), hypre_ParCSRMatrixRowStarts(A),
```
liruipeng, line 2159: Same question as above.

src/parcsr_mv/par_csr_matvec.c
```diff
@@ -849,3 +857,49 @@ hypre_ParCSRMatrixMatvec_FF( HYPRE_Complex       alpha,
 
    return ierr;
 }
+
```
liruipeng: Maybe it should be put in `par_csr_matop.c`. In `par_csr_matop_device.c`, there is one `hypre_ParCSRDiagScale` doing similar thing (with beta==0.0, alpha=1.0, also for CPU), so we should combine them.

src/seq_mv/csr_matvec.c
```diff
@@ -1126,3 +1126,186 @@ hypre_CSRMatrixMatvec_FF( HYPRE_Complex    alpha,
 
    return ierr;
 }
+
+/*--------------------------------------------------------------------------
+ * hypre_CSRMatrixInvDiagAxpy
+ *
+ * y = alpha*inv(A_D)*x + beta*y
+ *--------------------------------------------------------------------------*/
+
```
liruipeng: Combine with `hypreDevice_DiagScaleVector` that is the device version

src/parcsr_ls/par_amg_setup.c
```diff
@@ -1423,7 +1419,6 @@ hypre_BoomerAMGSetup( void               *amg_vdata,
                if (SN) hypre_ParCSRMatrixDestroy(SN);
                if (AN) hypre_ParCSRMatrixDestroy(AN);
                //hypre_TFree(CF_marker, HYPRE_MEMORY_HOST);
-               hypre_TFree(coarse_pnts_global, HYPRE_MEMORY_HOST);
```
liruipeng, line 1436: Why not freeing it?

src/parcsr_ls/gen_redcs_mat.c
```diff
@@ -303,7 +303,7 @@ hypre_seqAMGSetup( hypre_ParAMGData *amg_data,
 
             F_seq = hypre_ParVectorCreate(seq_comm, size, row_starts);
             U_seq = hypre_ParVectorCreate(seq_comm, size, row_starts);
-            hypre_ParVectorOwnsPartitioning(F_seq) = 0;
```
liruipeng, line 306: Why `ParVector` still keeps this information about who owns the partitioning?

liruipeng:
Hi @victorapm, this in general looks good to me. Thanks! I especially like imposing our code style. I left some comments and questions. There are a few functions we can combine with existing ones. I can work with you on the device ones.

Some major ones:
1. why not removing `hypre_ParVectorOwnsPartitioning` as well?
2. would it be better to have `row_starts` and `row_starts` on stack memory, i.e., `HYPRE_BigInt row_starts[2]`
3. some conflicts from previous PR need to be addressed.

victorapm:
Hi @liruipeng,

thanks a lot for your in-depth review! Addressing your comments:

1. I agree with it. The `OwnsPartitioning` data member isn't needed anymore for the same reason as `ParCSRMatrixOwns[Row,Col]Starts`. This change involves about 10 files, @ulrikeyang and @rfalgout, are you fine with that?
2. I think it's a good idea since it simplifies even more the code. This will involve some indirect code changes such as in `GenerateLocalPartitioning` and other places which use it, but I can do that.
3. Yes, I will merge master into this branch again.

Thanks!
Victor

src/parcsr_ls/par_amg_setup.c
```diff
@@ -1450,7 +1446,6 @@ hypre_BoomerAMGSetup( void               *amg_vdata,
                if (AN) hypre_ParCSRMatrixDestroy(AN);
                if (num_functions > 1) hypre_TFree(coarse_dof_func, HYPRE_MEMORY_HOST);
                hypre_TFree(CF_marker, HYPRE_MEMORY_HOST);
-               hypre_TFree(coarse_pnts_global, HYPRE_MEMORY_HOST);
```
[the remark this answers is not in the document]
victorapm: I moved this to `par_coarse_parms.c`: https://github.com/hypre-space/hypre/blob/dd34cf278bffcfb2f1f896a31c3e0bd83e4d547e/src/parcsr_ls/par_coarse_parms.c#L118

so that we don't need to remember freeing `coarse_pnts_global` in `BoomerAMGSetup` when it isn't needed anymore (in the original code, this was giving me some memleaks difficult to track).

Actually, I was thinking about your suggestion of moving `partitioning` to the stack. The same could be done here for `coarse_pnts_global`. What do you think?

src/parcsr_ls/gen_redcs_mat.c
victorapm: I agree with changing this as mentioned in the comment bellow

src/parcsr_ls/par_amg_setup.c
victorapm: For the same reason as in https://github.com/hypre-space/hypre/pull/392#discussion_r657303905

src/seq_mv/csr_matvec.c
victorapm: I was talking to @rfalgout and we will phase out the `InvDiag` functions that I added here and use the `DiagScaleVector` function instead to reduce code duplication. I think we still need some discussion on the name of the function that performs "element division" for vectors, right @rfalgout?

src/parcsr_mv/par_csr_matvec.c
victorapm: I think we need to decide whether maintaining a function to compute `y = inv(A_D)*x` in the `ParCSRMatrix` class or having only a `ParVector` function that does `y = alpha*z.*x + beta*b` (element-wise multiplication for `z` and `x`) that could be used for computing the former quantity. I think @rfalgout wanted to bring this for discussion.

src/parcsr_mv/par_csr_matrix.c
victorapm: Correct! Thanks for catching that. Moving to stack would solve this indirectly :)

src/parcsr_mv/par_csr_matrix.c
victorapm: Thanks! I totally agree with that!

src/parcsr_mv/par_csr_matvec.c
liruipeng: @victorapm Either way is fine. We have device version for both of them. So I would like to combine with the host function you added.

src/parcsr_ls/par_amg_setup.c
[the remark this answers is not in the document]
liruipeng: OK. I see how it works now, a bit tricky though. But it is fine. Thanks!

liruipeng:
> Hi @liruipeng,
> 
> thanks a lot for your in-depth review! Addressing your comments:
> 
> 1. I agree with it. The `OwnsPartitioning` data member isn't needed anymore for the same reason as `ParCSRMatrixOwns[Row,Col]Starts`. This change involves about 10 files, @ulrikeyang and @rfalgout, are you fine with that?
> 2. I think it's a good idea since it simplifies even more the code. This will involve some indirect code changes such as in `GenerateLocalPartitioning` and other places which use it, but I can do that.
> 3. Yes, I will merge master into this branch again.
> 
> Thanks!
> Victor

1. I think it is good to have the same change in parVector for the consistency with parCSRMatrix
2. This can reduce some amount of code but requires more changes. Hopefully not too complicated. It's up to your decision whether make the changes. I am fine with either way.

victorapm:
Thanks @liruipeng, 

I ran a grep to evaluate how much code I'd need to change, and I'd say it's something in the order of 100 lines. I'm willing to do that after finishing another work that I'm currently doing (SSAMG paper).

Thanks!

victorapm:
> Hi @victorapm, this in general looks good to me. Thanks! I especially like imposing our code style. I left some comments and questions. There are a few functions we can combine with existing ones. I can work with you on the device ones.
> 
> Some major ones:
> 
> 1. why not removing `hypre_ParVectorOwnsPartitioning` as well?
> 2. would it be better to have `row_starts` and `row_starts` on stack memory, i.e., `HYPRE_BigInt row_starts[2]`
> 3. some conflicts from previous PR need to be addressed.

Hi @liruipeng,

I addressed these points in the recent commits. Could you take a look again at this PR?

Thanks!

victorapm:
Just a detail: I forgot to mention that `machine-tux.sh` tests are running fine. Thanks!

liruipeng (approved):
Looks good to me. It removes the `OwnRow/ColStarts` totally, which makes the code a lot simpler. ~~I think you decided to keep `hypre_CSRMatrixInvDiagAxpy` and the `Parcsr` one, which is fine to me.~~ Thanks @victorapm

liruipeng:
> Just a detail: I forgot to mention that `machine-tux.sh` tests are running fine. Thanks!

Any thoughts on `machine-lassen.sh`? Thanks!

victorapm:
> Looks good to me. It removes the `OwnRow/ColStarts` totally, which makes the code a lot simpler. I think you decided to keep `hypre_CSRMatrixInvDiagAxpy` and the `Parcsr` one, which is fine to me. Thanks @victorapm

Thanks @liruipeng! I actually removed `hypre_(Par)CSRMatrixInvDiagAxpy` since they are no longer needed in the `recmat` branch

> Any thoughts on machine-lassen.sh? Thanks!

I'm running them just to make sure :)

liruipeng:
> > Looks good to me. It removes the `OwnRow/ColStarts` totally, which makes the code a lot simpler. I think you decided to keep `hypre_CSRMatrixInvDiagAxpy` and the `Parcsr` one, which is fine to me. Thanks @victorapm
> 
> Thanks @liruipeng! I actually removed `hypre_(Par)CSRMatrixInvDiagAxpy` since they are no longer needed in the `recmat` branch
> 
> > Any thoughts on machine-lassen.sh? Thanks!
> 
> I'm running them just to make sure :)



> > Looks good to me. It removes the `OwnRow/ColStarts` totally, which makes the code a lot simpler. I think you decided to keep `hypre_CSRMatrixInvDiagAxpy` and the `Parcsr` one, which is fine to me. Thanks @victorapm
> 
> Thanks @liruipeng! I actually removed `hypre_(Par)CSRMatrixInvDiagAxpy` since they are no longer needed in the `recmat` branch
> 

Yes, you did. I don't know why I saw those earlier, maybe clicked wrong links or something. Thanks!

victorapm:
Yes! That was probably the reason :) Thanks!!

PR #406, opened 2021-06-16
Fix some typos in .jobs and memory issues 

### LINKED ISSUE #404: Typos in .jobs
I've found the following typos in TEST_ij/*.jobs:

smoother.jobs: Test 19:  `-n 200 200 -P 2 2` should be `-n 200 200 1 -P 2 21`, as `ij` always reads in three parameters for `-n` and `-P`

solvers.jobs: Test 119: `ns 2` should be `-ns 2`

matrix.jobs: `-low_storage` should be `-storage_low` as per the option in `ij.c`

Note that for `-ns`, running the test on GPU currently produces an out of memory error.

AerysL:
This PR addresses #404. Includes the typo fixes, and changes to `.saved` due to inclusion of `-ns 2`. Also, fixed the memory issues after fixing `the .jobs`


liruipeng (approved):
Looks good to me. Thanks!

rfalgout (approved):
Looks good to me, too.  Thanks!

src/test/TEST_ij/smoother.jobs
```diff
@@ -84,8 +84,8 @@ mpirun -np 4  ./ij -rhsrand -solver 1 -rlx 16 -cheby_variant 1 -n 20 20 20 \
 mpirun -np 4  ./ij -solver 3 -rlx 16 -cheby_eig_est 0 -n 40 40 20 \
 -P 2 2 1 -difconv -a 10 10 10  > smoother.out.18
 
-mpirun -np 4  ./ij -rhsrand -solver 1 -rlx 16 -rotate -alpha 60 -eps 0.1 -cheby_fraction 0.2 -n 200 200 \
--P 2 2  > smoother.out.19
+mpirun -np 4  ./ij -rhsrand -solver 1 -rlx 16 -rotate -alpha 60 -eps 0.1 -cheby_fraction 0.2 -n 200 200 1 \
+-P 2 2 1 > smoother.out.19
```
ulrikeyang, line 89: Note that here the '1's are actually not needed, since this is a 2D problem, but they won't hurt either.

ulrikeyang (approved):
Looks good. Thanks

src/test/TEST_ij/smoother.jobs
AerysL: The reason I added them is because the argument parsing script reads three variables https://github.com/hypre-space/hypre/blob/9117fc20f2c4ed88700cd6cb1f5da2463b63c0fa/src/test/ij.c#L7785, so if it's the last argument, `argv[argc]` will be read (technically out of bounds of argv) and attempted to be converted to an int, which will be undefined (although I suspect unused). I agree that I don't think nz/R will be used in the 2D problem, so the different value won't matter.

liruipeng: > The reason I added them is because the argument parsing script reads three variables
> 
> https://github.com/hypre-space/hypre/blob/9117fc20f2c4ed88700cd6cb1f5da2463b63c0fa/src/test/ij.c#L7785
> 
> , so if it's the last argument, `argv[argc]` will be read (technically out of bounds of argv) and attempted to be converted to an int, which will be undefined (although I suspect unused). I agree that I don't think nz/R will be used in the 2D problem, so the different value won't matter.

It's good to fix. It's not fun to have `argv` out-of-range. Thanks!

PR #408, opened 2021-06-17
One based index

### LINKED ISSUE #396: Potential problem with par_vector.c? Running with IJ interface requires indexing starting at zero
I've recently updated to the latest hypre master from a January 2021 version. I'm calling various solvers using the IJ interface from fortran. For the longest time (several years), I indexed my cells starting at 1 (i.e., 1 to N) for a problem of size N for all my interactions with HYPRE, without any problem. It seems that the new version doesn't like that: I get an error that says `Index out of range! -- hypre_ParVectorGetValues`. I went into par_vector.c and printed out the index value that triggers the problem, and I found that my max index (equal to N) that is larger than HYPRE's expected max of N-1. I see that there were significant changes to memory handling/safety made on Feb 9th, so I'm assuming that along the way the support for a user-specified starting index was dropped, voluntarily or not.
It's not an enormous deal as I can simply start counting at 0 (which indeed fixes the issue), but I thought I should point it out here.

liruipeng:
This PR fixes #396, indexing problem in `HYPRE_IJVectorGetValues`

### DIFF
--- a/src/IJ_mv/IJVector_parcsr.c
+++ b/src/IJ_mv/IJVector_parcsr.c
@@ -655,6 +655,7 @@ hypre_IJVectorGetValuesPar(hypre_IJVector *vector,
    HYPRE_BigInt    *IJpartitioning = hypre_IJVectorPartitioning(vector);
    HYPRE_BigInt     vec_start;
    HYPRE_BigInt     vec_stop;
+   HYPRE_BigInt     jmin           = hypre_IJVectorGlobalFirstRow(vector);
    hypre_ParVector *par_vector     = (hypre_ParVector*) hypre_IJVectorObject(vector);
    HYPRE_Int        print_level    = hypre_IJVectorPrintLevel(vector);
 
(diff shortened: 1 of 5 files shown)


rfalgout (approved):
Looks good to me, @liruipeng .  Thanks!

rfalgout:
Hi @liruipeng .  Given that you fixed the 1-based indexing for GetValues on GPUs, I was curious if you were planning to change the Set/AddValues routines as well.

liruipeng:
> Hi @liruipeng . Given that you fixed the 1-based indexing for GetValues on GPUs, I was curious if you were planning to change the Set/AddValues routines as well.

@rfalgout We should fix those as well. Should I do it in this PR? ~~I will try to make it happen before the next release anyways.~~ Thanks!

rfalgout:
If you are going to fix them before the next release anyway, you may as well do it in this PR.  Your call.

liruipeng:
I will fix other parts of GPU IJ interface on this issue later after the release. So, merging it to master

PR #410, opened 2021-06-18
Fix potentially inconsistent eig estimates (#390)

### LINKED ISSUE #390: Discrepancy of max norm between distributed ranks
It appears to me that in `par_relax_more.c` that the returned `max_norm` may not be consistent between ranks in certain edge cases. Particulaily if the evaluation of the if statement on line 93 (https://github.com/hypre-space/hypre/blob/1d9411c7ab67600f1aa73febfff6aa658ab08ea2/src/parcsr_ls/par_relax_more.c#L93) is evaluated differently on each rank. While `max_norm` is reduced such that it is the same on all ranks on line 89 and 90 (https://github.com/hypre-space/hypre/blob/1d9411c7ab67600f1aa73febfff6aa658ab08ea2/src/parcsr_ls/par_relax_more.c#L89), the `pos_diag` and `neg_diag` variables are only ever local. If the input matrix has a significantly different spectrum rank to rank (e.g. strictly negative diagonal on one rank and not on another), the if statement could evaluate differently between ranks. 

I would expect that `max_eig` is used in such a way where one would want it to be consistent between ranks. Is this a bug?

Mentioning  @rfalgout @ulrikeyang  at @liruipeng 's request

AerysL:
This PR reimplements  `hypre_ParCSRMaxEigEstimate` using Gershgorin discs, which ensures that `max_eig` and `min_eig` are both `allreduced` across all ranks so that the return value of the function is the same for all ranks.

It also resolves the problem of solving symmetric negative definite (SND) systems, e.g. `-Ax=b` which should have **exactly** the same convergence profile to the solution, `-x`, as that of `Ax=b` using AMG starting with `x_0 = 0`. It is true for Jacobi and G-S smoothers. This PR fixes Chebyshev smoothers. For SND systems and Chebyshev, eigenvalue estimation can be done with `hypre_ParCSRMaxEigEstimate` but not `hypre_ParCSRMaxEigEstimateCG`. For the latter one, we can use a more general Lanczos algorithm, which has not been implemented in hypre.

TODO:
- [x] Test tux
- [x] Test lassen

Closes #390


src/parcsr_ls/par_relax_more.c
```diff
@@ -89,6 +89,12 @@ hypre_ParCSRMaxEigEstimate(hypre_ParCSRMatrix *A, /* matrix to relax with */
    hypre_MPI_Allreduce(&max_norm, &temp, 1, HYPRE_MPI_REAL, hypre_MPI_MAX, hypre_ParCSRMatrixComm(A));
    max_norm = temp;
 
+   hypre_MPI_Allreduce(&pos_diag, &temp, 1, HYPRE_MPI_INT, hypre_MPI_SUM, hypre_ParCSRMatrixComm(A));
```
rfalgout: Hi @AerysL .  I don't think we should do this because it introduces a lot of extra communication and it shouldn't be a problem for the definite cases, which we more-or-less require in AMG.  If we are going to change something here, then we can keep track of the upper and lower bounds of the Gershgorin circles and allreduce them both in one call.  The current `temp` variable is the upper bound (in the positive definite case), if we add the negative of the lower bound, then we can use MPI_MAX to do both at the same time.  The sign of the diagonal is not needed anymore if we do this.

Note that I am not even sure how to use these bounds for Chebyshev smoothing if we have an indefinite system.

liruipeng: Thanks @rfalgout . We redid it as per your suggestions. Please take another look. Thanks!

liruipeng (approved):

src/parcsr_ls/par_cheby.c
```diff
@@ -38,130 +38,136 @@ means half, and .1 means 10percent)
...
    /* we are using the order of p(A) */
-   cheby_order = order -1;
-   
-    /* make sure we are large enough -  Adams et al. 2003 */
-   upper_bound = max_eig * 1.1;
-   /* lower_bound = max_eig/fraction; */
-   lower_bound = (upper_bound - min_eig)* fraction + min_eig; 
+   cheby_order = order - 1;
 
+   if (min_eig >= 0.0)
```
rfalgout, line 92: I don't think this is quite right.  Maybe we should discuss.  Basically, what was identified as `upper_bound` in the previous code was actually the eigenvalue estimate (positive or negative) that corresponds to the eigenvalue with largest magnitude.  I believe the Chebyshev polynomial computations were correct in the previous code even in the negative definite case.  I don't think they are anymore here.  One other thing to fix is "`min_eig`", which I think we are better off setting to zero as in the previous code.

If you want to think about this ahead of time, just consider the simplest zero-th order case that basically produces the Richardson error propagator (referred to as the residual polynomial in the comments below), I - sA, where s is the inverse of the eigenvalue (positive or negative) with largest magnitude.  See `theta` and the resulting `coefs[0]` computation; the latter is s for Richardson.  This provides a good smoother when the problem is either positive or negative (semi-)definite.  I don't know what to do when the problem is indefinite.  Smoothers in that setting or more difficult and usually involve the normal equations.

rfalgout: I see that you are already setting one of the estimates to zero, so ignore that comment above.

rfalgout (approved):
This looks good to me.  I assume the tests worked out okay?  Thanks!

liruipeng:
Hi, @rfalgout, you were right about `upper_bound` and `lower_bound`. I actually made a mistake on them. But actually, switching `upper_bound` and `lower_bound` doesn't change standard Chebyshev polynomial (`cheby_variant = 0`). But does change slightly `cheby_variant = 1`. So the fix is still preferred to have consistent results with the SPD ones in all the cases.

Also, CG works fine with negative definite matrices to estimate eigenvalues. The change was on diagonal scaling itself.

Thanks again for your help!

AerysL:
@liruipeng  The tests are passing on Lassen. Can you test on tux? Then it should be ready for review @ulrikeyang 

Thanks!

ulrikeyang (approved):
Looks good. Thanks

src/parcsr_ls/par_relax_more.c
```diff
@@ -277,7 +282,7 @@ hypre_ParCSRMaxEigEstimateCGHost(hypre_ParCSRMatrix *A,     /* matrix to relax w
 
    if (scale)
    {
-      hypre_CSRMatrixExtractDiagonal(hypre_ParCSRMatrixDiag(A), ds_data, 3);
+      hypre_CSRMatrixExtractDiagonal(hypre_ParCSRMatrixDiag(A), ds_data, 4);
```
liruipeng, line 285: @AerysL Glad you found this error. Thanks!

PR #419, opened 2021-06-26
Check MSVC version before adding flags

ibaned:
This was originally started as an effort to
fix the Windows+CUDA build, in which case
these special flags need to be added differently
when CUDA is used.
However, they were there to workaround issues in
earlier MSVC versions that today's version doesn't
have, so the changes here actually just check
whether the MSVC being used is old enough to need
these workarounds.
Since using CUDA and MSVC through CMake is relatively
new, there shouldn't be a need to support these
workarounds with CUDA


osborn9 (approved):
Thanks @ibaned for the improvements! This looks good!

liruipeng (approved):
Looks good to me. Thanks!

PR #421, opened 2021-06-29
use FindCUDAToolkit.cmake when available

ibaned:
This is more work towards using CUDA on Windows. It hasn't actually been fully tested on Windows but should handle static libraries in an OS-generic way instead of the previous code that uses Linux-specific library names.


ibaned:
Update: with 3fa79f3 it has been tested on Windows with CUDA and the build succeeds

osborn9:
Apologies for my delay on reviewing this @ibaned! For the functionality, this looks good! 

For the style, please indent using 2 spaces the body of `if (CMAKE_VERSION VERSION_LESS 3.17)` / `else`.

ibaned:
Hi @osborn9 ! I definitely delayed this myself way longer! Sorry for it slipping through the cracks, let me know if there's anything else you'd like done. Thanks for reviewing.

osborn9:
Thanks @ibaned for making the formatting changes!

osborn9 (approved):

PR #423, opened 2021-07-01
cmake: fix config for CXX-only consumers

junghans:
Some consumers only have CXX enabled hence find_dependency(MPI) won't define a MPI::MPI_C target.
Explicitly enabling the C language will fix that.

Issue found here: https://github.com/ECP-copa/Cabana/pull/426#discussion_r661902293


junghans:
ping @osborn9

osborn9:
Thanks @junghans! This looks good!

osborn9:
Thanks @junghans! This looks good!

PR #431, opened 2021-07-29
Add SYCL support

waynemitchell:
This PR adds sycl and Intel oneAPI support to hypre with the goal of achieving good performance on Intel GPUs. In addition, some data structures, variables, and files are renamed from "cuda" to "device" since the underlying implementation may be in cuda/hip/sycl.

This first PR sets up some initial infrastructure (e.g. memory management, device setup, etc.) and implements the boxloops and structure solvers in sycl.

- [x] Configuration option
- [x] CMake configuration option
- [x] HYPRE_Init() and setup device and compute queue
- [x] Memory allocation/copy/free etc.
- [x] Boxloops for structured interface
- [x] Structured interface and solvers working on device
- [x] Tux regression tests
- [x] Lassen regression tests
- [x] Arcticus struct regression tests


waynemitchell:
@osborn9 , @rfalgout , I'm not that familiar with autoconf or cmake, and this is my first time digging into and modifying the configuration and compilation parts of hypre. So I wanted to make sure I've done this all correctly. 

For the autoconf side, I edited config/configure.in and config/Makefile.config.in and then ran config/update.sh. 

On the cmake side, I modified all the various CMakeLists.txt and also some changes to the cmake-related files in the config directory. 

Let me know if anything I've done doesn't make sense or could be improved. Thanks!

waynemitchell:
This should be a comprehensive list of the name changes present in this PR.
1. hypre_CudaData -> hypre_DeviceData
2. All macros hypre_CudaData* -> hypre_DeviceData* with the exception of those below with additional modifications
3. hypre_CudaDataCudaStream -> hypre_DeviceDataStream
4. hypre_CudaDataCudaComputeStream -> hypre_DeviceDataComputeStream
5. hypre_CudaDataCudaComputeStreamNum -> hypre_DeviceDataComputeStreamNum
6. hypre_CudaDataCudaReduceBuffer -> hypre_DeviceDataReduceBuffer
7. hypre_HandleCudaData -> hypre_HandleDeviceData
8. hypre_HandleCudaComputeStream -> hypre_HandleComputeStream
9. hypre_HandleCudaDevice -> hypre_HandleDevice
10. hypre_HandleCudaComputeStreamNum -> hypre_HandleComputeStreamNum
11. hypre_HandleCudaReduceBuffer -> hypre_HandleReduceBuffer
12. hypre_GetDefaultCUDABlockDimension -> hypre_GetDefaultDeviceBlockDimension
13. hypre_GetDefaultCUDAGridDimension -> hypre_GetDefaultDeviceGridDimension

Does anyone foresee any problems with these changes? I know that recently we've had issues with internal name/type changes affecting user codes, so I just did a grep through mfem, petsc, and trillinos looking for all these names and did not find them. Any other potential problems I could be overlooking here?

rfalgout:
> For the autoconf side, I edited config/configure.in and config/Makefile.config.in and then ran config/update.sh.

You did the right thing on the autoconf side.  Looks good to me.  Thanks!

rfalgout:
> This should be a comprehensive list of the name changes present in this PR.
> 
> 1. hypre_CudaData -> hypre_DeviceData
> 2. All macros hypre_CudaData* -> hyrpe_DeviceData* with the exception of those below with additional modifications
> [snip]
> 
> Does anyone foresee any problems with these changes?

I don't foresee any problems.  I assume `hyrpe` is just a mispelling in this list, right.  Thanks again!

waynemitchell:
> > This should be a comprehensive list of the name changes present in this PR.
> > 
> > 1. hypre_CudaData -> hypre_DeviceData
> > 2. All macros hypre_CudaData* -> hyrpe_DeviceData* with the exception of those below with additional modifications
> >    [snip]
> > 
> > Does anyone foresee any problems with these changes?
> 
> I don't foresee any problems. I assume `hyrpe` is just a mispelling in this list, right. Thanks again!

Oops... yup, just a mispelling here. Fixed now.

osborn9:
> On the cmake side, I modified all the various CMakeLists.txt and also some changes to the cmake-related files in the config directory.

@waynemitchell I just reviewed the CMake changes and everything looks good!

rfalgout (approved):
This looks good to me, Wayne.  Thanks!

victorapm (approved):
Thanks, @waynemitchell! I don't have access to an Intel GPU to test the code, however everything looks fine to me

osborn9 (approved):

src/struct_mv/boxloop_sycl.h
```diff
@@ -0,0 +1,429 @@
...
+
+/* Reduction BoxLoop1 */
+#define hypre_newBoxLoop1ReductionBegin(ndim, loop_size, dbox1, start1, stride1, i1, sum_var)         \
+{                                                                                                     \
+   hypre_newBoxLoopInit(ndim, loop_size);                                                             \
+   hypre_BoxLoopDataDeclareK(1, ndim, loop_size, dbox1, start1, stride1);                             \
+   HYPRE_Real *shared_sum_var = hypre_CTAlloc(HYPRE_Real, 1, HYPRE_MEMORY_DEVICE);                    \
+   hypre_TMemcpy(shared_sum_var, &sum_var, HYPRE_Real, 1, HYPRE_MEMORY_DEVICE, HYPRE_MEMORY_HOST);    \
+   ReductionBoxLoopforall( [=] (sycl::nd_item<1> item, auto &hypre_sycl_sum)                          \
+   {                                                                                                  \
```
liruipeng, line 341: @waynemitchell is it possible to use `sum_var` here, in the place of `hypre_sycl_sum`, so the interface can be simpler, i.e., don't need the `#if defined(HYPRE_USING_SYCL)` at calling reduction? I am not very familiar with sycl yet. Just curious.

ulrikeyang (approved):
This looks good, but I really don't know enough about SYCL to give you a reasonable review.

src/struct_mv/boxloop_sycl.h
waynemitchell: @liruipeng I couldn't find a cleaner way of implementing this (e.g. without the `#if defined(HYPRE_USING_SYCL)` macros at the reduction calls). `hypre_sycl_sum` here is actually a `sycl::reducer` variable. Inside the lambda (i.e. in the contents of the boxloop) you need to use the variable name of the `sycl::reducer`. I could see doing something similar to the cuda implementation here, i.e. at the call, setup a reducer object to pass to the boxloop. But this just moves where you have to put the `HYPRE_USING_SYCL` macro guards at the reduction call. I will probably do something like this in the future (in particular, if doing multiple reduction simultaneously works as expected with the newer intel compilers), but I think the current implementation works and is good enough for now.

liruipeng: @waynemitchell Maybe I wasn't clear about what I was asking about. Can we just write `auto &sum_var` instead of `auto &hypre_sycl_sum` here, if it makes sense at all? I **guess** `hypre_sycl_sum` is an argument (as you said, a reducer type) of the lambda and the name is up to our choice (did not find it used anywhere else). If this can work, do we still need the `#if defined` at the call? Please let me know if this makes sense. Thanks!

waynemitchell: @liruipeng If I understand you correctly, you're suggesting that we use `sum_var` as the name for the reducer variable. I expected this to cause conflicts since `sum_var` is also a variable name captured by the lambda, correct? I tried this, and it actually seems to work, however. I'm still not super comfortable working with all the details of lambdas... is this definitely safe?

waynemitchell: @liruipeng What about the first answer here:
https://stackoverflow.com/questions/58285806/how-to-pass-local-variable-and-argument-of-same-name-using-this-keyword-inside
I can do something like:
`[=,hypre_unused_var=sum_var] (sycl::nd_item<1> item, auto &sum_var)`
That way, there should be no conflict with the captured `sum_var` having the same name as the reducer argument passed to the lambda. Does that make sense or would you suggest something else?

liruipeng: @waynemitchell Yes, that's what I suggested. Thank you for trying that. It is interesting to know it "works" or at least it does compile. I was imagining (ab)use `sum_var` using the same variable name at different scopes. Essentially, we have something like
```
int var;
auto glambda = [](auto var) { var += ...};
auto b = glambda(...)

```
I guess the second `var` as the function arg. takes precedence over the first one inside the lambda (?). Maybe we should confirm this with Intel. But I am also happy with the current implementation. So, I approved this PR. Thank you! This is a nice work.

liruipeng (approved):
Looks great! Thank you!

src/struct_mv/boxloop_sycl.h
liruipeng: @waynemitchell Thank you for the reference. I think `[=,hypre_unused_var=sum_var] (sycl::nd_item<1> item, auto &sum_var)` should be the way to go. We will use `sum_var` safely in loop body. In this case, we don't actually have to capture the value of `sum_var` from outer scope, since I see an H2D copy is performed (is it right?). Thanks again!

waynemitchell: @liruipeng OK, I just pushed changes that implement this idea. This really cleans up the calls to the reduction boxloop, which I like a lot. Thanks for the suggestion!

liruipeng: Superb! Thank you for doing it.

PR #433, opened 2021-07-29
hypre's GPU SpGemm

liruipeng:
In this PR, we 
1. improved the performance of hypre's sparse matrix-matrix  on NVIDIA GPUs.
2. fixed hypre's sparse matrix-matrix implementation on AMD GPUs with hip. 

See the following performance report on multiplying $B=A^2$

#### Lassen (V100) with Umpire (time in seconds) 
| A | New hypre | hypre master | cusparse 10 |
| -  |--------------- | ----------------- | ----------------- |
| 150^3-27pt | 0.07 | 0.18 | 1.11|
| 200^3-7pt   | 0.04 | 0.11 | 0.49|
| 4096^2-9pt | 0.11 | 0.38 | 1.33|
| 4096^2-5pt | 0.08 | 0.16 | 0.63|

- M100 has more memory than V100 (32 GB vs. 16 GB), so we solved larger problems on M100;
- rocSPARSE's implementation of SpGEMM is much better than cuSPARSE, which has performance very close to our implementation on the tested problems (also algorithm-wise they are close);
- One downside of rocSPARSE's is it's less memory efficient than ours. E.g., it runs out-of-memory for $256^3$ 27pt problem but ours can compute it in 1.02(s).

##### Byproducts

1. User provides GPU allocator/deallocator  
```
typedef void (*GPUMallocFunc)(void **, size_t);
typedef void (*GPUMfreeFunc)(void *);

hypre_SetUserDeviceMalloc(GPUMallocFunc func)
hypre_SetUserDeviceMfree(GPUMfreeFunc func)
```

2. Bug fixes in `src/parcsr_ls/partial.c` and `src/parcsr_ls/par_interp_trunc_device.c`

- [x] tux
- [x] lassen
- [x] ray
- [x] redwood


liruipeng:
The benchmark problems on lassen have up to 10% speedup in AMG setup from this PR.

liruipeng:
@rfalgout @ulrikeyang @AerysL This PR is ready for review. Thanks!

src/utilities/handle.h
```diff
@@ -71,14 +74,17 @@ typedef struct
 #define hypre_HandleStructCommRecvBufferSize(hypre_handle)       hypre_CudaDataStructCommRecvBufferSize(hypre_HandleCudaData(hypre_handle))
 #define hypre_HandleStructCommSendBufferSize(hypre_handle)       hypre_CudaDataStructCommSendBufferSize(hypre_HandleCudaData(hypre_handle))
 #define hypre_HandleSpgemmUseCusparse(hypre_handle)              hypre_CudaDataSpgemmUseCusparse(hypre_HandleCudaData(hypre_handle))
-#define hypre_HandleSpgemmNumPasses(hypre_handle)                hypre_CudaDataSpgemmNumPasses(hypre_HandleCudaData(hypre_handle))
+#define hypre_HandleSpgemmAlgorithm(hypre_handle)                hypre_CudaDataSpgemmAlgorithm(hypre_HandleCudaData(hypre_handle))
```
rfalgout, line 77: Why is Spgemm stuff in utilities?

rfalgout:
Hi @liruipeng .  Why were there differences in the saved files?  Related to this (maybe), what were we using as the default Spgemm before and what are we using now?  I recall you saying at one point that the default should be CUDA.  Given these results, I'm not sure I agree, so I am curious.  Thanks!

pbauman:
Hi @liruipeng I don't mean to butt-in, but I pulled this branch wanting to test turning off the sorting in the rocSPARSE SpGEMM implementation (beyond what I've done already) in the hopes that we could maybe tack that on here, and I noticed that when I pass `-verify` to the `ij_mm` test on my AMD GPU, I get a segfault.

<pre>
./ij_mm -27pt -cusparse -verify
CPU #OMP THREADS 1
Segmentation fault (core dumped)
</pre>

If I don't pass the `-verify` flag, I'm fine:

<pre>
./ij_mm -27pt -cusparse
CPU #OMP THREADS 1
  Laplacian_27pt:
    (nx, ny, nz) = (10, 10, 10)
    (Px, Py, Pz) = (1, 1, 1)

Generate Matrix A:
Generate Matrix A:
  wall clock time = 0.000000 seconds
  wall MFLOPS     = 0.000000
  cpu clock time  = 0.000727 seconds
  cpu MFLOPS      = 0.000000

A 1000 x 1000
P 1000 x 85
Device Parcsr Matrix-by-Matrix, RAP2:
Device Parcsr Matrix-by-Matrix, RAP2:
  wall clock time = 0.000000 seconds
  wall MFLOPS     = 0.000000
  cpu clock time  = 0.003883 seconds
  cpu MFLOPS      = 0.000000

Done ...
</pre>

src/utilities/handle.h
liruipeng: These are for selecting algorithms and parameters for CSR GEMM. We don't pass in parameters to the routines since they are very deep in the call stack, so we reply on some global variable (hypre_handle) for it.

liruipeng:
> Hi @liruipeng . Why were there differences in the saved files? Related to this (maybe), what were we using as the default Spgemm before and what are we using now? I recall you saying at one point that the default should be CUDA. Given these results, I'm not sure I agree, so I am curious. Thanks!

OK. The default in the library is to use cuSPARSE, while the default in the driver `ij.c` is to use ours. We changed our algorithms, so that's why.

liruipeng:
> Hi @liruipeng I don't mean to butt-in, but I pulled this branch wanting to test turning off the sorting in the rocSPARSE SpGEMM implementation (beyond what I've done already) in the hopes that we could maybe tack that on here, and I noticed that when I pass `-verify` to the `ij_mm` test on my AMD GPU, I get a segfault.
> 
> ./ij_mm -27pt -cusparse -verify
> CPU #OMP THREADS 1
> Segmentation fault (core dumped)
> If I don't pass the `-verify` flag, I'm fine:
> 

No problem at all. It might be not very wise to write it in this way. But can you try `./ij_mm -27pt -cusparse -verify 1`? Thanks!

ulrikeyang:
Why are we still using cuSPARSE as default. It’s clearly slower and not as stable as Ruipeng’s spGEMM?

pbauman:
>  But can you try ./ij_mm -27pt -cusparse -verify 1? Thanks!

Ahhh, OK! I misread the code, sorry. That worked! I'll follow up in a bit with some tests on turning off the sorting in rocSPARSE. I think that would fit well with this PR (and will also pretty significantly speed up SpGEMM using rocSPARSE).

liruipeng:
> Why are we still using cuSPARSE as default. It’s clearly slower and not as stable as Ruipeng’s spGEMM? 

So far, as of CUDA 11, yes for CUDA. But for rocSPARSE it might be not or very close. So, I still think it's good to keep cuSPARSE/rocSPARSE as the default.

src/utilities/handle.h
rfalgout: Okay, so are these essentially "runtime configuration parameters"?

liruipeng:
> > But can you try ./ij_mm -27pt -cusparse -verify 1? Thanks!
> 
> Ahhh, OK! I misread the code, sorry. That worked! I'll follow up in a bit with some tests on turning off the sorting in rocSPARSE. I think that would fit well with this PR (and will also pretty significantly speed up SpGEMM using rocSPARSE).

Feel free to push to this PR. Thanks!

src/utilities/handle.h
liruipeng: Yes.

rfalgout:
> We changed our algorithms, so that's why.

Okay, thanks!

src/utilities/handle.h
rfalgout: Okay, that makes sense.  Thanks!

rfalgout (approved):

pbauman:
> Feel free to push to this PR. Thanks!

OK, I tried `-27pt`, `-9pt`, `-difconv`, with the change and all are passing with the `ij_mm` test at many different sizes. So I can push if you want (it's literally just deleting the two sorts) or if you'd rather it a separate PR after this, that's also fine, just let me know. But on `-27pt`, about 50% faster without the (unneeded!) sorts with rocSPARSE. I'm glad you brought that up!

liruipeng:
> > Feel free to push to this PR. Thanks!
> 
> OK, I tried `-27pt`, `-9pt`, `-difconv`, with the change and all are passing with the `ij_mm` test at many different sizes. So I can push if you want (it's literally just deleting the two sorts) or if you'd rather it a separate PR after this, that's also fine, just let me know. But on `-27pt`, about 50% faster without the (unneeded!) sorts with rocSPARSE. I'm glad you brought that up!

I think you can just push the changes. It won't affect the rest of this PR. Thanks!

pbauman:
Pushed  827e799 to this branch. Thanks!

ulrikeyang (approved):
Thanks.

PR #434, opened 2021-07-29
Add hypre_IntArray

waynemitchell:
This PR adds a hyrpe_IntArray data structure, which is very similar to something like a simplified hypre_Vector that stores integers instead of complex/real values. hypre_IntArray is a basic wrapper for HYPRE_Int* that contains the size and memory location with appropriate create, destroy, copy functions etc.

In addition, I have wrapped the AMG CFMarkerArray and DofFuncArray with hypre_IntArray as a starting test case for use of this data structure, and I have switched the default location of these arrays to the device as opposed to the host to prevent unnecessary copies.


waynemitchell:
@liruipeng @ulrikeyang @rfalgout , I am curious to get your thoughts on this PR. In particular:
1. What do you think of the general design of the hypre_IntArray, and do you think this is a desirable addition to the code in general?
2. I've started by just wrapping the CF marker and the DofFunc of BoomerAMG with hyrpe_IntArray. These are passed as arguments to lots of different functions as plain HYPRE_Int*, so rather than going through and replacing all instances of this with hypre_IntArray*, I've only made changes where the extra info provided by the hypre_IntArray is needed, mostly places where allocation happens (the coarsening routines, for example, may allocate the CF marker). Do you think it is worthwhile to go through and do a more thorough replacement of HYPRE_Int* with hypre_IntArray?
3. I noticed the hypre_ParAMGData members hypre_ParAMGDataDofPoint[Array] and hypre_ParAMGDataPointDofMap[Array] don't seem to be used anywhere? Are these deprecated/safe to remove?
4. Wrapping the DofFunc with hypre_IntArray is a bit tricky because there is a HYPRE_BoomerAMGSetDofFunc() function available, where the user just passes in a HYPRE_Int*. So unless we change the user interface (which I know is not preferred), we basically have to guess/infer the size and memory location when setting up the hypre_IntArray. My current workaround is to assume the memory location is host (which was the assumption up to now) and to set the size according to the matrix size when AMGSetup is called. Thoughts here?

src/parcsr_ls/par_lr_interp_device.c
```diff
@@ -26,6 +26,22 @@ __global__ void hypreCUDAKernel_compute_aff_afc_epe( HYPRE_Int nr_of_rows, HYPRE
 
 __global__ void hypreCUDAKernel_compute_dlam_dtmp( HYPRE_Int nr_of_rows, HYPRE_Int *AFF_diag_i, HYPRE_Int *AFF_diag_j, HYPRE_Complex *AFF_diag_data, HYPRE_Int *AFF_offd_i, HYPRE_Complex *AFF_offd_data, HYPRE_Complex *rsFC, HYPRE_Complex *dlam, HYPRE_Complex *dtmp );
 
+struct equal_pred : public thrust::unary_function<HYPRE_Int, bool>
```
liruipeng: Can we use https://github.com/hypre-space/hypre/blob/f4336b818dc73ddc6cf0b3fb1fe4a978edf7227e/src/utilities/cuda_utils.h#L849?

src/parcsr_ls/par_amg.c
```diff
@@ -3576,8 +3598,19 @@ hypre_BoomerAMGSetDofFunc( void     *data,
...
+   /* NOTE: size of hypre_IntArray will be set during AMG setup 
+    *       the memory location is assumed to be host */
+   if (dof_func == NULL)
+   {
+      hypre_ParAMGDataDofFunc(amg_data) = NULL;
+   }
+   else
+   {
+      hypre_ParAMGDataDofFunc(amg_data) = hypre_IntArrayCreate(-1);
+      hypre_IntArrayMemoryLocation(hypre_ParAMGDataDofFunc(amg_data)) = HYPRE_MEMORY_HOST;
```
liruipeng: Do we assume `dof_func` is always a host pointer?

src/parcsr_ls/par_coarse_parms.c
```diff
@@ -125,3 +125,37 @@ hypre_BoomerAMGCoarseParms(MPI_Comm       comm,
...
+                           hypre_IntArray  **coarse_dof_func_ptr,
+                           HYPRE_BigInt    **coarse_pnts_global_ptr) 
+{
+   HYPRE_Int ierr = 0;
+
+#if defined(HYPRE_USING_CUDA) || defined(HYPRE_USING_HIP)
+   HYPRE_Int use_device = hypre_IntArrayMemoryLocation(CF_marker) != hypre_MEMORY_HOST;
+   if (num_functions > 1)
+   {
+      use_device = (hypre_IntArrayMemoryLocation(dof_func) != hypre_MEMORY_HOST &&
```
liruipeng: How about `hypre_GetExecPolicy2(hypre_IntArrayMemoryLocation(dof_func), hypre_IntArrayMemoryLocation(CF_marker))`?

liruipeng:
> @liruipeng @ulrikeyang @rfalgout , I am curious to get your thoughts on this PR. In particular:
> 
> 1. What do you think of the general design of the hypre_IntArray, and do you think this is a desirable addition to the code in general?

Looks great!

> 2. I've started by just wrapping the CF marker and the DofFunc of BoomerAMG with hyrpe_IntArray. These are passed as arguments to lots of different functions as plain HYPRE_Int*, so rather than going through and replacing all instances of this with hypre_IntArray*, I've only made changes where the extra info provided by the hypre_IntArray is needed, mostly places where allocation happens (the coarsening routines, for example, may allocate the CF marker). Do you think it is worthwhile to go through and do a more thorough replacement of HYPRE_Int* with hypre_IntArray?

I think what we have now is good enough. We can make more replacement in the future if needed.

> 3. I noticed the hypre_ParAMGData members hypre_ParAMGDataDofPoint[Array] and hypre_ParAMGDataPointDofMap[Array] don't seem to be used anywhere? Are these deprecated/safe to remove?

Idk what they are. Maybe @ulrikeyang can comment. They don't seem to be used.

> 4. Wrapping the DofFunc with hypre_IntArray is a bit tricky because there is a HYPRE_BoomerAMGSetDofFunc() function available, where the user just passes in a HYPRE_Int*. So unless we change the user interface (which I know is not preferred), we basically have to guess/infer the size and memory location when setting up the hypre_IntArray. My current workaround is to assume the memory location is host (which was the assumption up to now) and to set the size according to the matrix size when AMGSetup is called. Thoughts here?

I have question about it. We generally assume that when hypre is configured with GPUs, all pointers passed in by users are GPU pointers. I am not sure if we should make exception for `dof_func`. The assumption on size is fine to me.

src/parcsr_ls/par_lr_interp_device.c
waynemitchell: Yes, thanks.

src/parcsr_ls/par_amg.c
waynemitchell: Per our discussion last week, I will change this to assume a device pointer.

src/parcsr_ls/par_coarse_parms.c
waynemitchell: Yup, thanks.

rfalgout (approved):
Looks good to me, @waynemitchell !  I will ask Daniel to take a look at the MGR changes.

rfalgout:
Hi @oseikuffuor1 .  Would mind taking a look at the MGR changes in this PR?  Thanks!

src/parcsr_ls/par_mgr_setup.c
```diff
@@ -1377,33 +1395,35 @@ hypre_MGRSetupFrelaxVcycleData( void *mgr_vdata,
...
           for (i = 0; i < local_size; i++)
           {
-            if (CF_marker_array[lev][i] == 1)
+            if (hypre_IntArrayData(CF_marker_array[lev])[i] == 1)
             {
               CF_marker_local[i] = 0;
             }
           }
-          CF_marker_array_local[lev_local] = CF_marker_local;
+          CF_marker_array_local[lev_local] = hypre_IntArrayCreate(local_size);
```
oseikuffuor1: @waynemitchell why is a new IntArray created here again? Is this different from what was created in line 1370?

waynemitchell: Ah, yes, this is redundant. Thanks for the catch.

oseikuffuor1 (approved):
Looks good to me. Thanks @waynemitchell !!

src/parcsr_ls/par_amg_setup.c
```diff
@@ -625,7 +637,7 @@ hypre_BoomerAMGSetup( void               *amg_vdata,
       k = num_functions - 1;
       for (j = indx-1; j > -1; j--)
       {
-         dof_func[j] = k--;
+         hypre_IntArrayData(dof_func)[j] = k--;
```
liruipeng: If `dof_func` is on device, this part needs to be ported as well?

waynemitchell: @liruipeng The CPU code here seems a bit over-complicated... the desired results is:
for (i = 0; i < local_size; i++)
   dof_func[i] = (i + first_local_row) % num_functions
Is that correct? Or am I misunderstanding how the dof_func should be set?

liruipeng: > @liruipeng The CPU code here seems a bit over-complicated... the desired results is:
> for (i = 0; i < local_size; i++)
> dof_func[i] = (i + first_local_row) % num_functions
> Is that correct? Or am I misunderstanding how the dof_func should be set?

I think it is correct. I also think CPU code is a bit over-complicated. Not sure why it was coded in that way. We may not use that in the GPU implementation. But first of all, is this part needed to port to GPUs? Just want to make sure if I understood the situation correctly. Thanks!

waynemitchell: I think this is a pretty cheap operation on the CPU, so this would probably be fine as is if you use unified memory. But, this is straightforward to port, so I'll go ahead and do it. 

By the way, in general, when porting to the GPU, do we expect ported features to work without unified memory? If so, then yes, this section of code definitely needs to be ported (I'd expect it to crash in the case where the user configured without unified memory and then did not pass in dof_func).

liruipeng: We want hypre to work without unified memory as much as we can. HIP has no UVM support as far as I know. So, we should port this as well. Thanks again @waynemitchell !

ulrikeyang (approved):
This looks good. Thanks

PR #441, opened 2021-08-05
Fix mingw

liruipeng:
This PR (#440) fixes portability issue of `ffs` on MinGW.

### DIFF
--- a/src/utilities/_hypre_utilities.h
+++ b/src/utilities/_hypre_utilities.h
@@ -2880,3 +2880,4 @@ hypre_UnorderedBigIntMapPutIfAbsent( hypre_UnorderedBigIntMap *m,
 #endif
 
 #endif
+
--- a/src/utilities/_hypre_utilities.h
+++ b/src/utilities/_hypre_utilities.h
@@ -1792,7 +1792,7 @@ static inline HYPRE_Int
 first_lsb_bit_indx( hypre_uint x )
 {
    HYPRE_Int pos;
-#if defined(_MSC_VER)
+#if defined(_MSC_VER) || defined(__MINGW64__)
    if (x == 0)
    {
       pos = 0;
--- a/src/utilities/hopscotch_hash.h
+++ b/src/utilities/hopscotch_hash.h
@@ -133,7 +133,7 @@ static inline HYPRE_Int
 first_lsb_bit_indx( hypre_uint x )
 {
    HYPRE_Int pos;
-#if defined(_MSC_VER)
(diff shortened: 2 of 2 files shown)


rfalgout:
I don't know anything about using the `__MINGW64__` flag, so if either of you (@liruipeng @victorapm) can enlighten me, that would be great.  I added @victorapm because he is familiar with the previous `_MSC_VER` guards in this code.  Thanks!

liruipeng:
> I don't know anything about using the `__MINGW64__` flag, so if either of you (@liruipeng @victorapm) can enlighten me, that would be great. I added @victorapm because he is familiar with the previous `_MSC_VER` guards in this code. Thanks!

I don't know much about it either. It is for gcc on Windows platforms.

src/utilities/_hypre_utilities.h
```diff
@@ -2880,3 +2880,4 @@ hypre_UnorderedBigIntMapPutIfAbsent( hypre_UnorderedBigIntMap *m,
 #endif
 
 #endif
+
```
prj-, line 2883:
```suggestion
```

prj-:
> I don't know anything about using the `__MINGW64__` flag, so if either of you (@liruipeng @victorapm) can enlighten me, that would be great. I added @victorapm because he is familiar with the previous `_MSC_VER` guards in this code. Thanks!

It's rather commonly paired with `_MSC_VER`, see, e.g., Valgrind [source code](https://sourceware.org/git/?p=valgrind.git;a=blob;f=include/valgrind.h;h=04a747c7a8f130c384a2a1acfe892fd4eab0ebca;hb=refs/heads/master#l6755).

src/utilities/_hypre_utilities.h
liruipeng: @prj- We don't directly edit this file which is generated by a script. Is there a problem for having an extra newline? Thanks!

victorapm:
Hi @rfalgout, `mingw-w64` is a compiler based on gcc for Windows 64 platforms. Its `string.h` do not support the function `ffs` that is used in `first_lsb_bit_indx`, so this fix is important because it will replace the call to `ffs` to another implementation of it: https://github.com/hypre-space/hypre/blob/837e8ea8e73333902f8fb68dcca7986eb637b260/src/utilities/_hypre_utilities.h#L1796-L1806

Thanks for the fix, @prj-

victorapm (approved):

rfalgout (approved):
Thanks all!

PR #444, opened 2021-08-07
Fix compile warnings

waynemitchell:
Minor fixes to get rid of some compiler warnings.


rfalgout (approved):
Looks good.  In general, we should move away from `ierr`, but I think that's a bigger undertaking.  Thanks!

PR #448, opened 2021-08-10
New Struct and SemiStruct multiplication interfaces

victorapm:
This PR contains the new struct and semi-struct matrix/matrix multiplication interfaces. The goal is to solve the issue with serializing matrix/matrix multiplication in parallel that happens for problems with multiple parts owned by different ranks.

TODO list:

- [x] Implement new `StructMatrixMult` and replace the old ones.
- [x] Check that the regression tests for `TEST_structmat` pass.
- [x] Finish new `SStructMatrixMult` and `SStructPMatrixMult` functions.
- [x] Fix regression tests and remaining memory leaks.
- [x] Implement the idea of aggregating communication packages needed for `SStructMatrixMult`
- [x] Update functions names as suggested by @rfalgout.
- [x] Run performance benchmarks with the new `SStructMatrixMult`.
- [x] Add new tests to `sstructmat`.


rfalgout:
Hi @victorapm .  Thanks for making this PR!  I have not finished looking at this, but wanted to make some name change suggestions.  I realize that we have inconsistencies in hypre wrt the matrix-vector linear algebra routines, so this is difficult.  I'll make additional comments in the source itself.  The short version is that I would like to retain `Matmult` for the general matrix multiplication routine (the one that takes an arbitrary number of matrices, your `Multgroup` I think) and use `Matmat` for the product of two matrices (to mirror `Matvec`).  I am okay with prefixing everything with `Matrix`.  That is probably the right thing to do from a design perspective.  Talk to you in a little while.  Thanks again!

src/sstruct_ls/ssamg_setup_rap.c
```diff
@@ -135,7 +135,7 @@ hypre_SSAMGComputeRAPNonGlk( hypre_SStructMatrix  *A,
             /* Use generic StructMatmult */
             hypre_StructMatrix *smatrices[3] = {sA, sP, sP};
 
-            hypre_StructMatmult(3, smatrices, 3, terms, trans, &sAc[part][vi][vi]);
+            hypre_StructMatrixMultGroup(3, smatrices, 3, terms, trans, &sAc[part][vi][vi]);
```
rfalgout: Change to `hypre_StructMatrixMatmult`

src/struct_mv/_hypre_struct_mv.h
```diff
@@ -2032,6 +2032,72 @@ hypre_BoxArrayBox(hypre_StructVectorDataSpace(vector), b)
...
+   HYPRE_Int       types[MAXTERMS]; /* types of computations to do for each term */
+   HYPRE_Complex  *tptrs[MAXTERMS]; /* pointers to matrix data for each term */
+   HYPRE_Complex  *mptr;            /* pointer to matrix data for M */
+} hypre_StructMMHelper;
+
+/*--------------------------------------------------------------------------
+ * StructMMData data structure
+ *--------------------------------------------------------------------------*/
+
+typedef struct hypre_StructMMData_struct
```
rfalgout: Nice job getting rid of the `void *` stuff.  I would call this `hypre_StructMatrixMatmultData_struct`.  It's long, but that's more a result of the term `StructMatrix` being long.  One thing we were doing in Struct before was to abbreviate to just `Struct` for `StructMatvec`, `StructMatmult`, etc.  Do something similar for the helper struct (replace `Data` with `Helper`).

src/struct_mv/_hypre_struct_mv.h
```diff
@@ -2282,7 +2348,16 @@ HYPRE_Int hypre_ReadBoxArrayData ( FILE *file , hypre_BoxArray *box_array , hypr
 HYPRE_Int hypre_ReadBoxArrayData_CC ( FILE *file , hypre_BoxArray *box_array , hypre_BoxArray *data_space , HYPRE_Int stencil_size , HYPRE_Int real_stencil_size , HYPRE_Int constant_coefficient , HYPRE_Int dim , HYPRE_Complex *data );
 
 /* struct_matmult.c */
-HYPRE_Int hypre_StructMatmult ( HYPRE_Int nmatrices , hypre_StructMatrix **matrices , HYPRE_Int nterms , HYPRE_Int *terms , HYPRE_Int *transposes , hypre_StructMatrix **M_ptr );
+hypre_StructMMData* hypre_StructMatrixMultCreate ( HYPRE_Int nmatrices_in , hypre_StructMatrix **matrices_in , HYPRE_Int nterms , HYPRE_Int *terms_in , HYPRE_Int *transposes_in );
+HYPRE_Int hypre_StructMatrixMultDestroy ( hypre_StructMMData *mmdata );
+HYPRE_Int hypre_StructMatrixMultSetup ( hypre_StructMMData  *mmdata , hypre_StructMatrix **M_ptr );
+HYPRE_Int hypre_StructMatrixMultCommunicate ( hypre_StructMMData *mmdata );
```
rfalgout: On second thought, maybe we should go back to `StructMatmult` without `Matrix`, at least for the internal routines, just for brevity.  The interfaces we expose to users can be longer.

src/struct_mv/_hypre_struct_mv.h
```diff
@@ -2282,7 +2348,16 @@ HYPRE_Int hypre_ReadBoxArrayData ( FILE *file , hypre_BoxArray *box_array , hypr
...
+HYPRE_Int hypre_StructMatrixMultCompute ( hypre_StructMMData *mmdata , hypre_StructMatrix *M );
+HYPRE_Int hypre_StructMatrixMultGroup ( HYPRE_Int nmatrices , hypre_StructMatrix **matrices , HYPRE_Int nterms , HYPRE_Int *terms , HYPRE_Int *trans , hypre_StructMatrix **M_ptr );
+HYPRE_Int hypre_StructMatrixMult ( hypre_StructMatrix *A , hypre_StructMatrix *B , hypre_StructMatrix **M_ptr );
```
rfalgout: `Matmat` instead of `Mult`

rfalgout:
Hi @victorapm .  The code looks good!  Hopefully this will help in SStruct.  Thanks!

victorapm:
Hi @rfalgout, the new interface for `SStructMatrix` multiplication solves the issue with parallelization. Considering a 2x2 parts problem solved with 4 ranks, one part per rank, I saw an speedup of nearly 2x over the old code (in recmat-merge). I'm investigating if I can do better. However, as of  the last commit d94c576, some tests are failing and there are still some memory leaks that I'm trying to solve. We talk later :)

victorapm:
Hi all, 

here are some results with the new `SStructMatMult` fur the `cubes` problem. `Old` refers to the code in `recmat-merge`. The first row is for 8 parts, while the second row is for 27 parts. I'm considering one rank per part.

|    |    |    |            |             | New     |  New     | New              | Old     | Old      | Old              |
|----|----|----|------------|-------------|---------|-------|---------------|---------|-------|---------------|
| Nx | Ny | Nz | Local DOFs | Global DOFs | SStructMatmult | SSAMG Setup | Matmult/Setup | SStructMatmult | SSAMG Setup | Matmult/Setup |
| 2  | 2  | 2  | 941,192    | 7,529,536   | 2.24    | 3.22  | 69.57%        | 6.24    | 7.1   | 87.89%        |
| 3  | 3  | 3  | 941,192    | 25,412,184  | 5.34    | 8.64  | 61.81%        | 19.7    | 23.1  | 85.28%        |

src/struct_mv/coarsen.c
```diff
@@ -478,6 +478,10 @@ hypre_StructCoarsen( hypre_StructGrid  *fgrid,
    /* create new coarse grid */
    hypre_StructGridCreate(comm, ndim, &cgrid);
 
+   /* Set global size to a number different than zero to
+      avoid its computation on hypre_StructGridAssemble. */
+   hypre_StructGridGlobalSize(cgrid) = -1;
```
rfalgout, line 487: Where will this be set to a correct value?

victorapm: Hi @rfalgout, it will be computed at `hypre_StructMatmultCommunicate`: https://github.com/hypre-space/hypre/blob/c7cab52acc7e8a20c0246f546133ff2c0a190a10/src/struct_mv/struct_matmult.c#L878

I believe it's a bit counter-intuitive to have a data member from `hypre_StructGrid` being computed at a `hypre_StructMatrix` function, but this solution was the best that I could find to reduce the number of `MPI_Allreduce` calls. What do you think about this?

rfalgout: The problem is that Coarsen should return a valid Grid.  Maybe we can redefine Coarsen to return an unassembled Grid, which would require us to explicitly call Assemble.  I don't know if that would help here, because I don't really understand how this reduces calls to Allreduce to begin with.  Can you explain in words or pseudocode?  I'm being lazy. :)

victorapm: I think it's a good idea to change the behavior of `hypre_StructCoarsen` to return an unassembled grid. 

Currently, `hypre_StructCoarsen` calls `hypre_StructGridAssemble`, which has four `MPI_Allreduce` calls (one from that function itself and the other 3 from `hypre_BoxManAssemble` that is called from GridAssemble). In the semi-struct matrix multiplication setup, we have as many calls to `hypre_StructCoarsen` as the number of parts multiplied by the number of variables. So, if we return an unassembled grid from `hypre_StructCoarsen`, we could avoid `nparts * nvars` calls to MPI_Allreduce to a single one with vector size `nparts * nvars` times larger than the respective one in the old code. Also, this would solve the blocking behavior in `SStructMatmultSetup` due to the collective call (allreduce). 

I'm not sure if that was sufficiently clear, we can meet otherwise :)

victorapm: PS: although the new interface for semi-struct matrix multiplication in this PR showed speedup wrt to the old, there's still some serialization happening in the setup function for semi-struct matrix multiplication (`hypre_SStructMatrixMultSetup`).

rfalgout: This definitely helps, Victor.  Thanks!  One more question.  I had misunderstood you to say that the global sizes for each of StructGrid grids were eventually computed, but that doesn't appear to be the case.  Is it really a good idea to have incomplete objects here?  Maybe we should discuss some more redesign ideas here.  For example, if I recall correctly, the SStructMatrix keeps a number of "variable grids" around that are reused in the different StructMatrix matrices to avoid replicating a bunch of grid objects (so each of the SStructMatrix matrices have references to possibly the same grids).  Are you doing anything like that here?  If not, maybe we can think about making some additional changes to the Matmult interface to help us have more control over grid construction in the SStruct code.

Actually, now that I look at this a bit closer, I wonder if we should maybe just remove the GlobalSize component of the grid altogether.  It doesn't look like it serves any useful purpose other than to help us compute flop information.  What do you think?  Or we could just turn it on when we have logging turned up high enough or something?

Are you planning to make the "Matmult" name change for SStructMatrix and SStructPMatrix as well (i.e., SStructMatrixMult -> SStructMatmult and SStructPMatrixMult -> SStructPMatmult)?  We should.

victorapm: Hi @rfalgout,

> I had misunderstood you to say that the global sizes for each of StructGrid grids were eventually computed, but that doesn't appear to be the case

In the case of a `hypre_StructGrid` generated from `hypre_SStructMatrixMultSetup`, its global size will be computed at https://github.com/hypre-space/hypre/blob/c7cab52acc7e8a20c0246f546133ff2c0a190a10/src/sstruct_mv/sstruct_grid.c#L474

> Is it really a good idea to have incomplete objects here? Maybe we should discuss some more redesign ideas here

I don't feel very comfortable with having incomplete objects either, but this was the best solution I could find to avoid having multiple calls to `MPI_Allreduce` from each of the struct grids belonging to a `hypre_SStructGrid`. Setting `global_size = -1` bypasses the computation of  `global_size` in `hypre_StructGridAssemble` and delays such calculation to `hypre_SStructGridComputeGlobalSize`, which has a single `MPI_Allreduce` that computes the global sizes of all `nparts*nvars` struct grids. Would you have some other idea in mind?

> Are you doing anything like that here?

If I understand the question correctly, the answer is yes. I do reuse the struct grids pointed by the `sgrids` member of a `hypre_SStructPGrid` in a `hypre_SStructMatrix` and its belonging struct matrices.

> It doesn't look like it serves any useful purpose other than to help us compute flop information. What do you think?

Good point, `global_size` is mostly used to compute flops. A more important use case though is on `hypre_PFMGComputeDxyz` and also in the SSAMG counterpart:
https://github.com/hypre-space/hypre/blob/c7cab52acc7e8a20c0246f546133ff2c0a190a10/src/struct_ls/pfmg_setup.c#L681

This is performed only on the fine grid matrix though.

> Are you planning to make the "Matmult" name change for SStructMatrix and SStructPMatrix as well

Yes,  I am planning. I started changed some names and pushed the changes, but haven't finished all the work yet :)

rfalgout: This all sounds good.  Regarding `global_size` and the use case you point out, I'm not sure we really need it there anyway.  I can take a look at some point to see, but I think the way you are handling this now is okay for the time being.  Let's keep in mind trying to remove this value altogether, though.  BTW, I think some of the logic in the Dxyz routines needs to be updated to match a recent corner case I fixed on master (unless you've already done that).  I also added a comment there that will help us to decode this routine more easily in the future. :)  Thanks again!

victorapm: Ok, thanks Rob! I updated the Dxyz routines in this branch according to master.

rfalgout: Awesome!

victorapm:
For documentation purposes, here are the most recent results for the `cubes` problem with one part per rank. Here, parts are added in the same direction. Those were ran in a single node of lassen.

| Nparts | Global DOFs | New setup | Old setup |
|--------|-------------|-----------|-----------|
| 1      | 941,192     | 0.87      | 0.87      |
| 2      | 1,882,384   | 1.1       | 1.71      |
| 4      | 3,764,768   | 1.37      | 3.15      |
| 8      | 7,529,536   | 1.71      | 5.85      |
| 16     | 15,059,072  | 2.33      | 11.3      |
| 32     | 30,118,144  | 3.66      | 22.2      |

victorapm:
Hi all, these are the last results considering the change on `hypre_StructCoarsen` for returning an unassembled grid (column newer). I show also the previous results from yesterday and respective speedups wrt to `recmat`. The last remaining step for this PR is adding tests for semi-struct matrix/matrix multiplication.

| Nparts | Global DOFs | Newer setup (f26e8f5) | New setup (d2a8c0a) | Old setup (e297c88) | Newer speedup | New speedup |
|--------|-------------|------------------------|----------------------|---------------------|---------------|-------------|
| 1      | 941,192     | 0.87                   | 0.87                 | 0.87                | 1.0           | 1.0         |
| 2      | 1,882,384   | 0.99                   | 1.1                  | 1.71                | 1.7           | 1.6         |
| 4      | 3,764,768   | 1.13                   | 1.37                 | 3.15                | 2.8           | 2.3         |
| 8      | 7,529,536   | 1.25                   | 1.71                 | 5.85                | 4.7           | 3.4         |
| 16     | 15,059,072  | 1.38                   | 2.33                 | 11.3                | 8.2           | 4.8         |
| 32     | 30,118,144  | 1.86                   | 3.66                 | 22.2                | 11.9          | 6.1         |

rfalgout:
> Hi all, these are the last results considering the change on `hypre_StructCoarsen` for returning an unassembled grid

Nice results, Victor!

src/sstruct_mv/HYPRE_sstruct_mv.h
```diff
@@ -530,6 +530,38 @@ HYPRE_Int
 HYPRE_SStructGridSetNumGhost(HYPRE_SStructGrid  grid,
                              HYPRE_Int         *num_ghost);
 
+/**
+ * Helper function for returning the box dimensions of a (part,var)-structured grid.
+ **/
+HYPRE_Int
+HYPRE_SStructGridGetVariableBox(HYPRE_SStructGrid  grid,
```
rfalgout, line 537: Should this be a user-level interface routine?

src/sstruct_mv/protos.h
```diff
@@ -30,7 +30,9 @@ HYPRE_Int HYPRE_SStructGridAssemble ( HYPRE_SStructGrid grid );
 HYPRE_Int HYPRE_SStructGridSetPeriodic ( HYPRE_SStructGrid grid , HYPRE_Int part , HYPRE_Int *periodic );
 HYPRE_Int HYPRE_SStructGridSetNumGhost ( HYPRE_SStructGrid grid , HYPRE_Int *num_ghost );
 HYPRE_Int HYPRE_SStructGridPrintGLVis ( HYPRE_SStructGrid grid, const char *meshprefix, HYPRE_Real *trans, HYPRE_Real *origin );
-HYPRE_Int hypre_SStructGridPrint ( hypre_SStructGrid *grid , const char *filename );
+HYPRE_Int HYPRE_SStructGridProjectBox ( HYPRE_SStructGrid grid , HYPRE_Int *ilower , HYPRE_Int *iupper , HYPRE_Int *origin , HYPRE_Int *stride );
```
rfalgout, line 33: I know we have this in a number of places in hypre, but the 'protos.h' file should not contain the prototypes for the user-level interface routines.  I don't know why we have these here.  It's totally possible that I was even the one who added them, but if I did, I shouldn't have. :)

rfalgout (approved):
This looks good to me @victorapm .  We can address my two comments later, especially the user interface design question.  Nice job on this!

victorapm:
Thanks @rfalgout!

ulrikeyang (approved):
This looks good. Thanks

PR #453, opened 2021-08-11
Change default timer to MPI_Wtime() when using MPI

rfalgout:
The timers in hypre date back a really long time and did not originally use MPI_Wtime().  This (finally) changes the default to be MPI_Wtime() whenever MPI is also being used.


liruipeng (approved):
Looks good to me. Thanks!

ulrikeyang (approved):
This looks good. Thanks

PR #458, opened 2021-08-20
Dcsrmv analysis

liruipeng:
This PR (by @pbauman #430) is a hook to be able to call rocsparse_dcsrmv_analysis when using rocSPARSE on AMD GPUs. This can give substantial performance benefits on ensuing SpMVs. Feel free to suggest/change the API name. I did not guard the API name, only internally in the function so that the user (or the library) can just drop this in their code and not worry about #ifdefing. I've added an example of using it in the ij test (after the 27pt stencil matrix construction). Discussions of how/where to plug this in can ensue, but at least this will be present.


src/test/ij.c
```diff
@@ -2265,6 +2265,8 @@ main( hypre_int argc,
    else if ( build_matrix_type == 4 )
    {
       BuildParLaplacian27pt(argc, argv, build_matrix_arg_index, &parcsr_A);
+
+      hypre_CSRMatrixGpuSpMVAnalysis(hypre_ParCSRMatrixDiag(parcsr_A));
```
liruipeng, line 2269: Is this what we want eventually? But it's fine for me to leave it as is for now for testing. Thanks!

liruipeng:
@pbauman This looks good to me. I had a minor question in the driver. Thanks!

src/test/ij.c
pbauman: This was what I had in mind and I put it here to serve as an example. I'm certainly open to suggestions/changes/etc.

pbauman: I should add it's what I had in mind as the low level, but user-facing API. Perhaps knobs could be added at the, e.g., `BoomerAMGSetup` level that will trigger calling this API on matrices used in the preconditioner.

rfalgout (approved):
Looks good to me.  Thanks!

pbauman (approved):
This at least adds the knob so users and the library can use it. Further use in the library, e.g. at the `BoomerAMGSetup` level so users can set a flag, is up for discussion, but could be addressed separately. Name changes etc. are also fine, but I think associating this with the CSR matrix makes sense since this is only for `csrmv`.

ulrikeyang (approved):
This looks good. Thanks

PR #461, opened 2021-08-24
Hypre petsc

liruipeng:
This PR has a number of bug fixes and user requests in the process of integrating GPU-hypre with PETSc.

1. Use host linker to link GPU version of hypre library. For testing, change ij drivers to pure host code and use host compiler to compile.
2. Bug fix for nan issues in GPU interpolation 
3. Separate `CUFLAGS` and `CXXFLAGS` in configure
4. Fixed compiler warnings of unused variables.
5. Other minor changes.

- [x] tux
- [x] ray
- [x] lassen
- [x] redwood

Update:
It turned out the bug fix in par_lr_interp_device.c changed saved results on GPUs (almost all of the new results are slightly better than the old), which means there were divided-by-zeros in the original code. However, we did not see nan or inf in the final results. I think that was just luck.


src/lib/Makefile
```diff
@@ -80,20 +102,7 @@ distclean: clean
...
-	${AR} $@ $(PARCSRLSFILES)
-	${AR} $@ $(PARCSRMVFILES)
-	${AR} $@ $(PARCSRBLOCKMVFILES)
-	${AR} $@ $(SEQMVFILES)
-	${AR} $@ $(SSTRUCTLSFILES)
-	${AR} $@ $(SSTRUCTMVFILES)
-	${AR} $@ $(STRUCTLSFILES)
-	${AR} $@ $(STRUCTMVFILES)
-	${AR} $@ $(UTILITIESFILES) $(BLASFILES) $(LAPACKFILES)
+	${AR} $@ $(FILES_HYPRE)
```
rfalgout: This was originally broken into smaller pieces because some systems didn't like the long list (I can't remember the specifics, but might be able to do a bit of research to figure it out).  Is there a reason you are doing it with one line?

src/parcsr_ls/par_amgdd_comp_grid.c
```diff
@@ -913,7 +913,7 @@ HYPRE_Int hypre_AMGDDCompGridFinalize( hypre_ParAMGDDData *amgdd_data )
    HYPRE_Int             real_cnt, ghost_cnt;
    HYPRE_Int             proc, outer_level, level, i, j;
 #if defined(HYPRE_USING_CUDA) || defined(HYPRE_USING_HIP)
-   HYPRE_Int             c_cnt, f_cnt;
+   //HYPRE_Int             c_cnt, f_cnt;
```
rfalgout: Why not just remove this code?

src/test/Makefile
```diff
@@ -131,59 +131,53 @@ distclean: clean
...
+ij_assembly: ij_assembly.o
 	@echo  "Building" $@ "... "
 	${LINK_CC} -o $@ $< ${LFLAGS}
 
 sstruct: sstruct.o
 	@echo  "Building" $@ "... "
 	${LINK_CC} -o $@ $< ${LFLAGS}
 
-struct: struct.${OBJ_SUFFIX}
+struct: struct.obj
```
rfalgout, line 148: I'm not following what's going on here.  Why are some of the drivers using `.o` extensions and others `.obj`?

liruipeng:
It turned out the bug fix in `par_lr_interp_device.c` changed saved results on GPUs (almost all of the new results are slightly better than the old), which means there were divided-by-zeros in the original code. However, we did not see `nan` or `inf` in the final results. I think that was just luck.

ulrikeyang (approved):

rfalgout (approved):
Looks good to me.  Thanks!

Issue #460, opened 2021-08-23
OpenMP error when compiling on Visual Studio HYPRE 2.22.1


nicola-giuliani:
Hi,

I downloaded the latests release (2.22.1) and I am trying to compile it in windows using visual studio 2017. If I enable the usage of OpenMP I get the following compiling error

hypre-2.22.1\src\parcsr_mv\par_csr_matop.c(5367): error C3036: 'max': invalid operator token in OpenMP 'reduction' clause 

Release 2.22.0 compiled fine so I digged to find what introduce the error. I found that the modification was done in commit https://github.com/hypre-space/hypre/commit/ffe4f7384bc9cb95195c04ae0aec1f3609237147 

The problem is that Visual Studio currently support only OpenMP 2. The performance loss disabling OpenMP would be too significant. Is there something I can do except commenting the problematic part?

Bests and thanks,
Nicola

liruipeng:
@nicola-giuliani For now, the best we can do might be commenting out this line https://github.com/hypre-space/hypre/blob/224bab21afd37fa577870f4f2fbc06bc14e3c6d3/src/parcsr_mv/par_csr_matop.c#L5367, before we implement a min/max reduction for MSVC. Thanks!

victorapm:
@nicola-giuliani, thanks for reporting this issue. Could you please test the branch `hypre:msvc-omp-fix` and let us know if that works for you? Thanks!

nicola-giuliani:
@victorapm thanks for the quick reply.  The new branch compiled fine. Thank you again :)

hbadi:
Hi @nicola-giuliani, you can also use oneAPI toolkit from Intel which includes the intel C++ compiler, VS integration and full openMP 4. I'm also looking to compile Hypre on native windows, can you tell me what options did you choose in cmake ?  
Does the `hypre:msvc-omp-fix` is now in the main branch ?

victorapm:
@hbadi, Yes, that fix is present in hypre's master branch.

hbadi:
@victorapm Great !

PR #463, opened 2021-08-27
Fix issue with reduction(max) in MSVC

victorapm:
MSVC does not support the `max` operation in OpenMP reduction clauses. This PR adds a new code branch for doing reduction in OpenMP regions by using a critical region instead. This fixes issue #460


liruipeng (approved):
Looks good. Thanks!

rfalgout (approved):
Looks good.  Thanks!

PR #470, opened 2021-09-01
Fortran gpu

liruipeng:
This PR adds GPU examples for FORTRAN users. Examples `ex5f.f` and `ex12f.f` were updated to demonstrate with 3 approaches

1. Allocate CUDA UVM with `cudaMallocManaged`. The pointers are handled with `C_F_POINTER`. It requires Fortran 2003.
2. OpenMP offload. It requires OMP target support.
3.  CUDA Fortran.

To compile, first build hypre with `./configure --with-cuda --enable-unified-memory --enable-debug --enable-shared`. Go to `src/examples`, type one of the following
```
make use_cuda=1
make use_domp=1
make use_cuf=1
```

#### Tests
- [x] tux
- [ ] ray
- [x] lassen
- [x] redwood


liruipeng:
@rfalgout @ulrikeyang I had to checkout this branch from `hypre-petsc`, so it has a lot of changes from that branch. It will be easier to compare against `hypre-petsc` to see the real changes for FORTRAN. Or, you can review after that branch is merged. Thanks!

src/examples/ex5f.f
```diff
@@ -40,8 +40,12 @@ program ex5f
 
       integer    MAX_LOCAL_SIZE
       integer    HYPRE_PARCSR
+      integer    HYPRE_MEMORY_DEVICE
+      integer    HYPRE_EXEC_DEVICE
 
       parameter  (MAX_LOCAL_SIZE=123000)
+      parameter  (HYPRE_MEMORY_DEVICE=1)
```
rfalgout: Why is this both here and in the `HYPREf.h` file?

rfalgout:
Hi @liruipeng .  I'll wait until `hypre-petsc` is merged before approving this, but a quick look at `HYPREf.h` and `ex5f.f` looks pretty good.  Thanks!

src/examples/ex5f.f
liruipeng: I don't see `HYPREf.h` is included. I might be wrong.

rfalgout: Oh, then we need to add the include.  That's why it's there.  It's in `ex12f.f`, but not this one.

ulrikeyang (commented):
Why are all the saved files changing?

liruipeng:
> Why are all the saved files changing?

Those were from `hypre_petsc`. We fixed some nan issue in GPU interpolation code.

liruipeng:
> Hi @liruipeng . I'll wait until `hypre-petsc` is merged before approving this, but a quick look at `HYPREf.h` and `ex5f.f` looks pretty good. Thanks!

@rfalgout @ulrikeyang @oseikuffuor1 You may review this PR now. I listed some questions and comments in the description of this PR. I am happy to chat if needed.

ulrikeyang (approved):
This looks good to me. Thanks

src/examples/ex5f.f
```diff
@@ -178,9 +206,15 @@ program ex5f
          endif
 
 !        Set the values for row i
+#if defined(HYPRE_USING_CUDA)
+         tmpi(1) = nnz-1;
```
p-costa: Sorry for commenting out of the blue. Just a nit from a Fortran and *hypre* user that is very interested in this PR! :)
```suggestion
         tmpi(1) = nnz-1
```

src/examples/ex12f.f
```diff
@@ -54,7 +54,15 @@ program ex12f
       integer    ent
       integer    nentries, nvalues, stencil_indices(5)
 
-      double precision  values(100), tol
+      double precision tol
+
+#if defined(HYPRE_USING_CUDA)
```
rfalgout: It looks like the Fortran standard does not support preprocessor macros, but this approach appears to be a common approach.  @p-costa , maybe you can comment on this?

src/examples/cuda_wrapper.c
```diff
@@ -0,0 +1,36 @@
...
+#if defined(HYPRE_USING_CUDA)
+
+#define CUDA_MALLOC_MANAGED  device_malloc_managed_
+#define CUDA_FREE            device_free_
+
+int CUDA_MALLOC_MANAGED (const int *nbytes, devptr_t *devicePtr)
+{
+    void *tPtr;
+    int retVal;
+    retVal = (int) cudaMallocManaged (&tPtr, *nbytes, cudaMemAttachGlobal);
```
rfalgout: Is this how Fortran users deal with allocating device memory, call a C interface?

src/examples/ex12f.f
p-costa: Sure! Yes, because in practice `cpp` marcos are supported by the compilers, typically with a `-cpp` flag, or changing the filename to `.F` (or `.F90`). But I'm not sure if the latter approach is as widely supported.

rfalgout: Sounds good, @p-costa.  Thanks!

src/utilities/F90_HYPRE_general.c
```diff
@@ -26,6 +26,30 @@ hypre_F90_IFACE(hypre_finalize, HYPRE_FINALIZE)
    *ierr = (hypre_F90_Int) HYPRE_Finalize();
 }
 
+void
+hypre_F90_IFACE(hypre_setmemorylocation, HYPRE_SETMEMORYLOCATION)
```
rfalgout, line 30: Hi @liruipeng .  It looks like these are not yet used in our examples, correct?  Would it be easy to add them to one of the example drivers (at some point, not necessarily now)?  Thanks!

rfalgout (approved):
This looks good.  Just a few minor outstanding questions.  Thanks @liruipeng !

src/utilities/F90_HYPRE_general.c
liruipeng: Hi, @rfalgout . They are used in ex5f https://github.com/hypre-space/hypre/blob/f51e1663ec33dc545af0a1d7805e9616a4751018/src/examples/ex5f.f#L109

rfalgout: Ahh.  Somehow I missed that.  Thanks!

src/examples/cuda_wrapper.c
liruipeng: I guess yes, although I am not an expert. There is a more neat approach that doesn't need C wrapper such as
(https://docs.nvidia.com/hpc-sdk/compilers/cuda-fortran-prog-guide/index.html#cfexam-matrix-mult)
```
real, device, allocatable, dimension(:,:) :: Adev,Bdev,Cdev
```
My understanding is that this requires special Fortran compilers. Maybe we can ask this to the users.

src/examples/ex5f.f
liruipeng: Thank you, @p-costa! We probably will need your input on how it works.

src/examples/cuda_wrapper.c
rfalgout: I saw that, but I agree it doesn't look to be a portable solution.  I also agree that we should get some input here from our Fortran GPU users.

p-costa: I have used CUDA Fortran, and OpenACC combined with some features of CUDA Fortran for managing memory. This is supported by the `nvfortran` compiler from the [NVIDIA HPC SDK](https://developer.nvidia.com/hpc-sdk). I'm not sure if there is an easy, portable solution for deploying Fortran code on GPUs...

Indeed one can declare and allocate device/managed memory variables as follows:
```fortran
  real, device, allocatable, dimension(:,:) a_d
  (...)
  allocate(a_d)
```
replacing `device` with `managed` would declare it as managed memory.

alternatively, one can do as follows, which is at times convenient:
```fortran
  real, allocatable, dimension(:,:) a
#if defined(USE_CUDA)
  attributes(managed) :: a ! or device
#endif
  (...)
  allocate(a)
```
(here is a short CUDA Fortran porting [document](https://www.pgroup.com/lit/literature/pgi-cuf-pg-2018.pdf)).

flow-phys: I'd be interested in getting an OpenMP4.5 version of one of these examples tested with this PR.  We use "!$omp map data..." pragmas to create and move data onto the GPU.  Its likely that users of this interface with have their own solution but so far the most portable solution for GPUs for fortran code has been OpenMP. 

```fortran
! Declare and allocate on the CPU
real, allocatable, dimension(:,:) a
allocate(a(n,m))

! Allocate device memory
#if defined(USE_GPU)
!$omp target enter data map(alloc:a)
#endif

! Use device here
...
!$omp target data use_device_ptr(a)
CALL call_hypre_routine( a ) 
!$omp end target data
...

! Clean-up device memory
#ifdefined(USE_GPU)
!$omp target exit data map(release:a)
#endif

```

rfalgout: Thanks @p-costa and @flow-phys for your advice and input on this!

src/examples/Makefile
```diff
@@ -9,32 +9,57 @@
...
+LINK_CC   = ${CC}
+LINK_CXX  = ${CXX}
+LINK_FC   = ${F77}
+
 HYPRE_DIR = ../hypre
 
+########################################################################
+# CUDA
+########################################################################
+ifeq ($(use_cuda), 1)
```
rfalgout, line 26: I wasn't sure about the portability of `ifeq` but there don't seem to be any complaints online about it.

src/examples/cuda_wrapper.c
```diff
@@ -1,36 +1,30 @@
...
 typedef size_t devptr_t;
 
-#if defined(HYPRE_USING_CUDA)
+#if defined(HYPRE_EXAMPLE_USING_CUDA)
+
+#include <cuda_runtime.h>
 
-#define CUDA_MALLOC_MANAGED  device_malloc_managed_
-#define CUDA_FREE            device_free_
+#define CUDA_MALLOC_MANAGED  device_malloc_managed
```
rfalgout: Presumably this line will change depending on the Fortran compiler used.  Users will just have to be aware of that.

rfalgout:
This looks good @liruipeng !  When you are ready, maybe you should invite our two Fortran user reviewers to comment again.  Thanks!

src/examples/Makefile
liruipeng: Not sure either. At least, it should be part of GNU make.

src/examples/cuda_wrapper.c
liruipeng: Yes. It's the `xl` Fortran name mangling (without an underscore after names) . Should we add comments in this file? Will that be enough?

rfalgout: Adding a brief comment in from of these lines is probably a good idea, just to let users know about the name mangling being compiler dependent.

src/examples/Makefile
rfalgout: I think it's fine.  If we have an issue in the future, we can revisit then, but that seems unlikely.

liruipeng:
@flow-phys @p-costa Please take a look at the changes we made to the Fortran examples, and let us know if you have suggestions. We now have 3 approaches to deal with Fortran + hypre (see the description of this PR). Thanks!

p-costa:
Thank you, @liruipeng! The examples are very clear. It's really great to see that the changes required on the user end for GPU offloading seem minimal.

I tested successfully the `*_cptr.f` and `*.cuf` examples on my workstation. Some minor points:

* In my case I had to change the line `#if 0` to `#if 1` in `cuda_wrapper.c` for the `cptr` case to compile successfully.

* Just a nit for your consideration, though it does not matter too much: line 84 of the file could be changed to `integer, external :: device_malloc_managed` to indicate that it is an external function.

* An alternative way to write the interface with the `CUDA C` API (or directly with the C functions `device_malloc_managed` and `device_free`), would be to write a Fortran module file (say `cuda_wrapper.f90`) using `BIND(C)` to interoperate with the C API (see one example [here](https://github.com/starkiller-astro/XNet/blob/master/source/cudaf.f90)), and import it in the Fortran example as, e.g., `use mod_cuda_wrapper, only: device_malloc_managed, device_free`. That would be a more "Fortranic" way of achieving the same, with some advantages in terms of clarity.

* Finally, I have a last question, if you don't mind. I guess that for some of the methods UVM is not required? I wonder how examples 5/12 would look like in `CUDA Fortran` without UVM, using explicitly host and device variables.

Thanks again!

liruipeng:
> Thank you, @liruipeng! The examples are very clear. It's really great to see that the changes required on the user end for GPU offloading seem minimal.
> 
> I tested successfully the `*_cptr.f` and `*.cuf` examples on my workstation. Some minor points:
> 
> * In my case I had to change the line `#if 0` to `#if 1` in `cuda_wrapper.c` for the `cptr` case to compile successfully.
> * Just a nit for your consideration, though it does not matter too much: line 84 of the file could be changed to `integer, external :: device_malloc_managed` to indicate that it is an external function.
> * An alternative way to write the interface with the `CUDA C` API (or directly with the C functions `device_malloc_managed` and `device_free`), would be to write a Fortran module file (say `cuda_wrapper.f90`) using `BIND(C)` to interoperate with the C API (see one example [here](https://github.com/starkiller-astro/XNet/blob/master/source/cudaf.f90)), and import it in the Fortran example as, e.g., `use mod_cuda_wrapper, only: device_malloc_managed, device_free`. That would be a more "Fortranic" way of achieving the same, with some advantages in terms of clarity.
> * Finally, I have a last question, if you don't mind. I guess that for some of the methods UVM is not required? I wonder how examples 5/12 would look like in `CUDA Fortran` without UVM, using explicitly host and device variables.
> 
> Thanks again!

@p-costa Thank you for the nice advice. I really liked the "interface with CUDA C" idea you pointed out, which has been implemented and pushed. With that, the C-wrapper is no longer needed. Please take another look. (@rfalgout please do as well).

To answer your question: For all the examples, UVM is not required **by the hypre library**. That is, you can pass in device pointers just as the UVM pointers. The reason of using UVM here is just for simplicity of the example, where we reuse the same CPU code to populate the arrays, for examples, interfacing with the library. We probably will have another simple example that uses device-only memory in the future. Thanks again!

p-costa:
Thank you, @liruipeng ! It is all clear.

I think the examples look great, very clear. One issue with the current example is that `8_8`, `4_8`, may not be very portable, because I believe it implies that the constant is of kind `integer(8)`, whose definition may vary between compilers. It should be more portable to write, `int(8,c_size_t)`.

In the new Fortran module, you can also create wrapper functions `device_malloc_managed` and `device_free` that do not expose the full CUDA API to the user, as you had in the previous wrapper. I attach the corresponding `Makefile`, `ex5f_cptr.f` and `cudaf.f` that do exactly that, so that you can see the differences with the current approach:
[files.zip](https://github.com/hypre-space/hypre/files/7417689/files.zip). I'm not sure which approach you prefer; one advantage of this one is that the integer inputs and outputs are now Fortran integer kinds. If you have any questions, please let me know!

As a side note of general advice, for your consideration perhaps for a later revision of the examples: you could consider moving from Fixed to [Free Source Format](http://www.mrao.cam.ac.uk/~pa/f90Notes/HTMLNotesnode44.html), which is what is typically used in new code.

Thanks again!

liruipeng:
> I think the examples look great, very clear. One issue with the current example is that `8_8`, `4_8`, may not be very portable, because I believe it implies that the constant is of kind `integer(8)`, whose definition may vary between compilers. It should be more portable to write, `int(8,c_size_t)`.
> 
> In the new Fortran module, you can also create wrapper functions `device_malloc_managed` and `device_free` that do not expose the full CUDA API to the user, as you had in the previous wrapper. I attach the corresponding `Makefile`, `ex5f_cptr.f` and `cudaf.f` that do exactly that, so that you can see the differences with the current approach: [files.zip](https://github.com/hypre-space/hypre/files/7417689/files.zip). I'm not sure which approach you prefer; one advantage of this one is that the integer inputs and outputs are now Fortran integer kinds. If you have any questions, please let me know!
> 

Thank you @p-costa for the suggestions. Your changes have been applied.

oseikuffuor1 (approved):
Thanks @liruipeng!!

liruipeng:
Retested today with `master` pulled in. All looked good but weird errors on ray.  Merge now and will rerun on ray at some point. Thanks to all again!

PR #473, opened 2021-09-03
cmake: REQUIRED packages HYPREConfig.cmake

junghans:
When a package uses a mpi-enabled hypre, mpi is actually required, same for lapack, blas and OpenMP.

@osborn9

### DIFF
--- a/src/config/HYPREConfig.cmake.in
+++ b/src/config/HYPREConfig.cmake.in
@@ -39,11 +39,11 @@ set(HYPRE_ENABLE_GPU_PROFILING @HYPRE_ENABLE_GPU_PROFILING@)
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}")
 
 if(NOT HYPRE_ENABLE_HYPRE_BLAS)
-  find_dependency(BLAS)
+  find_dependency(BLAS REQUIRED)
 endif()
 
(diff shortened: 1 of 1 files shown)


cedricchevalier19:
Just an user comment on this: I think it's good to have the `REQUIRED` concept in the client but for packaging, with Spack for example, it should be better to have the `REQUIRED` keyword when doing the actual `find_package` in Hypre's own `CMakeLists.txt`. `find_dependency` should properly forward the `REQUIRED` and `QUIET` kewords (https://cmake.org/cmake/help/latest/module/CMakeFindDependencyMacro.html).
The advantage of this is if user asks for Hypre with `OpenMP` and `OpenMP` is not found, `cmake` will fail. For automatic installations it is a lot easier to fix fail at configure time than try to figure out why some features are missing later.

osborn9:
Ok, so based on @cedricchevalier19 comment, since we are using the `REQUIRED` keyword within the `find_package` calls then these keywords should be correctly forwarded. 

@junghans Are there instances when the `find_dependency` isn't behaving as expected?

junghans:
It is mainly good practice, but you could construct a case where you e.g. compile a consumer code with a different compiler that doesn't support OpenMP and then `hypre` is looking for OpenMP and doesn't find. Throwing an error in these cases seem better.

PR #476, opened 2021-09-14
Fix for hip compilation

waynemitchell:
This is a small fix for a typo that was preventing me from compiling with hip.


liruipeng (approved):

liruipeng:
Thank you @waynemitchell  for fixing the bug, which was introduced by me.

Issue #475, opened 2021-09-09
Spock/Hip compilation help


PaulMullowney:
I tried adding -fPIC to my CXXFLAGS and CFLAGS and also to LDFLAGS but no luck.

Building libHYPRE_utilities.so ...
/opt/cray/pe/craype/2.7.8/bin/CC -shared  -o libHYPRE_utilities-2.22.1.so F90_HYPRE_error.o F90_HYPRE_general.o HYPRE_handle.o HYPRE_version.o amg_linklist.o binsearch.o exchange_data.o fortran_matrix.o ap.o log.o complex.o error.o hopscotch_hash.o merge_sort.o mpi_comm_f2c.o prefix_sum.o printf.o qsort.o utilities.o mpistubs.o qsplit.o random.o threading.o timer.o timing.o cuda_utils.obj general.obj handle.obj memory.obj omp_device.obj nvtx.obj -Wl,-soname,libHYPRE_utilities-2.22.1.so
ld.lld: error: can't create dynamic relocation R_X86_64_32 against symbol: __device_stub__hypreCUDAKernel_CompileFlagSafetyCheck(int*) in readonly segment; recompile object files with -fPIC or pass '-Wl,-z,notext' to allow text relocations in the output
>>> defined in cuda_utils.obj
>>> referenced by cuda_utils.c
>>>               cuda_utils.obj:(__device_stub__hypreCUDAKernel_CompileFlagSafetyCheck(int*))

PaulMullowney:
Doing:
export HIPCXXFLAGS="-fPIC"
seems to work

pbauman:
HI @PaulMullowney It looks like I haven't tried to build HIP in shared lib mode in awhile. I'll test it and submit a patch. As you've deduced, `HIPCXXFLAGS` is the configure variable you want to tickle to pass flags to HIP compilation. Sorry for the trouble.

liruipeng:
> HI @PaulMullowney It looks like I haven't tried to build HIP in shared lib mode in awhile. I'll test it and submit a patch. As you've deduced, `HIPCXXFLAGS` is the configure variable you want to tickle to pass flags to HIP compilation. Sorry for the trouble.

Hi @pbauman, Thank you for fixing this problem. Shall we let HIP build use CXXFLAGS as well? We recently refactored autoconf build for CUDA for PETSc since they like to control CXXFLAGS.

pbauman:
[...]

PR #477, opened 2021-09-15
Fix shared-libraries build for HIP

pbauman:
We're not sucking in CXXFLAGS because if the user enables --with-openmp,
this mucks up the compilation (with or without shared libs) because
the HIP compiler is clang under-the-hood and the clang openmp host
compilation pass does not play nice with the GCC openmp compilation
pass at link time. So instead we just manually enable the shared lib
flags for the HIP compiler.


pbauman:
This should close #475. I tested HIP compilation with (host) openmp and MPI (and shared libs) on my local workstation (RVII) as well as on Redwood. I tested CUDA compilation with CUDA 11.2 with MPI and openmp on Redwood.

PR #485, opened 2021-09-24
Restore versioncheck tests with fix

rfalgout:
This restores the versioncheck regression tests for the `HYPRE_DEVELOP_*` header file variables and fixes the case when no `.git` directory is present.


victorapm (approved):
Thanks, Rob!

liruipeng (approved):
Looks good. Thanks!

PR #486, opened 2021-09-24
Added more filtering to check-license.sh autotest script

rfalgout:
This modifies `check-license.sh` to ignore `HYPRE_config.h` and `.obj` files.


victorapm (approved):
Thanks, Rob!

PR #487, opened 2021-09-27
Remove '..' directory dependency in test Makefile

rfalgout:
This PR fixes a couple of issues that came up in the spack regression test:

- The `test/Makefile` had the '..' directory in the include path, which caused the `HYPRE_config.h` file to be included from two different places (in `..` and in `install`).  In the spack test, this caused a conflict.
- The `HYPRE_DEVELOP` macros should not be set if `git describe` fails.  This can happen even if the `.git` directory is present.

Note: The `versioncheck.sh` tests will not work in the spack regression tests, because there is a mismatch:
- the spack installation uses a shallow git repository (`--depth 1`) that breaks the `git describe` command (this is okay in general)
- the `versioncheck` tests use the current git repository, which is normally complete


rfalgout:
Hi @osborn9 .  The spack test seems to run cleanly with these changes, since we don't run the `versioncheck` tests currently.  This also runs fine inside a shallow clone of hypre, where `git describe` doesn't always work (depending on how shallow).

osborn9 (approved):
This looks good and the spack test runs clean on my tux box as well. Thanks @rfalgout!

PR #490, opened 2021-09-28
Multipass gpu

liruipeng:
This PR (copied from #489) adds GPU support for multipass interpolations.

> This is a pretty monstrous PR. In a nutshell, this PR takes @ulrikeyang's work on a CPU implementation of the multipass algorithm using SpGEMM and ports it to the GPU. Ulrike's original commits are at the base of the this PR. I tried to make it so that every commit in this history will build and run correctly. (This is how made sure my results were correct as I went through the porting.)
> 
> I did not try and harden the CPU code path. Rather, what I have here is for the "old" multipass implementation (`agg_interp_type =4`), I copy data from the device to the host, run the existing code on the host, and then copy back to the device. (This is in commit 6de667d). This allowed for direct comparison to the GPU porting (`agg_interp_type=9`). On HIP/AMD GPUs, I get identical preconditioners between the two, including in parallel (2,3,4 processes) and with `agg_num_levels <=2`. On NVIDIA GPUs, I don't get identical preconditioners, the number of nonzeros on some of the levels is not identical between the old and the new, but the convergence of the CG iteration seems unaltered. (FWIW, I traced this back to using cuSPARSE SpGEMM on the Pi matrices, on both CUDA 10.2 and CUDA 11.2. This with the V100s on redwood.)
> 
> @liruipeng  This includes your cleanup/algorithm modification commits you'd put on at the end of the old branch.
> 
> Ultimately, this gives a pretty good speedup over the CPU code path.
> 
> The incantation I used for testing on V100 and MI100, varying number of ranks, `-agg_nl`, and `-agg_interp`, is
> 
> <pre>
> mpiexec -np 4 ./ij -n 248 248 248 -pmis -keepT 1 -rlx 18 -exec_device -rap 1 -mod_rap2 1 -interptype 6 -solver 1  -printallstats -27pt -mxrs 0.9 -ns 2 -Pmx 8 -agg_interp 9 -agg_nl 2
> </pre>
> 
> I have *not* tried to update to the lasest master. I think at the very least the `CF_marker`/`hypreInt_Array` work will affect this PR (I assume `CF_marker` comes into this function on the host), but I'm sure there may be other conflicts.

- [x] lassen
- [x] ray
- [x] tux
- [x] redwood


rfalgout (approved):
This looks good to me, but @ulrikeyang and @liruipeng can judge better than me.  Thanks @pbauman !

src/parcsr_ls/par_mod_multi_interp.c
```diff
@@ -0,0 +1,1305 @@
...
+      for (i = 0; i < num_sends; i++)
+      {
+         startc = hypre_ParCSRCommPkgSendMapStart(comm_pkg, i);
+         for (j = startc; j < hypre_ParCSRCommPkgSendMapStart(comm_pkg, i+1); j++)
+         {
+            int_buf_data[index++] = pass_marker[hypre_ParCSRCommPkgSendMapElmt(comm_pkg,j)];
+         }
+      }
+
+      comm_handle = hypre_ParCSRCommHandleCreate( 11, comm_pkg, int_buf_data, pass_marker_offd);
```
liruipeng: @ulrikeyang This looks strange: `hypre_ParCSRCommHandleCreate` inside the if-statement. A process can have zero `num_cols_offd_A` but it may still need to participate in the communication. Am I right on this?

ulrikeyang: You are right. This should be changed to 
if (num_procs > 1)
to be safe.
Thanks for finding this.

ulrikeyang: Will pass_marker_offd set to NULL if num_cols_offd_A is 0? Or do we need to add extra precautions for this special case of num_Cols_offd_A = 0 and still needing to send data.be sent?

src/parcsr_ls/par_mod_multi_interp_device.c
```diff
@@ -0,0 +1,2162 @@
...
+      }
+
+      comm_handle = hypre_ParCSRCommHandleCreate( 11, comm_pkg, int_buf_data, dof_func_offd);
+
+      hypre_ParCSRCommHandleDestroy(comm_handle);
+
+      hypre_TFree(int_buf_data, HYPRE_MEMORY_HOST);
+   }
+
+   if (num_cols_offd_A)
```
ulrikeyang: I guess we need to fix this as well for the reasons mentioned above.

ulrikeyang (changes requested):
This is fine after the comments have been addressed

src/parcsr_ls/par_mod_multi_interp.c
liruipeng: @ulrikeyang If you were talking about this line
```
      pass_marker_offd = hypre_CTAlloc(HYPRE_Int, num_cols_offd_A, HYPRE_MEMORY_HOST);
```
yes, `pass_marker_offd` is `NULL` when `num_cols_offd_A==0`. I think after the change it will be fine even when `num_cols_offd_A==0`. More thoughts? Thanks!

src/parcsr_ls/par_mod_multi_interp.c
```diff
@@ -0,0 +1,1307 @@
...
+      for (i=0; i < n_fine; i++)
+      {
+         if (pass_marker[i] == color)
+         {
+            big_convert[i] = (HYPRE_BigInt)fine_to_coarse[i] + c_pts_starts[0];
+         }
+      }
+
+      num_cols_offd_P = 0;
+      if (num_cols_offd_A)
```
ulrikeyang: It looks like this should also be 
if (num_procs > 1)

src/parcsr_ls/par_mod_multi_interp.c
```diff
@@ -0,0 +1,1307 @@
...
+      {
+         if (pass_marker[i] == color)
+         {
+            big_convert[i] = (HYPRE_BigInt)fine_to_coarse[i] + c_pts_starts[0];
+         }
+      }
+
+      num_cols_offd_Q = 0;
+      if (num_cols_offd_A)
+      {
```
ulrikeyang, line 967: same here. I was really consistent :(

src/parcsr_ls/par_mod_multi_interp_device.c
```diff
@@ -0,0 +1,2156 @@
...
+      if (num_passes > 9)
+      {
+         hypre_error_w_msg(HYPRE_ERROR_GENERIC," Warning!!! too many passes! out of range!\n");
+         break;
+      }
+
+      pass_starts[num_passes] = cnt;
+
+      /* update pass_marker_offd */
+      if (num_cols_offd_A)
```
ulrikeyang: if (num_procs > 1)

src/parcsr_ls/par_mod_multi_interp_device.c
```diff
@@ -0,0 +1,2156 @@
...
+   {
+      f_pts_starts[0] = 0;
+      f_pts_starts[1] = num_points;
+      total_global_fpts = f_pts_starts[1];
+      total_global_cpts = c_pts_starts[1];
+   }
+
+   num_cols_offd_P = 0;
+
+   if (num_cols_offd_A)
```
ulrikeyang: if (num_procs > 1)

src/parcsr_ls/par_mod_multi_interp_device.c
```diff
@@ -0,0 +1,2156 @@
...
+   {
+      f_pts_starts[0] = 0;
+      f_pts_starts[1] = num_points;
+      total_global_fpts = f_pts_starts[1];
+      total_global_cpts = c_pts_starts[1];
+   }
+
+   num_cols_offd_Q = 0;
+
+   if (num_cols_offd_A)
```
ulrikeyang: if (num_procs > 1)

ulrikeyang (changes requested):
It looks like there were more instances of 
if (num_cols_offd_A) that need to be changed.
Once that has been done, it looks good to me.
Sorry I ddin't catch those earlier.

src/parcsr_ls/par_mod_multi_interp.c
```diff
@@ -0,0 +1,1307 @@
...
+      num_passes++;
+      if (num_passes > 9)
+      {
+         hypre_error_w_msg(HYPRE_ERROR_GENERIC," Warning!!! too many passes! out of range!\n");
+         break;
+      }
+      pass_starts[num_passes] = cnt;
+      /* update pass_marker_offd */
+      index = 0;
+      if (num_cols_offd_A)
```
liruipeng: @ulrikeyang Another one? fixed...

ulrikeyang (approved):
It looks like you got them all. Thank you

pbauman (approved):
All the changes you made look good to me, thanks! Glad to see it's in testing as well!

liruipeng:
Hi, All: everything looks good now. It is good to go on my end.

liruipeng:
@rfalgout @ulrikeyang @pbauman Any more comments, suggestions, questions before merging this PR?

rfalgout:
> @rfalgout @ulrikeyang @pbauman Any more comments, suggestions, questions before merging this PR?

None from me.  Thanks!

ulrikeyang:
This looks fine to me. Running quite well on Spock also 😊

liruipeng:
@ulrikeyang Thanks! Curious to see some performance numbers.

pbauman:
Nothing from me. I pulled this branch and the case I"m interested in ran in parallel without issue (and actually previous PRs work on MPI efficiency provided a noticeable speed up!). Thanks for cleaning it up and getting it ready @liruipeng!

PR #491, opened 2021-09-29
Fix memory leak for dof func

waynemitchell:
This fixes a memory leak of the DofFunc for BoomerAMG. I've also removed some code that has been commented out for a while and seems safe to remove.


rfalgout (approved):
Looks good to me.  Thanks @waynemitchell !

ulrikeyang (commented):
Looks good. Thank you!

PR #493, opened 2021-09-30
Add to protos and rerun headers

waynemitchell:
Some function declarations had been removed from the protos file in struct_ls and ./headers needed to be rerun in that directory.

### DIFF
--- a/src/struct_ls/_hypre_struct_ls.h
+++ b/src/struct_ls/_hypre_struct_ls.h
@@ -450,9 +450,9 @@ HYPRE_Int hypre_SparseMSGSetupRAPOp ( hypre_StructMatrix *R , hypre_StructMatrix
 /* sparse_msg_solve.c */
 HYPRE_Int hypre_SparseMSGSolve ( void *smsg_vdata , hypre_StructMatrix *A , hypre_StructVector *b , hypre_StructVector *x );
 
-
 #ifdef __cplusplus
 }
 #endif
 
 #endif
+
--- a/src/struct_ls/protos.h
+++ b/src/struct_ls/protos.h
@@ -5,6 +5,11 @@
  * SPDX-License-Identifier: (Apache-2.0 OR MIT)
  ******************************************************************************/
 
+/* coarsen.c */
+HYPRE_Int hypre_StructMapFineToCoarse ( hypre_Index findex , hypre_Index index , hypre_Index stride , hypre_Index cindex );
+HYPRE_Int hypre_StructMapCoarseToFine ( hypre_Index cindex , hypre_Index index , hypre_Index stride , hypre_Index findex );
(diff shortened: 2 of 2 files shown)


waynemitchell:
@victorapm It looks like the removal of the function declarations from protos happened in one of your PRs: 

https://github.com/hypre-space/hypre/pull/392

Did you intend to get rid of these? It seems that the only reason the code continue to compile correctly is that ./headers did not get rerun after the removal. But maybe I'm missing something here?

liruipeng:
I think we should have a way of automatically running `./headers` in various directories at compile. Or regression test this.

rfalgout:
I will put this on the list of things to add to regression testing and/or new github action testing.

victorapm:
Sorry all, this was my mistake. Those functions actually belong to `struct_mv`, and the moving has been done in the `recmat` branch but not in `master`. 

I agree it would be nice to run `headers` automatically

rfalgout (approved):

victorapm:
@rfalgout, should I move `coarsen.c`and its functions prototypes from `struct_ls` to `struct_mv` in this PR as we currently have in `recmat`?

rfalgout:
@victorapm.  If you move the prototypes, let's merge this PR after the release (not before).  I don't think the placement of these is super critical here, but if it's easier to keep track of things for you, then moving it makes sense.  Thanks!

waynemitchell:
> @victorapm. If you move the prototypes, let's merge this PR after the release (not before). I don't think the placement of these is super critical here, but if it's easier to keep track of things for you, then moving it makes sense. Thanks!

@rfalgout Is there a reason you don't want to move the coarsen.c file and corresponding prototypes from struct_ls to struct_mv before the release? Does that make things more complicated somehow? Seems straightforward to me, and I'm happy to go ahead and move those over. But will hold off if you prefer.

victorapm:
Ok! I agree it's not that critical so I will move those functions in the future. Thanks!

victorapm (approved):

rfalgout:
> @rfalgout Is there a reason you don't want to move the coarsen.c file and corresponding prototypes from struct_ls to struct_mv before the release? Does that make things more complicated somehow? Seems straightforward to me, and I'm happy to go ahead and move those over. But will hold off if you prefer.

This isn't necessary to compile the code, right?  Since we want to do the release tomorrow or Monday, I don't think we should merge anything in that isn't necessary, because it just increases the likelihood of something unforeseen breaking.  I realize this is trivial, but we've had "trivial" changes break things many times in the past, so I'd prefer to have a buffer.

rfalgout:
Just to be clear on my last comment, if we don't have to merge this PR prior to the release, we can go ahead and move the definitions of these functions to `struct_mv` and merge after the release is done.  I'll leave that up to you and @victorapm.

waynemitchell:
Sounds good. Thanks @rfalgout and @victorapm for the clarifications and input. I'll go ahead and merge this as is (I think it would be good to have this corrected before the release) and will leave the movement of the coarsen.c routines for the future.

PR #495, opened 2021-10-04
Fix memory leak

liruipeng:
This PR fixes memory leaks at various places (found by hypre memory tracker). Most of them are because of the previous changes made to `hypre_ParCSRMatrixRow/ColStarts`. With the old code, the `Row/ColStarts` arrays may or may not be freed depending on if they are owned or not by `hypre_ParCSRMatrix`, but with the new code they should be always freed. So, the leak happened. To avoid its happening again, all dynamic allocations such as `row_starts = hypre_CTAlloc(HYPRE_BigInt, 2, HYPRE_MEMORY_HOST)` were changed to `HYPRE_BigInt row_starts[2]`. It also simplifies the code and is potentially more efficient.

- [x] tux
- [x] ray
- [x] lassen
- [ ] redwood


src/parcsr_ls/par_coarse_parms.c
```diff
@@ -46,36 +35,34 @@
...
 #endif
 
-   HYPRE_Int      i;
-   HYPRE_Int      num_procs;
-   HYPRE_BigInt   local_coarse_size = 0;
-
-   HYPRE_Int     *coarse_dof_func;
-   HYPRE_BigInt  *coarse_pnts_global;
+   HYPRE_Int     i;
+   //HYPRE_Int     num_procs;
```
victorapm: Should this be deleted?

src/parcsr_ls/par_coarse_parms.c
```diff
@@ -46,36 +35,34 @@
...
+   HYPRE_BigInt  local_coarse_size = 0;
+   HYPRE_Int    *coarse_dof_func;
 
    /*--------------------------------------------------------------
     *----------------------------------------------------------------*/
 
-   hypre_MPI_Comm_size(comm,&num_procs);
+   //hypre_MPI_Comm_size(comm, &num_procs);
```
victorapm: Should this be deleted?

src/parcsr_ls/par_coarse_parms_device.c
```diff
@@ -61,23 +58,20 @@ hypre_BoomerAMGCoarseParmsDevice(MPI_Comm          comm,
...
-                                 HYPRE_BigInt    **coarse_pnts_global_ptr) 
+                                 HYPRE_BigInt     *coarse_pnts_global)
 {
 #ifdef HYPRE_PROFILE
    hypre_profile_times[HYPRE_TIMER_ID_COARSE_PARAMS] -= hypre_MPI_Wtime();
 #endif
 
    HYPRE_Int      ierr = 0;
-   HYPRE_Int	   num_procs;
+   //HYPRE_Int      num_procs;
```
victorapm: Delete this?

src/parcsr_ls/par_coarse_parms_device.c
```diff
@@ -61,23 +58,20 @@ hypre_BoomerAMGCoarseParmsDevice(MPI_Comm          comm,
...
    HYPRE_BigInt   local_coarse_size = 0;
 
-   HYPRE_BigInt  *coarse_pnts_global;
-
-
    /*--------------------------------------------------------------
     *----------------------------------------------------------------*/
 
-   hypre_MPI_Comm_size(comm,&num_procs);
+   //hypre_MPI_Comm_size(comm, &num_procs);
```
victorapm: Delete this?

victorapm (approved):
Thanks for catching all these additional cases that I missed, @liruipeng. This simplifies a lot the code. I am curious to know why our valgrind tests didn't point out the memory leaks that you are fixing. Also, how can I use the memory tracking feature that you mentioned?

Thanks!

src/parcsr_ls/par_coarse_parms.c
```diff
@@ -46,44 +35,40 @@
...
-                               hypre_IntArray   *CF_marker,
-                               hypre_IntArray  **coarse_dof_func_ptr,
-                               HYPRE_BigInt    **coarse_pnts_global_ptr) 
+hypre_BoomerAMGCoarseParmsHost(MPI_Comm         comm,
+                               HYPRE_Int        local_num_variables,
+                               HYPRE_Int        num_functions,
+                               hypre_IntArray  *dof_func,
+                               hypre_IntArray  *CF_marker,
+                               hypre_IntArray **coarse_dof_func_ptr,
+                               HYPRE_BigInt    *coarse_pnts_global)
```
rfalgout, line 51: Since this technically changes the API, albeit the internal one, we should be extra careful with regard to breaking MFEM and PETSc here.

victorapm: Yes! I see this function is used by MFEM but not by PETSc.

rfalgout: Thanks @victorapm !  Note that there are other API changes; I only pointed out the one.  We should do a more extensive test with the MFEM/PETSc codes.

rfalgout:
@liruipeng .  I also want to know what "hypre memory tracker" is.  Thanks!

rfalgout (approved):
This looks good.  Thanks!

src/parcsr_ls/par_coarse_parms.c
liruipeng: Good point. In mfem, I see `HYPRE_BigInt *cpts_global; hypre_BoomerAMGCoarseParms...  mfem_hypre_TFree(cpts_global);` The fix is simple:  `HYPRE_BigInt cpts_global[2]`. I am not sure how to deal with the version compatibility between hypre and mfem. Maybe @victorapm can help.

victorapm: Yes, I can prepare a PR for MFEM after we merge this one. The other API changes are for `hypre_BoomerAMGCoarseParmsHost` and `hypre_BoomerAMGCoarseParmsDevice` which aren't used directly by MFEM, but I can run both PETSc and MFEM tests to double check.

liruipeng:
@rfalgout @victorapm The memory tracker is some tool I wrote to track (GPU) memory allocation since mem-checker didn't give all the information I wanted. Nothing fancy. It just records all memory transactions inside hypre (alloc, free, copy on CPU and GPU) and prints a report at the end. See attached file. It reports a "leak" if a pointer is allocated but a corresponding free is not found. It's still experimental and probably not very "scalable", but helpful for me to find several leak problem on GPUs.

To use it, `./configure --with-memory-tracker`
 
[HypreMemoryTrack.log.00000.txt](https://github.com/hypre-space/hypre/files/7280133/HypreMemoryTrack.log.00000.txt)

src/parcsr_ls/par_coarse_parms.c
rfalgout: Thanks for offering to help with this @victorapm!  I would recommend connecting with Tzanio before we merge.  He may prefer to start a PR based on this branch, for example.

victorapm: Ok, @rfalgout! Will do that. Thanks!

victorapm:
Sounds cool, @liruipeng! Thanks for sharing this. I guess it doesn't have the same overhead as valgrind's memcheck, so this is very useful. Thanks!

waynemitchell:
@liruipeng Looks good to me, pending any compatibility issues with MFEM as pointed out above. Thanks!

waynemitchell (approved):

liruipeng:
> Sounds cool, @liruipeng! Thanks for sharing this. I guess it doesn't have the same overhead as valgrind's memcheck, so this is very useful. Thanks!

Well it's much less powerful though. The reason I needed this is (for GPU) was `cuda-memcheck` only tells there's a leak of certain bytes but does not say where it is from.

ulrikeyang (approved):
This looks fine

PR #498, opened 2021-10-05
Adding source file indentation with astyle

rfalgout:
This PR adds automatic indentation using Artistic Style (`astyle`):

  http://astyle.sourceforge.net/
  https://sourceforge.net/projects/astyle/files/astyle/astyle%203.1/
  http://astyle.sourceforge.net/install.html

If you want to use it, you can download via the second link above and follow the third link for install instructions.  Just make sure `astyle` ends up somewhere in your path.  The following script and option file determine the indentation strategy for hypre:

  https://github.com/hypre-space/hypre/blob/style/src/config/astyle-apply.sh
  https://github.com/hypre-space/hypre/blob/style/src/config/astylerc

A few notes:

- To run the script from the `src` directory, type `config/astyle-apply.sh .`
- The script will also run `headers` in all of the directories that automatically generate internal `_hypre_*.h` header files
- Much of this was borrowed from the MFEM project
- The options file is the same as the one in MFEM (which looks like it was based on the indentation style in hypre to begin with), except that the line length was increased to 100

Lots of files have changed, mostly in minor ways, but I have already done some manual reformatting in the `examples` directory to improve the readability of the indentation.  Please note any other files that need additional work.  Also note any other issues you have with the indenting.

To do:

- [x] Add a hook for git that can be used to indent on each commit
- [x] Write a Wiki on coding style for hypre


rfalgout:
FYI, it is useful to view the files with "hide whitespace changes" turned on.  This helps to identify where new lines/code have been introduced by `astyle` due to brace usage or lines that are longer than 100 characters.  These are the most likely scenarios where some manual editing could be a good idea.  You first might want to look at a few files without hiding whitespace changes, just to get a sense of what is changing overall.

victorapm:
This is a very nice idea, Rob! I will comment a few questions on the changes added here. Thank you for putting this together.

src/IJ_mv/HYPRE_IJVector.c
```diff
@@ -218,7 +218,7 @@ HYPRE_IJVectorSetValues( HYPRE_IJVector        vector,
 {
    hypre_IJVector *vec = (hypre_IJVector *) vector;
 
-   if (nvalues == 0) return hypre_error_flag;
+   if (nvalues == 0) { return hypre_error_flag; }
```
victorapm, line 221: Should the brackets be added on new lines?

src/IJ_mv/IJMatrix_parcsr.c
```diff
@@ -363,7 +368,7 @@ HYPRE_Int hypre_IJMatrixGetRowCountsParCSR( hypre_IJMatrix *matrix,
    hypre_MPI_Comm_rank(comm,&my_id);
 
 #ifdef HYPRE_USING_OPENMP
-#pragma omp parallel for private(i, row_index) HYPRE_SMP_SCHEDULE
+   #pragma omp parallel for private(i, row_index) HYPRE_SMP_SCHEDULE
```
victorapm, line 371: Do we want extra spaces here?

src/IJ_mv/IJMatrix_parcsr.c
```diff
@@ -2190,7 +2205,7 @@ hypre_IJMatrixAssembleOffProcValsParCSR( hypre_IJMatrix       *matrix,
 
    num_recvs = send_proc_obj.length;
    argsort_contact_procs = hypre_CTAlloc(HYPRE_Int,  num_recvs, HYPRE_MEMORY_HOST);
-   for(i=0; i < num_recvs; i++)
+   for (i=0; i < num_recvs; i++)
```
victorapm: Should we have spaces before and after the equal sign? (`i = 0`)

src/IJ_mv/IJ_vector.h
```diff
@@ -35,7 +35,7 @@ typedef struct hypre_IJVector_struct
 
    HYPRE_BigInt  global_first_row;  /* these for data items are necessary */
    HYPRE_BigInt  global_num_rows;   /* to be able to avoid using the global */
-                                    /* global partition */
+   /* global partition */
```
victorapm, line 38: Remove this comment?

src/IJ_mv/HYPRE_IJVector.c
rfalgout: I think it's okay to allow one-line if statements in certain cases where it helps readability.  The style file allows for this, but makes sure the parenthesis are there.

src/IJ_mv/IJMatrix_parcsr.c
rfalgout: I'm not sure why astyle indents `#pragma` this way (it doesn't do this for other preprocessor commands).  I looked around online, and some people prefer this approach.  Pre-ANSI standards used to require preprocessor commands to start at the beginning of the line, but that's not common practice today.  So, I think this is okay, and I don't think there is a nice way to change the style file to prevent this without having an impact on other preprocessor commands in a way that we probably wouldn't want.

src/IJ_mv/IJMatrix_parcsr.c
rfalgout: I'm not sure there is an option to enforce spaces before and after.  Let me know if you find one.  In general, I like to have the extra spaces, but it's not that critical I think.

src/IJ_mv/IJ_vector.h
rfalgout: I think there are lots of comments that could be improved, rewritten, etc.  This is one of them.

rfalgout: BTW, I don't think we need to get everything in pristine form in this PR, but addressing the more major formatting issues makes sense.

rfalgout:
Another note for everyone.  I've added an agenda item to our meeting tomorrow to discuss this PR.  The most important thing here is that we will have an enforced standard moving forward, which is a good thing for many reasons.  Part of the purpose of the PR is to settle on something that everyone is okay with.  The current style file is based on the one for MFEM which was based on the original style for hypre (from what I can tell).  The main thing that is different here is that I increased the line length from 80 to 100, and even that resulted in many line breaks.

src/IJ_mv/HYPRE_IJVector.c
victorapm: That makes sense, thanks!

src/IJ_mv/IJMatrix_parcsr.c
victorapm: Interesting, I didn't know that... Ok, I agree with this new approach

src/IJ_mv/IJMatrix_parcsr.c
victorapm: I found the option `--pad-oper / -p` at http://astyle.sourceforge.net/astyle.html for padding spaces around operators. Should we add it?

src/IJ_mv/IJ_vector.h
victorapm: I see! This is in fact out of scope. Thanks!

src/IJ_mv/IJMatrix_parcsr.c
rfalgout: I was going to vote "yes", but it looks like this applies to all operators, so I am on the fence now.  Anybody else have an opinion?  We can also discuss more tomorrow.

liruipeng: I vote yes. Looking at their example
```
if (foo==2)
    a=bar((b-c)*a,d--);
becomes:

if (foo == 2)
    a = bar((b - c) * a, d--);
```
which is exactly what I liked it to be.

Maybe there are other operators we don't like to apply it. Anyone can give an example?

waynemitchell: I vote yes on this.

osborn9: I also vote yes for this.

ulrikeyang: Yes

rfalgout: Looks like the yeses win.  I'll make the change a push a new version of the files soon.  Thanks!

rfalgout: Okay, I've made the change.  I will go through the examples codes to do some additional manual formatting.  If any of you want to take on any other manual indentation work, let the team know ahead of time so we don't step on toes.  Thanks!

rfalgout:
I just added a `pre-commit` git hook here: https://github.com/hypre-space/hypre/tree/style/src/config/githooks.  Another `pre-push` hook will come later in another PR to run a few autotest tests such as checking for license headers, etc.

rfalgout:
Hi All.  I wrote a draft Wiki for coding style at https://github.com/hypre-space/hypre/wiki/Coding-Style.  I have more to do on that, but I think this is enough for finishing up this PR.  I am changing the PR to a non-draft and will update with master soon as well.  Please take a look when you have a chance.  Thanks!

waynemitchell (approved):
All the style choices look good to me. I can't think of any issues with this. Seems very nice to have everything looking consistent. Thanks, Rob!

oseikuffuor1 (approved):
Thanks Rob! This looks great.

liruipeng (approved):
Looks good. Thanks!

rfalgout:
I added some things to the coding style Wiki.  Please take a look and let me know what you think.  Also, I could use some help on the Standards section at the end, especially with regard to GPU code.  Thanks!

liruipeng:
> I added some things to the coding style Wiki. Please take a look and let me know what you think. Also, I could use some help on the Standards section at the end, especially with regard to GPU code. Thanks!

GPU code is C++11 now, which is probably going to C++14.

osborn9 (approved):

src/parcsr_ls/par_ilu.h
```diff
@@ -39,7 +39,8 @@ typedef struct hypre_ParILUData_struct
    hypre_ParCSRMatrix      *P;
    hypre_Vector            *Ftemp_upper;
    hypre_Vector            *Utemp_lower;
-   HYPRE_Int               *A_diag_fake;//fake diagonal, used to pretend that the diagonal matrix is empty
+   HYPRE_Int
+   *A_diag_fake;//fake diagonal, used to pretend that the diagonal matrix is empty
```
ulrikeyang: It appears that having // immediately after the ; is causing this to be broken up strangely. It probably should be manually fixed.

src/seq_mv/vector_batched.c
```diff
@@ -23,107 +23,109 @@ hypre_SeqVectorMassAxpy8( HYPRE_Complex *alpha,
...
+      jstart = (k - 3) * size;
 #if defined(HYPRE_USING_OPENMP)
-#pragma omp parallel for private(i) HYPRE_SMP_SCHEDULE
+      #pragma omp parallel for private(i) HYPRE_SMP_SCHEDULE
 #endif
       for (i = 0; i < size; i++)
       {
-         y_data[i] += alpha[k-3] * x_data[jstart+i] + alpha[k-2] * x_data[jstart+size+i] + alpha[k-1] * x_data[(k-1)*size+i];
+         y_data[i] += alpha[k - 3] * x_data[jstart + i] + alpha[k - 2] * x_data[jstart + size + i] + alpha[k
+                                                                                                           - 1] *
```
ulrikeyang, line 76: another strange split

ulrikeyang (approved):
Overall this looks good. There are a few funny splits, but at least they are consistent.

src/parcsr_ls/par_ilu.h
rfalgout: I fixed this one.  In general, when the line is too long, astyle doesn't seem to make the best decision about how to break it.  I think we'll just have to manually fix these as we come across them.  Thanks!

src/seq_mv/vector_batched.c
rfalgout: There is a fair amount to do in this file (lots of long lines), so I didn't mess with it.

PR #504, opened 2021-10-22
[SYCL] convert sycl::device to sycl::device* for better handling 

abagusetty:


src/utilities/general.c
```diff
@@ -100,7 +100,43 @@ hypre_SetDevice(hypre_int device_id, hypre_Handle *hypre_handle_)
...
+         if (local_nDevices == device_id) {
+           hypre_HandleDevice(hypre_handle_) = &tile;
+         }
+         local_nDevices++;
+       }
+     }
+     // single-tile GPUs
+     else {
+       if (local_nDevices == device_id) {
+         hypre_HandleDevice(hypre_handle_) = &(gpu_devices[i]);
```
waynemitchell, line 124: If you set `hypre_HandleDevice(hypre_handle_)` here, you need to make sure it is not overwritten below (line 143). Also, if you want to use `hyrpe_SetDevice()`, you need to do corresponding implementation for `hypre_GetDevice()`, which still has no implementation for sycl.

src/utilities/general.c
```diff
@@ -100,7 +100,43 @@ hypre_SetDevice(hypre_int device_id, hypre_Handle *hypre_handle_)
...
+   }
+
+   HYPRE_Int local_nDevices=0;
+   for (int i = 0; i < gpu_devices.size(); i++) {
+     // multi-tile GPUs
+     if (gpu_devices[i].get_info<sycl::info::device::partition_max_sub_devices>() > 0) {
+       auto subDevicesDomainNuma = gpu_devices[i].create_sub_devices<sycl::info::partition_property::partition_by_affinity_domain>(sycl::info::partition_affinity_domain::numa);
+       for (auto &tile : subDevicesDomainNuma) {
+         if (local_nDevices == device_id) {
+           hypre_HandleDevice(hypre_handle_) = &tile;
```
waynemitchell, line 116: So does this consider each tile to be a separate device? What is the effect of this vs. setting the environment variable that Brian had mentioned: `export ZE_AFFINITY_MASK=0.0`? Does this do the same thing?

src/utilities/general.c
```diff
@@ -100,7 +100,43 @@ hypre_SetDevice(hypre_int device_id, hypre_Handle *hypre_handle_)
...
+   auto gpu_devices = platform.get_devices(sycl::info::device_type::gpu);
+   for (int i = 0; i < gpu_devices.size(); i++) {
+     if(gpu_devices[i].get_info<sycl::info::device::partition_max_sub_devices>() > 0) {
+       auto subDevicesDomainNuma = gpu_devices[i].create_sub_devices<sycl::info::partition_property::partition_by_affinity_domain>(sycl::info::partition_affinity_domain::numa);
+       nDevices += subDevicesDomainNuma.size();
+     }
+     else {
+       nDevices++;
+     }
+   }
```
waynemitchell: The code above looks like it replicates the new implementation in `hypre_GetDeviceCount()`. Can we just call that here?

src/utilities/general.c
abagusetty: It does the same but just generalizes. `ZE_AFFINITY_MASK=0.0` is designed to be set for device ID 0 and tile 0 and not to be a scalable solution.

src/utilities/general.c
abagusetty: Done

src/utilities/general.c
abagusetty: I agree, `hypre_GetDevice()` method is still undefined for SYCL. I guess `#if defined(HYPRE_USING_GPU) && !defined(HYPRE_USING_SYCL)` fixes the resetting the value.

src/utilities/general.c
waynemitchell: That sounds good. Is there ever a case where you would want to treat all tiles as a single device? Either way, I think I'm happy using this for now at least.

abagusetty: Treating a device as 2 separate tiles (explicit scaling) and device as a whole (implicit scaling) are two options. Since majority of the Apps interacting with Hypre are exploring explicit mode, hence the separate tiles. But I want to add implicit scaling too. I will add macros like the following under `HYPRE_USING_SYCL` for a user to choose either explicit or implicit. 

For e.g., `HYPRE_SYCL_EXPLICIT` and `HYPRE_SYCL_IMPLICIT`

abagusetty:
@waynemitchell Are there any changes for this PR

waynemitchell:
@abagusetty Looks good. I'll go ahead and bring this in.

PR #505, opened 2021-10-22
Add fflush in hypre_printf

liruipeng:
There have been cases where printing from hypre and printing from Fortran are misordered, like
```
    Cycle  5   6.268763e-08    0.064737     2.195967e-06 
    Cycle  6   3.996841e-09    0.063758     1.400105e-07 
 
 Iterations =  7
 Final Relative Residual Norm =   8.83917499E-09
 
    Cycle  7   2.523294e-10    0.063132     8.839175e-09 


 Average Convergence Factor = 0.070711

     Complexity:    grid = 1.719008
                operator = 2.360625
                   cycle = 4.716544
```
It makes the regression tests failed.

Adding `fflush(stdout)` in `hypre_printf` seems to fix this issue, whereas adding `FLUSH(output_unit)` or `FLUSH(6)`  in Fortran code doesn't. I am not sure if there are reasons that we don't want flushing at every `printf`.

### DIFF
--- a/src/utilities/printf.c
+++ b/src/utilities/printf.c
@@ -140,6 +140,8 @@ hypre_printf( const char *format, ...)
    free_format(newformat);
    va_end(ap);
 
+   fflush(stdout);
+
    return ierr;
 }
 


victorapm (approved):

oseikuffuor1 (approved):
Thanks @liruipeng this makes sense. The only other reason I I believe it may not be necessary is if it is printed to stderr instead of stdout, since stderr is not buffered (see [here](https://linux.die.net/man/3/stderr)). vprintf, which is used in the hypre_printf function writes to stdout (see [here](https://linux.die.net/man/3/vfprintf)), so what you are proposing makes sense. If there's any place where doing this seems unnecessary, then we should perhaps be using hypre_fprintf there instead (so we can specify the output stream)

rfalgout:
Out of curiosity, which driver generated the out of order output, and on which machine?  Also, note that there are calls to `fflush` in about 50 different file in hypre.  We should work on removing them if we do end up adding `fflush` to `printf`.  Don't do that yet, I'm still thinking about this. :)  Thanks!

waynemitchell (approved):

liruipeng:
> Out of curiosity, which driver generated the out of order output, and on which machine? Also, note that there are calls to `fflush` in about 50 different file in hypre. We should work on removing them if we do end up adding `fflush` to `printf`. Don't do that yet, I'm still thinking about this. :) Thanks!

@rfalgout this happened to `ex5f.f` and `ex12f.f` with GPU build of hypre and with `lrun` on LC machines. See the following message from John Gyllenhaal 

> This usually happens when different output streams are used (i.e., stdout and stderr) or they are being printed from different ranks.   Can you tell if that is the case here?
>  
> It could also in theory happen if you are mixing printf with cout, but I would expect the same issues with mpirun then.
>  
> It is caused by line buffering and transmitting the output over sockets, which can change relative timing.   Mpirun runs task0 locally, so no output out of order effects can occur, at least for task 0.
>  
> If you can figure out how those two lines are being output, you may be able to add flushes to get the output pattern you want.

rfalgout (approved):
This looks good, Ruipeng.  We should, over time, start to remove the `fflush` commands scattered throughout the library.  This is not even remotely urgent, though.  Thanks!

osborn9 (approved):

ulrikeyang (approved):

PR #510, opened 2021-10-27
Fix memoryLocation bug in parcsrmatrix_read.

oseikuffuor1:
This PR fixes a minor bug in the memory location assignment in parcsrmatrix_read.


liruipeng (approved):
Thanks!

PR #515, opened 2021-11-03
Benchmark IJ

liruipeng:
This PR adds `-pout 0` in all the jobs in `TEST_bench/benchmark_ij.jobs`, which removes the cost of copying AMG hierarchy from GPU to host that is only for printing statistics, to have a more meaningful timing for the benchmarks. The `.saved` results have been updated.


rfalgout (approved):
Looks good.  Thanks!

PR #518, opened 2021-11-08
Link to Caliper's CMake package

ibaned:
if HYPRE_WITH_CALIPER is enabled,
then actually find the CMake package
for caliper and link to it.
Without this, include files are not
found and the caliper library is not
properly linked.


osborn9 (approved):
Thanks @ibaned for this contribution!

PR #519, opened 2021-11-09
Deviceomp

liruipeng:
This PR fixes the build with Kokkos + OMP offload,  supports OMP offload without linking CUDA libraries, and supports OMP offload on Intel GPUs

- [x] tuxbox
- [x] lassen
- [x] ray
- [x] redwood
- [ ] jlse


src/utilities/_hypre_utilities.hpp
```diff
@@ -120,8 +120,16 @@ struct hypre_device_allocator
...
 /* WM: problems with this being inside extern C++ {} */
 /* #include <CL/sycl.hpp> */
+<<<<<<< HEAD
+
+#endif // defined(HYPRE_USING_CUDA)
+
+#if defined(HYPRE_USING_ROCSPARSE)
+#include <rocsparse.h>
+=======
 #if defined(HYPRE_USING_ONEMKLSPARSE)
```
ulrikeyang: looks like this hasn't been merged properly.

liruipeng: good catch. Sorry about that. Fixed.

src/utilities/hopscotch_hash.h
```diff
@@ -47,13 +47,13 @@
 #include <string.h>
 #include <stdio.h>
 #include <limits.h>
-#include <math.h>
+//#include <math.h>
```
ulrikeyang, line 51: I see you are commenting out various .h files. Are they no longer needed?

src/utilities/memory.c
```diff
@@ -62,12 +62,16 @@ hypre_DeviceMemset(void *ptr, HYPRE_Int value, size_t num)
...
+   for (size_t i = 0; i < num; i++)
+   {
+      ((unsigned char *) ptr)[i] = (unsigned char) value;
+   }
 #else
    memset(ptr, value, num);
    HYPRE_OMPOffload(hypre__offload_device_num, ptr, num, "update", "to");
 #endif
-   HYPRE_CUDA_CALL( cudaDeviceSynchronize() );
+   /* HYPRE_CUDA_CALL( cudaDeviceSynchronize() ); */
```
ulrikeyang, line 74: Here also a lot of calls are commented out. Leaving empty if statements in some cases. Are they no longer needed?

liruipeng: Maybe it's better to just delete? We cannot use CUDA routines in device OMP build.

src/utilities/hopscotch_hash.h
liruipeng: Good question. I forgot the reason why I wanted to comment out math.h here. It should compile fine without, otherwise I will add it back.

src/utilities/hopscotch_hash.h
```diff
@@ -47,13 +47,13 @@
...
 
 #ifdef HYPRE_USING_OPENMP
 #include <omp.h>
 #endif
 
-#include "_hypre_utilities.h"
```
liruipeng, line 56: But this looks strange. `_hypre_utilities.h` should contain this file, and this file includes `_hypre_utilities.h` back? Maybe I missed the point here?

src/utilities/memory.c
liruipeng: OK. I remembered something now. I did not find a portable way to deal with UVM (allocation, deallocation, etc) with device openmp that works everywhere. With the current PR as is, it just removes the capability of using UVM with device openmp, which seems not being used anywhere (?)

rfalgout (approved):
This all looks good to me, though I haven't done any testing.

src/utilities/hopscotch_hash.h
liruipeng: @rfalgout do you have comments on this line? Thanks!

rfalgout: It doesn't look necessary and probably should be deleted.  Is that the question you have?

liruipeng: Yes, just to confirm it's OK to delete. Thanks @rfalgout .

ulrikeyang (approved):
I can't say that I understand all this, but it looks fine, and all my questions were answered.

src/seq_mv/csr_matop_device.c
```diff
@@ -1558,7 +1558,7 @@ hypre_CSRMatrixTransposeDevice(hypre_CSRMatrix  *A,
 #elif defined(HYPRE_USING_ROCSPARSE)
       hypreDevice_CSRSpTransRocsparse(nrows_A, ncols_A, nnz_A, A_i, A_j, A_data, &C_i, &C_j, &C_data,
                                       data);
-#else
+#elif defined(HYPRE_USING_CUDA) || defined(HYPRE_USING_HIP)
```
waynemitchell: This should also include `|| defined(HYPRE_USING_SYCL)` since the function below has been ported to sycl (and is indeed the only device matrix transpose option available in that case since there's no oneMKL sparse implementation).

src/utilities/memory.c
```diff
@@ -495,12 +501,14 @@ hypre_Free_core(void *ptr, hypre_MemoryLocation location)
       return;
    }
 
+#if defined(HYPRE_USING_CUDA) || defined(HYPRE_USING_HIP)
```
waynemitchell: `hypre_GerPointerLocation()` is also available with sycl, so you could add `|| defined(HYPRE_USING_SYCL)` here.

waynemitchell (commented):
All looks good except the couple sycl related things I mention below.

src/seq_mv/csr_matop_device.c
liruipeng: Fixed. Thanks!

src/utilities/memory.c
liruipeng: Fixed. Thanks!

src/utilities/device_utils.h
```diff
@@ -227,13 +227,15 @@ struct hypre_DeviceData
    rocsparse_handle                  cusparse_handle;
 #endif
 
-#if defined(HYPRE_USING_CUDA) || defined(HYPRE_USING_DEVICE_OPENMP)
+#if defined(HYPRE_USING_CUDA_STREAMS)
```
waynemitchell, line 230: Adding this macro breaks the sycl build for me. There is a related change in `hypre_DeviceDataStream()` in `device_utils.c` where you've also moved the construction of the sycl queue behind this `HYPRE_USING_CUDA_STREAMS` macro. What does is this macro supposed to turn on/off?

src/utilities/device_utils.c
```diff
@@ -1174,13 +1139,31 @@ hypre_DeviceDataStream(hypre_DeviceData *data, HYPRE_Int i)
...
+         {
+            std::cout << "Caught asynchronous SYCL exception:" << std::endl
+                      << ex.what() << ", SYCL code: " << ex.code() << std::endl;
+         }
+      }
+   };
+
+   sycl::device* sycl_device = data->device;
+   sycl::context sycl_ctxt   = sycl::context(*sycl_device, sycl_asynchandler);
+   stream = new sycl::queue(sycl_ctxt, *sycl_device, sycl::property_list{sycl::property::queue::in_order{}});
```
waynemitchell, line 1166: This construction of the queue object needs to happen for the default queue regardless of whether `HYPRE_USING_CUDA_STREAMS` is set.

src/utilities/device_utils.h
liruipeng: `HYPRE_USING_CUDA_STREAMS` is whether hypre should create a new stream to run or use the default stream.

src/utilities/device_utils.c
liruipeng: Maybe a simple fix is to let `HYPRE_USING_CUDA_STREAMS` defined by default with sycl

waynemitchell: I'm fine with that solution if it makes sense to you. I guess there is no "default stream" in sycl, since you always have to construct a queue object.

waynemitchell:
Did some testing with sycl. All looks OK on my side.

waynemitchell (approved):

PR #530, opened 2021-11-19
fix cuSPARSE version conditional

ibaned:
I was using version 11.2.0 of cuSPARSE
and the constant CUSPARSE_SPMV_CSR_ALG2
was definitely not available.
Looking at the CUDA toolkit documentation,
the earliest release of the CUDA toolkit
that contained a cuSPARSE version that
supports CUSPARSE_SPMV_CSR_ALG2
was CUDA 11.2.1, with cuSPARSE version
11.4.0.135.
After this change, HYPRE only tries to
use CUSPARSE_SPMV_CSR_ALG2 for
cuSPARSE 11.4.0 and newer.


liruipeng (approved):
Looks good! Thank you @ibaned

Issue #531, opened 2021-11-21
Hypre hangs when calculating eigenvalues estimates


bkrasnopolsky:
Hello,

I've faced with hanging the calculations when estimating the min/max eigenvalues. This occurs in ```hypre_ParCSRMaxEigEstimateCGHost```. Specifically, CG provides after several iterations ```r``` vector equal to zero, and elements of ```tridiag``` and ```trioffd``` arrays become ```nan```. This leads to a hang inside the ```hypre_LINPACKcgtql1```. I suppose, after
https://github.com/hypre-space/hypre/blob/da32677ae067d63475b3793fa7c5fad2612fa848/src/parcsr_ls/par_relax_more.c#L371
we need to check if the norm of r is greater than 0 (or some small epsilon). And to break the ```while``` loop if that check fails.

Thanks.

liruipeng:
@bkrasnopolsky Thank you for reporting this issue. Is it possible that you can provide the matrix for which this problem occurs? Thanks!

bkrasnopolsky:
@liruipeng Would it helpful if I share the specific matrix, which causes the hang? This one is for the last level of MG hierarchy, and matrix has only 3 rows/columns distributed across 12 MPI processes (no OpenMP threads). The dumped matrix (using hypre_ParCSRMatrixPrint) is attached.
It should be noted that changing the number of MPI processes typically allows to overcome this issue (as the last level matrix slightly changes) and solve the system with the same parameters configuration.
[mat.tar.gz](https://github.com/hypre-space/hypre/files/7584898/mat.tar.gz)

liruipeng:
@bkrasnopolsky Can you please try if PR #534 fixes the issue? Thanks!

bkrasnopolsky:
@liruipeng Yes, that solves the problem! Thanks!

PR #534, opened 2021-11-25
Early break in CG Eig

liruipeng:
This PR adds early break in CG for eigenvalue estimations. The early break occurs when Krylov subspace is invariant with less than `n` iterations (`n` is the dimension of the matrix).  The current code hangs under certain circumstances, see #531.


ulrikeyang (approved):

rfalgout (approved):
Looks good

PR #537, opened 2021-12-02
WIP [SYCL] seq_mv 

abagusetty:
WIP


abagusetty:
@waynemitchell @liruipeng I will revert this PR to WIP since I think I have some left overs. I can push them by Mon (12/06). Just snowed under with multiple ECP reviews coming up.

PR #538, opened 2021-12-08
[SYCL] Seq mv sycl

abagusetty:


abagusetty:
@liruipeng @waynemitchell I have pushed all the relevant updates to this PR

PR #543, opened 2021-12-16
Replace deprecated cudaThreadSync

abagusetty:
replaces old, deprecated cudaThreadSync with cudaDeviceSync, with back compatibility


liruipeng:
This is handled in the new `spgemm` branch and is only for `HYPRE_PROFILE`. So, I close this PR, but thank you for reporting this issue.

PR #549, opened 2021-12-30
SYCL support for AMG solve phase

waynemitchell:
This PR adds SYCL support for the AMG solve phase for running on Intel GPUs. New features ported include:

1. matvec (via oneMKL sparse)
2. matrix transpose
2. vector operations (axpy, inner product, scale, etc.)

The following was also renamed throughout the code:

hypre_SyncCudaComputeStream() -> hypre_SyncComputeStream()

The IJ test suite now passes when compiling with SYCL, though AMG setup still takes place on the host.


src/utilities/Makefile
```diff
@@ -30,6 +30,7 @@ HEADERS =\
  mpistubs.h\
  threading.h\
  timing.h\
+ _hypre_onedpl.hpp\
```
rfalgout, line 33: I'm trying to remember if this is how we handle some of the other GPU ports.  Maybe @liruipeng can comment.

rfalgout (approved):
Looks good to me.  I am running the tux test now.

rfalgout:
Hi @waynemitchell .  The tux runs were clean except for one the test `check-mem.sh` which produced this:
```
./utilities/general.c:   free(hypre_handle_);
```
What's weird is that `general.c` doesn't show up in the changed files list (for me anyway).

rfalgout:
Ignore my previous comment.  I forgot to pull the branch before testing.  I'll rerun the tux tests now.  Sorry about that!

rfalgout:
One interesting thing I noticed while trying to figure out why `general.c` wasn't showing up is that a double-dot diff between the `sycl` branch and `master` shows 67 files changed (instead of 41):

   https://github.com/hypre-space/hypre/compare/master..sycl?w=1

GitHub uses a triple-dot diff by default (just add an extra "." in the url above), which is different in some way that I can't remember off the top of my head.  The double-dot is a "true" diff between the two branches, so it shows all of the differences.  I didn't check to see which files are not showing up, but it's definitely not `general.c`. :)

ulrikeyang (approved):
This looks good. Thanks

rfalgout:
FYI, the tux runs were clean (when run correctly :) ).  Thanks!

src/IJ_mv/aux_parcsr_matrix.c
```diff
@@ -264,12 +264,15 @@ hypre_AuxParCSRMatrixInitialize_v2( hypre_AuxParCSRMatrix *matrix,
       return 0;
    }
 
+   /* WM: Q - added the macro guards here (since IJ assembly not yet ported to sycl)... is this OK/correct? */
+#if defined(HYPRE_USING_CUDA) || defined(HYPRE_USING_HIP)
```
liruipeng, line 268: This is OK (for now).

src/seq_mv/csr_matop_device.c
```diff
@@ -1530,9 +1558,6 @@ hypre_CSRMatrixTransposeDevice(hypre_CSRMatrix  *A,
 #elif defined(HYPRE_USING_ROCSPARSE)
       hypreDevice_CSRSpTransRocsparse(nrows_A, ncols_A, nnz_A, A_i, A_j, A_data, &C_i, &C_j, &C_data,
                                       data);
-#elif defined(HYPRE_USING_ONEMKLSPARSE)
```
liruipeng, line 1533: Just want to make sure if it is supposed to be deleted, since I saw adding SYCL support for transpose on device in `csr_matop.c`.

src/utilities/Makefile
liruipeng: I was wondering if this file could go to `_hypre_utilities.hpp`. That's how we handled other cpp headers.

src/seq_mv/csr_matop_device.c
liruipeng: OK. maybe I understand now: this removed oneMKL version of the transpose

liruipeng (approved):
Looks good to me if the regressions are fine. Just had a few small questions. Thanks @waynemitchell

src/utilities/Makefile
waynemitchell: These headers are a bit of a strange special case. The oneDPL headers must be included before any C++ standard library headers in order to function correctly. I was having trouble finding an appropriate place in our current header setup that would yield the correct behavior (in particular, putting these in _hypre_utilities.hpp does not work), so I put them in their own header file that can be included at the very top of any source file where they are needed (e.g. seq_mv/vector.c). I'd be happy to hear thoughts on alternative ways of doing this.

src/seq_mv/csr_matop_device.c
waynemitchell: Correct, since oneMKL doesn't have a sparse transpose function (as of now).

src/utilities/Makefile
rfalgout: I don't have any bright ideas off the top of my head.  We can revisit this later.

liruipeng: I don't know much about about oneDPL. So even if you put it at the top of `_hypre_utilities.hpp`, it still did not work? It may be fine to me, just carrying another header file in `hypre/include`. Maybe @rfalgout can comment more. Thanks @waynemitchell .

waynemitchell: It's unclear to me whether this will always be true of the oneDPL headers or if they will be more flexible in the future. But for now, this was the cleanest quick solution I came up with. I think we can leave this as is for now and revisit later if oneDPL changes or I figure out a nicer way to include them.

liruipeng: Sounds good to me. Thanks!

src/seq_mv/csr_matvec_device.c
```diff
@@ -13,7 +13,6 @@
 
 #include "seq_mv.h"
 #include "_hypre_utilities.hpp"
-#include "seq_mv.hpp"
```
liruipeng, line 16: @waynemitchell Removing this line broke the GPU build with CUDA 11. What's the reason of removing it? for sycl? My reviewing did not catch this apparently.

flow-phys: Yes, this broke our builds as well.  Thanks for catching it, @liruipeng!

PR #562, opened 2022-01-21
Increase filename size for read/print of parcsr matrix/vector

waynemitchell:
I was running into issues with segfaults when trying to load matrix/vector files with long filenames (e.g. long absolute paths to some storage directory). This PR simply increases the possible size for the filenames, which fixed the issue for me.


rfalgout (approved):
Looks good to me.  -Rob

ulrikeyang (approved):
Looks good.

PR #563, opened 2022-01-22
Add convenient CMake alias

dbarbier:
Usually consumers of HYPRE call find_package(HYPRE) and depend on
HYPRE::HYPRE target.  But they could also want to use HYPRE via
add_directory (for instance via a git submodule)) or FetchContent,
in which case they have to depend on HYPRE target.

This alias makes this usage more consistent, all users could then
depend on HYPRE::HYPRE.  See for instance
  https://cmake.org/pipermail/cmake/2018-November/068629.html

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -549,4 +549,8 @@ export(EXPORT HYPRETargets
   NAMESPACE HYPRE::
   )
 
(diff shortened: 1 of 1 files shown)


osborn9:
Thanks @dbarbier for this addition!

osborn9 (approved):

PR #567, opened 2022-02-01
Fix to compile on cuda 11

PaulMullowney:


liruipeng:
this was broken by the very last PR. We will fix it. Thanks!

Issue #557, opened 2022-01-16
GPU support with single precision


metalcycling:
Hi Hypre Team,

I'm trying to compile Hypre to be used with single precision, but the code is failing on my end. Here is what I do:

```
$ ./configure --prefix=`pwd`/build/gpu/single/opt --with-cuda --enable-unified-memory --enable-single
checking for /gpfs/alpine/scratch/belloma2/csc262/Programs/Hypre/Master/src/../.git... yes
checking build system type... powerpc64-unknown-linux-gnu
checking host system type... powerpc64-unknown-linux-gnu
...
configure: creating ./config.status
config.status: creating config/Makefile.config
config.status: creating HYPRE_config.h
$ make install -j
Making blas ...
make[1]: Entering directory '/gpfs/alpine/csc262/scratch/belloma2/Programs/Hypre/Master/src/blas'
mpixlc -O2  -DHAVE_CONFIG_H  -I/sw/summit/cuda/11.0.3/include      -I.. -I../utilities -c dasum.c
...
csr_matop_device.c(1566): error: argument of type "HYPRE_Complex *" is incompatible with parameter of type "double *"

csr_matop_device.c(1572): error: argument of type "HYPRE_Complex *" is incompatible with parameter of type "double *"

csr_matop_device.c(1681): error: argument of type "HYPRE_Complex *" is incompatible with parameter of type "double *"
```

It seems to me that the CUDA functions that Hypre is using expect `double` arguments which are not valid when enabling single precision (`HYPRE_Complex = float`). Is this issue known? I couldn't find anything on the issues or pull requests. I tried compiling on Summit as well as my personal machine with the same results.

Thanks for you attention,
Pedro

PR #572, opened 2022-02-07
GPU support with single precision

liruipeng:
This PR fixes the GPU support with single precision, see #557. 

- [x] tux
- [x] ray
- [x] lassen
- [x] redwood


src/seq_mv/csr_matop_device.c
```diff
@@ -1741,10 +1741,8 @@ hypre_SortCSRCusparse(       HYPRE_Int      n,
    csru2csrInfo_t sortInfoA;
    HYPRE_CUSPARSE_CALL( cusparseCreateCsru2csrInfo(&sortInfoA) );
 
-   HYPRE_Int isDoublePrecision = sizeof(HYPRE_Complex) == sizeof(hypre_double);
-   HYPRE_Int isSinglePrecision = sizeof(HYPRE_Complex) == sizeof(hypre_double) / 2;
-
-   if (isDoublePrecision)
+#if !defined(HYPRE_COMPLEX)
```
rfalgout, line 1744: Is there no way to make the function calls themselves work under different precision scenarios, like we do in all of the CPU code?  We are starting to get lots of repeated #ifdef patterns in the GPU code.

rfalgout (approved):
Other than my question about the repeated #ifdef patterns, this looks good.  It would be nice if there were a cleaner approach, so I am curious to hear your answer on that.  Thanks!

src/seq_mv/csr_matop_device.c
liruipeng: @rfalgout I guess no (?). It's pretty much like when calling BLAS, ones has to call different versions (sgemm, dgemm, cgemm, zgemm) based on precisions. Of course, we can write another wrapper to hide all of these `#ifdef` inside. It's really a pain. Not sure if there can be a cleaner way, but I am happy to change if so.

rfalgout: I tend to lean towards writing wrapper functions.  How many distinct functions would need to be wrapped here?

liruipeng: A handful, currently. I think of 2 approaches 

1. We need to write a wrapper for each individual one. If we introduce more in the future, more wrappers will be needed. We probably will not save lines of code, but just rearrange things I guess.

2. Have a macro for all cusparse functions, and same for rocsparse mapping to different versions, since their names only differ by 1 letter (S, D, C, Z). I am not sure if this can work, but it's probably the cleanest way (albeit non-straightforward to read the code)

rfalgout: If we use these functions in more than one place in the code, then we save lines of code.  If they are only used once (here) and not expected to be used elsewhere, then we should leave things as you have them.

Just to be sure we are on the same page (I wasn't sure what you meant by "each individual one"), the wrapper function I am imagining would be able to call all precision versions and also switch between CUDA and rocsparse versions (all inside the one wrapper function).

liruipeng: We discussed this issue at https://teams.microsoft.com/l/message/19:D20F9e5I-A-kX560bS-UutIOIbDe2dcKrg7eYH1VzII1@thread.tacv2/1644943719570?tenantId=a722dec9-ae4e-4ae3-9d75-fd66e2680a63&groupId=f58d21a4-db04-469f-bd3a-7c278f424534&parentMessageId=1644943719570&teamName=HYPRE&channelName=General&createdTime=1644943719570.

ulrikeyang (approved):
This looks fine. Thank you.

PR #575, opened 2022-02-09
Invalid assumption on exclusive_scan

liruipeng:
This PR fixes a number of initialization problems with `exclusive_scan` on GPUs due to invalid assumptions of this function.


pwang234:
The changes look good to me.

ulrikeyang:
Ruipeng, does this PR need to be included in the new release?

PaulMullowney:
@liruipeng would you expect the incorrect initialization to produce inconsistent/unexpected results when using BoomerAMG? Would there be a difference between cuda 10 and 11?

liruipeng:
> @liruipeng would you expect the incorrect initialization to produce inconsistent/unexpected results when using BoomerAMG? Would there be a difference between cuda 10 and 11?

To clarify: we are chasing bugs with HIP. To do it, we wanted to clear all errors Compute Sanitizer throws. So far we found 1. uninitialization before `exclusive_scan`, which we believe it doesn't affect the correctness of the code but just make the sanitizer unhappy (it's basically adding garbage to results and take the same garbage out, so it shouldn't really matter, but still not a good thing to d0). 2. thrust bugs found in reduction with complex data types see https://github.com/NVIDIA/cccl/issues/860. It's unclear if it changes the results actually (all testings so far with CUDA suggest not). 3. There's a real bug in `src/parcsr_ls/par_lr_interp_device.c` that can be an issue, which is fixed in this PR.

ulrikeyang (approved):
Looks good. Thanks.

liruipeng:
Hi @rfalgout Can you take a look at this one? Thanks!

rfalgout (approved):
Looks good.  Thanks!

PR #577, opened 2022-02-10
Updated CMakeLists.txt for SYCL

mgrabban:
Hello,
I am using hypre as an workload here in Intel and so was trying to make it for Intel GPUs.
However, I was getting this error during make:

`no member named 'streams' in 'hypre_DeviceData'` from file src/utilities/device_utils.c

The reason for this is that SYCL (as well as HIP) code is inside a `#if defined(HYPRE_USING_CUDA_STREAMS)` block so the SYCL code is not being included. There are few other files inside utilities folder with the same 'issue'. To fix this, I just copied 3 lines from inside HYPRE_WITH_CUDA  section in CMakeLists.txt and pasted them inside HYPRE_WITH_SYCL section so HYPRE_USING_CUDA_STREAMS is set to ON for HYPRE_WITH_SYCL.

This compiles. But a better solution would probably be renaming HYPRE_USING_CUDA_STREAMS to something like HYPRE_USING_GPU_STREAMS.

Thanks,
Golam


liruipeng:
Yes, this makes sense to me. With autoconf, we enable CUDA stream by default when using sycl. Changing the name to a more appropriate one will be done in the future.

osborn9 (approved):
Thanks @mgrabban for the fix (and apologies for my delay in reviewing). 

Yes as @liruipeng mentioned, changing this to a more accurate name is a good idea.

PR #578, opened 2022-02-12
Fix checking Windows platform with preprocessor

Biswa96:
Check _WIN32 instead of _MSC_VER because the later one is for MSVC only.
Also __MINGW64__ is defined for 64 bit only. So, chekcing _WIN32 includes
all cases. This fixes building with 32 bit mingw-w64 toolchain.


Biswa96:
If required I can try to add GitHub Actions to build in all mingw environment.

Biswa96:
Any reply on this topic please?

PR #579, opened 2022-02-15
Improve SStruct RAP times

victorapm:
This PR updates the calculation of the unstructured component of a SStructMatrix triple-product. Specifically, this change targets the case when interpolation does not have an unstructured component. This ultimately benefits the setup time of SSAMG.


rfalgout (approved):
This looks good to me.  Ulrike should comment on the `seq_mv` and `parcsr_mv` file changes since they should also be changed on `master`.

ulrikeyang (approved):
This looks fine. Thanks

Issue #556, opened 2022-01-16
Searching for .git... when cross compiling?


artv3:
Hello, I've encountered an error I don't quite understand. 

When building on a blue_os system at LLNL I seem to hit the following build error with the latest hypre

```
src/../.git... configure: error: cannot check for file existence when cross compiling
```

Using hypre 2.22 built just fine. Has anyone seen this? Or have some insight on what might be going on?

artv3:
The source of the error seems to come from here: 

https://github.com/hypre-space/hypre/blob/514c72be691b8698df7e7069d54458831cbc531e/src/configure#L2690-L2692

which was introduced in PR https://github.com/hypre-space/hypre/pull/472

rfalgout:
Hi Arturo,

It looks like the autoconf macro `AC_CHECK_FILE` simply doesn't work in a cross-compile setting.  Can you say more about how we can replicate the problem?  Also, do you know if your build systems relies on having any of the `HYPRE_DEVELOP_*` variables?  Thanks!

-Rob

victorapm:
Hi all,

sorry for getting into the conversation but I implemented a fix for this issue on branch [fix-cross-comp](https://github.com/hypre-space/hypre/tree/fix-cross-comp). @artv3, could you try it and confirm if it works? 

Thanks!

rfalgout:
> sorry for getting into the conversation but I implemented a fix for this issue on branch [fix-cross-comp](https://github.com/hypre-space/hypre/tree/fix-cross-comp). @artv3, could you try it and confirm if it works?

That might just work, Victor!  Thanks!  Were you able to try it in a cross-compiling environment?

victorapm:
I didn't, Rob. The login and compute nodes of the machines I have access share the same architecture :(

victorapm:
Hi @artv3, I merged a fix for this issue on master. Let me know if you still have problems. 

Victor

artv3:
Thanks! I'll give it a try by the end of the week.

artv3:
Things are building now! Thanks!

PR #580, opened 2022-02-15
Fix cross-compilation problem

victorapm:
This PR fixes issue #556.

`AC_CHECK_FILE` was being used to test the existence of the `.git` folder. However, according to [Autoconf manual](https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Files.html):
```
 Be aware that, like most Autoconf macros, they test a feature of the host machine, and therefore, they die when cross-compiling
```
So, this PR implements another strategy for looking for the `.git` folder which works also when doing cross-compilation.

### DIFF
--- a/src/config/configure.in
+++ b/src/config/configure.in
@@ -105,7 +105,7 @@ AC_SUBST(HYPRE_BUGS)
 
 AC_SUBST(HYPRE_SRCDIR)
 
-AC_CHECK_FILE([$HYPRE_SRCDIR/../.git],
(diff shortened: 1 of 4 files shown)


rfalgout (approved):
This looks good, and I don't foresee any problems with it.  Hopefully it also works in cross compile settings too!  Thanks!

victorapm:
Thanks, Rob! I applied astyle in this branch and caught some style updates related to older commits.

PR #583, opened 2022-02-17
Add SStruct IO functions

victorapm:
The goal of this PR is to enable print/read functions for `SStructMatrix` and `SStructVector`. Additionally, public functions for printing and reading `StructMatrix` and `StructVector` are provided and a new regression test called `io` was added to the `TEST_sstruct` folder. FEM support for the `SStructGrid` is still missing on the print/read functions as well as CPU-GPU migration when executing hypre in device mode.


rfalgout (approved):
This looks really good, Victor!  I also like the regression tests you wrote.  Have you run the entire tux regression test yet (valgrind stuff included)?  Thanks!

victorapm:
Thanks, Rob! I ran the `TEST_sstruct` regression tests locally, including valgrind, and they passed. I'm running the full `machine-tux` test now.

ulrikeyang (approved):
This is fine, although it looks like there are some API changes, however I assume these functions haven't been used much.

rfalgout:
> This is fine, although it looks like there are some API changes, however I assume these functions haven't been used much.

Hi Ulrike.  I don't see any changes, only additions to the API, but maybe I missed something.  If existing interfaces don't change, I think we should be okay wrt backward compatibility, etc.

ulrikeyang:
It looks like HYPRE_StructMatrixPrint changed.

victorapm:
Hi all,

I double-checked the API of `HYPRE_StructMatrixPrint` on `master` and `sstruct-io` and they are equal. I didn't change the API on already existing functions in this PR.

The `machine-tux` tests passed, so this should be ready to be merged.

rfalgout:
I am happy with you merging, Victor.

PR #585, opened 2022-02-18
Small fix for mixedint mode

pbauman:
This should be a `HYPRE_BigInt`. I can run the `ij` test out to 64 nodes on crusher (~7B DoFs) with this change.


ulrikeyang (approved):
Good catch! That needs to be a BigInt!

liruipeng (approved):

PR #588, opened 2022-03-03
Need -O1 instead of -O0 for HIP in debug mode

pbauman:
Unfortunately, -O0 can cause ugly issues with the code generated
by the compiler so we need to use -O1 in debug mode (`--enable-debug`) at this time.


liruipeng (approved):
Looks good to me! Thanks @pbauman

PR #591, opened 2022-03-05
Nvcollab

liruipeng:
This PR contains various GPU optimizations in the collaboration with the NVIDIA team. The major changes include

1. Thrust sync stream issue. [Temporarily] replace with custom kernels for simple ones, axpy, scale, fill, etc...
2. CUDA mallocasync `--enable-device-malloc-async`. Vendor's memory pooling. Need CUDA 11.2
3. GPU-aware MPI: remove the sync stream. need to explicit add sync before doing MPI when needed. More performant but also more complicated to use
4. Move some memory to device (in par_amg.c)
5. Unify (cusparse, cublas....) function calls of versions with different precisions (code simplification)
6. Add options to use custom SpMV and SpTrans
7. ij.c: add `-benchmark` option. We should use this whenever timing performance. It always runs and times the 2nd "setup + solve" 
8. A lot more NVTX regions 

- [x] tux
- [x] ray
- [x] lassen
- [x] redwood
- [x] jlse


src/parcsr_mv/par_csr_matvec.c
```diff
@@ -268,7 +272,15 @@ hypre_ParCSRMatrixMatvecOutOfPlace( HYPRE_Complex       alpha,
...
+#endif
+
+   /* when using GPUs, start local matvec first in order to overlap with communication */
+#if defined(HYPRE_USING_GPU)
+   hypre_CSRMatrixMatvecOutOfPlace( alpha, diag, x_local, beta, b_local, y_local, 0 );
+#endif
+
+#ifdef HYPRE_PROFILE
    hypre_profile_times[HYPRE_TIMER_ID_HALO_EXCHANGE] -= hypre_MPI_Wtime();
 #endif
```
liruipeng, line 299: @pwang234 I don't recall why we did not need a `hypre_ForceSyncComputeStream` for the next `hypre_ParCSRCommHandleCreate_v2`. Is it because the above `thrust::gather` does a sync? If so, it is a dangerous assumption here. Thanks!

src/utilities/timing.c
```diff
@@ -278,6 +278,10 @@ hypre_EndTiming( HYPRE_Int time_index )
    hypre_TimingState(time_index) --;
    if (hypre_TimingState(time_index) == 0)
    {
+#if defined(HYPRE_USING_GPU)
```
liruipeng, line 281: @pwang234 Which is better here? SyncStream or SyncDevice?

rfalgout (approved):
This looks good to me.  One thing I would like to do at one of our future meetings is discuss the Zen behind the interfacing for the GPU code.  There are routines like `hypreDevice_*` and macros like `HYPRE_CUDA_*` and `#ifdef`-ing that is both scattered throughout the code and sometimes hidden behind macros.  Also, some of the C++ code lives in files with `.h` and `.c` extensions.

liruipeng:
@pwang234 Can you please run some testings on your side to check if the performance can still match what we recorded? Thanks!

oseikuffuor1 (approved):
Looks good to me. Thanks @liruipeng!

liruipeng:
@ulrikeyang @waynemitchell This PR should be good to go. Please take a look if you like. Thanks!

ulrikeyang (approved):
I approve this, but I don't think I understand all this. I agree with Rob that we need to talk about this mre.

src/parcsr_ls/par_relax_device.c
```diff
@@ -117,7 +117,7 @@ hypre_BoomerAMGRelaxTwoStageGaussSeidelDevice ( hypre_ParCSRMatrix *A,
    for (i = 0; i < num_inner_iters; ++i)
    {
       // 2) r = Lz
-      hypre_CSRMatrixSpMVDevice(1.0, A_diag, z_local, 0.0, r_local, -2);
+      hypre_CSRMatrixSpMVDevice(0, 1.0, A_diag, z_local, 0.0, r_local, NULL, -2);
```
waynemitchell, line 120: I'm just curious... why do we directly call the custom SpMV here instead of calling `hypre_CSRMatrixMatvecDevice` and letting it choose a default implementation?

liruipeng: We want to multiply the lower triangular part of a matrix (-2: strict lower), which is not supported by general MatvecDevice (using cusparse e.g.)

src/seq_mv/csr_matop_device.c
```diff
@@ -1807,15 +1827,24 @@ hypre_CSRMatrixTransposeDevice(hypre_CSRMatrix  *A,
    }
    else
    {
+      if ( !hypre_HandleSpTransUseVendor(hypre_handle()) )
```
waynemitchell, line 1830: Do we want to add an option to set this analogous to `HYPRE_SetSpMVUseVendor`?

liruipeng: Like this https://github.com/hypre-space/hypre/blob/7103dcc8ac71ec36c86db02964efc6819228aa8a/src/utilities/HYPRE_handle.c#L20?

waynemitchell: Yeah, but for selecting which transpose implementation to use. I see you added `HYPRE_SetSpMVUseVendor`, but nothing like `HYPRE_SetSpTransUseVendor`.

liruipeng: Oh. Sorry. I misunderstood the question (I was thinking about SpMV only). I can add that as well. Thanks!

waynemitchell:
@liruipeng I tested this on Arcticus, and everything looks good to me. Thanks!

waynemitchell (approved):

src/seq_mv/csr_matop_device.c
liruipeng: `HYPRE_SetSpTransUseVendor` has been added.

PR #599, opened 2022-03-12
(S)Struct IO on GPUs

victorapm:
This PR extends the (semi)-struct matrix/vector IO functions added on #583 with GPU support. Also, it fixes the regression tests on Lassen.


victorapm:
@liruipeng, I ran the regression tests on Lassen in the following way: `./test.sh machine-lassen.sh ../src` and saw some numerical differences in `TEST_ams/solvers.err` and `TEST_single/solvers_ij.err`. Does it hint there's still something to fix or should I have used some additional command line option when running the test? Thanks!!

.gitignore
```diff
@@ -12,3 +12,23 @@ config.status
 Makefile.config
 HYPRE_config.h
 autom4te.cache
+src/TAGS
```
rfalgout, line 15: What is this one?  I have not seen it before.  Would it be better in your global `.gitignore`?

.gitignore
```diff
@@ -12,3 +12,23 @@ config.status
...
+hypre/
+
+###############
+# Executables
+###############
+ams_driver
```
rfalgout: I think we should preface these with `src/test/` and `src/examples`?

rfalgout (approved):
Looks good to me.  Thanks!

.gitignore
victorapm: That makes sense! Fixed with [9d010e3](https://github.com/hypre-space/hypre/pull/599/commits/9d010e392c8e5cd4b00d3e5ad140d83b743a6841)

.gitignore
victorapm: Hi Rob, this file is generated by `make tags`, its purpose is to tell emacs, or other editors, where to find function definitions. See [this page](https://ftp.gnu.org/old-gnu/Manuals/emacs/html_node/emacs_338.html#:~:text=To%20select%20a%20tags%20table,variable%20tags%2Dfile%2Dname%20.) for instructions on how to configure it. For visiting a function definition, press `Alt + .` when the cursor is on top of it. I believe it makes sense to have it on `gitignore` since the contents of this file evolve with the library. But I'm fine with removing it :) let me know!

rfalgout: I didn't even realize we had a `tags` target. :)  Makes sense to keep this in `.gitignore`.  Thanks!

liruipeng:
> @liruipeng, I ran the regression tests on Lassen in the following way: `./test.sh machine-lassen.sh ../src` and saw some numerical differences in `TEST_ams/solvers.err` and `TEST_single/solvers_ij.err`. Does it hint there's still something to fix or should I have used some additional command line option when running the test? Thanks!!

How many nodes did you use? Did you follow https://github.com/hypre-space/hypre/wiki/Testing#running-on-lc-machines-with-gpu?

src/seq_mv/csr_matrix.c
```diff
@@ -333,18 +331,21 @@ hypre_CSRMatrixSetDataOwner( hypre_CSRMatrix *matrix,
...
+   HYPRE_Int   num_rows = hypre_CSRMatrixNumRows(matrix);
+   HYPRE_Int  *A_i = hypre_CSRMatrixI(matrix);
+   HYPRE_Int  *Arownnz = hypre_CSRMatrixRownnz(matrix);
 
-   HYPRE_Int i, adiag;
-   HYPRE_Int irownnz = 0;
+   HYPRE_Int   i;
+   HYPRE_Int   irownnz = 0;
+
+   if (Arownnz)
```
liruipeng: remove `if (Arownnz)`

src/IJ_mv/IJMatrix_parcsr.c
```diff
@@ -2912,6 +2912,11 @@ hypre_IJMatrixAssembleParCSR(hypre_IJMatrix *matrix)
       hypre_IJMatrixAssembleFlag(matrix) = 1;
    }
 
+   /* Generate the nonzero rows in the diag and offd matrices */
```
liruipeng: What's the reason of moving these two here?

victorapm: When we want to set the exact number of nonzeros of the diagonal and off-diagonal blocks of a `ParCSRMatrix`, which can be done via `HYPRE_IJMatrixSetDiagOffdSizes`, this function turns off the `NeedAux` flag, see: https://github.com/hypre-space/hypre/blob/f7787ab0aed34022503c656a3efc338e4616ff21/src/IJ_mv/IJMatrix_parcsr.c#L168

Fast-forward to `HYPRE_IJMatrixAssemble`, looking at the code on master, the following calls:
https://github.com/hypre-space/hypre/blob/f7787ab0aed34022503c656a3efc338e4616ff21/src/IJ_mv/IJMatrix_parcsr.c#L2865-L2866

are under the `if` statement https://github.com/hypre-space/hypre/blob/f7787ab0aed34022503c656a3efc338e4616ff21/src/IJ_mv/IJMatrix_parcsr.c#L2642

and thus are not reached. So, if want to build the `rownnz` info, we want them to be outside the `if` statement involving `NeedAux`. 

PS: It's actually a bit more correct to keep the `rownnz` functions inside the `if` block: https://github.com/hypre-space/hypre/blob/f7787ab0aed34022503c656a3efc338e4616ff21/src/IJ_mv/IJMatrix_parcsr.c#L2634

I implemented this minor fix on [352f74a](https://github.com/hypre-space/hypre/pull/599/commits/352f74a67a11d61ea473b74b990c8cd86d8527c9)

src/seq_mv/csr_matrix.c
victorapm: Thanks for catching it! Fixed with [f73e500](https://github.com/hypre-space/hypre/pull/599/commits/f73e50096c721bebdce8372b71174a391c56c98f)

victorapm:
> How many nodes did you use? Did you follow https://github.com/hypre-space/hypre/wiki/Testing#running-on-lc-machines-with-gpu?

Thanks for pointing this out. I used the wrong number of nodes... I will rerun the `machine-lassen.sh` now :)

src/IJ_mv/IJMatrix_parcsr.c
liruipeng: Thanks! I was just curious. Looks good!

liruipeng (approved):
Thank you @victorapm. All the changes look good once the regression tests are clean.

victorapm:
Thanks for the comments, Rob and Ruipeng! Both `machine-tux` and `machine-lassen` tests have passed, so I'm merging this into master.

PR #600, opened 2022-03-13
fix memory location

liruipeng:
This PR fixes a number of memory location issues in memory copy and memset. It also adds more strict checking in `memory.c` in the debug mode.

### DIFF
--- a/src/parcsr_ls/par_2s_interp_device.c
+++ b/src/parcsr_ls/par_2s_interp_device.c
@@ -53,7 +53,6 @@ hypre_BoomerAMGBuildModPartialExtInterpDevice( hypre_ParCSRMatrix  *A,
    HYPRE_Complex      *A_offd_data  = hypre_CSRMatrixData(A_offd);
    HYPRE_Int          *A_offd_i     = hypre_CSRMatrixI(A_offd);
    HYPRE_Int           A_offd_nnz   = hypre_CSRMatrixNumNonzeros(A_offd);
-   HYPRE_Int          *CF_marker_dev;
    HYPRE_Complex      *Dbeta, *Dbeta_offd, *rsWA, *rsW;
    hypre_ParCSRMatrix *As_F2F, *As_FC, *W, *P;
 
@@ -62,11 +61,6 @@ hypre_BoomerAMGBuildModPartialExtInterpDevice( hypre_ParCSRMatrix  *A,
    HYPRE_Int          *Soc_diag_j   = hypre_ParCSRMatrixSocDiagJ(S);
    HYPRE_Int          *Soc_offd_j   = hypre_ParCSRMatrixSocOffdJ(S);
 
-   CF_marker_dev = hypre_TAlloc(HYPRE_Int, A_nr_local, HYPRE_MEMORY_DEVICE);
-   hypre_TMemcpy(CF_marker_dev, CF_marker, HYPRE_Int, A_nr_local, HYPRE_MEMORY_DEVICE,
-                 HYPRE_MEMORY_HOST);
-
-   //TODO use CF_marker_dev
    /* As_F2F = As_{F2, F}, As_FC = As_{F, C2} */
    hypre_ParCSRMatrixGenerateFFFC3Device(A, CF_marker, num_cpts_global, S, &As_FC, &As_F2F);
 
@@ -113,7 +107,7 @@ hypre_BoomerAMGBuildModPartialExtInterpDevice( hypre_ParCSRMatrix  *A,
                       gDim, bDim,
                       A_nr_local,
                       A_offd_nnz > 0,
-                      CF_marker_dev,
+                      CF_marker,
                       A_diag_i,
                       A_diag_data,
                       Soc_diag_j,
@@ -127,7 +121,7 @@ hypre_BoomerAMGBuildModPartialExtInterpDevice( hypre_ParCSRMatrix  *A,
    HYPRE_Complex *new_end = HYPRE_THRUST_CALL( copy_if,
                                                rsWA,
                                                rsWA + A_nr_local,
-                                               CF_marker_dev,
+                                               CF_marker,
                                                rsW,
                                                equal<HYPRE_Int>(-2) );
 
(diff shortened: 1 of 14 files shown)


src/parcsr_ls/par_ilu_setup.c
```diff
@@ -3926,9 +3926,9 @@ hypre_ILUSetupMILU0(hypre_ParCSRMatrix *A, HYPRE_Int *permp, HYPRE_Int *qpermp,
          //hypre_TMemcpy(&(L_diag_j)[ctrL], iL, HYPRE_Int, lenl, HYPRE_MEMORY_DEVICE, HYPRE_MEMORY_HOST);
          //hypre_TMemcpy(&(L_diag_data)[ctrL], wL, HYPRE_Real, lenl, HYPRE_MEMORY_DEVICE, HYPRE_MEMORY_HOST);
          hypre_TMemcpy(&L_diag_j[ctrL], iL, HYPRE_Int, lenl,
-                       HYPRE_MEMORY_HOST, HYPRE_MEMORY_HOST);
+                       HYPRE_MEMORY_DEVICE, HYPRE_MEMORY_HOST);
```
liruipeng, line 3929: @oseikuffuor1 I restored the commented code (see above this line). Do you know why we did not want to use that?

liruipeng: @oseikuffuor1 And, also the other places in this file..

src/parcsr_ls/par_mgr_device.c
```diff
@@ -37,7 +37,7 @@ void hypreDevice_extendWtoP( HYPRE_Int P_nr_of_rows, HYPRE_Int W_nr_of_rows, HYP
 
 HYPRE_Int
 hypre_MGRBuildPDevice(hypre_ParCSRMatrix  *A,
-                      HYPRE_Int           *CF_marker_host,
+                      HYPRE_Int           *CF_marker,
```
liruipeng, line 40: @oseikuffuor1 Please double check if this is OK.

src/parcsr_ls/par_2s_interp_device.c
```diff
@@ -62,11 +61,6 @@ hypre_BoomerAMGBuildModPartialExtInterpDevice( hypre_ParCSRMatrix  *A,
    HYPRE_Int          *Soc_diag_j   = hypre_ParCSRMatrixSocDiagJ(S);
    HYPRE_Int          *Soc_offd_j   = hypre_ParCSRMatrixSocOffdJ(S);
 
-   CF_marker_dev = hypre_TAlloc(HYPRE_Int, A_nr_local, HYPRE_MEMORY_DEVICE);
```
liruipeng, line 65: @waynemitchell I think we missed the change here. With the IntArray, `CF_marker` should be on the device now. Does it look good to you?

src/parcsr_ls/par_strength2nd_device.c
```diff
@@ -21,7 +21,7 @@
 //-----------------------------------------------------------------------
 HYPRE_Int
 hypre_BoomerAMGCreate2ndSDevice( hypre_ParCSRMatrix  *S,
-                                 HYPRE_Int           *CF_marker_host,
```
liruipeng, line 24: @waynemitchell Another `CF_marker` change

liruipeng:
Hi @waynemitchell and @oseikuffuor1 Can you please take a look this PR, and let me know if it looks OK? I wanted to it merge it asap. Thanks!

src/parcsr_ls/par_2s_interp_device.c
waynemitchell: Yes, I agree that this extra copy to the device should be unnecessary now. Good catch!

src/parcsr_ls/par_strength2nd_device.c
waynemitchell: Yup, same as above. Looks good.

src/parcsr_ls/par_2s_interp_device.c
liruipeng: @waynemitchell Thanks! I am still puzzled why the following copy even worked. We were asking for an H2D copy but the memory is actually from device. Any ideas?

src/parcsr_ls/par_amgdd_helpers.c
```diff
@@ -2164,7 +2164,7 @@ hypre_BoomerAMGDD_PackSendBuffer( hypre_ParAMGDDData *amgdd_data,
          total_num_nodes = hypre_AMGDDCompGridNumOwnedNodes(compGrid[level]) +
                            hypre_AMGDDCompGridNumNonOwnedNodes(compGrid[level]);
 
-         hypre_Memset(add_flag[level], 0, sizeof(HYPRE_Int)*total_num_nodes, HYPRE_MEMORY_HOST);
+         hypre_Memset(add_flag[level], 0, sizeof(HYPRE_Int)*total_num_nodes, memory_location);
```
liruipeng, line 2167: @waynemitchell Does it look correct to you? Thanks

waynemitchell: Yes, I looked at this. I had allocated `add_flag` at `memory_location` above (even though this all only has host implementation right now), so this makes things consistent.

src/parcsr_ls/par_2s_interp_device.c
waynemitchell: Oh... that does seem suspicious. I'm not sure. Where all has this been tested? I remember you mentioning that lassen has a unique treatment of memory that allows for some bad behavior as far as accessing memory between host and device. I don't know whether that has anything to do with it.

waynemitchell (approved):
Looks good to me. Thanks @liruipeng .

src/parcsr_ls/par_2s_interp_device.c
liruipeng: @waynemitchell Yes, it has been tested in the regression. It may be because lassen can access host memory from device. But it doesn't explain the test on ray... Anyways, the copy is gone now :)

src/parcsr_ls/par_ilu_setup.c
oseikuffuor1: I am not sure exactly why. Must have been relying on UVM. These changes look good to me. Thanks.

src/parcsr_ls/par_mgr_device.c
oseikuffuor1: Looks good!

oseikuffuor1 (approved):

PR #604, opened 2022-03-18
hypre_ParPrintf

liruipeng:
This PR adds `hypre_ParPrintf`. Prints to standard out, only from the first processor in the communicator. Calls from other processes are ignored.

### DIFF
--- a/src/utilities/printf.c
+++ b/src/utilities/printf.c
@@ -128,28 +128,19 @@ hypre_ndigits( HYPRE_BigInt number )
 /* printf functions */
 
 HYPRE_Int
-hypre_vprintf( const char *format, va_list ap )
-{
-   HYPRE_Int ierr = 0;
-   char *newformat;
-
-   new_format(format, &newformat);
-   ierr = vprintf(newformat, ap);
-   free_format(newformat);
-   fflush(stdout);
-
-   return ierr;
-}
-
-HYPRE_Int
 hypre_printf( const char *format, ...)
 {
    va_list   ap;
+   char     *newformat;
    HYPRE_Int ierr = 0;
 
    va_start(ap, format);
-   ierr = hypre_vprintf(format, ap);
+   new_format(format, &newformat);
+   ierr = vprintf(newformat, ap);
+   free_format(newformat);
    va_end(ap);
+
+   fflush(stdout);
 
    return ierr;
 }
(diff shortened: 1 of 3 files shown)


src/utilities/printf.c
```diff
@@ -128,23 +128,32 @@ hypre_ndigits( HYPRE_BigInt number )
...
 }
 
+HYPRE_Int
+hypre_printf( const char *format, ...)
+{
+   va_list   ap;
+   HYPRE_Int ierr = 0;
+
+   va_start(ap, format);
+   ierr = hypre_vprintf(format, ap);
```
rfalgout: Call `vprintf` directly here.  I don't think there is any meaningful reason to write a `hypre_vprintf` function.  We won't use it anywhere else in the code, and it really only saves a few lines of code anyway.

src/utilities/printf.c
```diff
@@ -227,6 +236,27 @@ hypre_sscanf( char *s, const char *format, ...)
...
+   if (ierr)
+   {
+      return ierr;
+   }
+
+   if (!my_id)
+   {
+      va_list ap;
+      va_start(ap, format);
+      ierr = hypre_vprintf(format, ap);
```
rfalgout: Call `vprintf` directly here.

victorapm (approved):
Thanks, @liruipeng

I am curious to understand the reason for adding `hypre_vprintf`

src/utilities/printf.c
```diff
@@ -128,23 +128,32 @@ hypre_ndigits( HYPRE_BigInt number )
 /* printf functions */
 
 HYPRE_Int
-hypre_printf( const char *format, ...)
+hypre_vprintf( const char *format, va_list ap )
```
rfalgout: Remove this wrapper function.

rfalgout:
Hi @liruipeng .  Too bad it's not possible to pass the variable arguments for `hypre_ParPrintf` straight through to `hypre_printf`, but I don't think we should have the `hypre_vprintf` function.  I don't want to encourage its use elsewhere in the code, and it doesn't really save any lines of code (probably only creates more).

liruipeng:
> Hi @liruipeng . Too bad it's not possible to pass the variable arguments for `hypre_ParPrintf` straight through to `hypre_printf`, but I don't think we should have the `hypre_vprintf` function. I don't want to encourage its use elsewhere in the code, and it doesn't really save any lines of code (probably only creates more).

Sounds good. I will change it.

@victorapm The reason was I wanted to call `hypre_printf` from `hypre_ParPrintf` but did not find a way. So I introduced `hypre_vprintf`.

rfalgout (approved):
Looks great!  Thanks!

waynemitchell (approved):
Looks good to me. Thanks!

oseikuffuor1 (approved):
Looks great!!!

PR #605, opened 2022-03-18
Need to read start/end indices as HYPRE_BigInt

pbauman:
This fixes an issue with the ParCSRMatrixRead when compiled
with mixed-int enabled.


ulrikeyang (approved):
It took a while to find what %b actually means, but I guess that should work.

pbauman:
Sorry, I should have elaborated more.

`%b` is interpreted [here](https://github.com/hypre-space/hypre/blob/master/src/utilities/printf.c#L55) in the `new_format` function. And the format string is passed to that `new_format` function [here](https://github.com/hypre-space/hypre/blob/master/src/utilities/printf.c#L190).

liruipeng (approved):
Looks good to me! Thanks!

PR #606, opened 2022-03-22
fixed MM-multipass interpolation for case of no C-points

ulrikeyang:
This fixes the rare case of having no C-points globally. 
Note that having no C-points locally should not be an issue.


rfalgout (approved):
Looks good to me.  Thanks!

pbauman:
Hi @ulrikeyang  Will a similar change be needed on the device side as well? It seems so?

liruipeng:
> Hi @ulrikeyang Will a similar change be needed on the device side as well? It seems so?

I will fix that. Thanks!

ulrikeyang:
@pbauman , yes, but @liruipeng was planning to do so. Note that this is  bug is only triggered in very rare situations, which I would not expect to be the case for any of the problems you run.

src/parcsr_ls/par_mod_multi_interp_device.c
```diff
@@ -266,7 +272,7 @@ hypre_BoomerAMGBuildModMultipassDevice( hypre_ParCSRMatrix  *A,
                       HYPRE_Int(0) );
 
    /* contains beginning for each pass in pass_order field, assume no more than 10 passes */
-   pass_starts = hypre_CTAlloc(HYPRE_Int, 10, HYPRE_MEMORY_HOST);
```
liruipeng, line 269: @ulrikeyang I want to increase this by 1. It doesn't really matter at any rate. But semantically it would be better to allocate the 11th element, since later code syntactically tries to access `pass_starts[10]`. Thanks!

liruipeng:
When it breaks here https://github.com/hypre-space/hypre/blob/1074157200334943ff8a1da12a7a3959f9083838/src/parcsr_ls/par_mod_multi_interp_device.c#L410, I don't think continuing running the remaining code of the function is meaningful unless we make some special treatment for the last pass. But it's a rare case that running out of the 9 passes.

ulrikeyang (commented):
this looks good to me now.

liruipeng (approved):
Looks good. Thanks!

ulrikeyang:
Ok. This is not finished. The mod_multi_pass interpolation basically fails when there is a diagonal portion in the matrix.
Of course in that situation, one would just want to set that row in P to 0 and not incorporate this in a pass (or have a special pass to find those and set them).
So, before we merge this we should fix this.

ulrikeyang:
I think I fixed this now. I completely forgot that we had something called special fine points, which are isolated fine points and need to be treated differently. I pushed the fix.

ulrikeyang:
It turns out that this version still has problems with fine points connected only to fine and special fine points. I will provide a new version tomorrow that should handle that.

ulrikeyang:
Ok. I think this version should finally take care of all these issues. @liruipeng , you can now implement this on the GPU side.
Note that we have to check the global remaining points now to end early, which requires some bigints and global sum, vs max of local remaining points.

liruipeng:
> Ok. I think this version should finally take care of all these issues. @liruipeng , you can now implement this on the GPU side. Note that we have to check the global remaining points now to end early, which requires some bigints and global sum, vs max of local remaining points.

Sounds good. Thank you!

liruipeng:
@ulrikeyang I did the changes to the device code. Do you mind looking at it? I had to use slightly different implementations due to the way that the GPU code was written. Thanks!

ulrikeyang (commented):
Looks good. Thank you.

ulrikeyang:
@liruipeng Did those changes pass the  GPU regression tests?

liruipeng:
> @liruipeng Did those changes pass the GPU regression tests?

Not run yet. I am sending the branch to the testings .

src/test/TEST_ij/agg_interp.jobs
```diff
@@ -112,3 +112,5 @@ mpirun -np 8 ./ij    -rhsrand -n 30 29 31 -P 2 2 2 -agg_nl 1 -agg_interp 7 -inte
 mpirun -np 8 ./ij    -rhsrand -n 30 29 31 -P 2 2 2 -agg_nl 10 -agg_interp 7 -agg_Pmx 4 -agg_P12_mx 4 -solver 1 -rlx 6 \
  >> agg_interp.out.20
 
+mpirun -np 2 ./ij -fromfile tst -agg_nl 1 -agg_interp 8 \
+ >> agg_interp.out.21
```
liruipeng, line 116: @ulrikeyang where is the file `tst`?

ulrikeyang:
Sorry. I pushed the files.

liruipeng:
@ulrikeyang GPU regression is all good now. Thanks

PR #610, opened 2022-03-25
FSAI implementation on CPUs

victorapm:
Thir PR adds a factorized sparse approximate inverse (FSAI) implementation on hypre, which can be used as:

- Standalone solver
- Preconditioner to Krylov methods
- Complex smoother to BoomerAMG.

Particularly, we consider the adaptive algorithm version, where the sparsity pattern of the lower triangular factor `G` is built dynamically, i.e., during an iterative procedure that tries to find the best nonzero positions for a given row of `G`, see [this paper](https://dl.acm.org/doi/10.1145/2629475).

This implementation was performed on top of the `IJ` interface. It uses the diagonal portion of `A` for constructing `G`, i.e., it's a block-Jacobi method in the MPI sense. Lastly, both its setup and solve phases support threading via OpenMP. ~~A set of regression tests for FSAI remains to be added in `src/test/TEST_ij`~~.


src/blas/dasum.c
```diff
@@ -20,19 +20,19 @@ doublereal dasum_(integer *n, doublereal *dx, integer *incx)
     doublereal ret_val, d__1, d__2, d__3, d__4, d__5, d__6;
 
     /* Local variables */
-    static integer i, m;
-    static doublereal dtemp;
-    static integer nincx, mp1;
+    integer i, m;
```
rfalgout, line 23: Why did you remove all of the `static` keywords?  We generally try to change the blas code as minimally as possible.  We don't even apply astyle indentation to the files.

rfalgout (approved):
Other than my question about the blas/lapack changes, this looks good to me.  Thanks!

src/blas/dasum.c
victorapm: Hi Rob, this change was needed so that I'm able to call blas/lapack routines inside `omp parallel` regions. There are race conditions If these variable are `static` because they are initialized only once and shared between all threads executing that code. We can talk more if necessary :)

liruipeng: Why blas/lapack wanted to use `static` for local variables? It doesn't make sense to me. P.S. I understood this code was translated by F2C from Fortran. Is this because of Fortran?

rfalgout: I think `static` was used as an attempt to produce more optimized code, actually.  However, there will be an issue in openmp regions as Victor says.  Some of the early versions of this code were created via f2c, but I'm not sure about the CLAPACK release most recently used (see the `blas/README`).  Jack's group generally works hard at producing highly optimized code, so I wonder if it would be worthwhile to do a little more research about what's available for C implementations of blas, issues on thread safety, etc.  Maybe Victor has already done some of that?

liruipeng: So, are we using a non thread-safe version of CLAPACK? Shall we upgrade our local blas/lapack? I would like to use the latest version if it's worth the effort of changing these files. Thanks!

victorapm: Hi all, yes, I did some research on this. It seems that support for CLAPACK has been discontinued (the last version is from 2009). The reference LAPACK implementation has been including a set of C interfaces (LAPACKE) in recent releases, see [this](https://github.com/Reference-LAPACK/lapack#lapacke). To my knowledge, pure C implementations are not being actively supported. I remember seeing in a recent CHANGELOG of lapack that some thread safety issues have been fixed (I'll try to find that). So, it would be nice to upgrade our own distribution; however, we would need to compile Fortran code.

rfalgout: I don't want to require a Fortran compile for our provided version of blas/lapack.  Users can always link to external optimized versions of these routines.  The purpose of our provided code is to keep the compilation of hypre as simple as possible.  If there is a better C version of these routines that is thread safe, etc., then I am happy to switch to that.  I am also okay with keeping what is in this PR, but we should update the `README` file to mention removing `static` from new routines that get added in the future.

victorapm: I agree with you, Rob. Ruipeng, which functions are you planning to use? After the changes in this PR, I'm sure that `dpotrf` and `dpotrs` are thread-safe. Probably the more general dense direct solvers are thread-safe as well, but it would be good to double-check :).

ulrikeyang (approved):
Except for the 'static or non--static' discussion this looks good.
It seems that you only need a few of the lapack routines, but you changed all of them.

Heatherms27 (approved):

victorapm:
> Except for the 'static or non--static' discussion this looks good. It seems that you only need a few of the lapack routines, but you changed all of them.

Thanks, @ulrikeyang! Even though I needed that change for a few routines, it would be beneficial to all of them. That's why I applied the change everywhere. However, I can revert it to only the places that I needed if we decide so :)

src/parcsr_ls/CMakeLists.txt
```diff
@@ -47,6 +47,7 @@ set(SRCS
   HYPRE_parcsr_pcg.c
   HYPRE_parcsr_pilut.c
   HYPRE_parcsr_schwarz.c
+  HYPRE_parcsr_fsai.c
```
osborn9, line 50: Some new files are missing from this list 
```
  par_fsai.c
  par_fsai_setup.c
  par_fsai_solve.c
```
Maybe keep with the alphabetical ordering so after  `par_difconv.c` in the list

victorapm:
Hi all, I added user documentation and regression tests for FSAI in the recent commits. The regression tests `machine-tux` and `machine-lassen` are passing.

osborn9:
> Hi all, I added user documentation and regression tests for FSAI in the recent commits. The regression tests `machine-tux` and `machine-lassen` are passing.

@victorapm I've reviewed the rendered usr-manual documentation that you've added and it looks good!

osborn9 (approved):

victorapm:
Thanks, all! This has been merged into master. ~~I'll keep the `fsai` branch active for future developments~~. Thanks again!

PR #615, opened 2022-04-01
Fix copyright

victorapm:
I realized our Copyright headers are outdated, e.g., see https://github.com/hypre-space/hypre/blob/303457abae35c2cc8ce7e9264b06823e1b4f7b14/CHANGELOG#L1

This PR adds a check on `update.sh` for the finishing year used on the Copyright header. If outdated, that script updates every file of hypre that has the Copyright line with the new year.

I haven't run the `update.sh` script yet, but the change on the Copyright should apply for the next hypre release.


rfalgout:
Hi Victor,

Actually, I would like to change these headers once and for all by removing the end date.  It's been on my to do list for a while now.  I've already chatted with Todd Gamblin about what form that can take, etc.  Let's maybe do that here instead?  Let me know what you think and I'll give you more details.  Thanks!

-Rob

victorapm:
Sounds good, Rob! I agree with that approach. I'll move this PR to draft so that we can apply the next changes regarding Copyright here. Thanks!

rfalgout:
This is what I was planning to change it to:
```
# Copyright (c) 1998 Lawrence Livermore National Security, LLC and other
# HYPRE Project Developers. See the top-level COPYRIGHT file for details.
#
# SPDX-License-Identifier: (Apache-2.0 OR MIT)
```
Of course, the `#` characters don't apply to the other file types.  Basically, the text itself simply swaps `1998-2019` with `(c) 1998`.  The date (and even the copyright and the header itself, apparently) aren't even necessary according to Todd, but I like having the earliest date there so people have a sense of how long hypre has been around.  Also, I think it makes sense to only change the files minimally.

Comments from anyone on this before we make the change?

Victor: Would you be willing to use your scripts to make the change in this PR (let's wait for an answer to the above first)?

Thanks again!

victorapm:
Hi Rob,

I like the change you suggested. Also, I can update the files in hypre accordingly. Naturally, I would remove the changes I added to `update.sh` since they wouldn't apply anymore after this new idea. 

Thanks!

rfalgout:
Sounds good, Victor.

It looks like Ruipeng is okay with this change.  @ulrikeyang , @osborn9 , @oseikuffuor1 , @waynemitchell ?

rfalgout (approved):
Looks great, Victor!  Thanks!

PR #616, opened 2022-04-06
Fix complex build

victorapm:
This PR fixes build `basic--enable-complex` from the regression test suite `machine-tux`. It also adds `hypre_csqrt` for computing the square root of an `HYPRE_Complex` number. This macro/function works when `enable-complex` is turned on/off.


rfalgout (approved):
Looks good.  Thanks!

PR #622, opened 2022-04-18
Handle BigJ array reallocation in hypre_CSRMatrixResize

lahwaacz:
I hope that I got the change right... Let me know if it's acceptable or if I should improve anything.


liruipeng:
This looks good to me. We will test it before merging to the master. Thanks @lahwaacz !

PR #624, opened 2022-04-22
configure: update BLAS_SEARCH_DIRS and LAPACK_SEARCH_DIRS to include lib64

lahwaacz:
This should save users from having to specify `--with-blas-lib-dirs=/usr/lib64` etc. when the BLAS and Lapack libraries are located in `/usr/lib64` rather than `/usr/lib`.


lahwaacz:
Any feedback on this?

PR #629, opened 2022-05-04
Add header to remove header-transitivity issue

pbauman:
Future versions of Thrust are cleaning up their header transitivity and we need this header for `thrust::remove_if`.


liruipeng (approved):
looks good. Thanks!

PR #631, opened 2022-05-11
Hotfix for issue with dsuperlu in regression test.

oseikuffuor1:
Commented out unnecessary memory deallocation check.


osborn9 (approved):
Thanks @oseikuffuor1, this resolves the issue and the dsuperlu test passes.

PR #632, opened 2022-05-18
Runtest script

liruipeng:
1. Add an option of including a "script" in the command, as `gpu_tile_compact.sh` in `mpirun -n 4 gpu_tile_compact.sh a.out `
2. Move the `valgrind` and `mpibind` options out of  the `case $HOST`, so they become available for all platforms.


rfalgout (approved):
This looks good to me as long as you've tested it on all of the main platforms.  Thanks!

liruipeng:
tux and lassen tests looked good.

PR #633, opened 2022-05-20
Add -auxfromfile option to IJ driver

victorapm:
Add `-auxfromfile` option for reading an auxiliary matrix from file, which is then used to build the preconditioner. 

This is useful, for example, for the case that a filtered version of `A` is used to build the preconditioner.

### DIFF
--- a/src/test/ij.c
+++ b/src/test/ij.c
@@ -116,6 +116,8 @@ main( hypre_int argc,
    HYPRE_Int           omp_flag = 0;
    HYPRE_Int           build_matrix_type;
    HYPRE_Int           build_matrix_arg_index;
+   HYPRE_Int           build_matrix_M;
+   HYPRE_Int           build_matrix_M_arg_index;
    HYPRE_Int           build_rhs_type;
    HYPRE_Int           build_rhs_arg_index;
    HYPRE_Int           build_src_type;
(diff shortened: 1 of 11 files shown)


liruipeng:
So, the purpose is to build preconditioner from a matrix M and use it to precondition the original Ax=b? What is the application for this? Thanks!

victorapm:
@liruipeng, correct. One example of application is testing the separate displacement component strategy that geosx uses for solving linear elasticity.

rfalgout (approved):
Looks good to me.  Thanks!

oseikuffuor1 (approved):
Looks good. Thanks Victor!

src/test/ij.c
```diff
@@ -4148,11 +4220,11 @@ main( hypre_int argc,
...
       HYPRE_ParaSailsSetLogging(pcg_precond, ioutdat);
 
-      HYPRE_IJMatrixGetObject( ij_A, &object);
-      parcsr_mat = (HYPRE_ParCSRMatrix) object;
+      HYPRE_IJMatrixGetObject(ij_A, &object);
+      parcsr_A = (HYPRE_ParCSRMatrix) object;
 
-      HYPRE_ParaSailsSetup(pcg_precond, parcsr_mat, NULL, NULL);
-      HYPRE_ParaSailsBuildIJMatrix(pcg_precond, &ij_M);
+      HYPRE_ParaSailsSetup(pcg_precond, parcsr_A, NULL, NULL);
```
liruipeng: why it is not set up with `parcsr_M` here? It seems that `parcsr_M` is used in other setups

victorapm: I missed that! Thanks for catching!

liruipeng: so basically, from now on we should set up a preconditioner with `M` in the `ij` driver? If so, we should give all a heads-up

victorapm: Yes, good idea! I'll mention that on Teams

liruipeng (approved):
I am OK with these changes. We should keep in mind there is another matrix `M` to deal with and in most of the cases `M=A`.

liruipeng:
Please hold on the merge until I give green light. thanks!

liruipeng:
@victorapm You can merge in this PR now. (you have to deal with the conflicts. sorry)

PR #636, opened 2022-05-24
Thrust reduce header

liruipeng:
Copied from #629 "Future versions of Thrust are cleaning up their header transitivity and we need this header for thrust::remove_if."


waynemitchell (approved):
Looks fine to me. Thanks.

PR #643, opened 2022-05-31
Fix compilation warnings

victorapm:
This PR fixes compilation warnings obtained with `gcc-11`, `clang-12`, and `clang-14`.


rfalgout (approved):
This looks good to me.  Let me know if you need help running tests on tux.  Thanks!

victorapm:
>  Let me know if you need help running tests on tux

Thanks, @rfalgout. I appreciate if you could do that for me.

rfalgout:
> Thanks, @rfalgout. I appreciate if you could do that for me.

Running them now.

rfalgout:
The tux runs look good.  Only had the intermittent hopscotch error.

victorapm:
Thanks, Rob! That hopscotch error also bothers me. Do you believe it's safe to loose the residual norm comparison done over there so we don't have that error? I remember the difference was close to machine epsilon...

src/distributed_ls/Euclid/Vec_dh.c
```diff
@@ -29,7 +29,9 @@ void Vec_dhCreate(Vec_dh *v)
 void Vec_dhDestroy(Vec_dh v)
 {
   START_FUNC_DH
-  if (v->vals != NULL) FREE_DH(v->vals); CHECK_V_ERROR;
```
liruipeng, line 32: Did you change the "semantic" of this line? Do we want to change?

victorapm: Yes, I changed the "semantic". In the old version, `CHECK_V_ERROR` would be executed regardless of `FREE_DH(v->vals)` being executed or not. It seems that the error check makes sense only when the free operation is performed.

liruipeng: OK. I don't know the code. It seemed risky but I am fine with it if you know what it is doing. Thanks!

src/parcsr_ls/ams.c
```diff
@@ -409,12 +409,12 @@ hypreCUDAKernel_ParCSRMatrixFixZeroRows( HYPRE_Int      nrows,
 
    for (HYPRE_Int j = p1 + lane; j < q1; j += HYPRE_WARP_SIZE)
    {
-      l1_norm += fabs(A_diag_data[j]);
+      l1_norm += hypre_cabs(A_diag_data[j]);
```
liruipeng: `hypre_cabs` is a host function, one cannot use it in kernels. I understand you want it to work with complex, but we need to something else, if we want to support it.

src/parcsr_ls/par_relax_more_device.c
```diff
@@ -68,7 +68,7 @@ hypreCUDAKernel_CSRMaxEigEstimate(HYPRE_Int      nrows,
       }
       else
       {
-         row_sum_i += fabs(aij);
+         row_sum_i += hypre_cabs(aij);
```
liruipeng: same issue here.

src/seq_mv/csr_matop_device.c
```diff
@@ -995,7 +995,7 @@ hypreCUDAKernel_CSRMatrixFixZeroDiagDevice( HYPRE_Complex  v,
 
       if (find_diag)
       {
-         if (fabs(data[j]) <= tol)
+         if (hypre_cabs(data[j]) <= tol)
```
liruipeng: same here

src/distributed_ls/Euclid/ilu_mpi_bj.c
```diff
@@ -173,16 +173,16 @@ HYPRE_Int symbolic_row_private(HYPRE_Int localRow, HYPRE_Int beg_row, HYPRE_Int
 {
   START_FUNC_DH
   HYPRE_Int level = ctx->level, m = ctx->F->m;
-  HYPRE_Int *cval = ctx->F->cval, *diag = ctx->F->diag, *rp = ctx->F->rp; 
+  HYPRE_Int *cval = ctx->F->cval, *diag = ctx->F->diag, *rp = ctx->F->rp;
   HYPRE_Int *fill = ctx->F->fill;
   HYPRE_Int count = 0;
   HYPRE_Int j, node, tmp, col, head;
   HYPRE_Int fill1, fill2;
-  float val;
```
liruipeng, line 181: Do you know why did they specifically use single precision?

src/distributed_ls/Euclid/Vec_dh.c
rfalgout: I noticed this too, but since it is all on one line, I think the author originally intended it to be part of the `if`.  This is one of the reasons we enforce the use of squirly braces (the `distributed_ls` directory is excluded from our `astyle` indentation since it consists mainly of externally written software).

src/parcsr_ls/ams.c
victorapm: I wasn't aware of this, sorry! Thanks for your comment. I reverted to `fabs` inside device kernels on [292f62c](https://github.com/hypre-space/hypre/pull/643/commits/292f62cb7afb1f3b56472527f84116bf8efffcde)

src/distributed_ls/ParaSails/ParaSails.c
```diff
@@ -41,6 +41,7 @@ static void bzero(char *a, HYPRE_Int n) {HYPRE_Int i; for (i=0; i<n; i++) {a[i]=
 #endif
 #endif
 
+//#define PARASAILS_DEBUG
```
liruipeng: maybe move this line (as commented) to `Parasail.h`, so we can find it more easily to turn it on and off.

rfalgout:
> Thanks, Rob! That hopscotch error also bothers me. Do you believe it's safe to loose the residual norm comparison done over there so we don't have that error? I remember the difference was close to machine epsilon...

We've discussed this a number of times now as a team, and we haven't decided how best to deal with it.  I guess the issue is that the order of the computations is not guaranteed when using hopscotch?  We could loosen the tolerance, but it's nice to compare bitwise when the computations are deterministic.

src/distributed_ls/Euclid/ilu_mpi_bj.c
victorapm: I don't know, however this variable is used in https://github.com/hypre-space/hypre/blob/efe270582b8111877a0fc58ef849f04f62988afb/src/distributed_ls/Euclid/ilu_mpi_bj.c#L196 and `AVAL` is a `HYPRE_Real` (see function's definition). So, I figured out `val` should be a `HYPRE_Real`.

src/distributed_ls/ParaSails/ParaSails.c
victorapm: Sounds good to me. Done in [7a71fcb](https://github.com/hypre-space/hypre/pull/643/commits/7a71fcbe1a2bce167945dd7cfd4e4ae0dcf21235)

src/parcsr_ls/ams.c
```diff
@@ -3444,8 +3444,8 @@ HYPRE_Int hypre_AMSSetup(void *solver,
                {
                   GGt = hypre_ParCSRMatMat(ams_data -> G, Gt);
                }
-#endif
                else
+#endif
```
liruipeng, line 3448: This is a good catch. I was wondering how could it work before (with correct result)? Have you seen results changed with this fix?

victorapm: Yes, it was a good catch by the compiler... Imagine trying to debug this... I haven't seen CPU results changing after this fix. Maybe because https://github.com/hypre-space/hypre/blob/292f62cb7afb1f3b56472527f84116bf8efffcde/src/parcsr_ls/ams.c#L3429 is not true in the scenarios we test? So the following else block is always executed...  I'm not 100% sure, but that seems the case.

liruipeng: Glad you (or the compiler) found it which was indeed a dangerous mistake. Thanks!

liruipeng (approved):
A lot of good changes. Thanks! I didn't look at the blas/lapack changes which I assumed are fine.

liruipeng:
I also highly recommend running the GPU regression tests

victorapm:
The `machine-lassen` tests passed, so I'm merging this PR. Thanks!

PR #644, opened 2022-06-01
GPU-Accelerated ILU smoothers on AMD Hardware and Jacobi approximation to triangular solve

PaulMullowney:
This PR is coathored by Paul Mullowney and Steve Thomas. 

1) AMD and Nvidia architectures work for block ILU0 (ilu_type 0) and ILU0 with Schur compliment solves (ilu_type 10).

- Solvers work on Summit and Crusher without UVM. This requires data movement to/from GPU in various methods for computing permutations.
- There are certain cases that I haven't covered yet in AMD hardware, like ILUK, ILUT, ILU-RAP

2) Added Jacobi iterations as approximation to triangular solve with API for CPU/GPU. GPU Jacobi iterations works with any ilu factorization choice (Need UVM for choices other than ilu_type 0 or 10).

- When iterative tri solve (Jacobi Iterations) is chosen, one doesn't need to call solve analysis for the L and U factors. This yields nontrivial performance gain for Nvidia arches.
- default triangular solve parameter is direct (i.e. use Cusparse or Rocsparse)
- Added Lower/Upper Jacobi iteration parameters for more flexibility.
- Iterative solve works with Schur compliment

3) Added GPU implementation of Matrix Row/Column permutation (works for CUDA and HIP).

4) Added in hooks for CuSolver (used for RCM and other reorderings). This required significant reworking of the RCM code. Essentially I had to split hypre RCM methods into many submethods and then call cuSolver in the right place.

5) Lots of refactoring into *ilu*device.c files


PaulMullowney:
If it makes it easier, I could probably break this PR into 2
1) AMD support + no-UVM support
2) Jacobi Iteration approximation to triangular solve

liruipeng:
> If it makes it easier, I could probably break this PR into 2
> 
> 1. AMD support + no-UVM support
> 2. Jacobi Iteration approximation to triangular solve

Hi @PaulMullowney Yes, that will be easier for us. Thank you for the help!

PaulMullowney:
> > If it makes it easier, I could probably break this PR into 2
> > 
> > 1. AMD support + no-UVM support
> > 2. Jacobi Iteration approximation to triangular solve
> 
> Hi @PaulMullowney Yes, that will be easier for us. Thank you for the help!

Ok. I'll do that and let you know when its ready.

Issue #649, opened 2022-06-09
Write all the AMG matrices in a V-cycle to file


PaulMullowney:
Hi,
Is there a way to way to write matrices, in a BoomerAMG sequence, to disk. In particular, I want files for the matrices used in the smoothing operators.
-Paul

pbauman:
Hey @PaulMullowney. The (Par)CSRMatrix classes have methods for writing the matrices out. You can extract all the matrices from the `hypre_ParAMGData` (that is extracted from the, e.g., `hypre_PCGData`). You'd get one file for each of the matrices in the CSRMatrix case and nprocs*n_matrices in the ParCSRMatrix case. Is that what you're looking for?

PaulMullowney:
Hi @pbauman  Essentially yes. I was wondering if this had already been implemented somewhere which has already been exposed through some API/option.

pbauman:
Not AFAICT, but the HYPRE developers may know.. I just manually extracted the matrices and dumped them out.

liruipeng:
How about this https://github.com/hypre-space/hypre/blob/07a8def6f8b7dfdffa4208d962f08f5e5d80ccaa/src/parcsr_ls/par_amg_setup.c#L16? but I am not sure if it works with GPUs. If not, I can fix. Thanks!

PaulMullowney:
The issue seems to be that hypre_ParCSRMatrixPrintIJ assume either CPU matrices or GPU matrices in UVM.
Do you agree?

liruipeng:
> The issue seems to be that hypre_ParCSRMatrixPrintIJ assume either CPU matrices or GPU matrices in UVM. Do you agree?

Yes

PaulMullowney:
This worked for me:

````
(base) [mullowne@login2.summit src]$ git diff
diff --git a/src/parcsr_mv/par_csr_matrix.c b/src/parcsr_mv/par_csr_matrix.c
index cf108f904..56940bc95 100644
--- a/src/parcsr_mv/par_csr_matrix.c
+++ b/src/parcsr_mv/par_csr_matrix.c
@@ -672,6 +672,26 @@ hypre_ParCSRMatrixPrintIJ( const hypre_ParCSRMatrix *matrix,
       return hypre_error_flag;
    }

+#if defined(HYPRE_USING_GPU)
[...]

PR #655, opened 2022-06-20
hypre_ParCSRMatrixPrintIJ on device

victorapm:
This PR updates `hypre_ParCSRMatrixPrintIJ` to work on the device without the need of UVM. This also eliminates the need of migrating a `hypre_ParCSRMatrix` to host before calling the print routine and thus reduces some code duplication in hypre.

This addresses https://github.com/hypre-space/hypre/issues/649#issuecomment-1157008796

Note: This will be used by GEOSX after disabling UVM from their default hypre build.


liruipeng (approved):
Looks good to me. Thank you @victorapm

rfalgout (approved):
Looks good to me too.  Thanks!

PR #657, opened 2022-06-21
Chebyshev polynomial fix for nearly-singular matrices 

victorapm:
Our current Chebyshev implementation can fail for singular/nearly-singular matrices. This happens because the eigenvalue computation in 
https://github.com/hypre-space/hypre/blob/9b4c1336e7a0371ed1ffa168e8d6034aff2013d7/src/parcsr_ls/par_relax_more.c#L382
may return a negative `lambda_min` close to zero while `lambda_max > 0`, causing problems later on when setting `lower_bound`  and `upper_bound` at 
https://github.com/hypre-space/hypre/blob/b58585e0f041525a1579e046c2148a6ca189a318/src/parcsr_ls/par_cheby.c#L92-L103 (the bounds are never computed)

This PR fixes this issue by always computing the bounds regardless of the sign of `min_eig`

Question to @rfalgout: Should we add regression tests with singular matrices so that we can catch errors like this?

### DIFF
--- a/src/parcsr_ls/par_cheby.c
+++ b/src/parcsr_ls/par_cheby.c
@@ -68,7 +68,7 @@ hypre_ParCSRRelax_Cheby_Setup(hypre_ParCSRMatrix *A,         /* matrix to relax 
    hypre_CSRMatrix *A_diag       = hypre_ParCSRMatrixDiag(A);
    HYPRE_Real       theta, delta;
    HYPRE_Real       den;
-   HYPRE_Real       upper_bound = 0.0, lower_bound = 0.0;
+   HYPRE_Real       upper_bound, lower_bound;
    HYPRE_Int        num_rows     = hypre_CSRMatrixNumRows(A_diag);
    HYPRE_Real      *coefs        = NULL;
    HYPRE_Int        cheby_order;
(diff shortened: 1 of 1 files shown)


ulrikeyang (approved):
Looks good.

rfalgout (approved):
Looks good to me, too.  It's never really a bad idea to add new regression tests.  Your call.  Thanks!

liruipeng:
This looks OK to me. Another possible fix (together along with this one maybe) is to make sure the spectrum returned by estimation algorithms $[\lambda_{\min}, \lambda_{\max}]$ doesn't include 0, 
i.e., to not have $\lambda_{min} <0,  \lambda_{max} > 0$. We did it in the Gershgorin algorithm but not in CG. The old "if-else" statement was somewhat intentionally written as is, assuming this is true.

liruipeng (approved):

victorapm:
Thanks, all!

>  It's never really a bad idea to add new regression tests. Your call

I will put this in my todo list

> Another possible fix (together along with this one maybe) is to make sure the spectrum returned by estimation algorithms doesn't include 0

I agree with that Ruipeng, do you want to implement this additional check in this branch before I merge? If so, feel free to push changes here. 

Thanks!

liruipeng:
> Thanks, all!
> 
> > It's never really a bad idea to add new regression tests. Your call
> 
> I will put this in my todo list
> 
> > Another possible fix (together along with this one maybe) is to make sure the spectrum returned by estimation algorithms doesn't include 0
> 
> I agree with that Ruipeng, do you want to implement this additional check in this branch before I merge? If so, feel free to push changes here.
> 
> Thanks!

Let's do that at a later point, if we want, since I guess we are going to revisit those functions when doing more polynomial preconditioner work next year. Thanks!

PR #658, opened 2022-06-22
PrintIJ on GPUs (without UVM)

PaulMullowney:
Adding support for printing matrices when running on GPU without UVM


PaulMullowney:
Looks like you guys already have a PR for this

PR #664, opened 2022-06-28
Sycl pmis

waynemitchell:
This ports PMIS coarsening to SYCL.


ulrikeyang (approved):
This looks fine, as far as I can tell.

liruipeng (approved):
Looks good to me as far as I can tell, and I trust @waynemitchell on the sycl implementations. Thanks @waynemitchell .

src/test/TEST_ij/agg_interp.saved.arcticus
```diff
@@ -2,169 +2,169 @@
 
 
 Iterations = 10
-Final Relative Residual Norm = 5.547280e-09
+Final Relative Residual Norm = 4.982180e-09
```
rfalgout, line 5: These are pretty different numbers.  I assume you expect this?

waynemitchell: The .saved files for arcticus prior to this PR used the CPU implementation of PMIS and this new result comes from the GPU implementation. I assume the difference in convergence is due to non-trivial differences in the selection of C points (due to randomness in PMIS), which would be even more pronounced in an aggressive coarsening example like this one. Do you think this difference is too large to be accounted for by my explanation?

rfalgout: Sounds like a good reason for the difference.  Thanks!

PR #666, opened 2022-06-29
Interp truncation on device

liruipeng:
This PR optimizes interpolation truncation routines on GPUs. It adds faster routines for truncation $P_{\max} < 8$ and keeps the old routine as the "general" method to deal with any $P_{\max} \ge 8$

- [x] tux
- [x] lassen
- [x] tioga
- [x] jlse (@waynemitchell  Please help)


src/test/TEST_ij/elast.saved.lassen
```diff
@@ -28,15 +28,15 @@ Final Relative Residual Norm = 7.092339e-10
 
 # Output file: elast.out.7
 Iterations = 17
-Final Relative Residual Norm = 5.873774e-09
+Final Relative Residual Norm = 3.595815e-09
```
rfalgout, line 31: Why are these differences so large?  Are you expecting them to be?

rfalgout (approved):
This looks good to me modulo the question about the saved files differing by so much.  Thanks!

src/test/TEST_ij/elast.saved.lassen
liruipeng: Hi @rfalgout Which difference is "large"? I found the iteration counts were off by 1-2, and the norms were within the same order of magnitude. Maybe I missed something? Do you mean why there are so many? If so, since I changed the truncation routines, the way it breaks ties and the order of the remaining entries are different. So, I expected a lot of results would be changed. (there will be more in `sstruct`, I am running them).

rfalgout: Hi Ruipeng.  If the algorithms are "the same", then differences in the first digit would normally be considered quite "large".  It sounds like you are expecting this, though, so no worries.  I just wanted to check.  Thanks!

waynemitchell (approved):
@liruipeng, this all looks good to me. Since interpolation (and interpolation truncation) still does not occur on the device with the sycl backend as of master, this will not affect the sycl build at all. I think you can go ahead and merge this into master, and then I'll merge these changes into my upcoming PR porting interpolation stuff to sycl. It all looks good to go, requiring minimal changes to align with my branch.

PR #653, opened 2022-06-17
Cusolver

liruipeng:
Copied from #647, which adds the option of using cuSolver (https://docs.nvidia.com/cuda/cusolver/index.html), a GPU accelerated library for decompositions and linear system solutions from NVIDIA. We need the reordering algorithms (like RCM and AMD) on GPUs from cuSolver for ILU.

PR #671, opened 2022-07-07
config/update.sh

liruipeng:
The last PR #653 apparently did not run  `config/update.sh` to update `configure`.

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -1518,7 +1518,7 @@ Optional Features:
                           (default is NO).
   --enable-cuda-streams   Use CUDA streams (default is YES).
   --enable-cusparse       Use cuSPARSE (default is YES).
-  --enable-cusolver       Use cusolver (default is NO).
+  --enable-cusolver       Use cuSOLVER (default is NO).
   --enable-device-memory-pool
                           Use device pooling allocator (default is NO).
   --enable-device-malloc-async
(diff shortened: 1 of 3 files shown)


rfalgout:
Hi Ruipeng.  I don't see any real differences in these files, except for the order.  Is that all you wanted to change?

liruipeng:
> Hi Ruipeng. I don't see any real differences in these files, except for the order. Is that all you wanted to change?

Yes, @rfalgout. It appears just difference of the ordering. I was trying to make other changes to configure. I found running `config/update.sh` could change the files in the main branch. I guess it's still a good idea to make them consistent.

src/configure
```diff
@@ -1518,7 +1518,7 @@ Optional Features:
                           (default is NO).
   --enable-cuda-streams   Use CUDA streams (default is YES).
   --enable-cusparse       Use cuSPARSE (default is YES).
-  --enable-cusolver       Use cuSOLVER (default is NO).
+  --enable-cusolver       Use cusolver (default is NO).
```
rfalgout: Should the comment say "Use cuSOLVER"?  This would change `configure.in` as well.  The `configure` on `master` uses capital letters, but presumably `configure.in` does not?  I'm still confused about the inconsistencies here, but I guess this will clean everything up.

rfalgout (approved):
Looks good.

src/configure
liruipeng: Fixed. Thanks @rfalgout !

About the inconsistencies, I meant whenever we run the script `./config/update.sh`, the file `configure` should not be changed if it was generated by the script before. This was not true when I ran it with the latest master branch. I don't know how it could happen. I hope this gives what we want. Thanks again!

PR #675, opened 2022-07-08
Cmake cuda update

liruipeng:
This PR contains CMake changes for CUDA.


src/CMakeLists.txt
```diff
@@ -231,6 +226,12 @@ if (HYPRE_WITH_CUDA)
   # Check if CUDA is available, then enable it
   include(CheckLanguage)
   check_language(CUDA)
+
```
liruipeng, line 230: @osborn9 "For CUDA which can have an explicit host compiler, the cache `CMAKE_CUDA_HOST_COMPILER` variable will be set if it was required for compilation (and cleared if it was not)." It could be whether `check_language(CUDA)` decide whether `CMAKE_CUDA_HOST_COMPILER` is "required" can be system dependent.
In any case, it's not good to rely on `check_language(CUDA)` to decide whether `CMAKE_CUDA_HOST_COMPILER` is necessary if you know you need it.

osborn9: Thanks @liruipeng for the explanation / improvement.

osborn9 (approved):
Looks good! 

I made a minor change to mimic the configure functionality, so that cudaMallocAsync if off by default, but can be enabled by the user via the variable HYPRE_ENABLE_DEVICE_MALLOC_ASYNC.

PR #676, opened 2022-07-13
SYCL IJ without unified memory

waynemitchell:
This PR ports a few remaining routines required for running the ij.c driver without unified memory using the sycl backend (for some limited subset of options). Ported routines are:
1. IJ vector set/add values and assembly
2. Routines in par_coarse_parms_device.c
3. hypre_ParCSRComputeL1Norms()


waynemitchell:
Regression tests pass on tux, lassen, and jlse.

src/test/TEST_ij/solvers.saved.arcticus
```diff
@@ -134,8 +134,8 @@ Iterations = 19
 Final Relative Residual Norm = 3.502389e-09
 
 # Output file: solvers.out.107
-Iterations = 34
-Final Relative Residual Norm = 8.725157e-09
+Iterations = 29
```
ulrikeyang, line 137: Convergence improved a lot here. Was there a bug before?

ulrikeyang (approved):
This looks good as far as I can tell with limited SYCL knowledge .

rfalgout (approved):
Looks good to me too!

src/test/TEST_ij/solvers.saved.arcticus
waynemitchell: Yes, I think there was a previous bug, though I don't think it was fixed in this PR. When I reran master on this test I got the same (29 iteration) result shown here. So I guess I just hadn't updated the .saved file previously.

liruipeng (approved):
Looks good to me. Again, I trusted @waynemitchell on the sycl implementation details. Thanks!

PR #677, opened 2022-07-14
Fix GPU memory leak

liruipeng:
This PR fixes a memory leak on GPUs and adds regression tests to check.


waynemitchell (approved):

liruipeng:
Add a regression test which hopefully can catch memory leaks in the future.

ulrikeyang (approved):
This looks good. Thanks

PR #678, opened 2022-07-15
Backward compatible ROCm header update.

pbauman:
ROCm 5.2 changed the primary location of the headers, so updated
build system and a couple of include statements to reflect this.
The update is backward compatible to at least the previous three
ROCm releases (i.e. 4.5).


liruipeng (approved):
Looks good. Thanks!

PR #682, opened 2022-07-19
Umpire Pinned Pool Fix

PaulMullowney:
slight fix to compile with umpire and pinned memory pools


liruipeng (approved):
Looks good. Thanks!

Issue #681, opened 2022-07-18
HYPRE_Int -> HYPRE_BigInt in hypre_ParcsrGetExternalRowsDeviceInit?


pbauman:
Line in question:

https://github.com/hypre-space/hypre/blob/master/src/parcsr_mv/par_csr_matop_device.c#L714

Shouldn't this be a `HYPRE_BigInt`? 

@liruipeng @ulrikeyang 

Thanks!

liruipeng:
Yes, @pbauman . This should be BigInt. Thanks for catching it.

liruipeng:
Fixed. Thanks!

PR #684, opened 2022-07-20
fix bigint

liruipeng:
Fix BigInt issue #681


pbauman (approved):
Looks good! I should say credit goes to Kostas for finding this. Thanks!

ulrikeyang (approved):
Good Catch! Thanks!

PR #686, opened 2022-07-20
Conversions that may alter values

liruipeng:
[WIP]


pbauman:
Dup of #684?

liruipeng:
> Dup of #684?

No, there seem to be more real issues.

PR #689, opened 2022-07-25
Add `IJVectorSetConstantValues` for setting all values to a constant

fredrikekre:
This adds a new function:

```
HYPRE_IJVectorSetConstantValues( HYPRE_IJVector vector, HYPRE_Complex value )
```

for setting all values in an `IJVector` to a constant value, analoguous to `HYPRE_IJMatrixSetConstantValues`. Internally this simply generalizes the existing `hypre_IJVectorZeroValuesPar` function to work with any value, not just zero.

---

To provide some background information: I wanted a simple way to zero out a vector (for original allocation, and for reusing when solving the next time-step, for example). However, while working on this I noticed that
```
HYPRE_IJVectorInitialize(x)
HYPRE_IJVectorAssemble(x)
```
might be enough, but [`examples/ex5.c`](https://github.com/hypre-space/hypre/blob/ea3c43a6b091a4c1f23ec86cc346c635fae09d5f/src/examples/ex5.c#L285), where all values in the solution vector is explicitly set to 0, made the impression that this might not be the case. However, since there is a corresponding function for IJMatrix, perhaps this is patch is useful regardless. Opening as a draft PR (Edit: I miss clicked, and only maintainers can convert to a draft PR, apparently.) since I want feedback whether this patch make sense, and if it does I can add some tests. Cheers!

### DIFF
--- a/src/IJ_mv/HYPRE_IJVector.c
+++ b/src/IJ_mv/HYPRE_IJVector.c
@@ -319,6 +319,33 @@ HYPRE_IJVectorSetValues( HYPRE_IJVector        vector,
 }
 
 /*--------------------------------------------------------------------------
+ * HYPRE_IJVectorSetConstantValues
+ *--------------------------------------------------------------------------*/
+
+HYPRE_Int
+HYPRE_IJVectorSetConstantValues( HYPRE_IJVector vector, HYPRE_Complex value )
+{
+   hypre_IJVector *vec = (hypre_IJVector *) vector;
+
+   if (!vec)
+   {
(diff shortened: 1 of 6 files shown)


fredrikekre:
Bump. Is this something that is interesting/useful to have in hypre?

victorapm:
@fredrikekre I'm very sorry, I just merged a PR for this feature and just noticed now your PR. Sorry for missing this and thank you for your contribution. I'm closing this since the feature has been added

fredrikekre:
Great, thanks.

PR #695, opened 2022-07-27
[Multivec 4/5]: BoomerAMG partially supports multivectors

victorapm:
This is part of a series of PRs for enabling BoomerAMG to be applied to multivectors.

This PR extends BoomerAMG with L1-Jacobi smoothing to work with multivectors

Regression tests:

- [x] Tux
- [x] Lassen

### DIFF
--- a/src/parcsr_ls/par_amg.c
+++ b/src/parcsr_ls/par_amg.c
@@ -514,376 +514,343 @@ HYPRE_Int
 hypre_BoomerAMGDestroy( void *data )
 {
    hypre_ParAMGData  *amg_data = (hypre_ParAMGData*) data;
-   HYPRE_Int num_levels = hypre_ParAMGDataNumLevels(amg_data);
-   HYPRE_Int smooth_num_levels = hypre_ParAMGDataSmoothNumLevels(amg_data);
-   HYPRE_Solver *smoother = hypre_ParAMGDataSmoother(amg_data);
-   void *amg = hypre_ParAMGDataCoarseSolver(amg_data);
-   MPI_Comm new_comm = hypre_ParAMGDataNewComm(amg_data);
-   HYPRE_Int i;
-   HYPRE_Int *grid_relax_type = hypre_ParAMGDataGridRelaxType(amg_data);
 
    HYPRE_ANNOTATE_FUNC_BEGIN;
+   if (amg_data)
+   {
+      HYPRE_Int     num_levels = hypre_ParAMGDataNumLevels(amg_data);
+      HYPRE_Int     smooth_num_levels = hypre_ParAMGDataSmoothNumLevels(amg_data);
+      HYPRE_Solver *smoother = hypre_ParAMGDataSmoother(amg_data);
+      void         *amg = hypre_ParAMGDataCoarseSolver(amg_data);
+      MPI_Comm      new_comm = hypre_ParAMGDataNewComm(amg_data);
+      HYPRE_Int    *grid_relax_type = hypre_ParAMGDataGridRelaxType(amg_data);
+      HYPRE_Int     i;
 
 #ifdef HYPRE_USING_DSUPERLU
-   //   if (hypre_ParAMGDataDSLUThreshold(amg_data) > 0)
-   if (hypre_ParAMGDataDSLUSolver(amg_data) != NULL)
-   {
-      hypre_SLUDistDestroy(hypre_ParAMGDataDSLUSolver(amg_data));
-      hypre_ParAMGDataDSLUSolver(amg_data) = NULL;
-   }
+      // if (hypre_ParAMGDataDSLUThreshold(amg_data) > 0)
+      if (hypre_ParAMGDataDSLUSolver(amg_data) != NULL)
+      {
+         hypre_SLUDistDestroy(hypre_ParAMGDataDSLUSolver(amg_data));
+         hypre_ParAMGDataDSLUSolver(amg_data) = NULL;
+      }
 #endif
 
-   if (hypre_ParAMGDataMaxEigEst(amg_data))
-   {
-      hypre_TFree(hypre_ParAMGDataMaxEigEst(amg_data), HYPRE_MEMORY_HOST);
-      hypre_ParAMGDataMaxEigEst(amg_data) = NULL;
-   }
-   if (hypre_ParAMGDataMinEigEst(amg_data))
-   {
-      hypre_TFree(hypre_ParAMGDataMinEigEst(amg_data), HYPRE_MEMORY_HOST);
-      hypre_ParAMGDataMinEigEst(amg_data) = NULL;
-   }
-   if (hypre_ParAMGDataNumGridSweeps(amg_data))
-   {
-      hypre_TFree(hypre_ParAMGDataNumGridSweeps(amg_data), HYPRE_MEMORY_HOST);
-      hypre_ParAMGDataNumGridSweeps(amg_data) = NULL;
-   }
-   if (grid_relax_type)
-   {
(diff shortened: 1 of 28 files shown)


src/parcsr_ls/par_amg.c
```diff
@@ -514,376 +514,343 @@ HYPRE_Int
...
-   HYPRE_Int num_levels = hypre_ParAMGDataNumLevels(amg_data);
-   HYPRE_Int smooth_num_levels = hypre_ParAMGDataSmoothNumLevels(amg_data);
-   HYPRE_Solver *smoother = hypre_ParAMGDataSmoother(amg_data);
-   void *amg = hypre_ParAMGDataCoarseSolver(amg_data);
-   MPI_Comm new_comm = hypre_ParAMGDataNewComm(amg_data);
-   HYPRE_Int i;
-   HYPRE_Int *grid_relax_type = hypre_ParAMGDataGridRelaxType(amg_data);
 
    HYPRE_ANNOTATE_FUNC_BEGIN;
+   if (amg_data)
```
rfalgout, line 528: Is it always the case that if amg_data exists then all of the subcomponents do as well?  E.g., hypre_ParAMGDataLambda(amg_data)?

rfalgout (approved):
Looks good.  Thanks!

src/parcsr_ls/par_amg.c
victorapm: Not every AMG component (data structure member) will always exist but when they don't, they are set to `NULL`. E.g.,`hypre_ParAMGDataLambda(amg_data)` seems to be created only when using the Schwarz smoother option, otherwise it's `NULL`. 

Previously we had the check `if (hypre_ParAMGDataLambda(amg_data))` prior to the destroy call:
https://github.com/hypre-space/hypre/blob/922a3ce4df74d3b2b6593ff4d015531f147a8174/src/parcsr_ls/par_amg.c#L646-L649

but that check is not really necessary since `hypre_ParCSRMatrixDestroy` already does that internally, so that's why I removed some of the checks in `hypre_BoomerAMGDestroy`

rfalgout: Sounds good.  Thanks!

PaulMullowney:
@victorapm I'm delighted that you're doing this. ExaWind has an immediate use for this development. I have a few questions. 1) Do Krylov algorithms like GMRES work with the new multivectors? 2) It looks like csr_spmv_device.c was modified in another PR. Would supporting relaxation cases 11/12 on device be straightforward? We need a could of vector algorithms, but I think 
those should be easy. 3) Is the MPI packed into a single communication across all vector components in order to minimize the latency costs?

victorapm:
@PaulMullowney, glad to hear you are interested in this :)

Answering your questions:

1. Hypre's Krylov solvers do not support multivectors for now, but a lot of work needed for accomplishing that has been done in this series of PRs. We are not currently planning to extend multivector support to the Krylov solvers, but if ExaWind is interested, I would love to chat more about this.
2. It seems straightforward to me, The additional work would be extending `hypreDevice_DiagScaleVector[2]` to work with multivectors. As a matter of fact, the other project I'm working on might be interested in this. I'll consult them :)
3. Yes! That will come in the last PR of this series, which by the way I'm working right now :)

We can setup a call to chat more if you like.
Thanks!

ulrikeyang (approved):
This looks good.

PaulMullowney:
> @PaulMullowney, glad to hear you are interested in this :)
> 
> Answering your questions:
> 
> 1. Hypre's Krylov solvers do not support multivectors for now, but a lot of work needed for accomplishing that has been done in this series of PRs. We are not currently planning to extend multivector support to the Krylov solvers, but if ExaWind is interested, I would love to chat more about this.
> 2. It seems straightforward to me, The additional work would be extending `hypreDevice_DiagScaleVector[2]` to work with multivectors. As a matter of fact, the other project I'm working on might be interested in this. I'll consult them :)
> 3. Yes! That will come in the last PR of this series, which by the way I'm working right now :)
> 
> We can setup a call to chat more if you like. Thanks!

Hi Victor,
Let's talk after your PRs get merged.
-Paul

prj-:
Maybe this is not appropriate to ask here, but since there is already some discussion going on in this MR, here goes. I'd like to extend PETSc interface which already handles (pseudo-)block Krylov method so that it supports your multi-vectors. Would you by any chance have a high-level example so that I don't have to go through all MR to figure out how things are working, please?

victorapm:
Hi Pierre, 

the multivector development in hypre is not finished and I still need to add regression tests and an example for using that functionality. I will let you know once that's ready. What do you mean by (pseudo)-block methods? Can you point to the code region for that? In a block Krylov solver framework, do you need the ability to control which vector components are updated (e.g., in a multivector with 6 components, update only 2nd and 4th)? 

Thank you,
Victor

src/parcsr_ls/par_mgr_solve.c
```diff
@@ -787,7 +797,7 @@ hypre_MGRCycle( void               *mgr_vdata,
                hypre_ParCSRMatrixMatvecOutOfPlace(-1.0, A_array[level],
                                                   U_array[level], 1.0, F_array[level], Vtemp);
 
-               resnorm = hypre_ParVectorInnerProd(Vtemp, Vtemp);
+               resnorm = sqrt(hypre_ParVectorInnerProd(Vtemp, Vtemp));
```
liruipeng, line 800: @oseikuffuor1 This seems to be a bug before the fix. Do you agree? Thanks @victorapm

liruipeng (approved):
Looks good to me.

prj-:
Hello Victor,

> What do you mean by (pseudo)-block methods?

By (pseudo-)block Krylov methods, I mean block CG, block GMRES, their pseudo-block variant, and the likes. If you want to know more about this, here is a recent paper https://www.sciencedirect.com/science/article/abs/pii/S0898122121000055 on how it's done in PETSc and SLEPc (only section 2 is relevant).

> Can you point to the code region for that?

Here is a proof of concept: https://gitlab.com/petsc/petsc/-/commit/c56fff6745722f03374e954c0907002c0e1fe507. Once I'm aware of the API, I'll try to connect the dots.

> In a block Krylov solver framework, do you need the ability to control which vector components are updated (e.g., in a multivector with 6 components, update only 2nd and 4th)?

For block Krylov, you always iterate on the complete block. The size of the block may decrease throughout the restarts or the iterations if you do deflation (QR factorization of the block residual), see, e.g., section 8 of https://people.math.ethz.ch/~mhg/pub/delhipap.pdf. For pseudo-block variants, some systems may converge faster than others, so it could be worth only updating some components. However, IMHO, it requires too much book keeping, and you need to permute components back and forth if you want to keep everything contiguous to apply high-level BLAS or sparse BLAS operations. In SLEPc, for LOBPCG or CISS, there is the notion of [BVSetActiveColumns()](https://slepc.upv.es/documentation/current/docs/manualpages/BV/BVSetActiveColumns.html) which does this book keeping (it is much more important numerically here to filter out unwanted columns). In the first paper I'm mentioning, we used GAMG which did not handle (what you call) multivectors at that time inside LOBPCG (Stefano recently added this functionality so the figure should look much less dire now). I could use BoomerAMG as well in the eigensolver and show the performance improvement before your series of MR and after. That should be an easy >2x speedup.

victorapm:
Hi Pierre,

thanks for the info and references, those are very helpful. I'll contact you via e-mail so we can chat more.

oseikuffuor1 (approved):
Thanks Victor!!

src/parcsr_ls/par_mgr_solve.c
oseikuffuor1: Yes, subtle but important fix. Thanks @victorapm.

PR #710, opened 2022-08-04
Include missing header file for the CUDA install

ruohai0925:
We find the _hypre_utilities.hpp is not copied into the install directory when we build with CUDA support.


osborn9 (approved):
Thank you for the fix!

PR #716, opened 2022-08-15
Sycl matmat

waynemitchell:
Port of our custom spmv and spgemm routines to sycl. Note that this also involves significant updates to basic handling of kernel launches in sycl due to the need to support multi-dimensional kernels and the use of local shared memory.


src/struct_mv/_hypre_struct_mv.hpp
```diff
@@ -1395,7 +1395,7 @@ else                                                            \
 #define hypre_newBoxLoop0Begin(ndim, loop_size)                                                       \
 {                                                                                                     \
    hypre_newBoxLoopInit(ndim, loop_size);                                                             \
-   BoxLoopforall(hypre__tot, [=] (sycl::nd_item<1> item)                                              \
+   BoxLoopforall(hypre__tot, [=] (sycl::nd_item<3> item)                                              \
```
rfalgout, line 1398: Remind me what the 3 signifies again.

src/test/TEST_ij/amgdd.saved.florentia
```diff
@@ -1,64 +1,64 @@
...
 
 # Output file: amgdd.out.915
 GMRES Iterations = 5
-Final GMRES Relative Residual Norm = 7.111268e-09
+Final GMRES Relative Residual Norm = 7.894937e-09
 
 # Output file: amgdd.out.916
-GMRES Iterations = 21
-Final GMRES Relative Residual Norm = 1.801927e-03
+GMRES Iterations = 15
```
rfalgout, line 54: This is a pretty significant difference.  Are you comfortable that everything is still okay here?

rfalgout (approved):
This looks like it was a lot of work @waynemitchell !  It looks good to me, though I can't really speak to the details.  In addition to the questions I already asked, it looks like some of the changes might affect GPU code more generally.  If true, how much performance testing did you do here?  Thanks!

src/test/TEST_ij/amgdd.saved.florentia
waynemitchell: Yes, I think this is fine. I tested the cuda build on lassen and saw similarly significant differences when switching spgemm implementations (our custom vs. cusparse in that case). I was a bit surprised there were any differences at all when switching spgemm implementations, but there is a good explanation. Different implementations give different column index orderings, and this can cause slightly different behavior down the line for some pieces of AMG, notably coarsening. I've observed slightly different AMG hierarchies generated when doing the same run and just switching spgemm implementations. Note that AMG-DD convergence is particularly sensitive to the selection of coarse grids, thus the large difference here.

rfalgout: Sounds good.  Thanks!

src/struct_mv/_hypre_struct_mv.hpp
waynemitchell: It signifies a 3 dimensional kernel launch. In most cases throughout hypre, we just use 1D thread layouts, but in some of the spgemm implementation, we use higher dimensional layouts. This PR updates all the sycl kernel launches throughout hypre to be 3D in order to provide the same level of flexibility available in cuda/hip and to support the spgemm code.

waynemitchell:
> This looks like it was a lot of work @waynemitchell ! It looks good to me, though I can't really speak to the details. In addition to the questions I already asked, it looks like some of the changes might affect GPU code more generally. If true, how much performance testing did you do here? Thanks!

I ran the regressions on lassen and did see some deviation in the benchmark tests, but this went away on a subsequent run. I also saw that our regularly scheduled lassen regression tests had some reported deviations in the benchmark tests, so maybe there is either something going on with lassen or with the master branch of hypre. I will definitely investigate more and coordinate with @liruipeng to make sure this PR is not affecting performance before I merge.

src/seq_mv/csr_spgemm_device.h
```diff
@@ -22,13 +22,19 @@ constexpr HYPRE_Int SYMBL_HASH_SIZE[11] = { 0, 32, 64, 128, 256, 512, 1024, 2048
 constexpr HYPRE_Int NUMER_HASH_SIZE[11] = { 0, 16, 32,  64, 128, 256,  512, 1024, 2048, 4096,  8192 };
 #elif defined(HYPRE_USING_HIP)
 constexpr HYPRE_Int NUMER_HASH_SIZE[11] = { 0,  8, 16,  32,  64, 128,  256,  512, 1024, 2048,  4096 };
+/* WM: todo - what should this be for intel? */
```
liruipeng, line 25: Probably need to tune this with the information of Intel GPU shared memory size, etc.

src/seq_mv/csr_spgemm_device.h
```diff
@@ -181,43 +217,55 @@ T group_reduce_sum(hypre_DeviceItem &item, T in, volatile T *s_WarpData)
...
    const HYPRE_Int warp_lane_id = hypre_gpu_get_lane_id<2>(item);
+#endif
    const HYPRE_Int warp_id = hypre_gpu_get_warp_id<3>(item);
 
    if (warp_lane_id == 0)
    {
       s_WarpData[warp_id] = out;
    }
 
+#if defined(HYPRE_USING_SYCL)
```
liruipeng: Maybe we can unify this `#if #else` in a function call like `gpu_block_sync`.

src/struct_mv/_hypre_struct_mv.hpp
```diff
@@ -1242,12 +1242,12 @@ extern "C++"
       {
          return;
       }
-      const sycl::range<1> bDim = hypre_GetDefaultDeviceBlockDimension();
```
liruipeng, line 1245: Why not using `hypre_dim3`?

src/seq_mv/csr_spgemm_device.h
waynemitchell: Yes, I think there are a couple places where parameters need to be optimized for Intel hardware. But I plan to leave performance optimization for a future PR.

src/struct_mv/_hypre_struct_mv.hpp
waynemitchell: I'm not sure I understand the question. `hypre_dim3` is a simple wrapper function used for declaring a specific desired dim3 variable. But here, we use `hypre_GetDefaultDeviceBlockDimension()`, which will return the dim3 value we want. Note that `dim3` is still the type name, not `hypre_dim3`. Does that clarify this?

src/seq_mv/csr_spgemm_device.h
waynemitchell: Sure, that sounds good. I think I didn't bother with this at first because there aren't many use cases throughout the code. But I think it's a fine thing to add.

src/parcsr_ls/par_interp_trunc_device.c
```diff
@@ -229,7 +229,11 @@ hypreGPUKernel_InterpTruncationPass1_v1( hypre_DeviceItem &item,
    /* 2. save the largest max_elmts entries in sh_val/pos */
    const HYPRE_Int nt = hypre_gpu_get_num_threads<1>(item);
    const HYPRE_Int tid = hypre_gpu_get_thread_id<1>(item);
+#if defined(HYPRE_USING_SYCL)
+   HYPRE_Int *shared_mem = (HYPRE_Int*) shmem_ptr;
```
liruipeng, line 233: With SYCL, we can only use "dynamic" shared memory, right? I remembered seeing it somewhere. Just to learn this. Thanks!

src/seq_mv/csr_spgemm_device.h
```diff
@@ -48,27 +58,41 @@ constexpr HYPRE_Int T_GROUP_SIZE[11]    = { 0,  2,  4,   8,  16,  32,   64,  128
  * these are under the assumptions made in spgemm on block sizes: only use in csr_spgemm routines
  * where we assume CUDA block is 3D and blockDim.x * blockDim.y = GROUP_SIZE which is a multiple
  * of HYPRE_WARP_SIZE
```
liruipeng: My fault, we should have updated this comment: GROUP_SIZE is not necessarily a multiple of WARP_SIZE anymore. Maybe just delete "which is ...".

src/struct_mv/_hypre_struct_mv.hpp
liruipeng: Yes, I misread, sorry.

liruipeng (approved):
This is a lot of work. I am interested in knowing if the changes made for Sycl have impact on the performance of CUDA/HIP implementations. Other than that, everything looks good to me. Thanks @waynemitchell !

src/parcsr_ls/par_interp_trunc_device.c
waynemitchell: Yes, as far as I know, there is only dynamic shared memory. The only way I know to declare shared memory is through the use of a sycl accessor (this is a special sycl concept that I otherwise don't much use) with the sycl::target::local option. You can see the way this is done in the HYPRE_GPU_LAUNCH2 macro.

src/seq_mv/csr_spgemm_device.h
waynemitchell: I went ahead and fixed this comment.

PR #719, opened 2022-08-18
MGR - HIP support

victorapm:
This PR adds HIP support to MGR. Additionally:

1. Fixes a bug in the computation of `P_FF` on MGR when using GPUs.
2. Extends some NVTX regions to ROCTX


oseikuffuor1 (approved):
Thanks Victor!

rfalgout (approved):
Looks good.  Thanks!

src/parcsr_ls/par_amg_setup.c
```diff
@@ -226,7 +226,7 @@ hypre_BoomerAMGSetup( void               *amg_vdata,
    HYPRE_Int       dslu_threshold = hypre_ParAMGDataDSLUThreshold(amg_data);
 #endif
 
-#if defined(HYPRE_USING_NVTX)
+#if defined (HYPRE_USING_GPU)
```
liruipeng: Why changing them to `HYPRE_USING_GPU`? We only wanted them with NVTX turned on.

liruipeng (approved):
Looks good. I only had a minor question.

src/parcsr_ls/par_amg_setup.c
victorapm: To make it work for HIP. `hypre_GpuProfilingPushRange` is a no-op when using GPU but not `--with-gpu-profiling`: https://github.com/hypre-space/hypre/blob/614b2f181512f97bfae8f4722f16d0626dfbbeb6/src/utilities/nvtx.c#L89

Same for `hypre_GpuProfilingPopRange`, so I think this change (check `HYPRE_USING_GPU` instead of `HYPRE_USING_NVTX`) is fine

victorapm: As a matter of fact, I think it would be nice to embed these GPU profiling push/pop calls into the `HYPRE_ANNOTATE_` macros. It would simplify the code and hide the `#if defined (HYPRE_USING_GPU)` checks

liruipeng: How about `hypre_sprintf`? I don't see why you needed to make the change for HIP. Maybe I missed something.

src/parcsr_ls/par_amg_setup.c
```diff
@@ -3067,12 +3068,12 @@ hypre_BoomerAMGSetup( void               *amg_vdata,
...
       HYPRE_ANNOTATE_MGLEVEL_END(level);
-#if defined(HYPRE_USING_NVTX)
+#if defined (HYPRE_USING_GPU)
       hypre_GpuProfilingPopRange();
 #endif
       ++level;
       HYPRE_ANNOTATE_MGLEVEL_BEGIN(level);
-#if defined(HYPRE_USING_NVTX)
+#if defined(HYPRE_USING_GPU)
       hypre_sprintf(nvtx_name, "%s-%d", "AMG Level", level);
```
victorapm, line 3077: @liruipeng , do you mean this `sprintf`? If yes, note it's used in the `hypre_GpuProfilingPushRange` call below

src/parcsr_ls/par_amg_setup.c
victorapm: Continued discussion at https://github.com/hypre-space/hypre/pull/719#discussion_r1005015893

liruipeng: I meant that now when `HYPRE_USING_GPU` is on and GPU profiling is off, the `hypre_sprintf` will be called, which should not be (and was not under the #ifdef NVTX macro). I just don't understand why it's needed for HIP. What happens if you configure `--with-hip` with `--with-gpu-profiling`?

victorapm: If `--with-hip` and `--with-gpu-profiling` then `hypre_GpuProfilingPushRange` calls https://github.com/hypre-space/hypre/blob/614b2f181512f97bfae8f4722f16d0626dfbbeb6/src/utilities/nvtx.c#L107 same on master https://github.com/hypre-space/hypre/blob/6ca5cb542f6cff626a38034463c27c27436b98ed/src/utilities/nvtx.c#L109 Note that `roctxRangePush` expects the name of the region: https://github.com/ROCm-Developer-Tools/roctracer/blob/fb0c230b0bf4fd822d01aa9c2633d12f0e15029c/inc/roctx.h#L67

victorapm: > now when HYPRE_USING_GPU is on and GPU profiling is off

I see this now, I had read your original version. It doesn't seem to be a problem to have `nvtx_name` unused after `hypre_sprintf`. At least, this is not bothering the compiler (tested on debug mode with GCC). But, I'm fine if you prefer me to change `#if defined (HYPRE_USING_GPU)` to `#if defined (HYPRE_USING_NVTX) || defined(HYPRE_USING_ROCTX)`. I used the first to reduce code length :)

liruipeng: > > now when HYPRE_USING_GPU is on and GPU profiling is off
> 
> I see this now, I had read your original version. It doesn't seem to be a problem to have `nvtx_name` unused after `hypre_sprintf`. At least, this is not bothering the compiler (tested on debug mode with GCC). But, I'm fine if you prefer me to change `#if defined (HYPRE_USING_GPU)` to `#if defined (HYPRE_USING_NVTX) || defined(HYPRE_USING_ROCTX)`. I used the first to reduce code length :)

But we still need to pay the extra cost of the `hypre_sprintf` since `nvtx_name` is not used, right? Yes, I do prefer `#if defined (HYPRE_USING_NVTX) || defined(HYPRE_USING_ROCTX)`. Thanks!

victorapm: Sounds good! I made the change

victorapm:
Regression tests are clean and I'm merging this into master. Thanks, everyone!

Issue #513, opened 2021-11-02
Rename version to version.sh for compilation with c++20


Azrael3000:
We did a compile test recently of our software that includes hypre and if we enabled c++20 it failed because another software package uses  `#include <version>` which is part of the new standard ([0]). Since you have a `version` file [1] this caused the compiler to read the wrong file and fail. It is suggested that you rename the `version` file to `version.sh` for example to solve this issue.

[0] https://en.cppreference.com/w/cpp/header/version
[1] https://github.com/hypre-space/hypre/blob/master/src/utilities/version

rfalgout:
This file isn't part of the hypre installation, so I don't understand why there is a conflict.  You should only have the `include` directory for the hypre installation in your include path when compiling.  Also, you want the angle brackets in `#include <version>` to check system directories first, so you should ensure that you haven't changed this behavior by defining an include path environment variable or something.

liruipeng:
It seems to be related to #506

rfalgout:
Looks like there were similar issues here: https://github.com/spack/spack/issues/11515.  In particular, notice this comment:

  https://github.com/spack/spack/issues/11515#issuecomment-495626063

Finally, it looks like there is a legitimate issue in OpenMPI that is being fixed:

  https://github.com/open-mpi/ompi/issues/9122

From what I see so far, though, this doesn't look like an issue that hypre should try to address, since we don't include "version" (with any case-insensitive spelling) anywhere in our own source files, nor do we install a file of this type for users to accidentally include.

aaigner:
Maybe it depends on the way we build our program:
We reference hypre as a git submodule and add it via cmake `add_subdirectory`. Then we simply link the HYPRE target with our target.
The *utilities* directory that contains the *version* shell script is added to the hypre-target public include directories [here](https://github.com/hypre-space/hypre/blob/7f2762cffbfe0ec3bd2b718ce8baad929874a09f/src/CMakeLists.txt#L383)

I think it is also the combination of compiler and project combination. The used compiler was a non C++20 compiler, thus it has no std-header "version". Therefore, it picks what ever it finds in the include path. Something similar was also discussed in the [boost issue](https://github.com/boostorg/config/issues/345). (which is the source of all evil and unfortunately we can't avoid it)

and... the whole compilation is on Windows which is case-insensitive by the way.

rfalgout:
[...]

PR #725, opened 2022-08-30
Move utilities/version to utilities/version.sh

### LINKED ISSUE #723: `version` file in utilities clashes with c++ `<version>`
Newer C++ standards have a `<version>` header. Thus the `version` (bash) file in `utilities` will now collide with this. In particular, whenever dependent libraries (in this case `thrust`) `#include <version>`, either explicitly or implicitly, they are picking up the `version` file in `utilities` instead. The simplest fix is probably to rename `version` to `version.sh`, but it wasn't clear to me if the file was still used at all or not. Pinging @liruipeng @ulrikeyang @rfalgout

pbauman:
This avoid an #include issue with newer C++ compilers.

Fixes #723. It might be possible to just remove this file? I'm not entirely sure, but moving to .sh didn't break anything AFAICT. This fix is needed for forthcoming ROCm versions (and I imagine CUDA will need it at some point too as they start to support C++20).


ulrikeyang (approved):
This sounds like a good solution to this issue, for which we already had many complaints, to me.

rfalgout:
Hi @pbauman .  Please give issue #513 a quick read to get my take on this.  The entire software community seems to be rolling over and going with the "easy fix", but I think this sets a bad precedent.  Compiler standards have reserved words, but they should not be reserving file names.  This file is used internally by us.  It is not installed anywhere for others to potentially clash with (so we don't prefix it with `hypre_` for example).

Are you having trouble compiling hypre itself?

At some point I may roll over as well, maybe even now, but the whole thing really rankles me.

We would have to change at least one additional file in the `docs` directory.  Also, I think some of our users (e.g., MFEM) may use this script as part of their installation process (to get version information).

pbauman:
Ah, sorry I missed the other issues.

> Are you having trouble compiling hypre itself?

Yes. Compilation of HYPRE itself is failing internally with future ROCm builds. I'll elaborate with details in the issue.

pbauman:
Hi @rfalgout, I went ahead and added to the discussion in #513. TL;DR - I don't see how to avoid this without renaming the `version` file in HYPRE to something other than `version`. This failure will happen with a future version of ROCm (or the current master of rocPRIM/rocThrust).

rfalgout:
We will need to duplicate this on a branch in hypre and make additional changes.  Also, I need to check with the MFEM team and potentially some other users in case they are using this script in their build systems.

liruipeng:
> We will need to duplicate this on a branch in hypre and make additional changes. Also, I need to check with the MFEM team and potentially some other users in case they are using this script in their build systems.

I created another PR from a branch that has merged in this one. https://github.com/hypre-space/hypre/pull/731

PR #726, opened 2022-08-30
Offd num cols fix

liruipeng:
Modified from #718, this PR squashes out zero columns of `offd` of parcsr matrices.

The issue was in `offd` there exist empty columns (columns with no nonzeros), which correspond to "useless" entries in `col_map_offd`. This caused issues in at coarser grids in the communications with large number of ranks. We added a routine to compress the zero columns out and shorten `col_map_offd`. This should reduce communication cost even at higher levels.

Two sources of the empty columns have been located and fixed:
1. Truncation after building P
2. $P^T(AP)$: only the transpose multiplication part

- [x] tux
- [x] lassen
- [x] tioga
- [x] jlse


liruipeng:
There were small changes the regression test results, which were unexpected. I am still figuring out why this PR changes results. Please let me know if you have ideas.

ulrikeyang (approved):
This looks good. Thanks

liruipeng:
> This looks good. Thanks

Hi @ulrikeyang , I had to convert this to draft since there were result changes which I don't feel comfortable with even though they are small. I will ask your review later again if needed. Sorry for the confusion.

ulrikeyang:
Did you look at the setup stats? Since this changes the communication packages on coarse levels I would not be completely surprised by changes, but I agree that we should understand them.

liruipeng:
@ulrikeyang I did find small differences in the setup stats. As you can see in the PR, the parallel SPGEMM (with matrix transpose) results are changed and I know the reason, which is, as you expected, due to the communication packages. So, I can accept the fact that the regression results need to change.

liruipeng:
Hi @waynemitchell, Can you please help test this branch on jlse? Thanks!

waynemitchell (approved):
I added the relevant sycl implementation and tested this on JLSE. All looks good to me. Thanks!

rfalgout (approved):
Looks good.  Thanks!

PaulMullowney:
@liruipeng We are seeing problems on very large node counts (2350+) potentially in amg setup. I'm wondering if this PR is relevant to what we're seeing. What was the observed behavior prior to this fix; crashing code, wrong results, ... ? Is this a host, device, issue or both?

pbauman:
@PaulMullowney This PR addresses a performance issue rather than a "correctness" issue. I do not believe this is related to what you're referring to.

PR #727, opened 2022-08-31
new benchmark results on tioga

liruipeng:
The performance on tioga got significantly improved with new rocm.


ulrikeyang (approved):
Nice improvements!

PR #731, opened 2022-09-07
Fix file `version`

### LINKED ISSUE #723: `version` file in utilities clashes with c++ `<version>`

liruipeng:
copied from #725 by @pbauman 

> This avoid an #include issue with newer C++ compilers.
> Fixes https://github.com/hypre-space/hypre/issues/723. It might be possible to just remove this file? I'm not entirely sure, but moving to .sh didn't break anything AFAICT. This fix is needed for forthcoming ROCm versions (and I imagine CUDA will need it at some point too as they start to support C++20).


ulrikeyang (approved):
I approved this before.

rfalgout:
Hi @tzanio .  Does MFEM use the script `src/utilities/version` in any way?  We are changing the name to `version.sh`.  Thanks!

tzanio:
Nope, we only use the installed version of hypre.

rfalgout (approved):
I could not find any other use of this script in hypre, so this should be fine.  Thanks!

pbauman (approved):
Looks good, thanks!

PR #736, opened 2022-09-20
Multivec Krylov 2 Stage GS Relaxation

PaulMullowney:
This PR enables the 2stage GS relaxation to work properly with multivectors.

### DIFF
--- a/src/utilities/device_utils.c
+++ b/src/utilities/device_utils.c
@@ -2097,8 +2097,8 @@ __global__ void
 hypreGPUKernel_DiagScaleVector( hypre_DeviceItem &item,
                                 HYPRE_Int         num_vectors,
                                 HYPRE_Int         num_rows,
-                                HYPRE_Int        *A_i,
-                                HYPRE_Complex    *A_data,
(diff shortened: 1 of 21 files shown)


src/utilities/device_utils.c
```diff
@@ -2085,28 +2088,44 @@ template HYPRE_Int hypreDevice_CsrRowPtrsToIndicesWithRowNum( HYPRE_Int     nrow
...
-   if (i < n)
+   if (i < num_rows)
    {
+      val = 1.0 / A_data[A_i[i]];
+
       if (beta != 0.0)
       {
-         y[i] = x[i] / A_data[A_i[i]] + beta * y[i];
+#pragma unroll NV
+         for (j = 0; j < num_vectors; ++j)
```
victorapm, line 2117: @PaulMullowney, very minor question: is there a specific reason to use `++j` over `j++`? We prefer the latter in terms of style.

PaulMullowney: I think that's just an old habit of mine. You can change it back to j++

PR #741, opened 2022-09-26
Runtime switch and memory tracker

liruipeng:
This PR adds "runtime switch" feature to the solvers in hypre (`ilu` solvers do not have this feature yet). Using `HYPRE_SetMemoryLocation` and `HYPRE_SetExecutionPolicy`, users can choose to build solvers and solve on host or device with a GPU build of hypre.

A by product is a memory tracker in hypre, which helps developing this feature, that can track all the memory allocations and deallocations. Enabled with `--with-memory-tracker`, hypre drivers can dump a csv file that contains all the memory operations, and can detect memory leak if any (see attached sample file for an example).
[HypreMemoryTrack.log.00000.csv](https://github.com/hypre-space/hypre/files/9717352/HypreMemoryTrack.log.00000.csv)

Regressions:
- [x] tux
- [x] lassen
- [x] tioga
- [x] jlse


src/seq_mv/vector.c
```diff
@@ -903,52 +871,85 @@ hypre_SeqVectorInnerProd( hypre_Vector *x,
...
-   hypre_MemPrefetch(x_data, sizeof(HYPRE_Complex)*size, memory_location);
+#ifdef HYPRE_PROFILE
+   hypre_profile_times[HYPRE_TIMER_ID_BLAS1] += hypre_MPI_Wtime();
 #endif
 
-   return hypre_error_flag;
+   return sum;
 }
 
+
```
victorapm, line 935: Should we delete the code below?

src/sstruct_ls/HYPRE_sstruct_split.c
```diff
@@ -185,10 +185,17 @@ HYPRE_SStructSplitSetup( HYPRE_SStructSolver solver,
    nparts = hypre_SStructMatrixNParts(A);
    nvars = hypre_TAlloc(HYPRE_Int,  nparts, HYPRE_MEMORY_HOST);
    smatvec_data    = hypre_TAlloc(void ***,  nparts, HYPRE_MEMORY_HOST);
-   ssolver_solve   = (HYPRE_Int (***)()) hypre_MAlloc((sizeof(HYPRE_Int (**)()) * nparts),
-                                                      HYPRE_MEMORY_HOST);
-   ssolver_destroy = (HYPRE_Int (***)()) hypre_MAlloc((sizeof(HYPRE_Int (**)()) * nparts),
-                                                      HYPRE_MEMORY_HOST);
+
+   // RL: TODO TAlloc?
+   ssolver_solve   = (HYPRE_Int (***)()) hypre_MAlloc((sizeof(HYPRE_Int (**)()) * nparts), HYPRE_MEMORY_HOST);
```
victorapm, line 190: What is the reason for moving from `hypre_MAlloc` to `hypre_TAlloc` in some places? Just curious

src/struct_ls/HYPRE_struct_int.c
```diff
@@ -35,55 +36,36 @@ hypre_StructVectorSetRandomValues( hypre_StructVector *vector,
    hypre_SetIndex3(unit_stride, 1, 1, 1);
 
    boxes = hypre_StructGridBoxes(hypre_StructVectorGrid(vector));
+
+   if (hypre_GetExecPolicy1(memory_location) == HYPRE_EXEC_DEVICE)
```
victorapm, line 40: Is there a reason to not use cuRAND/rocRAND? Just curious, this question is out of scope for this PR

src/parcsr_ls/par_cgc_coarsen.c
```diff
@@ -14,6 +14,7 @@
 
 #include "_hypre_parcsr_ls.h"
 #include "../HYPRE.h" /* BM Aug 15, 2006 */
+#include "_hypre_IJ_mv.h"
```
victorapm, line 17: Maybe this can be removed?

victorapm: Nevermind, I see `hypre_IJMatrixMemoryLocation(ijmatrix)` now

src/sstruct_ls/HYPRE_sstruct_split.c
liruipeng: Since memory tracker can only track the `hypre_T` macros.

liruipeng: I am not familiar with the `HYPRE_Int (**)()` and `HYPRE_Int (***)()` syntax. Is this some C style? Because of it, I don't know how to write these mallocs with `TAlloc`.

src/struct_ls/HYPRE_struct_int.c
liruipeng: It is just to have exactly the same result on GPUs as CPUs. We thought it would be important to have identical results when developed GPU structured solvers. Maybe we can add an option to use curand later. Good point!

src/test/TEST_sstruct/maxwell.saved.lassen
```diff
@@ -1,12 +1,12 @@
 # Output file: maxwell.out.0
 Iterations = 10
-Final Relative Residual Norm = 2.480525e-09
+Final Relative Residual Norm = 2.480102e-09
```
liruipeng, line 4: the result changes are expected, since we move more computation of maxwell solvers to GPUs.

src/test/struct.c
```diff
@@ -20,7 +20,6 @@
 
 /* RDF: Why is this include here? */
 #include "_hypre_struct_mv.h"
```
liruipeng, line 22: Re "RDF comment" : because we need stuff like `hypre_BoxArrayBox` etc

src/test/struct.c
```diff
@@ -175,7 +178,17 @@ main( hypre_int argc,
    HYPRE_MatvecFunctions matvec_fn;
    /* end lobpcg */
 
+   /* default execution policy and memory space */
+#if defined(HYPRE_TEST_USING_HOST)
```
victorapm, line 182: @liruipeng Are you planning to implement in this PR the config file for setting additional driver options?

victorapm (approved):
Thanks, Rui Peng! These are very nice changes and a lot of effort. The new memory tracker capabilities are very cool.

src/multivector/seq_multivector.c
```diff
@@ -26,7 +26,7 @@ hypre_SeqMultivectorCreate( HYPRE_Int size, HYPRE_Int num_vectors  )
 {
    hypre_Multivector *mvector;
 
-   mvector = (hypre_Multivector *) hypre_MAlloc(sizeof(hypre_Multivector), HYPRE_MEMORY_HOST);
+   mvector = hypre_TAlloc(hypre_Multivector, 1, HYPRE_MEMORY_HOST);
```
rfalgout, line 29: Why are the TAlloc's needed here?

liruipeng: We want to use `hypre_T*` macros, so we can track all the memory operations using the tracker.

src/test/struct.c
liruipeng: @victorapm @rfalgout I like the idea of having a runtime config file. I have been thinking about what we want to have in the file and the file format. But I am not sure if we still have the time to do it for the release. I guess it's not of very high priority since it's just for the testings.

src/multivector/seq_multivector.c
liruipeng: If we find the tracker useful enough, we should always use `hypre_T*` macros to do the memory operations within hypre. There are a few places I mentioned in this PR that I don't know how to use the macros because of some C syntax, where I manually add the tracker insertions.

src/test/struct.c
rfalgout: This is not necessary for the release.  Thanks!

src/multivector/seq_multivector.c
rfalgout: Nevermind.  I confused myself about what TAlloc does, thinking that it was CTAlloc.  I don't think we should ever use Malloc directly anywhere in the code if we can help it.

src/parcsr_mv/par_csr_matrix.h
```diff
@@ -127,16 +127,31 @@ typedef struct hypre_ParCSRMatrix_struct
...
+   HYPRE_MemoryLocation memory_offd = offd ? hypre_CSRMatrixMemoryLocation(offd) : HYPRE_MEMORY_UNDEFINED;
+
+   if (diag && offd)
    {
-      hypre_printf("Warning: ParCSRMatrix Memory Location Diag (%d) != Offd (%d)\n", memory_diag,
-                   memory_offd);
-      hypre_assert(0);
+      if (memory_diag != memory_offd)
+      {
+         hypre_printf("Error: ParCSRMatrix Memory Location Diag (%d) != Offd (%d)\n", memory_diag, memory_offd);
```
rfalgout: Is this protected by a printlevel check?

src/multivector/seq_multivector.c
liruipeng: OK. Thanks! Btw, this the place where I don't know how to use TAlloc. https://github.com/hypre-space/hypre/pull/741/files#r988392174

src/sstruct_ls/HYPRE_sstruct_split.c
victorapm: Maybe it's for function pointers? I'm not that familiar either, but I guess we can take a look at this later

src/sstruct_mv/HYPRE_sstruct_grid.c
```diff
@@ -338,8 +338,9 @@ HYPRE_SStructGridSetFEMOrdering( HYPRE_SStructGrid  grid,
       }
    }
 
-   fem_vars    = hypre_TAlloc(HYPRE_Int,  fem_nvars, HYPRE_MEMORY_HOST);
-   fem_offsets = hypre_TAlloc(hypre_Index,  fem_nvars, HYPRE_MEMORY_HOST);
+   fem_vars    = hypre_TReAlloc(hypre_SStructGridFEMPVars(grid, part), HYPRE_Int, fem_nvars, HYPRE_MEMORY_HOST);
```
rfalgout, line 341: Why is this change needed?

src/sstruct_mv/sstruct_vector.c
```diff
@@ -843,3 +843,32 @@ hypre_SStructVectorClearGhostValues(hypre_SStructVector *vector)
 
    return hypre_error_flag;
 }
+
+HYPRE_MemoryLocation
+hypre_SStructVectorMemoryLocation(hypre_SStructVector *vector)
```
rfalgout, line 848: I would like to revisit the implementation of this after the release.

src/struct_mv/boxloop_dispatcher.h
```diff
@@ -0,0 +1,29 @@
...
+ * SPDX-License-Identifier: (Apache-2.0 OR MIT)
+ ******************************************************************************/
+
+#ifndef HYPRE_BOXLOOP_DISPATCHER_HEADER
+#define HYPRE_BOXLOOP_DISPATCHER_HEADER
+
+#define CONCAT2(x, y) x ## y
+#define XCONCAT2(x, y) CONCAT2(x, y)
+
+#define HYPRE_BOXLOOP(BEGIN, BEGINARGS, BODY, END, ENDARGS)                                  \
```
rfalgout: This is still just a draft idea, right?  Should we remove it from the PR?

src/test/Makefile
```diff
@@ -114,7 +114,7 @@ all++: ${HYPRE_DRIVER_CXX_EXECS}
 install:
 
 clean:
-	rm -f *.o *.obj
+	rm -f *.o *.obj *.csv
```
rfalgout, line 117: Where do the 'csv' files come from?

liruipeng: from the tracker's output.

src/struct_mv/boxloop_dispatcher.h
liruipeng: Yes, we should. Thanks!

src/sstruct_mv/sstruct_vector.c
liruipeng: Yes, I know we've talked about this. I think you suggested only `type == HYPRE_SSTRUCT` would be actually needed

src/sstruct_mv/HYPRE_sstruct_grid.c
liruipeng: There were memory leaks detected. I think ReAlloc can be a fix.

src/parcsr_mv/par_csr_matrix.h
liruipeng: Thanks for pointing it out. I should have added print_level guard.

src/test/Makefile
rfalgout: Ahh, okay.  Thanks!

src/sstruct_ls/HYPRE_sstruct_split.c
rfalgout: Function pointer syntax is strange, but we can definitely save this for later.  Thanks!

src/sstruct_mv/sstruct_matrix.c
```diff
@@ -1125,78 +1127,99 @@ hypre_SStructUMatrixSetBoxValues( hypre_SStructMatrix *matrix,
...
+                     }
+
+                     ijvalues[ci] = values[ei + vi * nentries];
+                     ncols[mi]++;
                   }
+                  hypre_BoxLoop2End(mi, vi);
+#undef DEVICE_VAR
+#define DEVICE_VAR
+               }
+               else
```
rfalgout, line 1195: Can you summarize why this if/else is needed here but not in the `sstruct_vector.c` file?

rfalgout (approved):
Hi @liruipeng .  This was a lot of work and looks great!  I am approving the PR but will follow any changes you make before merging.  Nice work!

src/struct_mv/struct_communication.c
```diff
@@ -762,6 +762,46 @@ hypre_CommTypeSetEntry( hypre_Box           *box,
...
+   }
+
+   return ptr;
+}
+
+HYPRE_Int
+hypre_StructCommunicationReleaseBuffer(HYPRE_Complex       *buffer,
+                                       HYPRE_MemoryLocation memory_location)
+{
+   if (hypre_GetActualMemLocation(memory_location) == hypre_MEMORY_HOST)
```
waynemitchell, line 797: Why are you only freeing this if it's host memory? Also, both here and above, why do you need to use hypre_GetActualMemLocation()?

src/sstruct_mv/sstruct_matrix.c
liruipeng: In the old code, we also had two code paths for GPU and CPU. We just runtime if-else now. The reason of the two code paths is the values of `hypre_Index rs, cs;` (`typedef HYPRE_Int  hypre_Index[HYPRE_MAXDIM];`) cannot be captured by the lambda, so we had to create `rs_0, ... rs_2, cs_0, ... cs_2` for them. It's not a nice way to do it, but I have better ideas. I don't think we have such issues in `sstruct_vector.c`.

src/struct_mv/struct_communication.c
liruipeng: These changes (GetBuffer and Release) are just for making the code cleaner, a little bit maybe. The logic is if it's CPU memory we allocate and deallocate every time, while for GPU memory, we allocate once and save it (reallocate 5x more if needed). So ReleaseBuffer is no-op for GPUs.

src/sstruct_mv/sstruct_matrix.c
rfalgout: I see that now.  The if/else doesn't quite line up with the previous #if/#else so I thought this was new.  Thanks again!

src/parcsr_mv/par_csr_matrix.h
liruipeng: Hi @rfalgout, `parcsr_matrix` doesn't have a `print_level`, so I just add `#if defined(HYPRE_DEBUG)` around the `hypre_printf`. Is it OK?

waynemitchell (approved):
I just pushed a couple easy fixes for the sycl build. Otherwise this looks good to me. Thanks @liruipeng for the huge amount of work here!

src/parcsr_mv/par_csr_matrix.h
rfalgout: Yes, that's okay.  Maybe we need a global runtime printlevel indicator that can be used in cases like this.  We can talk about that later though.  Thanks!

victorapm: Another approach would be to use `hypre_error_w_msg` like here: https://github.com/hypre-space/hypre/blob/ebc634a4e86d4b1aba97f61a8ba0aae701c9799b/src/seq_mv/csr_matrix.c#L177-L178

rfalgout: Good point.  Error messages should generally be managed through the error handling functions.

liruipeng: Correct! Thanks!

liruipeng:
Oct 7 Updates:
1. Fixed a few more leaks found by the tracker
2. Added new regression tests to lassen and tioga: build with GPU and run on CPU. The `.saved.lassen_cpu` results were generated with **CPU-only build** and compared against. No differences were found.

PR #742, opened 2022-09-28
HYPRE_USING_GPU and HYPRE_USING_DEVICE_OPENMP

waynemitchell:
This PR reorganizes the usage of the HYPRE_USING_GPU and HYPRE_USING_DEVICE_OPENMP macros. Up to now, enabling cuda, hip, sycl, or device openmp will also turn on the HYPRE_USING_GPU macro. The support for device openmp is much more limited than the other backends, however, with no current plans to expand that support. So in the interest of simplifying the usage of these macros and grouping the bulk of the GPU implementation under the single HYPRE_USING_GPU macro, this PR modifies the code such that using device openmp no longer enables this flag. Thus HYPRE_USING_GPU is now an umbrella over only the primary GPU backends used throughout the code: cuda, hip, and sycl.


src/struct_mv/struct_communication.c
```diff
@@ -843,17 +843,15 @@ hypre_InitializeCommunication( hypre_CommPkg     *comm_pkg,
...
    alloc_dev_buffer = (hypre_HandleStructExecPolicy(hypre_handle()) == HYPRE_EXEC_DEVICE);
 #elif defined(HYPRE_USING_DEVICE_OPENMP)
    alloc_dev_buffer = hypre__global_offload;
 #endif
-#endif
 
-#if defined(HYPRE_USING_GPU)
    if (alloc_dev_buffer)
    {
       send_buffers_device = hypre_TAlloc(HYPRE_Complex *, num_sends, HYPRE_MEMORY_HOST);
```
victorapm: This hasn't been touched in this PR, but I'm curious about this line. It seems we want to allocate memory for a device variable, but `HYPRE_MEMORY_HOST` is being passed. Shouldn't it be `HYPRE_MEMORY_DEVICE`? @liruipeng

src/utilities/_hypre_utilities.h
```diff
@@ -1308,6 +1308,11 @@ typedef struct
    HYPRE_MemoryLocation   memory_location;
    HYPRE_ExecutionPolicy  default_exec_policy;
    HYPRE_ExecutionPolicy  struct_exec_policy;
+   /* the device buffers needed to do MPI communication for struct comm */
```
victorapm, line 1483: What is the reason for moving these variables to `hypre_handle`? Thanks!

src/struct_mv/struct_communication.c
liruipeng: No. This is `HYPRE_Complex **`, i.e., pointer to pointers

victorapm: Oh yes! I missed that, thanks!

src/utilities/_hypre_utilities.h
waynemitchell: Moving these out of the hypre_DeviceData structure allows me to macro out hypre_DeviceData when using device openmp (i.e. all the hypre_DeviceData stuff is just behing HYPRE_USING_GPU and not DEVICE_OPENMP).

victorapm: Thanks for clarifying!

victorapm (approved):
This is a nice change! Thanks!!

src/utilities/protos.h
```diff
@@ -358,7 +358,6 @@ char *hypre_strcpy(char *destination, const char *source);
 HYPRE_Int hypre_SetSyncCudaCompute(HYPRE_Int action);
 HYPRE_Int hypre_RestoreSyncCudaCompute(void);
 HYPRE_Int hypre_GetSyncCudaCompute(HYPRE_Int *cuda_compute_stream_sync_ptr);
-HYPRE_Int hypre_SyncComputeStream(hypre_Handle *hypre_handle);
```
liruipeng, line 361: Why delete this proto?

src/seq_mv/csr_matvec_oomp.c
```diff
@@ -34,7 +34,7 @@ hypre_CSRMatrixMatvecOMPOffload( HYPRE_Int        trans,
 
    if (trans)
    {
-      hypre_CSRMatrixTransposeDevice(A, &B, 1);
+      hypre_CSRMatrixTranspose(A, &B, 1);
```
liruipeng, line 37: This will lead to use `hypre_CSRMatrixTransposeHost` I guess, which is fine since `hypre_CSRMatrixTransposeDevice` doesn't work with OMPOffload (no-op) either. We don't currently support this anyways.

src/utilities/protos.h
waynemitchell: This is redundant, I think. See line 272.

liruipeng: OK. I see. Thanks!

liruipeng (approved):
Looks good. Thanks!

rfalgout (approved):
Looks great, Wayne!  I think this cleans things up nicely!

PR #748, opened 2022-10-11
bugfix/tsuji1/dsuperlu

ptsuji:
- Update the FEI interface to SuperLU_Dist (use the same structs as src/parcsr_ls/dsuperlu.c)
- CMake support for building with the FEI and serial SuperLU
- Fix to prevent a floating point exception in hypre_BoomerAMGCGRelaxWt
- Fix to prevent the FEI solver from exiting with a convergence failure when the initial guess is the solution


ptsuji:
Any movement on reviewing this pull request @rfalgout @waynemitchell @osborn9? Is there anything I need to do before it can be reviewed?

rfalgout:
Hi Paul.  We spoke about it last week at our hypre meeting.  Sarah will look over the cmake changes.  I only had one question about a windows variable change that I'll ask separately.

src/matrix_matrix/HYPRE_BuildIJMatrixFromDistributedMatrix.c
```diff
@@ -16,9 +16,9 @@
 #endif
 
 #include <HYPRE_config.h>
-
+#ifndef WIN32
```
rfalgout: Why is this needed?

rfalgout (approved):
This looks good to me, modulo the question about the WIN32 ifdef.  Thanks!

src/matrix_matrix/HYPRE_BuildIJMatrixFromDistributedMatrix.c
ptsuji: Not sure, this was probably put in to get the Windows build with Visual Studio working. I can try removing it over on our side to see if it's still necessary.

rfalgout: My question would be, if `general.h` is not needed on the windows side, why is it needed at all?  Let me know what you find out.  Thanks!

ptsuji: I've removed the #ifndef, as it doesn't seem to be doing anything.

ptsuji:
@rfalgout thanks, I've added a few more changes that came from Tim Dunn, who found some issues when running incompressible flow problems in our code.

src/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ set(HYPRE_SRCDIR  "${PROJECT_SOURCE_DIR}")
 set(PROJECT_NAME HYPRE)
 project(${PROJECT_NAME}
   VERSION ${HYPRE_VERSION}
-  LANGUAGES C)
+  LANGUAGES C CXX)
```
osborn9:
```suggestion
  LANGUAGES C)
```
Instead, could we have this within the FEI block (so CXX isn't necessary unless required)

src/CMakeLists.txt
```diff
@@ -443,6 +478,24 @@ if (DSUPERLU_FOUND)
   set(HYPRE_USING_DSUPERLU TRUE)
 endif ()
 
+# Find FEI, if requested
+if (HYPRE_USING_FEI)
+  if (NOT TPL_FEI_INCLUDE_DIRS)
```
osborn9, line 486:
```suggestion
  enable_language(CXX)

  if (NOT TPL_FEI_INCLUDE_DIRS)
```

osborn9 (changes requested):
Thanks for the ping and the fix @ptsuji! This looks good to me (with the minor suggested fix for where the CXX language is specified).

src/CMakeLists.txt
ptsuji: Sounds good, I've removed this bit and added the enable_language(CXX) in the FEI part.

osborn9 (approved):

PR #758, opened 2022-10-18
Dev allocator

victorapm:
This PR updates `hypre_device_allocator` to use hypre's abstract memory model. This means that:

1. hypre configured **with** unified memory support: those allocations will be on managed memory.
2. hypre configured **without** unified memory support: those allocations will be on device memory


rfalgout (approved):
I don't know if there are any subtleties here to know about, but I'll leave that to Rui Peng and Wayne to comment on.  I am otherwise okay with this.  Thanks!

victorapm:
@rfalgout Thanks! Let me give some context: the idea is to allow for GPU memory over-subscription when hypre is configured with unified memory support. Previously, memory allocations from thrust, which uses `hypre_device_allocator`, were done directly on the device and this could cause out-of-memory issues. Now they can be done via managed memory, assuming hypre is configured with UVM, allowing memory over-subscription.

liruipeng (approved):
Looks good. Thanks!

waynemitchell (approved):
Looks good to me. Thanks, @victorapm!

PR #759, opened 2022-10-18
Add HYPRE_BRANCH_NAME

victorapm:
This PR extends #472 by adding a new variable for the name of the hypre's branch currently in use. Internal drivers have been updated to use this variable. Example with `ij` driver (first few lines):

```
Using HYPRE_DEVELOP_STRING: v2.26.0-4-gecb9a9ae2 (branch branch-name; not the develop branch)

  solver ID    = 0

Hypre init times:
```


src/test/TEST_ij/versioncheck.sh
```diff
@@ -19,7 +19,7 @@ if [ -d ../../../.git ]; then
   DEVBRANCH=`git rev-parse --abbrev-ref HEAD`
   if [ -n "$DEVSTRING" ]; then
     if [ "$DEVBRANCH" != "master" ]; then
-      echo "Using HYPRE_DEVELOP_STRING: $DEVSTRING (not main development branch)" \
+      echo "Using HYPRE_DEVELOP_STRING: $DEVSTRING (branch $DEVBRANCH)" \
```
rfalgout: Indicate in this message that this is not the main development branch.  You can still print the branch name.

src/test/ij.c
```diff
@@ -2391,15 +2391,19 @@ main( hypre_int argc,
...
-#ifdef HYPRE_DEVELOP_BRANCH
+#if defined(HYPRE_DEVELOP_STRING) && defined(HYPRE_DEVELOP_BRANCH)
       hypre_printf("\nUsing HYPRE_DEVELOP_STRING: %s (main development branch %s)\n\n",
                    HYPRE_DEVELOP_STRING, HYPRE_DEVELOP_BRANCH);
-#else
-      hypre_printf("\nUsing HYPRE_DEVELOP_STRING: %s (not main development branch)\n\n",
-                   HYPRE_DEVELOP_STRING);
-#endif
+
+#elif defined(HYPRE_DEVELOP_STRING) && !defined(HYPRE_DEVELOP_BRANCH)
```
rfalgout, line 2398: Same here.  Indicate this is not the main development branch.

src/test/TEST_ij/versioncheck.sh
victorapm: Done! Thanks!

src/test/ij.c
victorapm: Yes, thanks!

rfalgout:
Since we are printing the branch name every time now, maybe we should reverse the order of the information in the parenthesis to something like the following two options (trying to make it readable, but not too long):

    (branch $DEVBRANCH; the develop branch)
    (branch $DEVBRANCH; not the develop branch)

What do you think?

victorapm:
Yes, I agree with that. The format you suggested is easier to read. I'll do the update. Thanks!

rfalgout (approved):
Looks good to me.  Thanks!

PR #760, opened 2022-10-20
Minor fix to add HYPRE_BRANCH_NAME to CMake generated config file

osborn9:
This is a quick fix for the CMake build when not on the main dev branch.


victorapm (approved):
Thanks!

Issue #755, opened 2022-10-14
Cannot configure with recent Intel compilers


prj-:
It appears `configure` has been generated with a rather old `autoreconf` binary: https://github.com/hypre-space/hypre/blob/3c04212d77385869b7aa9a05fb07099c65a81b1b/src/configure#L4
It is causing grief with recent Intel compilers: https://gitlab.com/petsc/petsc/-/issues/1255#note_1135549645. Could `configure` but updated somehow and regenerated with a newer binary? Thanks.

rfalgout:
Hi @prj- .   There was too much going on in the petsc issue you pointed to above for me to see what's happening.  I would like to understand this before we try to update to a newer version of autoconf.  Could you provide a configure line with `config.log` etc. showing the problem?  Thanks!

prj-:
Sure, here goes (could probably be trimmed some more but I already removed a good amount of the PETSc-generated stuff).
```
$ export I_MPI_CC=icx   
$ export I_MPI_CXX=icpx
$ export I_MPI_F90=ifort
$ ./configure CC="mpiicc" CFLAGS="-fPIC -O3 -fopenmp" CXX="mpiicpc" CXXFLAGS="-O3 -std=gnu++17 -fPIC -fopenmp" F90="mpiifort" F90FLAGS="-fPIC -g -O3 -fopenmp" FFLAGS="-fPIC -g -O3 -fopenmp" FC="mpiifort" F77="mpiifort" FCFLAGS="-fPIC -g -O3 -fopenmp" --enable-shared --with-MPI-lib-dirs="" --with-MPI-libs=""
[...] // runs to completion, OK
$ export I_MPI_F90=ifx
$ ./configure CC="mpiicc" CFLAGS="-fPIC -O3 -fopenmp" CXX="mpiicpc" CXXFLAGS="-O3 -std=gnu++17 -fPIC -fopenmp" F90="mpiifort" F90FLAGS="-fPIC -g -O3 -fopenmp" FFLAGS="-fPIC -g -O3 -fopenmp" FC="mpiifort" F77="mpiifort" FCFLAGS="-fPIC -g -O3 -fopenmp" --enable-shared --with-MPI-lib-dirs="" --with-MPI-libs=""
[...]
checking for dummy main to link with Fortran libraries... unknown
configure: error: in `/local/jolivet/petsc/arch-intel-c-opt/externalpackages/git.hypre/src':
configure: error: linking to Fortran libraries from C fails
[...]

PR #766, opened 2022-10-31
Use $(MAKE) instead of (non-portable) make

rfalgout:
This is the same as PR #763.  Cf. #755 (comment).


rfalgout:
Hi @liruipeng and @waynemitchell .  Would you take a quick look at this one.  The fix is trivial and shouldn't have consequences anywhere since we use `$(MAKE)` already in our makefiles.  Thanks!

liruipeng (approved):
Looks good to me. Thanks!

waynemitchell (approved):
Looks fine to me. Thanks!

PR #778, opened 2022-11-18
SYCL regression testing

waynemitchell:
This PR has several bug fixes and small changes for the sycl build. The full suite of ij and struct regression tests now produce good outputs on the florentia nodes of JLSE. The machine-arcticus.sh autotest script (which was previously not passing) has been replaced with an updated machine-florentia.sh script. Running this script on a florentia node of JLSE is now sufficient for regression testing the SYCL build. The .saved files have also been updated accordingly (with name changes from .saved.arcticus to .saved.florentia).

Note that previously we also tested the device-openmp and kokkos/device-openmp builds on JLSE, but these are now broken, so those tests are commented out in machine-florentia.sh. Fixing these build on JLSE is left for a future PR.


src/test/TEST_ij/amgdd.saved.florentia
```diff
@@ -1,64 +1,64 @@
 # Output file: amgdd.out.900
-BoomerAMG-DD Iterations = 10
-Final Relative Residual Norm = 1.928307e-09
+BoomerAMG-DD Iterations = 9
```
rfalgout, line 2: Some of these differences are pretty substantial.  Do you have an explanation for them?

rfalgout (approved):
I didn't really follow any of the code changes, but nothing looked egregiously wrong either.  The fact that things like the '-g' option break the code is worrying, but maybe that's just the nature of working with sycl right now?  Thanks!

src/test/TEST_ij/amgdd.saved.florentia
waynemitchell: Most of the saved outputs are quite old. There have been several new parts of the code ported to the GPU (for the sycl backend), several bug fixes, and new oneAPI releases since the previous outputs were saved. I would not put much stock in the old saved outputs. I checked the new outputs against what we get with CPU and cuda builds, and the numbers look reasonable to me, so hopefully the new .saved files should be a good baseline going forward.

waynemitchell:
> I didn't really follow any of the code changes, but nothing looked egregiously wrong either. The fact that things like the '-g' option break the code is worrying, but maybe that's just the nature of working with sycl right now? Thanks!

The '-g' problem is a known issue with the current oneAPI that I have access to... I'm hoping this will be resolved soon, but until then there's not much I can do to circumvent it besides just leaving the -g option off for now.

src/utilities/device_utils.c
```diff
@@ -800,8 +800,11 @@ hypreDevice_CsrRowIndicesToPtrs_v2( HYPRE_Int  nrows,
                                     HYPRE_Int *d_row_ptr )
 {
 #if defined(HYPRE_USING_SYCL)
-   if (nnz <= 0 || nrows <= 0)
+   /* WM: if nnz <= 0, then dpl::lower_bound is a no-op, which means we still need to zero out the row pointer */
+   /* WM: question - does thrust lower_bound do this automatically? Otherwise, I'd expect this to break for cuda also. */
```
waynemitchell: @liruipeng , do you have an answer to this? I ran into an issue with the sycl build where hypreDevice_CsrRowIndicesToPtrs_v2() was called with nnz = 0 and nrows nonzero. In this case, the dpl lower_bound() call was a no-op, so d_row_ptr was left unchanged (with incorrect nonzero entries), leading to errors later on. What is the behavior of thrust's lower_bound in this case? If it is also a no-op, I would expect to need this check for nnz <= 0 for both cuda and sycl.

waynemitchell: OK, I tested this on lassen and verified that, for the cuda build, this function behaves correctly even when nnz = 0 and nrows is nonzero. That is, when nnz = 0, lower_bound sets the row pointer to 0 everywhere. This is NOT true for oneDPL's lower_bound, so for the sycl backend we need to explicitly take care of this case.

liruipeng: @waynemitchell Yes, you are right. I think `thrust::lower_bound` is consistent with C++ standard behavior (see, https://cplusplus.com/reference/algorithm/lower_bound/, look at the "The behavior of this function template is equivalent to:" part). That is, when `first == last`, it should return `first`, which is `0` here. But treating the special case (`nnz == 0`) separately as you did looks fine to me (even more clear).

liruipeng (approved):
Looks good. Thanks @waynemitchell

PR #779, opened 2022-11-22
Implement AddFEMBoxValues() routines

rfalgout:
This PR tracks the implementation of new AddFEMBoxValues() routines.

- [ ] Check regression tests
  - [x] machine-tux
  - [ ] machine-tux-spack
  - [ ] machine-rztopaz
  - [ ] machine-syrah
  - [x] machine-lassen


rfalgout:
I implemented the simplest thing I could think of to get this working and also get some added performance.  I haven't yet tested it, but will do that soon.

src/sstruct_mv/HYPRE_sstruct_matrix.c
```diff
@@ -617,6 +618,57 @@ HYPRE_SStructMatrixGetBoxValues2( HYPRE_SStructMatrix  matrix,
...
+   hypre_SStructGraph *graph        = hypre_SStructMatrixGraph(matrix);
+   hypre_SStructGrid  *grid         = hypre_SStructGraphGrid(graph);
+   HYPRE_Int           fem_nsparse  = hypre_SStructGraphFEMPNSparse(graph, part);
+   HYPRE_Int          *fem_sparse_i = hypre_SStructGraphFEMPSparseI(graph, part);
+   HYPRE_Int          *fem_entries  = hypre_SStructGraphFEMPEntries(graph, part);
+   HYPRE_Int          *fem_vars     = hypre_SStructGridFEMPVars(grid, part);
+   hypre_Index        *fem_offsets  = hypre_SStructGridFEMPOffsets(grid, part);
+   HYPRE_Complex      *tvalues;
+   hypre_Box          *box;
+   HYPRE_Int           s, i, d, vilower[3], viupper[3];
```
victorapm:
```suggestion
   HYPRE_Int           s, i, d, vilower[HYPRE_MAXDIM], viupper[HYPRE_MAXDIM];
```

src/sstruct_mv/HYPRE_sstruct_vector.c
```diff
@@ -557,6 +558,53 @@ HYPRE_SStructVectorGetBoxValues2(HYPRE_SStructVector  vector,
...
+                                   HYPRE_Complex       *values)
+{
+   HYPRE_Int           ndim         = hypre_SStructVectorNDim(vector);
+   hypre_SStructGrid  *grid         = hypre_SStructVectorGrid(vector);
+   HYPRE_Int           fem_nvars    = hypre_SStructGridFEMPNVars(grid, part);
+   HYPRE_Int          *fem_vars     = hypre_SStructGridFEMPVars(grid, part);
+   hypre_Index        *fem_offsets  = hypre_SStructGridFEMPOffsets(grid, part);
+   HYPRE_Complex      *tvalues;
+   hypre_Box          *box;
+   HYPRE_Int           i, d, vilower[3], viupper[3];
```
victorapm:
```suggestion
   HYPRE_Int           i, d, vilower[HYPRE_MAXDIM], viupper[HYPRE_MAXDIM];
```

src/sstruct_mv/HYPRE_sstruct_matrix.c
rfalgout: Done

src/sstruct_mv/HYPRE_sstruct_vector.c
rfalgout: Done.

rfalgout:
All of the regression tests that use the `sstruct` driver pass except for the following three:
````
TEST_sstruct/neighborpart.sh
TEST_sstruct/sharedpart.sh
TEST_fac/sstruct_fac.sh
````

The `neighborpart` and `sharedpart` differences are due to some rows of the final ParCSR matrix having different orderings of the column numbers (the matrix coefficients are the same bitwise).  This occurs along part boundaries due to the use of `SetNeighborPart` or `SetSharedPart` for some reason.  I wonder if we should try to produce matrices with sorted columns through our interfaces.  Comments?

Something more serious seems to go wrong in the `fac` test, though it may be for the same reason since FAC maps back and forth between matrix coefficients and the grid representation.  I have been thinking about dropping FAC anyway since I doubt it even works anymore.  Comments?

I was not able to do much with the GPU implementation, so it would be good to get some help.  I've noted a few places with `TODO` where things need to change.

**Comment by rfalgout** [comment]
FYI, I did a quick timing test on tux with the following run (compiler optimization turned on):

    sstruct -in TEST_sstruct/sstruct.in.bcube-fem -r 80 80 80 -solver 20

There is a speedup of about 8x for the `SStruct Interface` setup versus the code in master.

**Comment by rfalgout** [comment]
One more FYI...  The differences in column orderings have nothing to do with SetNeighborPart or SetSharedPart.  I added a simple cube test that can be run as follows (this generates a diagonally-dominant Laplacian-like operator on a 9x9 grid)

    sstruct -in TEST_sstruct/sstruct.in.cube-fem -r 4 4 4 -solver 20

Just add the `-print` option and diff the UMatrix with the one that comes from master.  The matrices are identical, but the column numbers are stored in a different order, so the solvers produce different results, especially AMG because it ends up coarsening differently.

**Comment by victorapm** [comment]
Hi @rfalgout @liruipeng , I've extended the new functions to work with device pointers. I'm seeing a speedup close to 150x wrt. master (pretty impressive) on a single GPU for `SStruct Interface` setup. The problem is
```
sstruct -in TEST_sstruct/sstruct.in.bcube-fem -r 20 20 20 -solver 20
```
BoomerAMG's Setup plus Solve is still 15x faster than the matrix assembly, I want to investigate this further. I believe we can get an even better speedup for assembly.

**Comment by rfalgout** [comment]
Awesome!  Thanks, @victorapm !

File: src/sstruct_mv/HYPRE_sstruct_matrix.c [hunk 3]
```diff
@@ -617,6 +618,76 @@ HYPRE_SStructMatrixGetBoxValues2( HYPRE_SStructMatrix  matrix,
...
+HYPRE_SStructMatrixAddFEMBoxValues(HYPRE_SStructMatrix  matrix,
+                                   HYPRE_Int            part,
+                                   HYPRE_Int           *ilower,
+                                   HYPRE_Int           *iupper,
+                                   HYPRE_Complex       *values)
+{
+   HYPRE_Int             ndim            = hypre_SStructMatrixNDim(matrix);
+   hypre_SStructGraph   *graph           = hypre_SStructMatrixGraph(matrix);
+   hypre_SStructGrid    *grid            = hypre_SStructGraphGrid(graph);
+   HYPRE_MemoryLocation  memory_location = hypre_IJMatrixMemoryLocation(
```
liruipeng: Can we use `hypre_SStructMatrixMemoryLocation`? Is it better to use?

victorapm: We can't because it returns `HYPRE_MEMORY_UNDEFINED` if called prior to the matrix assembly. However, I think it's more appropriate to do as you said.

To make that work, we would need to move
https://github.com/hypre-space/hypre/blob/467e3f3d9002fae4b71f46b07876884a23a8e7e0/src/sstruct_mv/sstruct_matrix.c#L1301-L1302 
from `hypre_SStructUMatrixAssemble` to after the line

https://github.com/hypre-space/hypre/blob/467e3f3d9002fae4b71f46b07876884a23a8e7e0/src/sstruct_mv/sstruct_matrix.c#L778
which is at `hypre_SStructUMatrixInitialize`. Then, `hypre_SStructMatrixMemoryLocation` would return the correct memory location when called prior to assembling the matrix. Let's do this change?

liruipeng: Got it. Good point. Not sure about the changes you suggested, which seem to be big. I am not as familiar with SStruct, so I will rely on the expert decisions from others. The current code looks fine to me though. Thank you!

victorapm: @rfalgout can you comment on my suggestion above? It seems safe to move the GetObject call to right after initialize. This change will allow using `HYPRE_SStructMatrixMemoryLocation` in the new BoxValues routines.

rfalgout: Yes, this is the right place to put it anyway I think.  Thanks!

victorapm: >It seems safe to move the GetObject call to right after initialize

Done in [5138246](https://github.com/hypre-space/hypre/pull/779/commits/5138246d287c1513edc75759d7c2cf111daa6d2e)

src/test/sstruct.c
```diff
@@ -3249,6 +3250,24 @@ main( hypre_int argc,
                }
             }
          }
+#else    // Use AddFEMBoxValues
+         /* TODO: There is probably a smarter way to do this copy */
+         for (i = 0; i < data.max_boxsize; i++)
+         {
+            j = i * data.fem_nsparse;
+            hypre_TMemcpy(&d_values[j], data.fem_values, HYPRE_Real,
```
victorapm, line 3258: @rfalgout this memcpy is taking most of the assembly time. I have some ideas to refactor this, later we can discuss. Thanks!

liruipeng: This copy is just for the simplicity of our test driver. The real apps should have their fem data on device already. Should we be concerned by the cost?

victorapm: I agree the apps will already have their data on the device. However, I think we should give "a good example" here if users want to see how to employ the new AddFEMBoxValues routines. Moreover, I think we should consider passing in data relative to a FEM stencil in a contiguous fashion rather than interleaved as we are doing currently. This would eliminate strided copies.

rfalgout: I think this can wait for a later PR, but let's discuss at one of our upcoming meetings.  Thanks for taking a close look at the timing breakdown!

rfalgout:
Hi @liruipeng and @victorapm .  I am going to remove the draft status of this PR.  Are there any other things we need to do coding-wise here?  I think this is probably optimized enough for now.  I will merge master into this branch again, and we should definitely also do some testing.  Let me know.  Thanks!

liruipeng:
> Hi @liruipeng and @victorapm . I am going to remove the draft status of this PR. Are there any other things we need to do coding-wise here? I think this is probably optimized enough for now. I will merge master into this branch again, and we should definitely also do some testing. Let me know. Thanks!

This PR looks good to me. I agreed with the merging. Thanks!

liruipeng (approved):
Looks good! Thanks @rfalgout !

rfalgout:
There is a problem with `TEST_sstruct/io.sh` that appeared after the merge of master into this branch.  Any ideas what may have happened?  The test was clean before.  Thanks!

liruipeng:
> There is a problem with `TEST_sstruct/io.sh` that appeared after the merge of master into this branch. Any ideas what may have happened? The test was clean before. Thanks!

on tux? Let me try `master` now. I believe I tested before I merged my PR.

rfalgout:
> > There is a problem with `TEST_sstruct/io.sh` that appeared after the merge of master into this branch. Any ideas what may have happened? The test was clean before. Thanks!
> 
> on tux? Let me try `master` now. I believe I tested before I merged my PR.

Yes, on tux.  Note that I merged master before you merged PR #702.  I probably should do another merge now.

liruipeng:
> > There is a problem with `TEST_sstruct/io.sh` that appeared after the merge of master into this branch. Any ideas what may have happened? The test was clean before. Thanks!
> 
> on tux? Let me try `master` now. I believe I tested before I merged my PR.

`io.sh` test is clean in `master`.

rfalgout:
Hi @victorapm .  It looks like your commit 5138246 (move GetObject calls) breaks the io tests.  Any idea why?  It seemed like a safe change to make at the time.  Let me know.  Thanks!

victorapm:
Hi Rob, sorry about that. Just fixed it with [893b7b9](https://github.com/hypre-space/hypre/pull/779/commits/893b7b968cdd49d7fb7cb11d64a42392b2b9688b)

victorapm (approved):

rfalgout:
Thanks @victorapm !

rfalgout:
The tux runs are now mostly clean.  The `TEST_sstruct/sharedpart.sh` test breaks (just a couple of runs) for the `openmp` and `hopscotch` configurations, so it is apparently an openmp problem.  I'll need to look into it further.

rfalgout:
I have not found the problem yet, but here are a few data points.  Let me know if you have any ideas.

If I compile `--with-openmp` (with or without debugging) and run the following on my tux box with two threads, it crashes:

    mpirun -np 3  ./sstruct -in TEST_sstruct/sstruct.in.reduced3D -solver 28 -pooldist 1

If I change the `#if 0` to `#if 1` [here](https://github.com/hypre-space/hypre/blob/6bde373fd1aa5f50768658a51007cf21702688cd/src/test/sstruct.c#L3228) it runs just fine.  The crash appears to be associated with the AddFEMBoxValues() call for part 2.  What's weird is that the new routine was implemented by calling an underlying SetBoxValues() routine that has been around for a while.

I'll do some more debugging later, but if you have any thoughts, let me know.  Thanks!

victorapm:
@rfalgout The crash with OpenMP was due to a bug on `hypre_IJMatrixAddToValuesOMPParCSR`: the pointer `my_offproc_cnt` was being realloced, but its new address was not being saved at `offproc_cnt[my_thread_num]`. I fixed this in the commit [a610454](https://github.com/hypre-space/hypre/pull/779/commits/a610454c89a020c3c9340b9328efd3bcba3f4e0b).

rfalgout:
> @rfalgout The crash with OpenMP was due to a bug on `hypre_IJMatrixAddToValuesOMPParCSR`: the pointer `my_offproc_cnt` was being realloced, but its new address was not being saved at `offproc_cnt[my_thread_num]`.

Thanks @victorapm for fixing this!

PR #782, opened 2022-11-30
Fixed int types discrepancies in MPI calls.

TotoGaz:
Related to https://github.com/GEOSX/GEOSX/issues/2174 where we use the `--enable-mixedint` option when building `hypre`.

I've also written comments where I think there may be something tricky, but I do not know how to fix (neither if it's a real issue).

@victorapm

PR #783, opened 2022-11-30
Bug fixes involving mixedint and CUDA builds

victorapm:
Copied from #782:
> Related to https://github.com/GEOSX/GEOSX/issues/2174 where we use the --enable-mixedint option when building hypre.
> I've also written comments where I think there may be something tricky, but I do not know how to fix (neither if it's a real issue).

Additionally, this PR:

1. Fixes a bug in the creation of the Schur complement matrix for `hypre_ILU`
2. Finishes the TODO's added in the original PR #782 
3. Fixes the device build when using `CUDA-11.1.1` as pointed out by the CI tests of PR https://github.com/GEOSX/thirdPartyLibs/pull/205


src/parcsr_ls/par_ilu_setup.c
```diff
@@ -3353,20 +3353,17 @@ hypre_ILUSetupRAPILU0Device(hypre_ParCSRMatrix *A, HYPRE_Int *perm, HYPRE_Int n,
...
       {
          HYPRE_BigInt global_start;
-         hypre_MPI_Scan( &big_m, &global_start, 1, HYPRE_MPI_BIG_INT, hypre_MPI_SUM, comm);
-         S_row_starts[0] = global_start - m;
+         hypre_MPI_Scan(&big_m, &global_start, 1, HYPRE_MPI_BIG_INT, hypre_MPI_SUM, comm);
+         S_row_starts[0] = global_start - big_m;
          S_row_starts[1] = global_start;
       }
 
-      S_row_starts[1] = S_total_rows;
```
victorapm, line 3367: @oseikuffuor1 these are the lines I told you about

src/parcsr_ls/par_relax.c
```diff
@@ -393,9 +393,10 @@ hypre_BoomerAMGRelax1GaussSeidel( hypre_ParCSRMatrix *A,
       num_sends = hypre_ParCSRCommPkgNumSends(comm_pkg);
       num_recvs = hypre_ParCSRCommPkgNumRecvs(comm_pkg);
 
-      v_buf_data = hypre_CTAlloc(HYPRE_Real, hypre_ParCSRCommPkgSendMapStart(comm_pkg, num_sends),
+      v_buf_data = hypre_CTAlloc(HYPRE_Complex,
+                                 hypre_ParCSRCommPkgSendMapStart(comm_pkg, num_sends),
                                  HYPRE_MEMORY_HOST);
-      v_ext_data = hypre_CTAlloc(HYPRE_Real, num_cols_offd, HYPRE_MEMORY_HOST);
+      v_ext_data = hypre_CTAlloc(HYPRE_Complex, num_cols_offd, HYPRE_MEMORY_HOST);
```
ulrikeyang, line 399: Should we really use complex here? We sometimes use real and sometimes complex. Do we have any rules about this?

src/utilities/_hypre_utilities.hpp
```diff
@@ -95,7 +95,7 @@ using hypre_DeviceItem = void*;
 #endif
 
 #define CUSPARSE_NEWAPI_VERSION 11000
-#define CUSPARSE_NEWSPMM_VERSION 11201
+#define CUSPARSE_NEWSPMM_VERSION 11401
 #define CUDA_MALLOCASYNC_VERSION 11020
```
ulrikeyang, line 95: I have no idea how these numbers come about, so hopefully someone else has.

ulrikeyang (commented):
Apart from my comments this looks fine.

src/utilities/_hypre_utilities.hpp
victorapm: `11401` refers to the cuSPARSE version in which a new algorithm (CUSPARSE_SPMM_CSR_ALG3) was added for Sparse Matrix - Matrix Multiplication (cusparseSpMM). That algorithm shows better performance for small matrices. cuSPARSE `11401` was introduced in CUDA `11.2.1`, see: https://docs.nvidia.com/cuda/archive/11.2.1/cuda-toolkit-release-notes/index.html

src/parcsr_ls/par_relax.c
victorapm: The main reason for this change is because `v_ext_data` is declared as `HYPRE_Complex`, see https://github.com/hypre-space/hypre/blob/e042698bc5a1cc6b44274ce419335bd526753b59/src/parcsr_ls/par_relax.c#L379

Now, regarding the second question: we use `HYPRE_Complex` for storing the data array of our vectors and matrices, so I think it's correct to use `HYPRE_Complex` for other data arrays like `v_ext_data` to make it consistent. If we decide to use `HYPRE_Real` for matrix and vector data instead, then `v_ext_data` should be `HYPRE_Real`. At the end of the day, `HYPRE_Real` is the same as `HYPRE_Complex` in the way hypre is usually built. 

Side question not for this PR: should we have something like `HYPRE_Scalar` to minimize confusion? I like this idea, however the number of changes would be large.

ulrikeyang (approved):
okay then. :)

liruipeng (approved):
Looks good to me. Thanks!

rfalgout (approved):
Looks good

oseikuffuor1 (approved):
Looks good. Thanks Victor!

PR #788, opened 2022-12-07
Add new search path for the NVIDIA math libs

victorapm:
This PR adds two new search paths (`${CUDA_PATH}/../math_libs` and `${CUDA_PATH}/../../math_libs`) for the NVIDIA math libraries (cuSPARSE, cuBLAS, cuSOLVER) to the configure script. 

This PR fixes build issues on Polaris and Perlmutter, see https://github.com/xsdk-project/xsdk-issues/issues/207


ulrikeyang (approved):
I don't know much about this, but according to Pieter it fixes our issues on those machines. Thanks for putting this together.

osborn9 (approved):
Thanks for fixing this issue @victorapm! This looks good to me.

rfalgout (approved):
Thanks Victor!

liruipeng:
Just out of my curiosity why `extra-ldpath` or something like it doesn't work. The solution in this PR is fine to me, which however is specific to those two machines. Thanks!

victorapm:
Hi Rui Peng, that solution also works and it's in fact more general, however it requires input from the user and changes in other places such as the spack package. The solution implemented in this PR has the advantage of working automatically for these machines. I think it's good to have both

liruipeng:
> Hi Rui Peng, that solution also works and it's in fact more general, however it requires input from the user and changes in other places such as the spack package. The solution implemented in this PR has the advantage of working automatically for these machines. I think it's good to have both

OK. Thanks, Victor! I am fine with this PR.

PR #790, opened 2022-12-09
Fix for Intel GPUs

liruipeng:
This PRs is a (temporary) fix for device openmp "zero-reduction-results" issue on JLSE Intel GPU machines. It seems that the driver `struct.c` needs to be compiled as a "device code" with the openmp-target flag. This is not needed for other platforms, CUDA or HIP. A workaround used was to compile this file into `struct.obj`.


waynemitchell (approved):
Thanks for the simple fix!

PR #793, opened 2022-12-14
Update CMake version in INSTALL directions

osborn9:
This corrects the minimum version of CMake required and removes reference to src/test/cmbuild (which no longer exists).


rfalgout (approved):
Looks good.  Thanks!

PR #796, opened 2022-12-15
Update mac regression test

rfalgout:
This PR updates the mac regression test.  The test is really simple right now.  The most important thing to test here are the changes to the `runtest.sh` script.

The expression `\t` for the "tab" character does not work on the mac, so the whitespace expressions were changed to use the posix standard `[[:space:]]`.  This should be (more) portable.  Verify by running basic tests:

- [x] macbook
- [x] tux
- [x] lassen


victorapm (approved):
Thanks! This runs fine on my Mac Catalina (10.15.7)

liruipeng:
Hi @rfalgout Thank you for putting these together. I don't have fortran compilers on my macbook, so I always configure with `-enable-fortran=no`. But it may not be the case for your test platform. Also, I have a lot of warnings like `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libHYPRE.a(csr_spgemm_device_numer6.obj) has no symbols`. I am not sure if they need to be filtered out in the regression. Thanks!

rfalgout:
Hi @liruipeng and @oseikuffuor1 .  Since this PR is pretty straightforward, I'd like to merge it.  Would you please take a look and give it a review?

We can make changes to the `machine-mac.sh` test later after we discuss a bit more.

oseikuffuor1 (approved):

liruipeng (approved):
Looks good. Thanks!

PR #798, opened 2022-12-17
Benchmark

liruipeng:
This PRs updates the amg benchmark and the associated functions that are needed in hypre.

- [x] tux
- [x] lassen
- [x] tioga


src/seq_mv/HYPRE_seq_mv.h
```diff
@@ -80,6 +80,8 @@ HYPRE_Int HYPRE_VectorDestroy( HYPRE_Vector vector );
 HYPRE_Int HYPRE_VectorInitialize( HYPRE_Vector vector );
 HYPRE_Int HYPRE_VectorPrint( HYPRE_Vector vector, char *file_name );
 HYPRE_Vector HYPRE_VectorRead( char *file_name );
+HYPRE_Int HYPRE_VectorGetData( HYPRE_Vector vector, HYPRE_Complex **data);
```
rfalgout, line 83: Why is this something we want to expose to users?

rfalgout (approved):
Looks good, modulo the question about the user-level GetData() interface.

ulrikeyang (commented):
Do we really want to include the amg driver in hypre?
This driver is supposed to be given to the vendors separately with a Makefile, which will allow them to build it with a hypre installation.
Unless there is overall agreement that it should be in hypre, I suggest to remove it.

rfalgout:
> Do we really want to include the amg driver in hypre? This driver is supposed to be given to the vendors separately with a Makefile, which will allow them to build it with a hypre installation. Unless there is overall agreement that it should be in hypre, I suggest to remove it.

I agree.  The benchmarks (in this case, the driver) should be separate from hypre.

liruipeng:
> > Do we really want to include the amg driver in hypre? This driver is supposed to be given to the vendors separately with a Makefile, which will allow them to build it with a hypre installation. Unless there is overall agreement that it should be in hypre, I suggest to remove it.
> 
> I agree. The benchmarks (in this case, the driver) should be separate from hypre.

OK. I guess there was some misunderstanding about it. `amg.c` was pushed to the branch and the we talked about PR from the branch, so I thought we wanted to have it in the release. I can remove it, along with all the regression test on it. This makes the work on this PR easier, which is good. Thanks!

src/seq_mv/HYPRE_seq_mv.h
liruipeng: This is due to how `amg.c` was written: it doesn't include any developer level head files, only user level `HYPRE_*.h` instead, which I assumed was required by the benchmark. If this is the case, we have to have some user level functions for `HYPRE_Vector` (especially for making the code portable on GPUs and simple) for which we basically don't provide anything currently, other than create and destroy (which also seems weird).

ulrikeyang:
Yes, the PR was not supposed to be done from the benchmark branch. We just need to get those functions into hypre for the benchmark driver to work for the vendors.

ulrikeyang:
Yes, I was mostly trying to avoid having to include _hypre .h files. I also think that the new routines can be useful for users.

src/seq_mv/HYPRE_seq_mv.h
liruipeng: If we are allowed to include any headers in the benchmark, I can remove all these, which basically leaves this PR nothing except for a simple mistake fix made in `_hypre_utilities.h` from earlier merges.

liruipeng: > copied here from @ulrikeyang: Yes, I was mostly trying to avoid having to include _hypre .h files. I also think that the new routines can be useful for users.

rfalgout: The issue I have here is that GetData() exposes the underlying data storage to the user, so we lock ourselves into a specific format if we want to be backward compatible.  In the case of a simple sequential vector, it's probably okay to assume this is just an array of doubles (HYPRE_Complex) (though there is additional complexity to consider when the vector is a true multivector).  The reason we are doing this here is to simplify the CPU/GPU complexity for setting values in a double array, which is otherwise just a trivial `for` loop (at least that's my take on what's happening in `amg.c`).

I agree that we don't want to include `_hypre_*.h` files in the benchmark driver.

Anyway, since this is an interface that isn't really widely used or advertised to users, I am willing to go along with it just to make life simple.  Thanks!

ulrikeyang: I see your argument. We actually already have _hypre_utilities.h in the driver for the timing routines. So, we could add one more, if you want to avoid this issue.

liruipeng: OK. This sounds good to me. So, I guess I can just close this PR without merging in anything. I will change `amg.c` accordingly, which will be much easier to do. Thanks again!

PR #802, opened 2022-12-20
Update User Manual to reference the API chapter

rfalgout:
This PR hopefully clears up confusion about the Reference Manual.  Now the User Manual references the API chapter and the description of the Reference Manual in the Introduction comments that it is equivalent to the API chapter.


ulrikeyang (approved):
Good idea to change that. Thanks

rfalgout:
> Good idea to change that. Thanks

This was Jerome Solberg's suggestion from more than a year ago.  I am just now getting around to it. :(

PR #806, opened 2023-01-03
Fix RAP memory leak

victorapm:
This PR fixes a memory leak in the parallel `RAP` routine in the following scenario:

- A single task is used.
- Matrix `R` already contains its `diagT` member at input.


rfalgout (approved):
Looks good to me.  Thanks!

liruipeng:
Looks good. A question is in which case, R has the transpose on input? Thanks

victorapm:
> in which case, R has the transpose on input?

That's useful for MGR, where we compute `diagT` when building the restriction operator. Later on, the coarse grid matrix `RAP` is computed

liruipeng (approved):

PR #807, opened 2023-01-05
Add new hypre_IntArray functions

victorapm:
This PR adds new routines to `hypre_IntArray`:

- `hypre_IntArrayMigrate`: migrates the array to a new memory location
- `hypre_IntArrayPrint`: prints the array to file
- `hypre_IntArrayRead`: reads the array from file
- `hypre_IntArrayInverseMapping`: computes the inverse mapping `b[a[i]] = i` out-of-place (host and device implementations).


src/utilities/_hypre_utilities.h
```diff
@@ -1948,18 +1948,23 @@ HYPRE_Int hypre_IntArrayCopy( hypre_IntArray *x, hypre_IntArray *y );
...
+HYPRE_Int hypre_IntArrayMigrate( hypre_IntArray *v, HYPRE_MemoryLocation memory_location );
+HYPRE_Int hypre_IntArrayPrint( MPI_Comm comm, hypre_IntArray *array, const char *filename );
+HYPRE_Int hypre_IntArrayRead( MPI_Comm comm, const char *filename, hypre_IntArray **array_ptr );
 HYPRE_Int hypre_IntArraySetConstantValuesHost( hypre_IntArray *v, HYPRE_Int value );
 HYPRE_Int hypre_IntArraySetConstantValues( hypre_IntArray *v, HYPRE_Int value );
 HYPRE_Int hypre_IntArrayCountHost( hypre_IntArray *v, HYPRE_Int value,
                                    HYPRE_Int *num_values_ptr );
 HYPRE_Int hypre_IntArrayCount( hypre_IntArray *v, HYPRE_Int value,
                                HYPRE_Int *num_values_ptr );
+HYPRE_Int hypre_IntArrayInverseMapping( hypre_IntArray *v, hypre_IntArray **w_ptr );
```
rfalgout, line 1962: Should this be part of the `IntArray` "class"?  It only makes sense if the original IntArray defines a permutation (a 1-1 mapping of the set of indices).  So maybe this should be renamed?

rfalgout (approved):
Looks good modulo the naming question.  Thanks!

src/utilities/_hypre_utilities.h
victorapm: I see the permutation as a special case of an `IntArray`, so I think the naming makes sense. We could perhaps add a new flag member to `IntArray` that indicates it's a permutation, something like `is_permutation`. If that flag is false, then `IntArrayInverseMapping` is a no-op. What is your opinion?

victorapm: Another idea is to rename it to `hypre_IntArrayPermInverseMapping` to indicate it operates on a special type of `IntArray`

rfalgout: No need for a new argument.  This makes sense as a specialization of IntArray.  I'm okay with it as is.

src/utilities/int_array.c
```diff
@@ -138,6 +138,135 @@ hypre_IntArrayCloneDeep( hypre_IntArray *x )
...
+   {
+      hypre_error_w_msg(HYPRE_ERROR_GENERIC, "Error: can't open output file\n");
+      return hypre_error_flag;
+   }
+
+   /* Print to file */
+   hypre_fprintf(file, "%d\n", size);
+   for (i = 0; i < size; i++)
+   {
+      hypre_fprintf(file, "%d %d\n", i, data[i]);
```
waynemitchell: @victorapm , is there a particular reason you want to print the index, i, to file? Doesn't seem necessary to me, but maybe you have a good reason for it.

waynemitchell (approved):
Looks good to me other than my question below. Thanks.

src/utilities/int_array.c
victorapm: Hi Wayne, there was no particular reason and I agree it's unnecessary. I'll remove it. Can you check if this builds fine with SYCL? Thanks!

src/utilities/int_array_device.c
```diff
@@ -39,6 +39,55 @@ hypre_IntArraySetConstantValuesDevice( hypre_IntArray *v,
    return hypre_error_flag;
 }
 
+/*--------------------------------------------------------------------------
+ * hypreGPUKernel_IntArrayInverseMapping
+ *--------------------------------------------------------------------------*/
+
+__global__ void
+hypreGPUKernel_IntArrayInverseMapping( hypre_DeviceItem  &item,
```
liruipeng, line 47: You can do this with thrust, but I am also fine with writing a kernel. Thanks!

src/utilities/int_array.c
waynemitchell: @victorapm , I just checked, and it builds OK with sycl. I see no issues with the implementation for sycl.

PR #809, opened 2023-01-05
Access pointer data only if it is not NULL

### LINKED ISSUE #808: Seg fault when passing null pointer to HYPRE_SStructMatrixDestroy or uninitialized HypreVector to HYPRE_SStructVector Destroy
#define HYPRE_DEVELOP_NUMBER 29
#define HYPRE_DEVELOP_STRING "v2.26.0-29-gf9f5b4de8"

HYPRE_SStructMatrixDestroy issues a seg fault when a null pointer is passed in.  The seg fault occurs at the 9th line of the function:

HYPRE_MemoryLocation    memory_location = hypre_SStructMatrixMemoryLocation(matrix);

which is outside the "if(matrix)" null pointer test on the rest of the function.

 There is a similar problem with HYPRE_SStructVectorDestroy, which issues a seg fault when an uninitialized vector is passed in.

victorapm:
This PR solves #808


rfalgout (approved):
Looks good.  Thanks @victorapm !

PR #810, opened 2023-01-06
Add axpyz kernels

victorapm:
This PR adds sequential `SeqVector` and parallel `ParVector` kernels for the operation `z = a * x + b * y`.

This contains both host and device implementations.


src/seq_mv/vector_device.c
```diff
@@ -151,6 +151,48 @@ hypre_SeqVectorAxpyDevice( HYPRE_Complex alpha,
...
+   HYPRE_Int       size        = hypre_VectorSize(x);
+   HYPRE_Int       total_size  = size * num_vectors;
+
+#if defined(HYPRE_USING_CUDA) || defined(HYPRE_USING_HIP)
+   hypreDevice_ComplexAxpyzn(total_size, x_data, y_data, z_data, alpha, beta);
+
+#elif defined(HYPRE_USING_SYCL)
+   HYPRE_ONEDPL_CALL( std::transform,
+                      x_data, x_data + total_size, y_data, z_data,
+                      [alpha, beta](HYPRE_Complex x, HYPRE_Complex y) -> HYPRE_Complex { return alpha * x + beta * y; } );
```
victorapm: @waynemitchell Does this make sense?

rfalgout (approved):
Looks good to me.  Thanks!

liruipeng:
Thanks @victorapm. Maybe we should talk about the strategy of adding more kernels like this. This one obviously can be done by copy, axpy and scaling, which is less efficient of course but doesn't need new code.

victorapm:
Sounds good @liruipeng, I found this kernel useful because it simplifies the code (one call instead of three) + performance reasons. I think these two reasons pay the cost of maintaining extra code.

src/seq_mv/vector_device.c
waynemitchell: @victorapm , this looks correct and compiles for me on JLSE, but I'm curious why you make a special case for sycl here? The call to hypreDevice_ComplexAxpyzn() above will also work for sycl, so I think it's better to have all backends (cuda/hip/sycl) call the same code for simplicity.

victorapm: Oh yes! You are correct, it's definitely better to only have `hypreDevice_ComplexAxpyzn` (which by the way needs a name update). Not sure why I put that onedpl call there. I'll remove it, thanks!!

src/seq_mv/vector_device.c
```diff
@@ -151,6 +151,43 @@ hypre_SeqVectorAxpyDevice( HYPRE_Complex alpha,
...
+   HYPRE_Int i;
+
+   #pragma omp target teams distribute parallel for private(i) is_device_ptr(z_data, y_data, x_data)
+   for (i = 0; i < total_size; i++)
+   {
+      z_data[i] = alpha * x_data[i] + beta * z_data[i];
+   }
+#endif
+
+   hypre_SyncComputeStream(hypre_handle());
```
victorapm, line 186: @liruipeng I put this synchronization here since I saw it's being used in the other functions as well. Thinking twice, it would be good to remove it and deal with synchronization outside of the function so it becomes more generic. What are your thoughts?

victorapm:
@liruipeng @waynemitchell Can you take another look at this? Do you find the addition of `hypre_GetExecPolicy3` something useful here? Thanks!!!

src/seq_mv/vector_device.c
liruipeng: Which place outside the functions do you want to put the sync?

victorapm: I would put in the function that calls `hypre_ParVectorAxpyz` depending on whether a synchronization is necessary or not. The idea would be to make these basic linear algebra functions more flexible by giving the user the ability to decide whether a synchronization must be called or not. To make it safe to external users we could impose the synchronization in the public `HYPRE_` functions (when they exist). The downside of this approach would be that ,inside hypre, we should not take it for granted that device synchronization is called anymore in these BLAS-like functions. Thus, we would need to explicitly write the synchronization calls when necessary... In summary, I see advantages and disadvantages and I believe we can discuss that :) @rfalgout

liruipeng:
> @liruipeng @waynemitchell Can you take another look at this? Do you find the addition of `hypre_GetExecPolicy3` something useful here? Thanks!!!



> @liruipeng @waynemitchell Can you take another look at this? Do you find the addition of `hypre_GetExecPolicy3` something useful here? Thanks!!!

I am not sure about `hypre_GetExecPolicy3`. We have functions like matvec which also have 3 operands but we're still using `hypre_GetExecPolicy1`. So I don't know, and  `hypre_GetExecPolicy3` seems complicated.

src/seq_mv/vector_device.c
liruipeng: > I would put in the function that calls `hypre_ParVectorAxpyz` depending on whether a synchronization is necessary or not. The idea would be to make these basic linear algebra functions more flexible by giving the user the ability to decide whether a synchronization must be called or not. To make it safe to external users we could impose the synchronization in the public `HYPRE_` functions (when they exist). The downside of this approach would be that ,inside hypre, we should not take it for granted that device synchronization is called anymore in these BLAS-like functions. Thus, we would need to explicitly write the synchronization calls when necessary... In summary, I see advantages and disadvantages and I believe we can discuss that :) @rfalgout

The reason of having `hypre_SyncComputeStream` in most of the functions of `seq_mv` is for UVM and mixed CPU-GPU computing. It's complicated. Maybe it's better to talk about it in an SLS meeting. Thanks!

victorapm:
@liruipeng @waynemitchell I removed `hypre_GetExecPolicy3` from this PR. Can you take another look? Thanks!

liruipeng (approved):
Looks good to me. Not sure if regression testing is necessary (if not done).

victorapm:
> Not sure if regression testing is necessary

They are clean

waynemitchell (approved):
The latest version of this PR looks good to me. Thanks, @victorapm !

PR #814, opened 2023-01-10
Add matrix scaling function

victorapm:
This PR adds `hypre_ParCSRMatrixDiagScale` for computing left and right parallel matrix scaling. The function also works when one of the scaling factors are not present. The scaling factors are stored as vectors. Regression tests have been added for this new function.


ulrikeyang (approved):
This looks fine. Thanks

src/parcsr_mv/HYPRE_parcsr_matrix.c
```diff
@@ -408,3 +408,18 @@ HYPRE_ParCSRMatrixMatvecT( HYPRE_Complex      alpha,
                alpha, (hypre_ParCSRMatrix *) A,
                (hypre_ParVector *) x, beta, (hypre_ParVector *) y) );
 }
+
+/*--------------------------------------------------------------------------
+ * HYPRE_ParCSRMatrixDiagScale
+ *--------------------------------------------------------------------------*/
+
+HYPRE_Int
+HYPRE_ParCSRMatrixDiagScale(HYPRE_ParCSRMatrix A,
```
rfalgout: This is an example of an external matrix interface that we should discuss as a team before making it available.

victorapm: Yes, same as `axpyz`. I'll remove this public API from this PR

rfalgout (approved):
This looks good modulo the question about the external user interface.

src/parcsr_mv/HYPRE_parcsr_matrix.c
victorapm: I'm using it in the IJ driver, should I replace it with `hypre_ParCSRMatrixDiagScale`?

rfalgout: Yes.  We try to use the drivers as a means to test user interfaces, but this is an exception (for now, since we may end up adding the external interface later).

victorapm: Ok! Thanks!!

liruipeng (approved):
Looks good. This is a nice extension from the CSR-level to ParCSR. Thank you @victorapm

PR #815, opened 2023-01-10
Add CSRMatrix functions

victorapm:
This PR adds `hypre_CSRMatrixPermute` for computing the permutation `B = A(perm, qperm)` in which `perm` defines the row permutation and `qperm` defines the column permutation. This function expects the inverse permutation vector `rqperm` at input. The device implementation is inspired by PR #792 from @PaulMullowney.

This PR also adds:

-  `hypre_CSRMatrixMigrate` for migrating a `CSRMatrix` to a given memory location.
-  `hypre_CSRMatrixPrintIJ` for printing to file a `CSRMatrix` in IJ format


rfalgout (approved):
Looks fine to me

liruipeng:
Hi @victorapm this looks fine to me. But I believe there should be easier and more portable solution, which might be less efficient but permutation is not in the critical path of performance of anything we have (unless we need to reorder some matrix over and over...) . So I still want to talk about it.

victorapm:
Sure, we can chat more about this. 

The main use case now is for ILU. In GPU runs, GEOSX needs to compute the ILU factorization with RCM enabled for matrices of ~1M DOFs. The old permutation code (running on the CPU) becomes a bottleneck for them. Their use case requires matrix permutations multiple times during a simulation.

liruipeng:
> Sure, we can chat more about this.
> 
> The main use case now is for ILU. In GPU runs, GEOSX needs to compute the ILU factorization with RCM enabled for matrices of ~1M DOFs. The old permutation code (running on the CPU) becomes a bottleneck for them.

Yes, I understand the need. What I was saying is once this is on GPU, a "suboptimal" implementation but much simpler and portable should remove the bottleneck. Permutation should be fast relative to the other parts of ILU and usually we just need to permute the matrix once. So we may not care too much if we don't achieve the best performance.

victorapm:
Ah ok, thanks! I see your point now. By easier implementation you mean replacing the GPU kernels added here with thrust calls?

liruipeng:
> Ah ok, thanks! I see your point now. By easier implementation you mean replacing the GPU kernels added here with thrust calls?

We can chat offline on the details. If it's right, we can do it with a handful of existing hypre util calls and thrust calls, no new kernels. Also, I am not sure about the need of reverse perm array, just after a quick look of what you have, but I could be wrong.

victorapm:
Sounds good! Thanks again for looking at this!

liruipeng:
At the ECP meeting, many code teams talk about code complexity and portability. Some other code doesn't write any raw kernels for the reason. We are not that strict for sure as we already have many raw CUDA kernels already. But I think it will be good to start think about it. Thanks!

victorapm:
Hi @rfalgout @liruipeng, can you take another look at this PR?

We have two implementations for `hypre_CSRPermuteDevice`: one using thrust and the other using hypre GPU kernels. I tested performance, and the latter is slightly faster. It reaches a speedup of `1.4x` for a 27-pt 3D Poisson test case. This operation is about `10%` of ILU's setup time (excluding RCM computation, which is done on the host), thus, the final speedup is not very impressive. Although the GPU kernels implementation gives us full control of what is happening, one drawback is that it requires more lines of code. I am leaned to keep thrust's implementation and if `CSRPermute` ever becomes an important bottleneck, we can return to this PR and restore the GPU kernels implementation.

What are your thoughts? Thank you @liruipeng for contributing with the thrust's implementation!!

rfalgout:
> Hi @rfalgout @liruipeng, can you take another look at this PR?
> 
> We have two implementations for `hypre_CSRPermuteDevice`: one using thrust and the other using hypre GPU kernels. 
> 
> What are your thoughts? Thank you @liruipeng for contributing with the thrust's implementation!!

Hi @victorapm .  Since the performance difference is very small here, I would suggest going with the more portable and maintainable route.  I'll leave that to you and @liruipeng to decide.  Thanks!

rfalgout (approved):

victorapm:
@liruipeng I removed the old CSRPermute implementation. Does this look good for merging into master now? Thanks!

src/seq_mv/csr_matop_device.c
```diff
@@ -1912,6 +1912,102 @@ hypre_CSRMatrixDiagScaleDevice( hypre_CSRMatrix *A,
...
+   HYPRE_Int        *B_j          = hypre_CSRMatrixJ(B);
+   HYPRE_Complex    *B_a          = hypre_CSRMatrixData(B);
+
+   HYPRE_Int        *B_ii;
+
+   /* Build B_i */
+   HYPRE_THRUST_CALL(gather,
+                     perm,
+                     perm + num_rows,
+                     thrust::make_transform_iterator(thrust::make_counting_iterator(0), adj_functor(A_i)),
```
liruipeng, line 1985: @victorapm Have you checked if this works for Sycl and HIP? I didn't. Thanks!

victorapm: It works for HIP and this implementation is not meant to work with SYCL as is. In fact, it's under an `ifdef HYPRE_USING_CUDA || HYPRE_USING_HIP` block. Do you want to address the SYCL implementation in this PR?

liruipeng: No, we don't need to. I didn't realized it was not under HYPRE_USING_SYCL. Thanks!

liruipeng (approved):
Looks good to me, assuming the regressions have been all clean.

PR #816, opened 2023-01-10
ILU device support without UVM

victorapm:
Expanded from #792. This PR has three main goals:

- Enable ILU to work without UVM on device builds.
- Add the runtime switch option to ILU. Thus, in device builds, one can choose now whether to execute ILU's setup and solve on the device or host.
- Improve error handling. e.g, when trying to execute on the device ILU types that are not supported yet.

Also, ILU0's performance on the device has been improved, e.g., for a 27-pt 3D Poisson test case, ILU setup times on a V100 card are give below
| Size        | master 53dfbe360f582b7d23f56a56c36e6b120f9f165f | Ilu-uvm 16c186798fbfc045ef5c299affac904a80b25a0c|
| ----------- | ------ | ------- |
| 64x64x64    | 0.394  | 0.296   |
| 92x92x92    | 1.194  | 0.951   |
| 128x128x128 | 3.408  | 2.773   |
| 160x160x160 | 6.956  | 5.632   |
| 192x192x192 | 12.551 | 10.15   |

Note: this PR depends on #815.

PR #817, opened 2023-01-11
New features for MGR

victorapm:
This PR adds device support to various MGR options:

1. Non-galerkin coarse grid correction options (except for option 4)
2. Block diagonal interpolation (`interp_type = 12`)
3. Block Jacobi relaxation (`level_smooth_type = 0` for global relaxation and `interp_type = 12` for F-relaxation)

Additionally, this PR improves the performance of MGR in device runs.

Note: This PR depends on #816


src/test/TEST_ij/ilu.saved.lassen
```diff
@@ -44,11 +44,11 @@ Final Relative Residual Norm = 6.699477e-09
 
 # Output file: solvers.out.311
 hypre_ILU Iterations = 35
-Final Relative Residual Norm = 8.236857e-09
+Final Relative Residual Norm = 6.392185e-09
```
rfalgout: These are pretty significant differences.  Is there a good explanation?

rfalgout (approved):
Looks good to me.  That was a lot of work.  Thanks!

src/test/TEST_ij/ilu.saved.lassen
victorapm: Yes, some residuals are quite different. I've looked into this and found out it's due to a few floating-point operations during ILU's setup that were previously carried out on the CPU and are now being done on the GPU. I guess these differences are acceptable in this context.

rfalgout: Sounds good.  Thanks!

src/parcsr_ls/HYPRE_parcsr_ls.h
```diff
@@ -4119,15 +4154,15 @@ HYPRE_Int HYPRE_MGRBuildAff(HYPRE_ParCSRMatrix A,
  * Currently uses BoomerAMG.
  * The default, if not set, is BoomerAMG with default options.
  *
- * @param solver [IN] solver or preconditioner object
+ * @param solver [IN] MGR solver/preconditioner object
  * @param coarse_grid_solver_solve [IN] solve routine for BoomerAMG
  * @param coarse_grid_solver_setup [IN] setup routine for BoomerAMG
- * @param coarse_grid_solver [IN] BoomerAMG solver
+ * @param coarse_grid_solver [OUT] coarse grid solver object
```
oseikuffuor1: Is this really an output variable?

victorapm: No, it isn't. Thanks for catching!

src/parcsr_ls/par_mgr_solve.c
```diff
@@ -627,188 +643,293 @@ hypre_MGRCycle( void               *mgr_vdata,
...
+               HYPRE_Int  nrows     = hypre_ParCSRMatrixNumRows(A_ff_array[fine_grid]);
+               HYPRE_Int  n_block   = nrows / blk_size[fine_grid];
+               HYPRE_Int  left_size = nrows - n_block * blk_size[fine_grid];
+               for (i = 0; i < nsweeps[fine_grid]; i++)
                {
-                  // We need to first compute residual to ensure that
-                  // F-relaxation is reducing the global residual
-                  alpha = -1.0;
-                  beta = 1.0;
-                  hypre_ParCSRMatrixMatvecOutOfPlace(alpha, A_array[fine_grid], U_array[fine_grid], beta,
```
oseikuffuor1, line 728: Has this been removed? It may still be necessary to do so since the block Jacobi scheme does not (implicitly) compute the global residual like the other F-relaxation schemes do

src/parcsr_ls/par_mgr_solve.c
```diff
@@ -220,30 +226,21 @@ hypre_MGRSolve( void               *mgr_vdata,
...
-         if (rhs_norm > HYPRE_REAL_EPSILON)
-         {
-            rel_resnorm = resnorm / rhs_norm;
-         }
-         else
-         {
-            rel_resnorm = resnorm;
-         }
-
+         conv_factor = (old_resnorm > HYPRE_REAL_EPSILON) ? resnorm / old_resnorm : resnorm;
```
oseikuffuor1: This meaning has changed, but I think it is OK. Should the change be consistent with line 972 then?

src/parcsr_ls/par_mgr_setup.c
```diff
@@ -1444,80 +1750,66 @@ hypre_MGRSetup( void               *mgr_vdata,
...
+#if MGR_DEBUG_LEVEL == 2
+   wall_time = time_getWallclockSeconds() - wall_time;
+   hypre_ParPrintf(comm, "Proc = %d - Coarse grid setup: %f\n", my_id, wall_time);
+#endif
 
+   /* Allocate l1_norms when necessary
+      TODO (VPM): move this block closer to global smoother setup */
    for (j = 0; j < num_c_levels; j++)
    {
-      HYPRE_Int frelax_type = Frelax_type[j];
```
oseikuffuor1, line 1487: Are the following changes consistent with the relax_order condition?

src/parcsr_ls/par_mgr_solve.c
victorapm: I messed this up! Thanks for catching :)

src/parcsr_ls/par_mgr_solve.c
victorapm: Yes, it should! Thanks!

src/parcsr_ls/par_mgr_setup.c
victorapm: It's not totally consistent. Let me fix

src/parcsr_mv/par_csr_matop_device.c
```diff
@@ -1328,6 +1328,14 @@ hypre_ParCSRMatrixDropSmallEntriesDevice( hypre_ParCSRMatrix *A,
    HYPRE_Real      *elmt_tols_diag = NULL;
    HYPRE_Real      *elmt_tols_offd = NULL;
 
+   /* Exit if tolerance is zero */
+   if (tol < HYPRE_REAL_MIN)
```
liruipeng, line 1332: We can prune numerical zeros by setting tol = 0.0. With this change, it becomes a no-op.

liruipeng (approved):

oseikuffuor1 (approved):
Thanks Victor. This was a lot of work. Other than the comments I made, it looks good to me, thanks.

src/parcsr_mv/par_csr_matop_device.c
victorapm: I'll revert this change if you are fine

liruipeng: I would like to hear what others think about it.

victorapm:
Thanks, all! @liruipeng, I will need to merge #873 before this PR since it fixes some memory leaks appearing in this PR. Can you take a look at 873 when you have time? Thanks!

src/parcsr_mv/par_csr_matop_device.c
victorapm: @rfalgout @oseikuffuor1 In your opinion, What should be the behavior of `hypre_ParCSRMatrixDropSmallEntries` when `tol = 0.0` is passed in?
1. Exit the function
2. Try to remove coefficients that are actually equal to zero from the sparsity pattern

oseikuffuor1: It appears the behavior of the parent function [hypre_ParCSRMatrixDropSmallEntries](https://github.com/hypre-space/hypre/blob/58d1abbc274a550da5a1c34c536e2d4a364c1920/src/parcsr_mv/par_csr_matop.c#L6215) returns hypre_error_flag if (tol <=0.0). This means we cannot prune numerical zeros anyway with this function. As Rui Peng noted, I think it is a good idea to be able to do that (I thought we were able to do this), but the change would have to happen at a higher level than this device code (i.e. in par_csr_matop.c)

victorapm: Good point, Daniel! I'm wondering if we should have a different function for the purpose of pruning zeros from the sparsity pattern. That would allow the DropSmallEntries function to behave as no-op in the case of `tol = 0.0` and I think that's important because it would provide an efficient way to turn off the dropping of small entries. I say efficient because in the `tol = 0` scenario, the function would just exit.

victorapm: Another idea is: have a single [hypre_ParCSRMatrixDropSmallEntries](https://github.com/hypre-space/hypre/blob/58d1abbc274a550da5a1c34c536e2d4a364c1920/src/parcsr_mv/par_csr_matop.c#L6215) function, make it drop numerical zeros when `tol = 0` and make it a no-op when `tol < 0`. Finally, make this behavior consistent in the host and device implementations

rfalgout: After looking through some of the calling code, I am thinking that your last idea @victorapm is the right way to go, i.e., have a separate routine for dropping zero entries.  This gives the most flexibility since: it allows a no-op; it still allows us to prune zeros from matrices if we want to; and it also lets us keep the zeros (we may want to have a fixed graph structure for some reason, even if some of the entries are zero).

rfalgout: Note that by "last idea" in my previous comment, I really meant the idea before your last. :)  So, no `tol<0` option.

liruipeng: If I read correctly, I like the idea of tol < 0 being no-op and tol = 0.0 to prune numerical zeros. I was wondering relaxing the if-condition in the parent function will do so. At least the GPU code seems fine which I checked yesterday. Thanks for the discussion.

liruipeng: OK. I guess Rob referred to the other idea Victor suggested...

oseikuffuor1: I think if we are going with Victor's former suggestion (i.e. new functionality to specifically sparsity by dropping zeros), then we need not change anything in the parent function, and this new code under discussion in the device code should simply be removed. I believe that will make both host and device code behave the same way (i.e. a no-op if tol <= 0.0). Then we can add new functionality for dropping zeros.

liruipeng: Agreed. The new function, say named `hypre_ParCSRMatrixDropZeros`, is essentially the same as this one, removing the `if (tol <=0.0) return` of course, at least for the GPU code.

victorapm: Thanks everyone for the inputs! If we have a`hypre_ParCSRMatrixDropZeros`, how many places in hypre would we need to include calls to it? Otherwise, if we go with the single drop function idea with no-op given by `tol < 0`, we would not have to include calls to DropZeros throughout the code. Hmmm, that's tricky! There are advantages in both approaches... Nice discussion!

rfalgout: It doesn't look like the current code would remove zeros anyway with `tol=0`.  The thrust calls deep down in `csr_matop_device.c` use `less_than`.  Maybe I'm misreading it.  Anyway, Daniel's suggestion is probably best for now so we can discuss this more.  It might well be that a special routine for dropping zeros can be done by simply calling a version of this routine that uses the `tol<0` approach, or we can skip the special routine altogether.  We would want to minimize writing similar code as much as possible in either case.

rfalgout: One other comment about the `tol<0` no-op idea.  This would require the calling code to know ahead of time that it wants a no-op (it requires passing in a negative number and most of these tolerances are assumed to be `>=0`).  If that's the case and you have an extra `if` anyway, why not just skip calling the function altogether instead of passing a negative tolerance?  I think this needs more thought.

liruipeng: I think we missed taking absolute values in `csr_matop_device.c`? Oops, this is a bigger problem..

And Rob was right on the `less_than`

rfalgout: I was wondering about the absolute value question, but I was guessing you intended to only pass matrices with non-negative entries to it (e.g., strength matrices).  Thanks!

liruipeng: It should be a bug. We should add `thrust::make_transform_iterator(A_data,  absolute_value<HYPRE_Real>())`. Sorry.

victorapm: Sounds great! Shall I leave `hypre_ParCSRMatrixDropSmallEntriesDevice` as is in this branch and open another branch to continue working on the dropping strategy, e.g., adding a new `hypre_ParCSRMatrixDropZeros`, make adjustments to `csr_matop_device.c` as talked above and possibly others?

victorapm: Hi all, so I'll make changes related to matrix truncation in another PR. Thanks!

PR #824, opened 2023-01-19
Extend hypre complex

oseikuffuor1:
This PR minimally extends the complex type in hypre to allow for single and long double complex types.


src/utilities/HYPRE_utilities.h
```diff
@@ -83,9 +83,44 @@ typedef double HYPRE_Real;
...
 #if defined(HYPRE_COMPLEX)
+/* support for float double and long double complex types */
+#if defined(HYPRE_SINGLE)
+typedef float _Complex HYPRE_Complex;
+#define HYPRE_MPI_COMPLEX MPI_C_FLOAT_COMPLEX
+#ifdef HYPRE_SEQUENTIAL
+#define MPI_C_FLOAT_COMPLEX hypre_MPI_COMPLEX
+#endif
+/* wrappers to some general complex functions */
+#define hypre_complex_abs cabsf
```
rfalgout: Since we are already using function calls defined in `complex.c` for these, we should put the `#if defined` details inside the functions.  We definitely don't want lower-case `hypre_` stuff in this file.

src/utilities/HYPRE_utilities.h
```diff
@@ -83,9 +83,44 @@ typedef double HYPRE_Real;
 #endif
 
 #if defined(HYPRE_COMPLEX)
+/* support for float double and long double complex types */
+#if defined(HYPRE_SINGLE)
+typedef float _Complex HYPRE_Complex;
+#define HYPRE_MPI_COMPLEX MPI_C_FLOAT_COMPLEX
+#ifdef HYPRE_SEQUENTIAL
+#define MPI_C_FLOAT_COMPLEX hypre_MPI_COMPLEX
```
rfalgout: This should go in `mpistubs.h` (and hence also in `_hypre_utilities.h`).

src/utilities/general.h
```diff
@@ -47,5 +47,17 @@ typedef double                 hypre_double;
 #define hypre_pow2(i)  ( 1 << (i) )
 #endif
 
+#include <HYPRE_config.h>
```
rfalgout: You shouldn't need this line since this is only used to help build `_hypre_utilities.h` which should already have it by way of `HYPRE_utilities.h`.

src/utilities/_hypre_utilities.h
```diff
@@ -1599,10 +1610,10 @@ HYPRE_Real    hypre_cimag( HYPRE_Complex value );
 HYPRE_Complex hypre_csqrt( HYPRE_Complex value );
 #else
 #define hypre_conj(value)  value
-#define hypre_cabs(value)  fabs(value)
+#define hypre_cabs(value)  hypre_abs(value)
```
rfalgout: If we leave this as `fabs` we get some type checking.  If we use `hypre_abs` then `hypre_cabs` would work for `int` as well.

src/utilities/complex.c
```diff
@@ -14,31 +14,31 @@
 HYPRE_Complex
 hypre_conj( HYPRE_Complex value )
 {
-   return conj(value);
+   return hypre_conjugate(value);
```
rfalgout: See my other comment about putting the ifdefs here

src/utilities/general.h
oseikuffuor1: Thanks @rfalgout, this makes sense.

rfalgout: You are right that a bunch of header files could be deleted from the `Makefile` dependency list since they are only used to generate `_hypre_utilities.h`.  You still should be able to delete this line though, I think.

src/utilities/HYPRE_utilities.h
oseikuffuor1: @rfalgout, just to be clear should the whole highlighted block be moved or just the new portion with HYPRE_SEQUENTIAL?

src/utilities/HYPRE_utilities.h
oseikuffuor1: Thanks

src/utilities/HYPRE_utilities.h
rfalgout: All of the `MPI_` defines should be moved.  Sorry, I keep assuming that when I click on a line in github to make a comment, that the line is somehow highlighted.  I'll have to try to remember that's not the case.

src/utilities/_hypre_utilities.h
```diff
@@ -1599,10 +1610,10 @@ HYPRE_Real    hypre_cimag( HYPRE_Complex value );
...
 #define hypre_creal(value) value
 #define hypre_cimag(value) 0.0
-#define hypre_csqrt(value) sqrt(value)
+#define hypre_csqrt(value) hypre_sqrt(value)
```
rfalgout: How did `hypre_csqrt` work before when using "single" or "quad"?  Maybe we don't even use this function anywhere yet? ... Looks like we only use it in FSAI at the moment, but maybe we don't test that in single or quad yet.

src/utilities/_hypre_utilities.h
liruipeng: `hypre_abs` is a macro, so it works everywhere (CPU and GPU) and for everything(double, float, int, long, etc). `fabs` is a CPU function and available on GPUs as well, which accepts only `double`.

src/utilities/_hypre_utilities.h
oseikuffuor1: @falgout the current HYPRE_COMPLEX is defined to only support double precision, so it would be fine if hypre is built with --enable-complex. However, without this build option, I would expect there to have been a warning, since sqrt is only for double precision. Using hypre_sqrt should fix that issue. I see hypre_csqrt being used in par_fsai_setup.c.

rfalgout: Good point that you would expect to get a warning when `par_fsai_setup.c` was compiled with "single" or "quad" turned on.  It's not just a question of testing.  Hmmm...  In any case, we need the functionality of this `hypre_sqrt` macro to handle all of the real cases, so this is okay.

src/utilities/_hypre_utilities.h
rfalgout: I knew it was a macro, but I forgot that we need it for "single" and "quad" too (even though I talked about this very thing in my comment on `hypre_sqrt`).  I still wonder about the type checking question though.  Is that important to us?

oseikuffuor1: @falgout, what's your take on this one? I like the generic-ness of the macro, but I also like the safety of type checking. An alternative is to expressly define hypre_cabs to take <fabsf, fabs, or fabsl>, though this may have portability issues on GPUs.

rfalgout: Just leave it as `hypre_abs` for now.  I'm not sure about the type checking question or the GPU side yet.

rfalgout:
Hi @liruipeng .  Given your comment on GPUs, are there special functions for `fabs`, `sqrt`, etc. on GPUs that would differ from the CPU case?  Same question for `conj`, `cabs`, `creal`, `cimag`, and `csqrt` (see `complex.c`).

src/utilities/mpistubs.h
```diff
@@ -55,7 +55,13 @@ extern "C" {
 #define MPI_CHAR            hypre_MPI_CHAR
 #define MPI_LONG            hypre_MPI_LONG
 #define MPI_BYTE            hypre_MPI_BYTE
+#if defined (HYPRE_SINGLE)
```
rfalgout: You don't actually need the ifdef stuff here since each of these names is distinct.

rfalgout (approved):
This looks good @oseikuffuor1 !  Be sure to run all of our regression tests.  Thanks!

liruipeng:
> Hi @liruipeng . Given your comment on GPUs, are there special functions for `fabs`, `sqrt`, etc. on GPUs that would differ from the CPU case? Same question for `conj`, `cabs`, `creal`, `cimag`, and `csqrt` (see `complex.c`).

`fabs`, `sqrt` are available on GPUs with the same function signatures as the CPU ones.

Complex numbers are handled differently on GPUs by `cuComplex` which is compatible with C99 _Complex and we may need to use things like `cuConj` for them.

liruipeng (approved):
Looks good to me. Thanks!

PR #826, opened 2023-01-31
Binary I/O functions for IJ matrices and vectors

victorapm:
This PR adds new Print and Read functions for matrices and vectors to be stored/read in binary format.

Two new options have been added to the IJ driver for testing purposes:

- Print an `IJMatrix` A and `IJVector` x0, x and b to file in binary format: `mpirun -np 1 ./ij -printbin`
- Read an `IJMatrix` A and `IJVector` x0 and b from file stored in binary format: `mpirun -np 1 ./ij -frombinfile IJ.out.A -rhsfrombinfile IJ.out.b -x0frombinfile IJ.out.x0`

Additionally, new regression tests have been added under `TEST_ij/io.jobs` to test the new functionalities


src/IJ_mv/protos.h
```diff
@@ -189,9 +191,12 @@ HYPRE_Int HYPRE_IJMatrixSetDiagOffdSizes ( HYPRE_IJMatrix matrix, const HYPRE_In
 HYPRE_Int HYPRE_IJMatrixSetMaxOffProcElmts ( HYPRE_IJMatrix matrix, HYPRE_Int max_off_proc_elmts );
 HYPRE_Int HYPRE_IJMatrixRead ( const char *filename, MPI_Comm comm, HYPRE_Int type,
                                HYPRE_IJMatrix *matrix_ptr );
+HYPRE_Int HYPRE_IJMatrixReadBinary ( const char *filename, MPI_Comm comm, HYPRE_Int type,
```
rfalgout, line 196: Why is `HYPRE_IJMatrixReadBinary` not in `HYPRE_IJ_mv.h`?

victorapm: Nice catch! I forgot to put it there... Fixed with [478857a](https://github.com/hypre-space/hypre/pull/826/commits/478857a9411676cbb407a53e0ed07f1234783ee7)

src/utilities/HYPRE_utilities.h
```diff
@@ -35,19 +35,34 @@ extern "C" {
 #if defined(HYPRE_BIGINT)
 typedef long long int HYPRE_BigInt;
 typedef long long int HYPRE_Int;
+
+#define HYPRE_BIG_INT_MAX INT64_MAX
```
rfalgout: I'm not sure this `HYPRE_BIG_INT_MAX` define and the others like it belong in this external user header file.  I think this is something we only need internally for the binary read/write.  Also, I am not sure about explicitly associating 32 and 64 to Int and BigInt.  Maybe these should be removed and the one `if` statement in `IJMatrix.c` can be dealt with differently.

src/parcsr_mv/par_csr_matrix.c
```diff
@@ -785,6 +785,311 @@ hypre_ParCSRMatrixPrintIJ( const hypre_ParCSRMatrix *matrix,
...
+   header[6] = (hypre_uint64) jupper;
+   header[7] = (hypre_uint64) 1;
+   if (fwrite((const void*) header, sizeof(hypre_uint64), count, fp) != count)
+   {
+      hypre_error_w_msg(HYPRE_ERROR_GENERIC, "Could not write all header entries\n");
+      return hypre_error_flag;
+   }
+
+   /* Allocate memory for buffers */
+   if (sizeof(HYPRE_BigInt) == sizeof(hypre_uint32))
```
rfalgout, line 934: I like the way you check the sizes explicitly and store it in header information written in a fixed 64 bit format.

src/utilities/HYPRE_utilities.h
victorapm: Yes, I actually think this whole block belongs to `general.h`, which already contains a lot of typedefs for integer and floating-point types. Should we move this block to `general.h`?

Regarding the other question, I agree this needs to be changed

src/parcsr_mv/par_csr_matrix.c
victorapm: That's the best solution I could think of for reaching portability across different platforms using the `ILP64`, `LP64`, `LLP64` and `LP32` models :)

src/IJ_mv/IJMatrix.c
```diff
@@ -252,3 +242,259 @@ hypre_IJMatrixRead( const char     *filename,
...
+
+   /* Exit if trying to read from big-endian machine */
+   if ((*(char*)&header[7]) == 0)
+   {
+      hypre_error_w_msg(HYPRE_ERROR_GENERIC, "Support to big-endian machines is incomplete!");
+      return hypre_error_flag;
+   }
+
+   /* Check for integer overflow */
+   if (header[2] > HYPRE_INT_MAX)
```
rfalgout: I don't understand this `if` statement as written.  I get the idea of checking for potential overflow, but why doesn't this check against the size of `BigInt`?  I also don't understand what `INT_MAX` or `INT_MIN` actually mean (same for the others).  And as mentioned before, we should NOT associate specific numbers like 32 and 64 with `Int` or `BigInt`.  In these Print/ReadBinary routines, you are explicitly checking the sizes against 32 and 64, which totally makes sense and doesn't link 32 to `Int` or 64 to `BigInt`.

rfalgout (approved):
Hi @victorapm .  This looks good to me, except for my comments about the one `if` statement and the new macros you added.  You did a nice job of thinking about portability here!  Thanks!

src/IJ_mv/IJMatrix.c
victorapm: Thanks @rfalgout!! Regarding the first question: I need to check against `HYPRE_Int` because the variable in question `nrows` will be used as input to `HYPRE_IJMatrixSetValues`, which expects it to be a `HYPRE_Int`. Now, there's the question: should it really be a `HYPRE_Int` there? I think a `HYPRE_BigInt` is more appropriate. But as long as we don't make this change, we'll need to do the overflow check with `INT_MAX`.

Second question: I agree that's wrong. I implemented a fix on [c35b5a0](https://github.com/hypre-space/hypre/pull/826/commits/c35b5a0aa7bca181ae2107816a0ec5fd645773db) to use the appropriate limit values according to the data types: `int` and `long long int`. Let me know if that's Ok! There's still the question about moving that block of code to some other place...

rfalgout: Well, I answered my question about what `MAX` and `MIN` mean here.  I was confused because you are introducing these to avoid numbers that are too large, whereas `MIN` is a negative number.  I guess I don't really think you need to introduce these in the first place.  If you do, however, they should be in `general.h` and start with lower-case `hypre_`.

Good question about `SetValues`.  My guess is we thought this would never be larger than 2 billion since it refers to the local number of rows on the processor.  Maybe that's not a good assumption?  I think I agree with you.

But, since you are casting `nrows` to `HYPRE_Int`, why not just check against `sizeof(HYPRE_Int)`?

rfalgout: Never mind my question about `sizeof`...

rfalgout: I also see that we have `HYPRE_REAL_MAX` etc. in `HYPRE_utilities.h`.  Maybe users could potentially use these?  Hold off of moving stuff for the moment...

rfalgout: Okay, sorry for all the fuss.  I think you have it right just the way it is. :)  I would only suggest moving the `HYPRE_MPI_BIG_INT` and `HYPRE_MPI_INT` defines next to each other instead of interleaving them with the other 4 new defines (put the `MPI` ones at the bottom like we do for the Real stuff).

There is still the question about the `SetValues` interface.  We can maybe take that offline with the rest of the team for now.

victorapm: Thanks for all the comments, Rob! I'll move those macros close to each other. And I agree we can talk about `SetValues` later :)

victorapm:
@rfalgout Can you take a look again at this?

I have finished implementing the public APIs and internal functions for reading vectors and added regression tests. Thanks!

src/parcsr_mv/HYPRE_parcsr_mv.h
```diff
@@ -70,6 +70,7 @@ HYPRE_Int HYPRE_ParVectorDestroy( HYPRE_ParVector vector );
 HYPRE_Int HYPRE_ParVectorInitialize( HYPRE_ParVector vector );
 HYPRE_Int HYPRE_ParVectorRead( MPI_Comm comm, const char *file_name, HYPRE_ParVector *vector );
 HYPRE_Int HYPRE_ParVectorPrint( HYPRE_ParVector vector, const char *file_name );
+HYPRE_Int HYPRE_ParVectorPrintIJ( HYPRE_ParVector vector, const char *file_name );
```
rfalgout: Should this be PrintBinaryIJ?

victorapm: Thanks! Just fixed

victorapm:
@rfalgout I was thinking about something here... Maybe we need to add some kind of version info to the header metadata. In the future, if we want to change the data that is printed to the binary files, we will be able to identify those changes by looking at the version of the header. Just trying to anticipate a very likely problem... 

Do you support the idea of including a header version info?

rfalgout:
> @rfalgout I was thinking about something here... Maybe we need to add some kind of version info to the header metadata. In the future, if we want to change the data that is printed to the binary files, we will be able to identify those changes by looking at the version of the header. Just trying to anticipate a very likely problem...
> 
> Do you support the idea of including a header version info?

That makes sense to me.

rfalgout:
I was thinking about how you are doing testing.  I am wondering if we really want to keep binary saved files around.  I don't think this is portable across platforms (some quick googling seems to confirm that).

Also, it doesn't look like you are testing the print routines right now?

I wonder if doing something like this would be better:

- run with input from ascii saved files and print in binary
- run with binary files and print in ascii
- compare the original ascii files to the ones printed out

One problem I see with this is we could change the binary read/write routines to use a different output format so that they still work symmetrically, and we wouldn't see a failed test here.  If we know we are changing the file format, then it's not a problem, but if we accidentally do this somehow, we'd want to detect it in the testing.

Just thinking out loud...

victorapm:
Thanks for the feedback, Rob!

The rationale behind adding binary files here is to test the functionality of reading data generated from different hypre builds. Currently 4 combinations are considered:

- default build: 32-bit integers, 64-bit floats
- `--enable-single`: 32-bit integers, 32-bit floats
- `--enable-mixedint`/`--enable-bigint`: 64-bit integers, 64-bit floats
- `--enable-single --enable-mixedint`/`--enable-single --enable-bigint`: 64-bit integers, 32-bit floats

The idea is to make sure we are able to read and process information that was produced from these different builds regardless of the build type that is currently running the test. I think this test is important for portability. What do you think?

> I don't think this is portable across platforms (some quick googling seems to confirm that).

I'm struggling to understand why the current binary files would not be portable across platforms since the metadata was designed to have a fixed format and length and also because it dictates the sizes of the the data types that are read below. The only portability problem that I can see is trying to read those files from a big-endian machine. That would be possible with some byte swapping, but I decided to not implement it (these machines are not very common in the HPC environment) and instead just throw an error. Perhaps, I'm missing something here and there's another reason for not achieving portability easily... If that is the case, please let me know!

> It doesn't look like you are testing the print routines right now?

Yes, good catch. That needs to be included! I would do it similarly to the idea you provided:

- First stage: Run a small poisson 7-pt problem and print the system to binary files
- Second stage: Read the binary files, run the problem with same solver, and compare the number of iterations with the first run.

Do you think this approach is good?

Thanks!!

rfalgout:
> The idea is to make sure we are able to read and process information that was produced from these different builds regardless of the build type that is currently running the test. I think this test is important for portability. What do you think?

Sounds good.

> > I don't think this is portable across platforms (some quick googling seems to confirm that).
> 
> I'm struggling to understand why the current binary files would not be portable across platforms

I brought this up based on my own, old, almost certainly outdated, experience. :)  It appears you have already thought about portability issues, so I am good to go with this.

> > It doesn't look like you are testing the print routines right now?
> 
> Yes, good catch. That needs to be included! I would do it similarly to the idea you provided:
> 
> * First stage: Run a small poisson 7-pt problem and print the system to binary files
> * Second stage: Read the binary files, run the problem with same solver, and compare the number of iterations with the first run.
> 
> Do you think this approach is good?

Yes, perfect.  Thanks!

victorapm:
Thanks! I'll work on adding tests for the print routines

liruipeng:
Little and Big Endian is the only thing I can imagine that prevents the portability of binary files (vs. text files).

oseikuffuor1 (approved):
Looks good. Thanks Victor

PR #827, opened 2023-01-31
Fix floating point conversion warnings

oseikuffuor1:
This PR cleans up several floating point conversion warnings when using special gcc compiler flags (-Wfloat-conversion and -fsingle-precision-constant) for strict checking. The PR touches several files, but many of these changes are simple and straightforward. The gcc flag -Wfloat-conversion throws several warnings, some of which are benign, such as:
   float tol = 1.0e-2; 
since the standard assumes floating point constants and literals are in double precision. The -fsingle-precision-constant flag provides a way to get around this issue by representing all constants as floats. For the purpose of adding type conversion checking to our strict-checking option, this workaround seems reasonable.

### DIFF
--- a/src/krylov/lobpcg.c
+++ b/src/krylov/lobpcg.c
@@ -146,7 +146,7 @@ lobpcg_MultiVectorImplicitQR(
 }
 
 static void
-lobpcg_sqrtVector( HYPRE_Int n, HYPRE_Int* mask, HYPRE_Real* v )
+lobpcg_hypre_sqrtVector( HYPRE_Int n, HYPRE_Int* mask, HYPRE_Real* v )
 {
 
    HYPRE_Int i;
@@ -154,7 +154,7 @@ lobpcg_sqrtVector( HYPRE_Int n, HYPRE_Int* mask, HYPRE_Real* v )
    for ( i = 0; i < n; i++ )
       if ( mask == NULL || mask[i] )
       {
-         v[i] = sqrt(v[i]);
+         v[i] = hypre_sqrt(v[i]);
       }
 }
 
@@ -653,7 +653,7 @@ lobpcg_solve( mv_MultiVectorPtr blockVectorX,
                                           NULL, sizeX,
                                           utilities_FortranMatrixValues( residualNorms ) );
 
-         lobpcg_sqrtVector( sizeX, NULL,
+         lobpcg_hypre_sqrtVector( sizeX, NULL,
                             utilities_FortranMatrixValues( residualNorms ) );
 
          if ( lambdaHistory != NULL )
@@ -961,7 +961,7 @@ lobpcg_solve( mv_MultiVectorPtr blockVectorX,
       mv_MultiVectorByMultiVectorDiag(   blockVectorR, blockVectorR,
                                          activeMask, sizeX,
                                          utilities_FortranMatrixValues( residualNorms ) );
-      lobpcg_sqrtVector(  sizeX, activeMask,
+      lobpcg_hypre_sqrtVector(  sizeX, activeMask,
                           utilities_FortranMatrixValues( residualNorms ) );
 
       i = *iterationNumber + 1;
--- a/src/krylov/lobpcg.c
+++ b/src/krylov/lobpcg.c
@@ -146,7 +146,7 @@ lobpcg_MultiVectorImplicitQR(
 }
 
 static void
-lobpcg_hypre_sqrtVector( HYPRE_Int n, HYPRE_Int* mask, HYPRE_Real* v )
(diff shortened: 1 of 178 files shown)


victorapm (approved):
That was a lot of work and some bugs fixed! Thanks!

src/lapack/dlasq1.c
```diff
@@ -157,7 +157,7 @@ extern "C" {
 
     eps = dlamch_("Precision");
     safmin = dlamch_("Safe minimum");
-    scale = sqrt(eps / safmin);
+    scale = hypre_sqrt(eps / safmin);
```
rfalgout: You don't need to change all of the blas/lapack files to use 'hypre_sqrt', and we generally try to minimize changes to these files directly since they are not written by us.  Instead, use 'f2c.h' to change 'sqrt' appropriately.  It is already defined there, actually.  This will automatically update all of the '.c' files since they all include 'f2c.h'.  The 'blas/README' file explains how to add new files to these directories, FYI.

src/lapack/f2c.h
```diff
@@ -22,11 +22,13 @@
 #include "_hypre_utilities.h"
 #include "math.h"
 
+/*
```
rfalgout: Why did you comment this out?  Why not use it instead and redefine with 'hypre_sqrt'?

rfalgout (approved):
Hi @oseikuffuor1 .  This was a lot of work!  And you definitely fixed what looks like a number of bugs, so that's awesome too!  This all looks good to me, modulo my comments about the blas and lapack directories.  Thanks!

src/parcsr_ls/par_scaled_matnorm.c
```diff
@@ -35,7 +35,7 @@ hypre_ParCSRMatrixScaledNorm( hypre_ParCSRMatrix *A, HYPRE_Real *scnorm)
    HYPRE_BigInt           *row_starts = hypre_ParCSRMatrixRowStarts(A);
    HYPRE_Int       num_rows = hypre_CSRMatrixNumRows(diag);
 
-   hypre_ParVector      *dinvsqrt;
+   hypre_ParVector      *dinvhypre_sqrt;
```
ulrikeyang: Shouldn't we revert this to dinvsqrt?  This is the name of a vector.

src/seq_mv/csr_matop.c
```diff
@@ -1910,8 +1910,8 @@ hypre_CSRMatrixComputeRowSum( hypre_CSRMatrix *A,
  * type 0: diag
  *      1: abs diag
  *      2: diag inverse
- *      3: diag inverse sqrt
- *      4: abs diag inverse sqrt
+ *      3: diag inverse hypre_sqrt
```
ulrikeyang: This also should just be sqrt, since it just describes what is done.

ulrikeyang (approved):
Thanks for doing this. This was a lot of work.

src/lapack/dlasq1.c
oseikuffuor1: Thanks @falgout. I'll make that change

src/lapack/f2c.h
oseikuffuor1: Good point.

src/seq_mv/csr_matop.c
oseikuffuor1: Thanks for catching these @ulrikeyang. My script was a bit too aggressive :-).

liruipeng (approved):
A lot of work, which looks good to me. Thank you @oseikuffuor1 !

src/krylov/lobpcg.c
```diff
@@ -146,15 +146,15 @@ lobpcg_MultiVectorImplicitQR(
 }
 
 static void
-lobpcg_sqrtVector( HYPRE_Int n, HYPRE_Int* mask, HYPRE_Real* v )
+lobpcg_hypre_sqrtVector( HYPRE_Int n, HYPRE_Int* mask, HYPRE_Real* v )
```
waynemitchell: Here's another unnecessary name change similar to what Ulrike pointed out elsewhere.

liruipeng: I thought @oseikuffuor1  wanted the change, but it does seem unnecessary to me too.

oseikuffuor1: Thanks @waynemitchell, it was a well-hidden change I missed. Thanks for catching it.

waynemitchell (approved):
Looks good except my one minor comment above. Thanks, @oseikuffuor1 !

osborn9 (approved):
Looks good to me!

PR #829, opened 2023-02-03
Add vector resizing functions

victorapm:
This PR adds `hypre_SeqVectorResize` and `hypre_ParVectorResize` for resizing sequential and parallel vectors, respectively. Data reallocation takes place when the total size (`size * num_vectors`) passed as input is greater than the space currently allocated for the data member of a `hypre_Vector`.

This is useful for block-Krylov solvers/eigensolvers using BoomerAMG and multicomponent vectors.

cc @prj-

### DIFF
--- a/src/seq_mv/vector.c
+++ b/src/seq_mv/vector.c
@@ -162,7 +162,7 @@ hypre_SeqVectorSetSize( hypre_Vector *vector,
  * hypre_SeqVectorResize
  *
  * Resize a sequential vector by either changing its number of components
- * or size.
+ * or size (per component).
  *--------------------------------------------------------------------------*/
 
 HYPRE_Int
--- a/src/seq_mv/vector.c
+++ b/src/seq_mv/vector.c
@@ -161,20 +161,18 @@ hypre_SeqVectorSetSize( hypre_Vector *vector,
 /*--------------------------------------------------------------------------
  * hypre_SeqVectorResize
  *
- * Resize a sequential vector by either changing its number of components
- * or size (per component).
+ * Resize a sequential vector when changing its number of components.
  *--------------------------------------------------------------------------*/
 
 HYPRE_Int
 hypre_SeqVectorResize( hypre_Vector *vector,
-                       HYPRE_Int     num_vectors_in,
-                       HYPRE_Int     size_in )
+                       HYPRE_Int     num_vectors_in )
 {
    HYPRE_Int  method        = hypre_VectorMultiVecStorageMethod(vector);
    HYPRE_Int  size          = hypre_VectorSize(vector);
    HYPRE_Int  num_vectors   = hypre_VectorNumVectors(vector);
    HYPRE_Int  total_size    = num_vectors * size;
-   HYPRE_Int  total_size_in = num_vectors_in * size_in;
+   HYPRE_Int  total_size_in = num_vectors_in * size;
 
    /* Reallocate data array */
    if (total_size_in > total_size)
@@ -188,7 +186,6 @@ hypre_SeqVectorResize( hypre_Vector *vector,
    }
 
    /* Update vector info */
-   hypre_VectorSize(vector) = size_in;
    hypre_VectorNumVectors(vector) = num_vectors_in;
    if (method == 0)
    {
--- a/src/seq_mv/vector.c
+++ b/src/seq_mv/vector.c
@@ -159,7 +159,48 @@ hypre_SeqVectorSetSize( hypre_Vector *vector,
 }
 
 /*--------------------------------------------------------------------------
- * ReadVector
+ * hypre_SeqVectorResize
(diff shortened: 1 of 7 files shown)


src/seq_mv/vector.c
```diff
@@ -159,7 +159,53 @@ hypre_SeqVectorSetSize( hypre_Vector *vector,
...
+ * hypre_SeqVectorResize
+ *
+ * Resize a sequential vector by either changing its number of components
+ * or size (per component).
+ *--------------------------------------------------------------------------*/
+
+HYPRE_Int
+hypre_SeqVectorResize( hypre_Vector *vector,
+                       HYPRE_Int     num_vectors_in,
+                       HYPRE_Int     size_in )
```
rfalgout: What happens if `size_in` is different from `size`?  Won't the data layout be messed up for the existing vector components?  Do you really want to allow the size to change?  If not, then you probably should remove it from the argument list.  For `num_vectors`, do you want to allow this to both increase and decrease in size?  I suppose there is no harm in that.

victorapm: Hi Rob, thank you for the comments!

> What happens if size_in is different from size? Won't the data layout be messed up for the existing vector components?

If `size_in > size`, the data array will be reallocated; if `size_in <= size`, the data array won't be reallocated because we already have enough space. In both cases, the members `idxstride` and `vecstride` are updated according to the storage (row-major/col-major). So, the data layout would not be messed up.

> Do you really want to allow the size to change?

I think it makes sense to allow the size to change, because, in theory, a vector could be resized due to two reasons: change in size or change in number of components.

>  For `num_vectors`, do you want to allow this to both increase and decrease in size?

If I understand the question correctly, yes, we need to support that. We need to  allow increase because BoomerAMG can be setup with `NULL` pointers for the vectors, which assumes vectors to have a single component. Later when calling `BoomerAMGSolve` from a block Krylov solver, we need to adjust temporary work vectors to have multiple components, thus increase in size. Now, for decrease in size: some block Krylov solver implementations reduce the space of vectors when individual components converge, this means decreasing `num_vectors` along the solve. Note I chose to not reallocate space for the data array in this scenario since we already have enough space. However, this approach is questionable because we might be keeping memory space that we don't need.

victorapm: Q1. Sorry! I see what you mean, what I've done is not sufficient. I would need to implement some data copies to make sure the data layout is not messed up

rfalgout: Maybe I am missing something, but imagine you have 2 vectors of size 10 in the vector originally, and the data is stored consecutively in the array (i.e., 10 doubles for the first vector followed directly by the second 10).  Now, call Resize with `size_in=15` and `num_vectors=2`.  The first vector has 10 values from the previous vector plus another 5 from the previous second vector, and the second vector has the last 5 from the previous second vector and uninitialized memory for its last 10 values  There is really no reason to do a realloc in this case, since the data is not useful (it would be cheaper to just do free/alloc).

In the case of `num_vectors_in != num_vectors`, realloc is meaningful.

Let me know if I am misunderstanding.  Thanks!

rfalgout: Our comments crossed. :)  You just need to decide if you want this routine to always keep the data intact somehow.  I don't see how that is possible in all cases when changing `size`.

victorapm: Yes! Let me think more about whether I want this routine to support increase in sizes without messing up with existing data or not :)

I hope the answers to the other questions make sense!

rfalgout: My guess is that changing `size` is not a use case we will come across (unlike in the semi-structured code where we want to add/remove ghost regions).  If you can't think of a good reason to do this, then just remove `size_in` from the argument list instead of trying to implement a bunch of data copying.  Just my 2 cents.  Thanks!

rfalgout (approved):
Looks good modulo our discussion about `size_in`.  Thanks!

src/seq_mv/vector.c
victorapm: Yes, that makes a lot of sense. Trying to solve a problem that doesn't exist is not a great idea :)

prj-:
Thanks for the PR. For reference, here is the corresponding PETSc MR: https://gitlab.com/petsc/petsc/-/merge_requests/6022.

src/seq_mv/vector.c
victorapm: Hi Rob, I couldn't think of a good reason to support changing on `size` now, so I updated the `VectorResize` functions to depend only on `num_vectors` [fb91729](https://github.com/hypre-space/hypre/pull/829/commits/fb9172902d2f273aa985cc0c8b0ac42e1935c733). Thanks!

ulrikeyang (approved):
This looks to me. Thanks.

PR #833, opened 2023-02-07
Update Tioga Regressions

waynemitchell:
The way we are passing the mpibind option to slurm for the tioga regressions is no longer correct and causing issues. Slurm is now functioning as a wrapper for flux on Tioga, and plain srun commands (without any mpibind arguments) produces the correct behavior. In addition, I found that some of the benchmark runs are now going consistently faster than the recorded .saved files for tioga, so these have been updated as well.


liruipeng (approved):
Looks good. Thank you!

PR #837, opened 2023-02-15
Report error on hypre calls outside of HYPRE_Init() - HYPRE_Finalize()

### LINKED ISSUE #832: Hypre segfault when calling hypre_DeviceDataStream when using SYCL
When using Hypre with SYCL 

If one calls : HYPRE_Finalize()

before calling: hypre_PCGDestroy (or similar)

https://github.com/hypre-space/hypre/blob/832ad94de321b583424a958b4659a7bbe7204676/src/utilities/device_utils.c#L415

the sycl_device will be invalid and result in a seg fault down stream.

The issue manifests itself when one relies on reference counted pointers to hold Hypre structures which may be released data AFTER HYPRE_Finalize() is called.

Fix: An assert should be called that catches that the sycl device is invalid (nullptr) due more than likely because HYPRE_Init() has not been called or HYPRE_Finalize() has already been called.

This error does not manifest itself with CUDA or HIP only with SYCL.

waynemitchell:
This changes the behavior of calls to hypre_handle() to report an error if the handle is not initialized. Note that there are many hypre functions that do not call hypre_handle() underneath, so there is no easy way to do error reporting for these cases.

This also calls HYPRE_Init() instead of only hypre_HandleCreate() in the case where hypre_handle() is called when the handle is not initialized. This eliminates a seg fault in the case of sycl, where the GPU device needs to be setup (which happens in HYPRE_Init() but not hypre_HandleCreate()).


victorapm (approved):
Thank you!!

oseikuffuor1 (approved):
Looks good to me. Thanks.

rfalgout (approved):
Looks good.  Thanks!

osborn9 (approved):

liruipeng (approved):

ulrikeyang (approved):
Thanks. this looks good.

prj-:
How to check whether HYPRE has been initialized/finalized or not to know whether it's legal or not to call routines other than `HYPRE_Init()`? I cannot find a routine such as `MPI_Initialized()` or `PetscInitialized()`, so how is this MR supposed to work?

knepley:
@liruipeng As Pierre says above, it looks like we need a `HypreInitialized()` to obey the new rules in the PETSc interface.

liruipeng:
> @liruipeng As Pierre says above, it looks like we need a `HypreInitialized()` to obey the new rules in the PETSc interface.

Sounds good. This shouldn't be hard to have. Thanks @knepley !

PR #844, opened 2023-02-23
Fix memory tracker with omp

liruipeng:
This PR fixes hypre memory tracker with OMP threading. It seems that adding `omp critical` fixes the problem.

### DIFF
--- a/src/utilities/memory_tracker.c
+++ b/src/utilities/memory_tracker.c
@@ -16,6 +16,25 @@
 
 #if defined(HYPRE_USING_MEMORY_TRACKER)
 
+hypre_MemoryTracker *_hypre_memory_tracker = NULL;
+
+/* accessor to the global ``_hypre_memory_tracker'' */
+hypre_MemoryTracker*
+hypre_memory_tracker(void)
+{
+#ifdef HYPRE_USING_OPENMP
+#pragma omp critical
+#endif
+   {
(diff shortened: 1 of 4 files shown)


ulrikeyang (approved):
This looks good. Thanks

victorapm (approved):

rfalgout (approved):
Looks good.  Thanks!

src/utilities/memory_tracker.c
```diff
@@ -252,44 +271,49 @@ hypre_MemoryTrackerInsert2(const char           *action,
...
+      sprintf(entry->filename, "%s", filename);
+      sprintf(entry->function, "%s", function);
+      entry->line = line;
+      entry->pair = (size_t) -1;
 
 #if 0
-   HYPRE_Int myid;
-   hypre_MPI_Comm_rank(hypre_MPI_COMM_WORLD, &myid);
-   if (myid == 0 && entry->time_step == 28111) {assert(0);}
+      HYPRE_Int myid;
```
victorapm, line 306: Do we want to keep this?

oseikuffuor1 (approved):
Looks good. Thanks.

src/utilities/memory_tracker.c
liruipeng: yeah. maybe. some debugging code that is useful when finding leak.

PR #849, opened 2023-03-08
Add hypre_assert_exit

victorapm:
This PR proposes an implementation for `hypre_assert_exit` as discussed [here](https://github.com/hypre-space/hypre/pull/840#discussion_r1113709360). If hypre hits an error, e.g., when doing a CUDA call, it will exit from the application only when `HYPRE_DEBUG` is defined instead of doing so always as we have now.

### DIFF
--- a/src/utilities/_hypre_utilities.h
+++ b/src/utilities/_hypre_utilities.h
@@ -231,27 +231,37 @@ void hypre_error_handler(const char *filename, HYPRE_Int line, HYPRE_Int ierr, c
 #if defined(HYPRE_DEBUG)
 /* host assert */
 #define hypre_assert(EX) do { if (!(EX)) { fprintf(stderr, "[%s, %d] hypre_assert failed: %s\n", __FILE__, __LINE__, #EX); hypre_error(1); assert(0); } } while (0)
+
+/* host assert and exit */
+#define hypre_assert_exit(comm, EX) do { if (!(EX)) { hypre_assert(EX); hypre_MPI_Abort(comm, 1); } } while (0)
+
 /* device assert */
 #if defined(HYPRE_USING_CUDA)
 #define hypre_device_assert(EX) assert(EX)
-#elif defined(HYPRE_USING_HIP)
+
 /* FIXME: Currently, asserts in device kernels in HIP do not behave well */
+#elif defined(HYPRE_USING_HIP)
 #define hypre_device_assert(EX) do { if (0) { static_cast<void> (EX); } } while (0)
+
 #elif defined(HYPRE_USING_SYCL)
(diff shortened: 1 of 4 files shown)


rfalgout:
Hi @victorapm .  In looking at the definition of `assert` in C, it looks like it always exits when it evaluates to `false`.  So, I wonder if this should be the normal behavior of the `hypre_assert` macro (instead of adding a new one).

src/utilities/_hypre_utilities.h
```diff
@@ -231,27 +231,37 @@ void hypre_error_handler(const char *filename, HYPRE_Int line, HYPRE_Int ierr, c
 #if defined(HYPRE_DEBUG)
 /* host assert */
 #define hypre_assert(EX) do { if (!(EX)) { fprintf(stderr, "[%s, %d] hypre_assert failed: %s\n", __FILE__, __LINE__, #EX); hypre_error(1); assert(0); } } while (0)
+
+/* host assert and exit */
+#define hypre_assert_exit(comm, EX) do { if (!(EX)) { hypre_assert(EX); hypre_MPI_Abort(comm, 1); } } while (0)
```
rfalgout, line 236: Since `hypre_assert` calls `assert(0)` which exits, then it's too late to call `MPI_Abort` right?

victorapm:
Hi @rfalgout, thanks for the additional info. I had missed the fact that `assert` was already calling abort. So, the work in this PR is not really necessary. Perhaps we can simply remove the instances of `exit` from the code and that's enough to be in compliance with policy R3?

rfalgout:
> Hi @rfalgout, thanks for the additional info. I had missed the fact that `assert` was already calling abort. So, the work in this PR is not really necessary. Perhaps we can simply remove the instances of `exit` from the code and that's enough to be in compliance with policy R3?

Maybe so.  I am a little uneasy about `HYPRE_DEBUG` serving as the switch here even for `hypre_assert`.  Sometimes it's useful to add the `-g` flag to the compile line even when optimization is turned on.  But, I suppose that can be done without using the `--enable-debug` configure option, so I guess this is probably okay.

victorapm:
Yes, I also find it useful to have `CFLAGS='-O2 -g'` without necessarily having `--enable-debug`. Sometimes I do this when debugging. If you are fine with it, I will close this PR and let the work on removing `exit` calls from hypre to another PR.

rfalgout:
> Yes, I also find it useful to have `CFLAGS='-O2 -g'` without necessarily having `--enable-debug`. Sometimes I do this when debugging. If you are fine with it, I will close this PR and let the work on removing `exit` calls from hypre to another PR.

Sounds good.  Thanks!

Issue #858, opened 2023-03-16
A misprint in the manual


Tomcat4891:
Hi.

On this manual page:

https://hypre.readthedocs.io/en/latest/solvers-ams.html

In the sentence
********************************
The Poisson matrices  A_alpha and A_beta , corresponding to assembling of the forms (alpha grad u, grad v)+(beta grad u, grad v)
.......
********************************
The variational definition of the A_alpha matrix is wrong. It should be  
(alpha grad u, grad v)+(beta u, v)

See the paper:

https://global-sci.org/intro/article_detail/jcm/8592.html

page 614 , case A

BR, 
Andrew

tzanio:
Good catch @Tomcat4891, this should indeed be `(alpha grad u, grad v)+(beta u, v)`

rfalgout:
Thanks @Tomcat4891 .  We will fix.

Issue #857, opened 2023-03-16
Incorrect data type in a function


Tomcat4891:
Hi.

Please check the function below ( ams.c file  )  . 
Why is the third parameter (A_cheby_fraction)  of type INTEGER?

HYPRE_Int hypre_AMSSetChebySmoothingOptions(void *solver,
                                            HYPRE_Int A_cheby_order,
                                            HYPRE_Int A_cheby_fraction)  //    <<<<<<< WHY INTEGER?
{
   hypre_AMSData *ams_data = (hypre_AMSData *) solver;
   ams_data -> A_cheby_order =  A_cheby_order;
   ams_data -> A_cheby_fraction =  A_cheby_fraction; //         <<<<<<< (ams_data -> A_cheby_fraction)  IS REAL !

   return hypre_error_flag;
}

BR,
Andrew

victorapm:
@Tomcat4891 thank you for pointing out the issues!

PR #859, opened 2023-03-16
Auxiliary space preconditioners fix

victorapm:
This PR fixes the data type of `cheby_fraction` used as input in the `SetChebySmoothingOptions` functions of AMS and ADS. It also fixes a comment about AMS in the documentation.

This PR addresses GitHub issues #857 and #858.


src/docs/usr-manual/solvers-ams.rst
```diff
@@ -68,9 +68,9 @@ In addition to the above quantities, AMS can utilize the following (optional)
 information:
 
 * The Poisson matrices :math:`A_\alpha` and :math:`A_\beta`, corresponding to
```
rfalgout: I suppose you can remove this change now from the PR

rfalgout (approved):
Looks good.  Should we add a few tests before merging this?

src/docs/usr-manual/solvers-ams.rst
victorapm: Yes, done at [79d5d84](https://github.com/hypre-space/hypre/pull/859/commits/79d5d84f3a91b65b9acb9888c15476d5b542b874)

victorapm:
Thanks! I was thinking about another PR for the tests. It will involve adding a public `HYPRE` function, but I am fine adding the tests in this PR

rfalgout:
> Thanks! I was thinking about another PR for the tests. It will involve adding a public `HYPRE` function, but I am fine adding the tests in this PR

It's fine to add tests in another PR.  Why do you need a new function (and what will it do)?

victorapm:
Thank you! I need to add `HYPRE_AMSSetChebySmoothingOptions`. Note we already have this for ADS: https://github.com/hypre-space/hypre/blob/c3ce7b9f17cbf8019727dd4a4214bdd62815ef05/src/parcsr_ls/HYPRE_parcsr_ls.h#L2548

It is setting the input parameters to Chebyshev. Since we are already looking at this, I think it is a good idea to revise the interface. The function does not consider all input parameters that can affect Chebyshev setup, e.g., it does not have `cheby_variant`. In BoomerAMG, we have separate functions for setting each of the input parameters (order, variant, fraction...). Should we have the same for AMS and ADS? Or add a general function that sets all chebyshev parameters for BoomerAMG as well? I think we can consider these questions.

victorapm:
Note: The function I was talking about already exists here https://github.com/hypre-space/hypre/blob/c3ce7b9f17cbf8019727dd4a4214bdd62815ef05/src/parcsr_ls/HYPRE_ams.c#L232

Just need to add it to `HYPRE_parcsr_ls.h`. Still, I think we can discuss the questions above :)

rfalgout:
At a minimum, adding the existing function to the documentation makes sense.  We should bring some others into the discussion about changing the interface (and maybe add it to the agenda for our next meeting).

victorapm:
Yes, makes sense. Thank you!

PR #869, opened 2023-03-22
superlu_dist support on GPUs

liruipeng:
This PR adds superlu_dist support on GPUs.

- [x] tux
- [x] lassen
- [x] tioga
- [x] sunspot


rfalgout (approved):
Looks good to me.  Should we add a GPU test as well, like what we do in `machine-tux-spack.sh` currently?  This might be a bit more complicated though.  Just thinking out loud.

liruipeng:
> Looks good to me. Should we add a GPU test as well, like what we do in `machine-tux-spack.sh` currently? This might be a bit more complicated though. Just thinking out loud.

That will be great to have. I will work with @osborn9 on it. Thanks!

ulrikeyang (approved):
This looks good. Thanks.

osborn9 (approved):
I've been able to successfully run the ij driver using superlu-dist as coarse grid solver on Lassen. Nice work!! 

The outstanding issue (which I'm working on but don't have a complete solution yet) is adding a lassen-spack test to test with the most recent superlu-dist master. This can be added as a separate MR.

oseikuffuor1 (approved):
Changes look fine to me. Thanks Rui.

PR #879, opened 2023-04-11
Cuda versions

liruipeng:
This PR adds support and regression tests for all the versions from CUDA 9.0 to 12.0.

- [x] tux
- [x] lassen
- [x] tioga
- [x] sunspot

A new test `machine-lassen-cuda.sh` is added, which takes about 1 hour on lassen. We should run it in the regression to make sure all versions of CUDA are OK.


liruipeng:
Should go after #871. Maybe also better to review after #871 being merged.

rfalgout (approved):
Looks good to me.  Thanks!

victorapm (approved):
Thanks, Rui Peng!

src/test/TEST_error/error.saved
victorapm, line 1: @liruipeng I also need these changes in the `ilu-amd` branch. Are you close to merge `cuda_versions` into master? If that's the case, I would wait for your merge

liruipeng: will do it tonight . Thanks!

PR #880, opened 2023-04-12
Remove comparisons between HYPRE_Real and HYPRE_Complex

victorapm:
This fixes the `machine-tux` regression tests

### DIFF
--- a/src/parcsr_ls/par_mgr_solve.c
+++ b/src/parcsr_ls/par_mgr_solve.c
@@ -128,7 +128,7 @@ hypre_MGRSolve( void               *mgr_vdata,
       if (logging > 1)
       {
          hypre_ParVectorCopy(F_array[0], residual );
-         if (tol > fp_zero)
+         if (tol > hypre_cabs(fp_zero))
          {
             hypre_ParCSRMatrixMatvec(fp_neg_one, A_array[0], U_array[0], fp_one, residual);
          }
@@ -137,7 +137,7 @@ hypre_MGRSolve( void               *mgr_vdata,
       else
       {
          hypre_ParVectorCopy(F_array[0], Vtemp);
-         if (tol > fp_zero)
+         if (tol > hypre_cabs(fp_zero))
          {
             hypre_ParCSRMatrixMatvec(fp_neg_one, A_array[0], U_array[0], fp_one, Vtemp);
          }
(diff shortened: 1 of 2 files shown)


rfalgout (approved):
Looks good.  Thanks!

src/parcsr_ls/par_mgr_solve.c
```diff
@@ -128,7 +128,7 @@ hypre_MGRSolve( void               *mgr_vdata,
       if (logging > 1)
       {
          hypre_ParVectorCopy(F_array[0], residual );
-         if (tol > fp_zero)
```
liruipeng, line 131: Maybe it's too late to ask: why `fp_zero` is not `HYPRE_Real`?

victorapm: Because it's used as input argument to functions that expect `HYPRE_Complex`. I could have written `0.0` explicitly here to avoid adding `hypre_cabs`. I ended up doing the former one, but both solutions looked fine to me

victorapm: BTW, we could have something like `HYPRE_COMPLEX_ONE`, `HYPRE_REAL_ONE`, `HYPRE_COMPLEX_ZERO`, and so on to avoid having to define constants like this... What do you think?

oseikuffuor1: I think if you are defining the variable as HYPRE_Complex it is best to stick with using hypre_cabs, otherwise it may trigger warnings with-strict-checking on. I also think it helps to be consistent in that regard (hypre_cabs for HYPRE_Complex variables -- even if they are used as HYPRE_Real). Defining special variables HYPRE_COMPLEX_ONE, etc as you suggest seems reasonable to me. I am not sure if they are used a lot. We just need to be careful that their use is actually for a constant, and not a variable if making changes.

victorapm: Good point about the strict checking test. My motivation for defining `HYPRE_COMPLEX_ONE` and others alike is that we create local variables for assigning constant values such as 0 and 1 in multiple parts of hypre, e.g., pretty much all of the solve functions for solvers and preconditioners. If we have a `#define HYPRE_REAL_ONE 1.0` for example, we can avoid having to define local variables for constants in many places of hypre.

rfalgout: After looking at this a bit more closely, I think this should just be written out as `tol > 0.0` because tolerances such as `tol` should always be real.  I can't think of a situation where it should be anything else, anyway.  I'm not necessarily against the idea of introducing some of the other constants, but I'm not sure it's needed for this PR.  Just my 2 cents.

victorapm: Sounds good! I have another PR about MGR coming soon and I can make the `tol > 0.0` change over there. Let me know if you guys find the constants idea I gave above to be useful and I can create a separate PR for doing that

liruipeng: Agreed with changing to `tol > 0.0`.

PR #899, opened 2023-05-08
Remove deprecated warning from Init

victorapm:
When using `HYPRE_Init`, the compiler no longer gives a deprecated function warning message.


rfalgout (approved):
Looks good.  Thanks!

PR #900, opened 2023-05-09
Change sh to bash

### LINKED ISSUE #881: shell scripts use `function` keyword which is not POSIX compliant
The shell scripts in hypre, e.g. [runtest.sh](https://github.com/hypre-space/hypre/blob/master/src/test/runtest.sh) use the `function` keyword, which is not in the POSIX shell standard. `function` was first defined in ksh, and is used by bash and zsh.

The scripts use the hashbang `#!/bin/sh`.  This creates a problem if `/bin/sh` points at a POSIX-strict shell such as dash, which is the case for debian-based linux systems for instance.  In this case runtest.sh fails with
```
./runtest.sh: 30: function: not found
```
(meaning the command "function" was not found)

I encountered the problem running hypre tests via runtest.sh on a Debian system, but all hypre scripts (*.sh and *job) may be affected, not just runtest.sh.

Should the `function` keyword be dropped and other necessary edits done to make all scripts POSIX-compliant?
Or should the hashbang be changed to `#!/bin/bash`?

victorapm:
Change shell scripts from `#!/bin/sh` to `#!/bin/bash`

Tests:
- [x] tux
- [x] lassen
- [x] tioga
- [x] sunspot
- [x] Debian
- [x] Mac


rfalgout:
I'm a little nervous about changing all these files.  I thought we would only change the 3 in `src/test`.  On the surface, I don't see an issue using `/bin/bash` but we really need to test this, just in case I'm missing something.

rfalgout:
> I'm a little nervous about changing all these files. I thought we would only change the 3 in `src/test`. On the surface, I don't see an issue using `/bin/bash` but we really need to test this, just in case I'm missing something.

FYI, I'm not suggesting we should change anything back just yet.

victorapm:
I understand! My reasoning was to make the change everywhere since we were already dealing with it for a few files. Tux tests are passing, should I try the other ones?

rfalgout:
> I understand! My reasoning was to make the change everywhere since we were already dealing with it for a few files. Tux tests are passing, should I try the other ones?

Yes, we should test everything.  We should also ask the person who submitted the Debian issue to test at some point (not yet).  I can help with running tests.

I've also asked @oseikuffuor1 to review, since he has done some script writing in the past and had to deal with various portability issues.

victorapm:
@waynemitchell can you double-check this on florentia?

dbarbier:
Hello, another option is to fix bashisms: in checktest.sh and cleantest.sh, you can replace shopt by
```sh
for testdir in ${@:-TEST*}
do
   [ -d "$testdir" ] || continue
```

In runtest.sh, replace all function definitions like
```sh
function usage
```
by
```
usage ()
```
There is another caveat in this file, type in dash does not support -p option:
```sh
$ type -p mpirun
-p: not found
mpirun is /usr/bin/mpirun
```
and thus RunPrefix contains garbage.  But I do not understand why you don't simply set `RunPrefix=mpirun`, it will just work.  If you really need full path,
```sh
RunPrefix=`command -v mpirun`
```
is more portable.

victorapm:
Hi @dbarbier, thanks for the info! Can you test this branch in your system and let us know if it works fine?

drew-parsons:
This patch passes Debian tests when applied to 2.29.0 source.

waynemitchell (approved):

dbarbier:
> Hi @dbarbier, thanks for the info! Can you test this branch in your system and let us know if it works fine?

It works, but I do not understand why you prefer modifying 240 files instead of fixing bashisms in 2 files.

victorapm:
@dbarbier,

Thank you for taking your time and proposing another fix to this issue.

I agree that fixing "bashisms" and keeping `/bin/sh` is a more general fix than using `/bin/bash`. However, all HPC systems targeted by hypre, for testing purposes, have `/bin/bash` installed. Thus, the fix in this branch seems enough to me. Having said that, I will leave to the reviewers of this PR, @oseikuffuor1 and @rfalgout, to decide which fix to keep.

Lastly, I wanted to mention that these changes impact mostly developers, who run the regression tests in hypre very often, and not general users, who are mostly interested in building hypre and using it. In fact, the `configure` script used for building hypre still uses `/bin/sh`.

Thank you!

rfalgout (approved):
Looks great!  Thanks!

oseikuffuor1 (approved):
Thanks Victor!

PR #904, opened 2023-05-13
SYCL MGR, AMS, AIR, etc.

waynemitchell:
This PR enables more functionality with sycl:
1. MGR
2. AMS/ADS/AME
3. AIR restriction


src/test/TEST_ij/matrix.saved.sunspot
```diff
@@ -24,11 +24,11 @@ Final Relative Residual Norm = 4.233149e-01
 
 # Output file: matrix.out.6
 Iterations = 5
-Final Relative Residual Norm = 4.233149e-01
+Final Relative Residual Norm = 3.515462e-01
 
 # Output file: matrix.out.7
 Iterations = 5
-Final Relative Residual Norm = 4.233149e-01
+Final Relative Residual Norm = 3.515462e-01
```
rfalgout, line 31: These changes seem large.  Is this expected?

rfalgout (approved):
Looks good to me.  Thanks!

src/test/TEST_ij/matrix.saved.sunspot
waynemitchell: This is due to a bug fix in IJ matrix assembly for sycl:
https://github.com/hypre-space/hypre/pull/904/files#diff-7ddf8e4e5bf557900adda9b6a7bf028a61225dbbbc09b2018c575174aa717470L304
I previously had some arguments reversed... I guess that the correctness of the IJ matrix assembly is not tested in the regressions? If you run the ij driver with -test_ij, I guess it just tests that the code runs, not necessarily that you get the correct result, since I had missed this bug up to now.

liruipeng: Yes, good catch. We should add some regression tests. We can use ij.c or ij_assembly.c that is just for assembly. Thanks!

src/parcsr_ls/ads.c
```diff
@@ -5,22 +5,32 @@
...
 #include "_hypre_parcsr_ls.h"
 #include "float.h"
 #include "ams.h"
 #include "ads.h"
 #include "_hypre_utilities.hpp"
 
-#if defined(HYPRE_USING_CUDA) || defined(HYPRE_USING_HIP)
+#if defined(HYPRE_USING_GPU)
+#if defined(HYPRE_USING_SYCL)
+SYCL_EXTERNAL
```
liruipeng, line 17: Do we need `SYCL_EXTERNAL` because the kernels are called from another file?

liruipeng (approved):
A lot of work! Thanks, @waynemitchell.

src/parcsr_ls/ads.c
waynemitchell: Yes, since these kernels are defined in `ams.c`, if I don't include `SYCL_EXTERNAL` here, I get:
error: SYCL kernel cannot call an undefined function without SYCL_EXTERNAL attribute

PR #905, opened 2023-05-13
oneDPL fixes and Sunspot regressions

waynemitchell:
This PR contains some fixes needed due to recent changes with oneDPL. I've also moved sycl regression testing to the Sunspot platform and updated the regression script and results accordingly.


src/seq_mv/csr_spgemm_device_symbl.h
```diff
@@ -389,8 +389,8 @@ hypre_spgemm_symbolic( hypre_DeviceItem             &item,
          }
          else
          {
-            failed = (char) group_reduce_sum<hypre_int, NUM_GROUPS_PER_BLOCK, GROUP_SIZE>(item,
```
liruipeng, line 397: Why do we need to change to `HYPRE_Int`?

rfalgout (approved):
Looks good to me.  Thanks!

src/seq_mv/csr_spgemm_device_symbl.h
waynemitchell: This call takes in `s_HashKeys` as an argument, which has type `HYPRE_Int`. In the case of the bigint build, this is not the same as `hypre_int`, so you get an error at compile time without this change.

liruipeng: @waynemitchell This may be fine for now.  The reason I don't like it is due to the fact that the reduction type is just a `char`. Casting it to `int` is an overkill already, I guess. But we can leave the change as is, since it only takes effect for `BigInt` build on GPUs, which I plan to work on later. Thank you for clarifying it for me.

liruipeng (approved):
Looks good to me. Thanks!

PR #910, opened 2023-05-23
Add soname for shared library

### LINKED ISSUE #907: Set SOVERSION and VERSION when building as shared library in CMake
For shared library, Adding VERSION and SOVERSION can specify the build version and api version. It is useful to indicate that the ABI compatibility, will be beneficial for the packaging of various Linux distributions.

Ref:
https://tldp.org/HOWTO/Program-Library-HOWTO/shared-libraries.html
https://docs.fedoraproject.org/en-US/packaging-guidelines/#_downstream_so_name_versioning
https://www.debian.org/doc/debian-policy/ch-sharedlibs.html#run-time-shared-libraries
https://en.opensuse.org/openSUSE:Shared_library_packaging_policy#Underlying_principles

topazus:
resolve #907
Ref: https://tldp.org/HOWTO/Program-Library-HOWTO/shared-libraries.html

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -422,6 +422,11 @@ endif ()
 # Create the HYPRE library object
 add_library(${PROJECT_NAME})
 
+set_target_properties(${PROJECT_NAME} PROPERTIES
+  VERSION "${HYPRE_VERSION}"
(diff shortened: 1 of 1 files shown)


victorapm:
Thank you for your contribution! The contents of this PR have been included in #1181, which has been merged into master

PR #913, opened 2023-06-03
Block inversion with rocSOLVER in MGR

victorapm:
This PR adds HIP support to exact block inversion in MGR via rocSOLVER's batched routines.


oseikuffuor1 (approved):
Looks good. Thanks Victor.

PR #920, opened 2023-06-07
New error handling feature to print messages to memory

rfalgout:
Hi All,

This allows users to direct hypre's error messages to a memory buffer instead of stderr.  One of our users here requested this feature because it allows more control over the output, especially when there are large numbers of MPI tasks where things can get unwieldy.  Here are the new user API routines:
```
/* Set behavior for printing errors: mode 0 = stderr, mode 1 = memory buffer */
HYPRE_Int HYPRE_SetPrintErrorMode(HYPRE_Int mode);

/* Return a buffer of error messages and clear them in hypre */
HYPRE_Int HYPRE_GetErrorMessages(char **buffer, HYPRE_Int *bufsz);

/* Print the error messages and clear them in hypre */
HYPRE_Int HYPRE_PrintErrorMessages(MPI_Comm comm);
```
With this, there are now three basic ways to use hypre when configured `--with-print-errors`:
1. Default (mode 0): Errors are printed immediately to stderr (there is no processor information available in this print).
2. Store errors in memory (mode 1) and call PrintErrorMessages to print them.
3. Store errors in memory (mode 1) and call GetErrorMessages to manage the error messages however you like.

I modified the `ij.c` driver to test these functions and added a regression test in `src/test/TEST_error` and `AUTOTEST/machine-tux.sh`.  Note that this test stores the error messages from hypre in the `.saved` file, and these have source code line numbers in them.  This could become a pain for us to maintain every time we modify the various files and the line numbers change.  On the other hand, it's nice to have a regression test for this stuff.  Let me know what you think.

This uses the C99 function `snprintf`, and that function is used regardless of the error mode setting (even when the messages print to stderr as before).  I did this just to simplify the code a little bit.  Note that `snprintf` is recommended over `sprintf` and some C++ compilers are even printing warnings that `sprintf` is deprecated.  It's a nice function because it gives better control over the buffer size.

Thanks!

- [x] tux


src/utilities/error.c
```diff
@@ -69,19 +126,81 @@ void HYPRE_DescribeError(HYPRE_Int ierr, char *msg)
...
+{
+   hypre_Error err = hypre__global_error;
+
+   char *msg = err.memory;
+
+   hypre_MPI_Barrier(comm);
+
+   while (msg < (err.memory + err.msg_sz))
+   {
+      hypre_fprintf(stderr, "%s", msg);
```
victorapm: Should we also print the rank ID from where this error message was originated?

victorapm (approved):
This is a nice feature, thanks!

src/utilities/error.c
rfalgout: I had thought about that, but originally I wasn't passing `comm` to `PrintErrorMessages`.  It's probably a good idea to add this feature, so I'll go ahead and do that.  I created the Print routine just to give users another option.  So, there are three ways to use hypre when configured `--with-print-errors`:

1. Default: Errors are printed immediately to stderr (there is no processor information available to use in this print).
2. Store errors in memory and call `PrintErrorMessages` to print them.
3. Store errors in memory and call `GetErrorMessages` to manage the error messages however you like.

Thanks for the suggestion!

victorapm: Good! Thanks for the changes and summary!

rfalgout:
Hi @hypre-space/developers .  It looks like we only configure `--with-print-errors` in our tux regression tests.  We probably should add this flag to one of the compiles in a few of our other test scripts.  What do you think and what specifically would you suggest?  BTW, I don't think we necessarily need to add the `TEST_error` run-test to any of the other scripts besides tux, but adding the configuration flag seems like a good idea.

victorapm:
I think it would be good to add `--with-print-errors` to at least one of the machines testing the CUDA, HIP, and SYCL backends. So, I suggest adding this flag to at least one of the tests on `machine-lassen`, `machine-tioga`, and `machine-sunspot`.

rfalgout:
> I think it would be good to add `--with-print-errors` to at least one of the machines testing the CUDA, HIP, and SYCL backends. So, I suggest adding this flag to at least one of the tests on `machine-lassen`, `machine-tioga`, and `machine-sunspot`.

Which test would you add it to in `machine-lassen`?

ulrikeyang (approved):
This looks good. Thank you.

victorapm:
Hi @rfalgout 

I would add `--with-print-errors` to the following build:

https://github.com/hypre-space/hypre/blob/2794aa3b287f13ee91846f39bbeb52e16d0b4649/AUTOTEST/machine-lassen.sh#L53

I thought twice and maybe it's still early to add it to `machine-tioga` and `machine-sunspot`. But I think it's important for the future :)

rfalgout:
> Hi @rfalgout
> 
> I would add `--with-print-errors` to the following build:

Hi @victorapm .  I have made this change (not yet committed) and I am running the updated regression test to make sure I did it right.  When compiling with these flags manually and running the TEST_error test, interestingly we get 4 fewer error messages on Lassen then we do on Tux.  But, there's no reason to expect that is incorrect behavior.  The main point is to check that the logging and printing of the messages works correctly.

Hmmm...  Actually, I probably should do a quick run with the current source to see if it gives the same errors.

src/test/TEST_error/error.saved
```diff
@@ -0,0 +1,58 @@
...
+0: hypre error in file "pcg.c", line 782, error code = 256 - Reached max iterations in PCG before convergence
+0: hypre error in file "pcg.c", line 782, error code = 256 - Reached max iterations in PCG before convergence
+1: hypre error in file "par_amg_solve.c", line 339, error code = 256
+1: hypre error in file "pcg.c", line 709, error code = 256 - Subnormal gamma value in PCG
+1: hypre error in file "pcg.c", line 709, error code = 256 - Subnormal gamma value in PCG
+1: hypre error in file "pcg.c", line 709, error code = 256 - Subnormal gamma value in PCG
+1: hypre error in file "pcg.c", line 782, error code = 256 - Reached max iterations in PCG before convergence
+1: hypre error in file "pcg.c", line 782, error code = 256 - Reached max iterations in PCG before convergence
+1: hypre error in file "pcg.c", line 782, error code = 256 - Reached max iterations in PCG before convergence
+1: hypre error in file "pcg.c", line 782, error code = 256 - Reached max iterations in PCG before convergence
```
waynemitchell, line 58: If you have error messages coming from multiple ranks, are these guaranteed to be printed in the same order every time? Or could you randomly get failed regressions because things print in a different order?

waynemitchell: Oh, I think I see where you are sorting things in the `error.sh` script, but still want to double check with you that this should be deterministic.

rfalgout: Yes, they are sorted so it's deterministic.  Thanks!

waynemitchell (approved):
Looks great! Thanks, Rob!

src/test/TEST_error/error.saved
```diff
@@ -0,0 +1,58 @@
...
+                   cycle = 5.332187
+
+# Output file: error.out.10
+
+
+Iters       ||r||_2     conv.rate  ||r||_2/||b||_2
+-----    ------------   ---------  ------------ 
+    1    5.298880e+01    1.675653    1.675653e+00
+    2    5.298880e+01    1.000000    1.675653e+00
+0: hypre error in file "par_amg_solve.c", line 339, error code = 256
```
oseikuffuor1, line 43: @rfalgout just to clarify what I am seeing and my understanding here. Are these errors that are typically generated for these problems, but we ignore or print to stderr?

rfalgout: This is just part of the output that hypre spits out when I run the tests that cause errors to occur.  I basically chose certain parameter options that I knew would cause problems, for example, setting max iterations to a value that is too small so the run doesn't have a chance to converge, etc.  If you look at `error.jobs` you can see the run lines.

oseikuffuor1 (approved):
Thanks @rfalgout. This looks good.

osborn9 (approved):
Nice, looks good!

rfalgout:
Unfortunately, I can't just add the `--with-print-errors` to the lassen test you suggested @victorapm , because there are two other tests that generate error messages when this is on.  One is `TEST_ij/elast.sh` and the other is `TEST_ams/solvers.sh`.  The former produces errors by way of the hypre error stuff because of error codes thrown in `csr_matop_device.c`.  The latter by way of a `printf` at the very bottom of the file `utilities/device_utils.c`. 

All: Should we try to clean this up?  I can just go ahead without the additional regression test for now, or maybe I can just add a separate test that only runs the new `error.sh` tests for now.  I think I'll move forward with the last idea while waiting to hear back from everyone.

Thanks!

victorapm:
I see it, Rob! I vote for leaving this work for a separate PR, i.e., adding `--with-print-errors` to `machine-lassen` and making proper changes to the two tests you mentioned so that they run successfully :)

rfalgout:
Okay, this change to the lassen test works.  What fails is if we add `--with-print-errors` to the configure line for the very first test `basic-cuda-um`.  I agree with Victor that we can revisit this in a separate PR.  It occurs to me that we might want to check all of our tests with the additional `--with-print-errors` turned on (and maybe keep it on)`.

I am going to pull in the changes from `master` and do one last test on tux before merging the PR.  Thanks!

PR #923, opened 2023-06-12
Fix warning messages about set-but-not-used variables

rfalgout:
I updated my Mac this weekend, and the compiler version must have changed, because a number of warning messages are now being generated about variables that are set but not used.  This PR fixes the warnings.  In most cases, I just commented lines out.  I am not against deleting some of these variables either, but I wasn't sure if any of you would prefer to keep this code around for debugging reasons.

Thanks!

### DIFF
--- a/src/IJ_mv/IJMatrix_parcsr.c
+++ b/src/IJ_mv/IJMatrix_parcsr.c
@@ -2984,7 +2984,7 @@ hypre_IJMatrixSetValuesOMPParCSR( hypre_IJMatrix       *matrix,
 
    HYPRE_Int print_level = hypre_IJMatrixPrintLevel(matrix);
    //HYPRE_Int max_num_threads;
-   HYPRE_Int error_flag = 0;
+   //HYPRE_Int error_flag = 0;
 
    /*HYPRE_Complex *off_proc_data;*/
    hypre_MPI_Comm_size(comm, &num_procs);
@@ -3094,7 +3094,7 @@ hypre_IJMatrixSetValuesOMPParCSR( hypre_IJMatrix       *matrix,
 #ifdef HYPRE_USING_OPENMP
                   #pragma omp atomic
 #endif
-                  error_flag++;
+                  //error_flag++;
                   if (print_level)
                   {
                      hypre_printf (" row %b too long! \n", row);
@@ -3122,7 +3122,7 @@ hypre_IJMatrixSetValuesOMPParCSR( hypre_IJMatrix       *matrix,
 #ifdef HYPRE_USING_OPENMP
                         #pragma omp atomic
 #endif
-                        error_flag++;
+                        //error_flag++;
                         if (print_level)
                         {
                            hypre_printf (" Error, element %b %b does not exist\n",
@@ -3146,7 +3146,7 @@ hypre_IJMatrixSetValuesOMPParCSR( hypre_IJMatrix       *matrix,
 #ifdef HYPRE_USING_OPENMP
                         #pragma omp atomic
 #endif
-                        error_flag++;
+                        //error_flag++;
                         if (print_level)
                         {
                            hypre_printf (" Error, element %b %b does not exist\n",
@@ -3166,7 +3166,7 @@ hypre_IJMatrixSetValuesOMPParCSR( hypre_IJMatrix       *matrix,
 #ifdef HYPRE_USING_OPENMP
                         #pragma omp atomic
 #endif
-                        error_flag++;
+                        //error_flag++;
                         if (print_level)
                         {
                            hypre_printf (" Error, element %b %b does not exist\n",
@@ -3194,7 +3194,7 @@ hypre_IJMatrixSetValuesOMPParCSR( hypre_IJMatrix       *matrix,
 #ifdef HYPRE_USING_OPENMP
                         #pragma omp atomic
 #endif
-                        error_flag++;
+                        //error_flag++;
                         if (print_level)
                         {
                            hypre_printf (" Error, element %b %b does not exist\n",
@@ -3453,7 +3453,7 @@ hypre_IJMatrixSetValuesOMPParCSR( hypre_IJMatrix       *matrix,
 #ifdef HYPRE_USING_OPENMP
                               #pragma omp atomic
 #endif
-                              error_flag++;
+                              //error_flag++;
                               if (print_level)
                               {
                                  hypre_printf("Error in row %b ! Too many elements!\n",
@@ -3489,7 +3489,7 @@ hypre_IJMatrixSetValuesOMPParCSR( hypre_IJMatrix       *matrix,
 #ifdef HYPRE_USING_OPENMP
                               #pragma omp atomic
 #endif
-                              error_flag++;
+                              //error_flag++;
                               if (print_level)
                               {
                                  hypre_printf("Error in row %b ! Too many elements !\n",
--- a/src/parcsr_ls/par_cr.c
+++ b/src/parcsr_ls/par_cr.c
@@ -2075,7 +2075,9 @@ hypre_BoomerAMGIndepPMIS( hypre_ParCSRMatrix    *S,
          }
       }
 
-      //iter++;
+#if 0 /* debugging */
+      iter++;
+#endif
       /*------------------------------------------------
        * Set C-pts and F-pts.
        *------------------------------------------------*/
@@ -2685,7 +2687,9 @@ hypre_BoomerAMGIndepPMISa( hypre_ParCSRMatrix    *S,
          }
       }
 
-      //iter++;
+#if 0 /* debugging */
+      iter++;
+#endif
       /*------------------------------------------------
        * Set C-pts and F-pts.
        *------------------------------------------------*/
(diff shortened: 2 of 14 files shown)


src/IJ_mv/IJMatrix_parcsr.c
```diff
@@ -2984,7 +2984,7 @@ hypre_IJMatrixSetValuesOMPParCSR( hypre_IJMatrix       *matrix,
 
    HYPRE_Int print_level = hypre_IJMatrixPrintLevel(matrix);
    //HYPRE_Int max_num_threads;
-   HYPRE_Int error_flag = 0;
+   //HYPRE_Int error_flag = 0;
```
victorapm, line 2987: I'm wondering if it makes more sense to remove this and other all together instead of leaving them commented out

src/parcsr_ls/par_multi_interp.c
```diff
@@ -320,13 +322,15 @@ hypre_BoomerAMGBuildMultipassHost( hypre_ParCSRMatrix  *A,
...
    n_coarse_offd = 0;
-   n_SF_offd = 0;
 #ifdef HYPRE_USING_OPENMP
+   n_SF_offd = 0;
    #pragma omp parallel for private(i) reduction(+:n_coarse_offd,n_SF_offd) HYPRE_SMP_SCHEDULE
 #endif
    for (i = 0; i < num_cols_offd; i++)
       if (CF_marker_offd[i] == 1) { n_coarse_offd++; }
+#ifdef HYPRE_USING_OPENMP
       else if (CF_marker_offd[i] == -3) { n_SF_offd++; }
```
victorapm: I thought this line should always happen, regardless of turning on OpenMP support. Let's double-check?

src/IJ_mv/IJMatrix_parcsr.c
victorapm: Oops, just read it more slowly and realized you are not against of deleting it :) I'm in favor of deleting

src/parcsr_ls/par_multi_interp.c
rfalgout: I didn't check to see who wrote this code, so I wanted to be careful not to delete something that was meant to be kept around.  Please feel free to delete if you know the situation better than me.  Thanks!

victorapm: Good! I don't know it better, but according to git, @ulrikeyang implemented this 18 years ago :) Perhaps she can verify the desired behavior here. Ulrike, shouldn't the `else if` condition above be executed regardless of using OpenMP or not? If that's the case, then the new `ifdef HYPRE_USING_OPENMP` should be removed

rfalgout: If you remove the `HYPRE_USING_OPENMP` guard, the compiler warnings will come back unless you get rid of the `n_SF_offd++` stuff altogether (or comment it out some other way).  Just FYI.

ulrikeyang: I don't know how this OPENMP got in there, but it shouldn't be.

victorapm: Ok, now I see that `n_SF_offd` isn't used after being computed. That is probably the primary reason for the warnings. So, would it make sense removing `else if (CF_marker_offd[i] == -3) { n_SF_offd++; }`? and the `n_SF_offd` variable all together?

src/IJ_mv/IJMatrix_parcsr.c
ulrikeyang: We probably should get rid of this and replace it with the proper hypre_error messages.

src/parcsr_ls/par_multi_interp.c
rfalgout: I added the OPENMP guard because that was the only context where it was being used (in a `pragma`).  The warning messages were all about variables being set but not used.

src/parcsr_ls/par_cr.c
ulrikeyang, line 1: It looks like this iter was used for debugging purposes. Who is still using this routine? If so, we might keep it, but include it inside the debug option.

victorapm (approved):
Thanks, Rob!! My review comments are minor...

src/IJ_mv/IJMatrix_parcsr.c
rfalgout: The reason I wasn't sure what to do here is because, at the end of the routine, ther is an `if (error_flag)` then `return hypre_error_flag` that was commented out (I assume for a good reason).  So, should that `return` be replaced with a `hypre_error_w_msg` or something (and uncommented)?  Any advice here is welcome.

src/parcsr_ls/par_multi_interp.c
ulrikeyang: It would make sense :)

ulrikeyang (approved):
This looks fine except for the fact that we need to resolve the comments.

src/parcsr_ls/par_multi_interp.c
rfalgout: @ulrikeyang Are we good leaving this one as is then?  Thanks!

src/parcsr_ls/par_cr.c
rfalgout: Okay, I now use a "debugging" comment in the same way as `iter` was defined to begin with.

src/parcsr_ls/par_multi_interp.c
ulrikeyang: No, we should remove all references to n_SF_offd, since it is not used at all. That way we avoid unnecessary if checks.

rfalgout: Okay, done.

ulrikeyang (approved):
Looks good. Thanks

rfalgout:
@liruipeng @waynemitchell : If you could take a quick look at this, I'd like to merge it soon since it's trivial compiler warning stuff (and astyle indentation fixes).  Let me know.  Thanks!

src/parcsr_ls/par_cr.c
```diff
@@ -2076,7 +2075,9 @@ hypre_BoomerAMGIndepPMIS( hypre_ParCSRMatrix    *S,
          }
       }
 
+#if 0 /* debugging */
```
waynemitchell, line 2078: Is there a reason you are guarding this with `#if 0` instead of just eliminating (or commenting out) as you've done in other places?

waynemitchell (approved):

liruipeng (approved):

src/parcsr_ls/par_cr.c
rfalgout: Yes.  Ulrike commented that this looked like debugging stuff and should be marked as such.  So, I used the same `#if 0` line that was already there in other parts of the file.  It would be better if there was a `DEBUG` flag, but I took the path of least effort and followed what was already there. :)

src/IJ_mv/IJMatrix_parcsr.c
rfalgout: @ulrikeyang @victorapm : Any suggestions about what to do with this file, or should we just leave it as is?  Thanks again!

ulrikeyang: We cannot just comment this out, because we are adding on this variable, and so this would break the code. We would have to remove all the error_flag++ occurrences. The best thing would be to clean it up using hypre_error stuff. Right now there are also print statements depending on the print_level.

rfalgout: This doesn't break the code.  `error_flag` is never used in this routine.  It's set and incremented, and that's it.  There is a block of code at the bottom of the routine that does use `error_flag`, but it's commented out (not by me).  These changes are guaranteed safe and won't change the behavior of the code.

I don't know how to clean up the error stuff here, because the commented-out block near the bottom was apparently commented out for a reason, and I don't understand what that reason was.  From my visual inspection, I would say we should keep the `error_flag` stuff and uncomment the `if (error_flag)` thing at the bottom, but that may cause issues with regression tests and users.

rfalgout: Hi @ulrikeyang .  It looks like you commented that `if (error_flag)` stuff out 5 years ago when making some BigInt changes.  So, let me know if you want to leave this as is or uncomment that `if` statement (and the other stuff).  Thanks!

ulrikeyang: It doesn't look commented out to me. Are we looking at different files? What about line 3097 etc?

rfalgout: 3097 is commented out:
https://github.com/hypre-space/hypre/blob/cccdab8543c9972c2d697b10700812f02e78dae8/src/IJ_mv/IJMatrix_parcsr.c#L3097

The `if` statement I mention is at the bottom of the routine here:
https://github.com/hypre-space/hypre/blob/cccdab8543c9972c2d697b10700812f02e78dae8/src/IJ_mv/IJMatrix_parcsr.c#L3558-L3569

ulrikeyang: Ok. If it's all commented out it's fine.

PR #924, opened 2023-06-12
ame/ams to use Jacobi on GPUs

liruipeng:
This PR updates the ams driver to use Jacobi smoothers on GPUs.

- [x] lassen
- [x] tioga
- [x] tux
- [x] sunspot

PR #877, opened 2023-04-07
Extend ILU support to AMD GPUs

victorapm:
Extended from #867

This PR adds HIP support to `hypre_ILU` (setup and solve phases):

1. Algorithm type 0 (BJ-ILU0)
2. Algorithm type 10 (GMRES-ILU0)
3. Iterative triangular solves for backward and forward substitutions.

TODOs for following PRs:
- ILU code refactoring
- Add regression tests

Regression tests:
- [x] tux
- [x] lassen
- [x] tioga
- [x] sunspot

PR #925, opened 2023-06-14
Fix device build

victorapm:
A change from PR #923 broke our device build, see the compilation error below. Essentially, we cannot use `hypre_error_w_msg` inside GPU kernels or lambdas. Speaking more broadly, I think we should decide on how to capture errors from situations like this. In the meantime, I'm providing a fix with this PR :)

```
hipcc  -O2 -x hip -std=c++14   -DHAVE_CONFIG_H -I.. -I. -I./.. -I./../multivector -I./../utilities -I./../krylov -I./../struct_mv -I./../seq_mv -I./../parcsr_mv -I./../IJ_mv -I./../parcsr_block_mv -I./../sstruct_mv -I./../struct_ls -I./../parcsr_ls     -I/opt/rocm-5.4.3/include       -I/opt/cray/pe/mpich/8.1.23/ofi/crayclang/10.0/include -c node_relax.c -o node_relax.obj
node_relax.c:769:22: error: reference to __host__ function 'hypre_error_handler' in __host__ __device__ function
                     hypre_error_w_msg(HYPRE_ERROR_GENERIC, "Error in gselim!\n");
                     ^
./../utilities/_hypre_utilities.h:295:39: note: expanded from macro 'hypre_error_w_msg'
#define hypre_error_w_msg(IERR, msg)  hypre_error_handler(__FILE__, __LINE__, IERR, msg)
                                      ^
./../struct_mv/_hypre_struct_mv.hpp:788:10: note: called by 'forall_kernel<(lambda at node_relax.c:738:16)>'
         loop_body(idx);
         ^
./../utilities/_hypre_utilities.h:292:6: note: 'hypre_error_handler' declared here
void hypre_error_handler(const char *filename, HYPRE_Int line, HYPRE_Int ierr, const char *msg);
     ^
```

### DIFF
--- a/src/sstruct_ls/node_relax.c
+++ b/src/sstruct_ls/node_relax.c
@@ -765,6 +765,7 @@ hypre_NodeRelax(  void                 *relax_vdata,
                    *----------------------------------------------*/
                   hypre_gselim(A_loc, x_loc, nvars, err);
                   (void) err;
+                  /* TODO (VPM): need a way to check error codes on device */
 
                   /*------------------------------------------------
                    * Copy solution from local storage.
--- a/src/sstruct_ls/node_relax.c
+++ b/src/sstruct_ls/node_relax.c
@@ -764,10 +764,8 @@ hypre_NodeRelax(  void                 *relax_vdata,
                    * Invert intra-nodal coupling
                    *----------------------------------------------*/
                   hypre_gselim(A_loc, x_loc, nvars, err);
-                  if (err)
-                  {
-                     hypre_error_w_msg(HYPRE_ERROR_GENERIC, "Error in gselim!\n");
-                  }
+                  (void) err;
+
                   /*------------------------------------------------
                    * Copy solution from local storage.
                    *----------------------------------------------*/
@@ -971,10 +969,8 @@ hypre_NodeRelax(  void                 *relax_vdata,
                    * Invert intra-nodal coupling
                    *----------------------------------------------*/
                   hypre_gselim(A_loc, x_loc, nvars, err);
-                  if (err)
-                  {
-                     hypre_error_w_msg(HYPRE_ERROR_GENERIC, "Error in gselim!\n");
-                  }
+                  (void) err;
+
                   /*------------------------------------------------
                    * Copy solution from local storage.
                    *----------------------------------------------*/
@@ -1170,4 +1166,3 @@ hypre_NodeRelaxSetTempVec( void                 *relax_vdata,
 
    return hypre_error_flag;
 }
-


src/sstruct_ls/node_relax.c
```diff
@@ -764,10 +764,8 @@ hypre_NodeRelax(  void                 *relax_vdata,
                    * Invert intra-nodal coupling
                    *----------------------------------------------*/
                   hypre_gselim(A_loc, x_loc, nvars, err);
-                  if (err)
-                  {
-                     hypre_error_w_msg(HYPRE_ERROR_GENERIC, "Error in gselim!\n");
-                  }
+                  (void) err;
```
rfalgout, line 767: Hi @victorapm .  Sorry for breaking this.  Didn't realize it was device code.  And you are correct that we need to devise a way to capture errors on the device.

Can you add a comment after each of these `(void) err;` lines that says something like, "need a way to check error codes on device"?  Otherwise, this looks fine to me and still compiles on the Mac.  Thanks!

rfalgout (approved):
Looks good, thanks!

src/sstruct_ls/node_relax.c
victorapm: Sure, no worries! I can do that :) BTW, the `void (err)` line was a trick to avoid the set-but-not-used warning :)

rfalgout: Yep, I figured that.  And it works.

pbauman:
FWIW, asserts have a different device version: https://github.com/hypre-space/hypre/blob/master/src/utilities/error.h#L48 Might need similar treatment for device functions.

I'll caution though that you probably only want asserts in device kernels (i.e. they normally do not get compiled, only in debug mode) and any runtime error checking done outside of GPU kernels. This is because error handling and printing in device kernels can affect device code compilation in sometimes disastrous ways. My two cents.

victorapm:
Thanks for the input, Paul! We will definitely consider these points

victorapm:
@liruipeng I would like to merge this before #924 and #877

src/sstruct_ls/node_relax.c
liruipeng: How about `hypre_device_assert`?

liruipeng (approved):

src/sstruct_ls/node_relax.c
victorapm: That works but just when `HYPRE_DEBUG` is defined. We might need something in release mode as well. Let's discuss :)

PR #926, opened 2023-06-14
Update mac autotest to use a tolerance

rfalgout:
After upgrading to Ventura on my mac, the runs in `machine-mac.sh` no longer match the default `.saved` files.  This PR simply adds a tolerance to the test.

An alternative would be to add `.saved.mac` files, but we would probably have to set those files up to match one specific mac (mine, I guess), and it's doubtful they would be okay for others on the team.  I'm not sure it's worth doing this.

Opinions?  It would be nice if all of us who have macs could run this test.  That may be possible with the tolerance approach, but that's not clear either.  Let me know what you think.


victorapm (approved):
Thanks! I assume the compiler version has changed?

rfalgout:
> Thanks! I assume the compiler version has changed?

Probably.  I didn't look and can't remember what I was using before. :)

victorapm:
I ran `machine-mac.sh` of this branch on the macOS Monterey and it worked fine. It used `clang-14`. I prefer the tolerance approach over the `.saved.mac` option

rfalgout:
Hi @liruipeng and @waynemitchell .  Any opinions on this PR?  Thanks!

waynemitchell:
@rfalgout , I've not run this on my mac before. What mpi do you use? I have the homebrew mpich installed, and on configure, I get:
configure: error: cannot compile a simple Fortran program
It finds mpif77 for the fortran compiler. I think when I've built hypre on my mac before, I've just disabled fortran. Advice?

rfalgout:
> @rfalgout , I've not run this on my mac before. What mpi do you use? I have the homebrew mpich installed, and on configure, I get: configure: error: cannot compile a simple Fortran program It finds mpif77 for the fortran compiler. I think when I've built hypre on my mac before, I've just disabled fortran. Advice?

Hi @waynemitchell .  I mainly wanted your opinion about using the tolerances here, so you don't necessarily have to get it running.  However, I use the homebrew open-mpi install.  Configure also finds mpif77, which is in /opt/homebrew/bin/mpif77 on my machine.  Disabling fortran is an option, but it would be better if a "working" Fortran compiler is installed.

waynemitchell:
> > @rfalgout , I've not run this on my mac before. What mpi do you use? I have the homebrew mpich installed, and on configure, I get: configure: error: cannot compile a simple Fortran program It finds mpif77 for the fortran compiler. I think when I've built hypre on my mac before, I've just disabled fortran. Advice?
> 
> Hi @waynemitchell . I mainly wanted your opinion about using the tolerances here, so you don't necessarily have to get it running. However, I use the homebrew open-mpi install. Configure also finds mpif77, which is in /opt/homebrew/bin/mpif77 on my machine. Disabling fortran is an option, but it would be better if a "working" Fortran compiler is installed.

Hmm... I tried open-mpi with the same results. I guess my underlying fortran compiler/libraries are probably broken? Not sure. Anyway, this seems like more trouble than it is worth for me to figure out right now. Conceptually, I don't have any issues with this PR.

waynemitchell (approved):

liruipeng (approved):
Looks good to me. Thank you!

PR #930, opened 2023-06-20
Fix valgrind issue in AIR

victorapm:
This fixes an issue reported from PETSc, see https://gitlab.com/petsc/petsc/-/merge_requests/6622#note_1438503216


lindsayad:
thanks @victorapm !

PR #932, opened 2023-06-21
Add an example for AIR

lindsayad:
I've been trying to get AIR [working in PETSc](https://gitlab.com/petsc/petsc/-/merge_requests/6622). I realized that my first task is to come up with a test/example that demonstrates a correct AIR implementation, and I think this example does that. (I modeled it after the example in pyamg). It converges in one iteration. Switching to a 0 restriction (transpose of interpolation operator), the example takes multiple iterations.

I think having this example in the hypre repository will help people curious about AIR and curious to try hypre's implementation.

@bensworth figure it's worth pinging you here. I think armed with this hypre-only example I'll be able to determine very quickly any issues with the corresponding PETSc AIR wrapper.


victorapm:
Hi @lindsayad, thanks for your contribution!

As explained [in this page](https://github.com/hypre-space/hypre/wiki/Git-Workflow#pull-requests-from-users-outside-the-main-development-team), I will create a development branch from yours (#933) and make some minor changes. Then, we will proceed with the review.

Thanks!!

PR #939, opened 2023-06-24
Improve ILU documentation

victorapm:
- Add warnings for Euclid and PILUT redirecting users to hypre-ILU
- Rewrite hypre-ILU overview section
- Add new sections to hypre-ILU documentation: "User-level functions", "ILU as smoother for BoomerAMG", and "GPU support"
- Include info about new iterative ILU options
- Update BoomerAMG complex smoothers section

A read-the-docs compile of this branch can be found here: https://hypre.readthedocs.io/en/ilu-docs/

### DIFF
--- a/src/docs/usr-manual/solvers-hypre-ilu.rst
+++ b/src/docs/usr-manual/solvers-hypre-ilu.rst
@@ -9,8 +9,8 @@ hypre-ILU
 ==============================================================================
 
 hypre-ILU is a suite of parallel incomplete LU factorization algorithms featuring dual
-threshold (ILUT) and level-based (ILUK) variants. Its implementation is based on a domain
-decomposition framework for achieving distributed parallelism. It can be used as
+threshold (ILUT) and level-based (ILUK) variants. The implementation is based on a domain
+decomposition framework for achieving distributed parallelism. hypre-ILU can be used as
 preconditioners for Krylov subspace methods, or smoothers for
 multigrid methods such as BoomerAMG and MGR.
 
--- a/src/docs/usr-manual/solvers-hypre-ilu.rst
+++ b/src/docs/usr-manual/solvers-hypre-ilu.rst
@@ -90,8 +90,8 @@ A short explanation for each of those functions and its parameters is given next
   * 11: GMRES with ILUT (GMRES-ILUT).
   * 20: NSH with ILUK (NSH-ILUK).
   * 21: NSH with ILUT (NSH-ILUT).
-  * 30: RAS with ILUK (GMRES-ILUK).
-  * 31: RAS with ILUT (GMRES-ILUT).
+  * 30: RAS with ILUK (RAS-ILUK).
+  * 31: RAS with ILUT (RAS-ILUT).
   * 40: ddPQ-GMRES with ILUK (ddPQ-GMRES-ILUK).
   * 41: ddPQ-GMRES with ILUT (ddPQ-GMRES-ILUT).
   * 50: GMRES with RAP-ILU0 with modified ILU0 (GMRES-RAP-ILU0).
--- a/src/docs/usr-manual/solvers-hypre-ilu.rst
+++ b/src/docs/usr-manual/solvers-hypre-ilu.rst
@@ -3,30 +3,31 @@
 
    SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+.. _hypre-ilu:
 
 hypre-ILU
 ==============================================================================
 
-The hypre-ILU solver is a parallel ILU solver based on a domain decomposition 
+The hypre-ILU solver is a parallel ILU solver based on a domain decomposition
 framework. It may be use iteratively as a standalone solver or smoother, as well as a
-preconditioner for accelerators like Krylov subspace methods. This solver 
-implements various parallel variants of the dual threshold (truncation) incomplete 
+preconditioner for accelerators like Krylov subspace methods. This solver
+implements various parallel variants of the dual threshold (truncation) incomplete
 LU factorization - ILUT, and the level-based incomplete LU factorization - ILUk.
 
 
 Overview
 ------------------------------------------------------------------------------
-The parallel hypre-ILU solver follows a domain decomposition approach for solving 
-distributed sparse linear systems of equations. The strategy is to decompose the 
-domain into interior and interface nodes, where an interface node separates two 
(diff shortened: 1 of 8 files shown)


adam-sim-dev:
Hi @victorapm , as a user, I am not clear which parameters/setting functions are for ilu(k) and ilut, respectively.

victorapm:
Thanks for the feedback, we will include particular instructions for ILUK and ILUT parameters

rfalgout:
Hi @victorapm .  I added a compile to read-the-docs while we work through this PR.  See the main comment above.

victorapm:
@rfalgout That's great! I was going to ask you about that :)

rfalgout (approved):
This looks good to me, @victorapm !  The only question I have is about the name 'hypre-ILU'.  The use of the 'hypre-' prefix seems redundant and we don't do that for any other solvers.  Simply using 'ILU' would still differentiate it from PILUT and Euclid.  But, let's see what the other reviewers think.  Nice job!  Thanks!

victorapm:
I agree with your comment about the name, Rob!

liruipeng (approved):
Looks good to me. I made a few minor changes on the description of hypre-ILU. Thanks!

victorapm:
Thanks!!

src/docs/usr-manual/solvers-hypre-ilu.rst
```diff
@@ -10,8 +10,8 @@ hypre-ILU
 
 hypre-ILU is a suite of parallel incomplete LU factorization algorithms featuring dual
 threshold (ILUT) and level-based (ILUK) variants. Its implementation is based on a domain
-decomposition framework for achieving distributed parallelism. It can be used as a
-standalone iterative solver, preconditioner for Krylov subspace methods, or smoother for
+decomposition framework for achieving distributed parallelism. It can be used as
+preconditioners for Krylov subspace methods, or smoothers for
```
victorapm: @liruipeng Shouldn't we mention that it can be used as a standalone solver (even though that's uncommon)? 

Also: the phrase starts with "It", so I think it makes sense to have preconditioner and smoother in the singular form to keep the right verb conjugation

liruipeng: I made some grammar changes. Hope it looks better now.

I haven't seen anyone used ILU a "standalone solver". Technically it can; if you set the tolerance to 0, it becomes LU. Maybe it's true that it can used as a crude inner solver of something. But I just felt this could others wrong information about hypre-ILU. Just my 2 cents. 

Thanks!

victorapm: Thanks, Rui Peng! @rfalgout Should we mention, in the documentation, the possibility of using preconditioners in general (ILU, MGR, FSAI...) as standalone solvers (even though this is uncommon)?

victorapm:
@liruipeng @oseikuffuor1, as for the method's name, do you have any preference? We are currently calling it as `hypre-ILU`, but as Rob mentioned above, maybe it makes more sense to change it to just `ILU` altogether

src/docs/usr-manual/solvers-hypre-ilu.rst
rfalgout: Absolutely.  It makes more sense to use it as a preconditioner in practice, IMO, but we should let people know.  Thanks!

rfalgout: So, I didn't see Rui Peng's comment before I responded.  Is the ILU solver in hypre not already written to be a standalone solver?  I thought that's how we did all of our preconditioning (that was our Babel design anyway).  To ask another way, is it not implemented as x_{k+1} = x_k + M^{-1} r_k, where M is the ILU approximation?

victorapm: Hi Rob, hypre's ILU is already written to be a standalone solver.

I believe Rui Peng's point was more in the sense: if we mention that it can be used as a standalone solver, people might want to use it in that way. However, that wouldn't be our recommendation (it makes much more sense to use it as a preconditioner or smoother, the same applies to other methods as well, e.g., BoomerAMG, MGR, FSAI...). So, this could possibly confuse users as to what way ILU should be used.

Now, my opinion is: maybe non-advanced users could be confused, but still, I believe we should mention that it is possible to use it as a standalone solver. Lastly, to eliminate confusion, we can add a "Hint" box as we already have in the GPU support section saying that we don't recommend it being used as standalone solver. Something along these lines...

Sorry for the long comment!

rfalgout: Why can't we just explicitly say "it can be used as a standalone solver (this is not recommended)"?  I'm not sure a separate hint is needed, but that's okay with me too.  I'm also okay with leaving this out as Rui Peng suggested, but I think it's important that users understand it can technically be used as a standalone solver.

Since preconditioners and smoothers in hypre are all (standalone) solvers (ILU *is* a solver by design), it's a bit redundant to repeat it everywhere, though we need to be sure we say it *somewhere*).  It's probably better for the documentation of individual solvers to discuss recommended use in general.  Maybe this is yet another topic to discuss as a team when we go through our documentation review. :)

liruipeng: Saying it as "it can be used as a standalone solver (this is not recommended)" sounds good to me. Thanks!

oseikuffuor1: The short statement suggested by @rfalgout and @liruipeng  sounds good to me too. Thanks!

oseikuffuor1:
> 

@victorapm dropping the "hypre-" is fine with me. Eventually PILUT and EUCLID will be deprecated so it makes sense to me.

src/docs/usr-manual/solvers-hypre-ilu.rst
victorapm: Thanks, everyone! Rob raised a good point in his last paragraph. I agree we should talk about this and make a decision, which will be helpful to know when updating the documentation of other methods

victorapm:
@liruipeng Is the change `hypre-ILU` -> `ILU` fine for you?

liruipeng:
> @liruipeng Is the change `hypre-ILU` -> `ILU` fine for you?

Yes, it is good for me.

oseikuffuor1 (approved):
Thanks Victor!

PR #739, opened 2022-09-22
FSAI support with CUDA and HIP

victorapm:
Add CUDA and HIP support to FSAI, which can be also used as a preconditioner for BoomerAMG

PS: I left a few TODO items in `par_fsai_device.c` for better performance which I plan to implement on another PR.

PR #940, opened 2023-06-27
Add MAGMA option to FSAI

victorapm:
- Allow the use of MAGMA as local linear solver for FSAI.
- Add `HYPRE_FSAISetLocalSolveType` for choosing the local linear solve type used in FSAI.
- Add `HYPRE_BoomerAMGSetFSAILocalSolveType` for the same purpose as above when FSAI is used as a smoother to BoomerAMG.

This PR will be merged into master after #739


osborn9 (approved):
Thanks for adding the runtime option @victorapm. Looks good!

src/utilities/_hypre_utilities.hpp
```diff
@@ -602,14 +602,6 @@ using hypre_DeviceItem = sycl::nd_item<3>;
       hypre_assert(0); exit(1);                                                              \
    } } while(0)
 
-#define HYPRE_CUSOLVER_CALL(call) do {                                                       \
```
rfalgout, line 605: What's this change about @victorapm ?

rfalgout (approved):
Looks good.  Just the one question.  Thanks!

src/utilities/_hypre_utilities.hpp
victorapm: We had this defined twice, so I removed the redundant one

rfalgout: Got it.  Thanks!

PR #941, opened 2023-06-28
Use config flag for locality-aware mpi

geraldc-unm:
Add configure option for locality-aware MPI.

When creating Comm Pkg, must set `comm_pkg->use_neighbor = 1`, defaults to 0.
Since we observed that node aware performs worse at finest levels, added a lvl threshold setting that controls a switchover to use node aware MPI. By default, all levels will use node_aware.

### DIFF
--- a/src/parcsr_ls/par_cycle.c
+++ b/src/parcsr_ls/par_cycle.c
@@ -286,6 +286,15 @@ hypre_BoomerAMGCycle( void              *amg_vdata,
    hypre_GpuProfilingPushRange(nvtx_name);
    while (Not_Finished)
    {
+#ifdef HYPRE_USING_NODE_AWARE_MPI
+      if (level >= hypre_HandleNodeAwareSwitchoverThreshold(hypre_handle()))
+      {
+         hypre_HandleUsingNodeAwareMPI(hypre_handle()) = 1;
+      } else
+      {
+         hypre_HandleUsingNodeAwareMPI(hypre_handle()) = 0;
+      }
+#endif
       if (num_levels > 1)
       {
          local_size = hypre_VectorSize(hypre_ParVectorLocalVector(F_array[level]));
--- a/src/parcsr_mv/_hypre_parcsr_mv.h
+++ b/src/parcsr_mv/_hypre_parcsr_mv.h
@@ -23,6 +23,10 @@ extern "C" {
 #ifndef HYPRE_PAR_CSR_COMMUNICATION_HEADER
 #define HYPRE_PAR_CSR_COMMUNICATION_HEADER
 
+#ifdef HYPRE_USING_NODE_AWARE_MPI
+#include "mpi_advance.h"
+#endif
+
 /*--------------------------------------------------------------------------
  * hypre_ParCSRCommPkg:
  *   Structure containing information for doing communications
@@ -59,6 +63,9 @@ typedef struct
    void                 *recv_data_buffer;
    HYPRE_Int             num_requests;
    hypre_MPI_Request    *requests;
+#ifdef HYPRE_USING_NODE_AWARE_MPI
+   MPIX_Request         *Xrequest;
+#endif
 } hypre_ParCSRCommHandle;
 
 typedef hypre_ParCSRCommHandle hypre_ParCSRPersistentCommHandle;
@@ -66,6 +73,10 @@ typedef hypre_ParCSRCommHandle hypre_ParCSRPersistentCommHandle;
 typedef struct _hypre_ParCSRCommPkg
 {
    MPI_Comm                          comm;
+#ifdef HYPRE_USING_NODE_AWARE_MPI
+   MPIX_Comm                        *neighbor_comm;
+   MPIX_Comm                        *neighborT_comm;
+#endif
    HYPRE_Int                         num_components;
    HYPRE_Int                         num_sends;
    HYPRE_Int                        *send_procs;
@@ -75,6 +86,11 @@ typedef struct _hypre_ParCSRCommPkg
    HYPRE_Int                         num_recvs;
    HYPRE_Int                        *recv_procs;
    HYPRE_Int                        *recv_vec_starts;
+   HYPRE_Int                         use_neighbor;
+#ifdef HYPRE_USING_NODE_AWARE_MPI
+   long                             *global_send_indices;
+   long                             *global_recv_indices;
+#endif
    /* remote communication information */
    hypre_MPI_Datatype               *send_mpi_types;
    hypre_MPI_Datatype               *recv_mpi_types;
--- a/src/parcsr_mv/new_commpkg.c
+++ b/src/parcsr_mv/new_commpkg.c
@@ -11,6 +11,9 @@
  *-----------------------------------------------------*/
 
 #include "_hypre_parcsr_mv.h"
+#ifdef HYPRE_USING_NODE_AWARE_MPI
+#include <mpi.h>
+#endif
 
 /* some debugging tools*/
 #define mydebug 0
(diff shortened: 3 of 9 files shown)


geraldc-unm:
I can remove the print statements, I was using them to debug

victorapm:
@geraldc-unm, this looks nice! 

Which path should I pass to `--with-node-aware-mpi-include` to config hypre with node-aware MPI support? i.e., Which is the dependency library here?

Thanks!

src/parcsr_mv/_hypre_parcsr_mv.h
```diff
@@ -75,6 +86,11 @@ typedef struct _hypre_ParCSRCommPkg
    HYPRE_Int                         num_recvs;
    HYPRE_Int                        *recv_procs;
    HYPRE_Int                        *recv_vec_starts;
+   HYPRE_Int                         use_neighbor;
+#ifdef HYPRE_USING_NODE_AWARE_MPI
+   long                             *global_send_indices;
```
victorapm, line 91: Maybe it's better to use `hypre_uint64`, see https://github.com/hypre-space/hypre/blob/36ab29ba0977294a741987c4f921fe02840dff59/src/utilities/general.h#L33 or `hypre_longint` here

src/parcsr_mv/new_commpkg.c
```diff
@@ -750,3 +753,58 @@ hypre_FillResponseIJDetermineSendProcs(void       *p_recv_contact_buf,
...
+                            HYPRE_BigInt *col_map_offd,
+                            MPI_Comm comm,
+                            hypre_ParCSRCommPkg *comm_pkg) {
+   HYPRE_Int num_sends = hypre_ParCSRCommPkgNumSends(comm_pkg);
+   HYPRE_Int num_recvs = hypre_ParCSRCommPkgNumRecvs(comm_pkg);
+   HYPRE_Int *send_map_starts = hypre_ParCSRCommPkgSendMapStarts(comm_pkg);
+   HYPRE_Int *recv_vec_starts = hypre_ParCSRCommPkgRecvVecStarts(comm_pkg);
+   HYPRE_Int *send_map_elmts =  hypre_ParCSRCommPkgSendMapElmts(comm_pkg);
+
+   int *sendcounts = (int *)malloc(num_sends * sizeof(int));
```
victorapm: We prefer using `hypre_TAlloc`, see https://github.com/hypre-space/hypre/blob/36ab29ba0977294a741987c4f921fe02840dff59/src/utilities/memory.h#L143 for allocating memory. In addition, it's preferable to use `hypre_` datatypes internally. So, `hypre_int` here.

src/parcsr_mv/par_csr_communication.c
```diff
@@ -60,11 +64,46 @@ hypre_ParCSRPersistentCommHandleCreate( HYPRE_Int job, hypre_ParCSRCommPkg *comm
...
+   HYPRE_Int *send_sizes;
+   HYPRE_Int *recv_sizes;
+   HYPRE_BigInt num_send_elmts;
+   HYPRE_BigInt num_recv_elmts;
+   MPIX_Request *Xrequest;
+
+   HYPRE_Int node_aware_on = hypre_HandleUsingNodeAwareMPI(hypre_handle());
+   if (comm_pkg->use_neighbor && node_aware_on) {
+      if (comm_pkg->neighbor_comm == NULL) {
+         hypre_printf("Trying to communicate with a NULL communicator\n");
```
victorapm: Please, update error handling. Check here for an example: https://github.com/hypre-space/hypre/blob/36ab29ba0977294a741987c4f921fe02840dff59/src/parcsr_mv/par_csr_matrix.c#L712-L713

src/test/ij.c
```diff
@@ -1561,6 +1561,19 @@ main( hypre_int argc,
          arg_index++;
          snprintf(mem_tracker_name, HYPRE_MAX_FILE_NAME_LEN, "%s", argv[arg_index++]);
       }
+#endif
+#if defined(HYPRE_USING_NODE_AWARE_MPI)
+      else if ( strcmp(argv[arg_index], "-node_aware_lvl_threshold") == 0 )
+      {
+         arg_index++;
+         hypre_HandleNodeAwareSwitchoverThreshold(hypre_handle()) = atoi(argv[arg_index++]);
+         if (hypre_HandleNodeAwareSwitchoverThreshold(hypre_handle()) > 0) {
```
victorapm, line 1570: Should `hypre_HandleNodeAwareSwitchoverThreshold` be an option internal to BoomerAMG instead of a global variable?

geraldc-unm:
> @geraldc-unm, this looks nice!
> 
> Which path should I pass to `--with-node-aware-mpi-include` to config hypre with node-aware MPI support? i.e., Which is the dependency library here?
> 
> Thanks!

As of right now, MPI Advance must be separately cloned and built. I run `CC=mpicc CXX=mpicxx ./configure --enable-persistent --with-MPI-include=<path to mpi advance>/src --with-MPI-lib-dirs=<path to mpi advance>/build/src/ --with-MPI-libs=mpi_advance --with-node-aware-mpi`

And I build MPI advance with `mkdir build; cd build; CC=mpicc CXX=mpicxx cmake ..; make`

src/test/ij.c
geraldc-anl: This was a first approach to have the value accessible. I'm not sure how we could have it as part of BoomerAMG but yes it should only be used for BoomerAMG.

victorapm:
Thanks @geraldc-unm!

Where can I find the MPI advance library?

src/test/ij.c
victorapm: Thank you! @liruipeng, do you support the idea of moving this switch threshold inside BoomerAMG?

geraldc-unm:
> Thanks @geraldc-unm!
> 
> Where can I find the MPI advance library?

https://github.com/mpi-advance/locality_aware

Thanks so much for the review

src/test/ij.c
liruipeng: Hi @victorapm I totally agreed with you. It would be much better if we can move it into the AMG struct. But a technical issue is that we need to access this value in some functions involved in the solve phase but the AMG struct is not passed along to those functions. So we ended up having a global variable for its simplicity. Any suggestions? Thanks.

src/parcsr_ls/par_cycle.c
```diff
@@ -286,6 +286,15 @@ hypre_BoomerAMGCycle( void              *amg_vdata,
    hypre_GpuProfilingPushRange(nvtx_name);
    while (Not_Finished)
    {
+#ifdef HYPRE_USING_NODE_AWARE_MPI
+      if (level >= hypre_HandleNodeAwareSwitchoverThreshold(hypre_handle()))
```
victorapm, line 290: @liruipeng This switch threshold could be moved inside the AMG struct

src/parcsr_ls/par_cycle.c
```diff
@@ -286,6 +286,15 @@ hypre_BoomerAMGCycle( void              *amg_vdata,
...
+      {
+         hypre_HandleUsingNodeAwareMPI(hypre_handle()) = 1;
```
victorapm, line 292: @liruipeng however, this other global variable (which turns off/on node aware MPI) cannot. One idea is to move it to the comm_pkg struct. 

PS: we could have it defaulted to zero, and implement something like `hypre_CommPkgSetNodeAwareMPI` to turn `comm_pkg->node_aware_mpi` on/off

src/test/ij.c
victorapm: @liruipeng thanks for providing the context. Now, I can see it better! See my follow-up comments at https://github.com/hypre-space/hypre/pull/941#discussion_r1331547752 and https://github.com/hypre-space/hypre/pull/941#discussion_r1331548539
Sorry to bug with this, but I think it's better if we have less global variables

src/parcsr_ls/par_cycle.c
geraldc-anl: I think to do that, comm_pkg would need to be accessible within `hypre_BoomerAMGCycle` which I don't think it is unless I'm mistaken. I don't see it in hypre_ParAMGData. Maybe it could be added as a field there but would need to be properly set and maintained.

src/parcsr_ls/par_cycle.c
geraldc-anl: For this one, I think it could be in the AMG struct as long as it can still be set when parsing command line args, where the value comes from

src/parcsr_ls/par_cycle.c
victorapm: You can use https://github.com/hypre-space/hypre/blob/36ab29ba0977294a741987c4f921fe02840dff59/src/parcsr_mv/par_csr_matrix.h#L108

e.g., `hypre_ParCSRMatrixCommPkg(A_array[lvl])`

liruipeng: I could be wrong...but to what granularity do we want to control `hypre_HandleUsingNodeAwareMPI`? If we put in commPkg, it would be per commPkg/Parcsrmatrix. I thought it's a more global thing, like the whole AMG solver. Again, putting it into the AMG struct is ideal, if we can find a clean way to do it.

src/parcsr_ls/par_cycle.c
liruipeng: OK. We have the access to `hypre_ParAMGData` in this function. Do we need `hypre_HandleNodeAwareSwitchoverThreshold` in other functions that don't have the access?

victorapm: I can only see it being used at `par_cycle.c`

src/parcsr_ls/par_cycle.c
victorapm: As far as I can see, the node aware option has been implemented only for the IJ interface. If we want it on the other interfaces in the future, we could extend their respective comm_pkg structs to handle this option.

Note I'm suggesting to put `using_node_aware_mpi` under the comm_pkg struct here, i.e., not under the AMG struct as for the other case :)

src/parcsr_ls/par_cycle.c
geraldc-anl: So I didn't check these in as permanent changes to the ij driver but to properly use node aware, you must set this hypre_HandleNodeAwareSwitchoverThreshold otherwise it will unoptimally use node aware mpi for every amg level. 

This brings up 2 things:
1. `NodeAwareSwitchoverThreshold` must be accessible to set like from the ij.c driver. It's quite nice to have it as a command line arg to test different thresholds without recompiling or anything.

2. We probably want to document this and how to connect to MPI advance. I can compile the steps and maybe we could include it somewhere in the wiki or something? But I don't know how to contribute to the wiki.

src/parcsr_ls/par_cycle.c
geraldc-anl: This may still work but it's my understanding that we have multiple matrices per level that each have a comm pkg. The `using_node_aware` toggle logically applies to a whole level with multiple matrices. So I think we would need to get the comm pkg for A and P and set them both to use node aware in their comm pkgs but I think that would work. What do y'all think?

victorapm: Thanks @geraldc-unm! I spoke to Rui Peng and shared that's a better approach in my opinion. Let's think about this (moving the variable to inside comm_pkg) a bit more...

src/parcsr_ls/par_cycle.c
victorapm: 1. Yes, there should be a Set routine for that, e.g., `HYPRE_BoomerAMGSetNodeAwareSwitchLevel`
2. Great idea, I'm not sure about the access but if you prepare something in markdown format, we can add it to the wiki.

liruipeng: Documenting it in the Wiki is a good idea.

PR #942, opened 2023-06-29
Skipcheck

ulrikeyang:
Hi all,
this PR is about two things:
1. Adding the ability the ability to skip certain breaks that check whether gamma and alpha are very close to zero
2.  Fixing some issues when recomputing the residual, either as restart or at seeming convergence.

I am just looking for comments at this point. I haven't really been able to test this, since our matrices are too nice and don't trigger this.


src/utilities/HYPRE_utilities.h
```diff
@@ -61,6 +61,7 @@ typedef int HYPRE_Int;
 typedef float HYPRE_Real;
 #define HYPRE_REAL_MAX FLT_MAX
 #define HYPRE_REAL_MIN FLT_MIN
+#define HYPRE_TRUE_MIN FLT_TRUE_MIN
```
victorapm: Should this be called `HYPRE_REAL_TRUE_MIN` for consistency with the above defines? Same for the other definitions below

ulrikeyang: It is a matter of taste. I just wanted to keep it short, but either one is fine with me.

victorapm: Yes, this was a minor comment

rfalgout: Even though I tend to like short names as well (a carryover from my fortran days when 8 characters was the limit :), I vote for inserting the 'REAL_' prefix, since it matches the C naming convention more closely.  2 cents.

src/krylov/HYPRE_pcg.c
```diff
@@ -246,6 +246,24 @@ HYPRE_PCGGetRecomputeResidualP( HYPRE_Solver  solver,
...
+
+HYPRE_Int
+HYPRE_PCGSetSkipBreak( HYPRE_Solver solver,
+                     HYPRE_Int    skip_break )
+{
+   return ( hypre_PCGSetSkipBreak( (void *) solver, skip_break ) );
+}
+
+HYPRE_Int
+HYPRE_PCGGetSkipBreak( HYPRE_Solver  solver,
```
rfalgout, line 261: This should be documented in HYPRE_krylov.h along with the 'Set' routine.  We have a bunch of replicated prototypes in krylov.h that we should clean up at some point.

src/krylov/pcg.c
```diff
@@ -540,11 +562,31 @@ hypre_PCGSolve( void *pcg_vdata,
...
+      if (! (alpha >= HYPRE_REAL_MIN) )
       {
          hypre_error_w_msg(HYPRE_ERROR_CONV, "Subnormal alpha value in PCG");
          if (i == 1) { i_prod = i_prod_0; }
-         break;
+         if (!skip_break)
+         {
+            break;
+         }
+         else if (! (alpha  >= HYPRE_TRUE_MIN))
```
rfalgout: I think both the 'alpha >= TRUE_MIN' and 'alpha <= 0.0' tests should always be done here.  That way we have that extra information when --with-print-errors is on, even when skip_break is not set.  Maybe add a boolean variable called `breakout` (or something) that is local to this REAL_MIN check that you set instead of calling `break`, then check its value at the bottom to decide whether or not to break.

src/krylov/pcg.c
```diff
@@ -98,7 +98,10 @@ hypre_PCGCreate( hypre_PCGFunctions *pcg_functions )
    (pcg_data -> rel_change)   = 0;
    (pcg_data -> recompute_residual) = 0;
    (pcg_data -> recompute_residual_p) = 0;
+   (pcg_data -> rtol)         = 0.0;
```
rfalgout: This line and the next are already set above (lines 95 and 97)

rfalgout: BTW, I'm not sure what `rtol` actually does.  It looks like it is only used in the `ams_driver.c` file.  I wonder if anybody really uses this.  It checks the difference in successive residuals, so it's similar to rel-change in the sense that it is looking for the residual norm to flatten out.

src/krylov/pcg.c
```diff
@@ -703,12 +789,28 @@ hypre_PCGSolve( void *pcg_vdata,
...
       if (! (gamma > HYPRE_REAL_MIN) )
       {
          hypre_error_w_msg(HYPRE_ERROR_CONV, "Subnormal gamma value in PCG");
-
-         break;
+         if (!skip_break)
+         {
+            break;
+         }
+         else if (! (gamma >= HYPRE_TRUE_MIN))
```
rfalgout: Same comment here as above on wanting to always do these tests to get more error message info.

rfalgout:
Hi @ulrikeyang .  This looks great modulo the comments!  It looks like a significant part of the work you did was related to this `rtol` thing (set by calling `HYPRE_PCGSetResidualTol`).  This is only available in PCG and only used by the AMS driver.  The documentation for it in `HYPRE_krylov.h` says it's "useful when trying to converge to very low tolerances, in order to bail-out before roundoff errors affect the approximation".  This is something we care about, but I'm not sure this is the right way to do it or if anybody even uses it.  If it's useful, we should add it to the other Krylov routines as well.  If not, maybe we should remove it altogether (in a future PR).  Sounds like an agenda item for our meeting. :)

ulrikeyang:
I have incorporated the comments. I also added a flexible CG option, which uses the Polak-Ribiere update for beta and provides more robustness, but also requires one extra dotproduct per iteration

src/krylov/HYPRE_krylov.h
```diff
@@ -178,7 +178,18 @@ HYPRE_Int HYPRE_PCGSetRecomputeResidualP(HYPRE_Solver solver,
...
 /**
- * (Optional) Skips submormal alpha and gamma values in CG.
+ * (Optional) Setting this to 1 allows use of Polak-Ribiere Method (flexible)
+ * this incrceases robustness, but adds an additional dot product per iteration
+ **/
+HYPRE_Int HYPRE_PCGSetFlex(HYPRE_Solver solver,
+                           HYPRE_Int    flex);
+
+/**
+ * (Optional) Skips subnormal alpha, gamma and iprod values in CG.
```
rfalgout, line 188: We probably should also remark that if set to 4 or higher, will never break.

src/krylov/pcg.c
```diff
@@ -565,35 +588,53 @@ hypre_PCGSolve( void *pcg_vdata,
...
-         }
+	 }
       }
 
       gamma_old = gamma;
 
       /* x = x + alpha*p */
       (*(pcg_functions->Axpy))(alpha, p, x);
 
+      if (flex) (*(pcg_functions->CopyVector))(r, r_old); /*save old residual */
```
rfalgout, line 636: You probably should run this through 'astyle'.  I think one-line 'if' statements are okay, but they might require the squirly braces.  I'd have to check again.

rfalgout (approved):
This looks great, Ulrike!  I had not heard of the Polak-Ribiere update for beta, but it's nice to have another option to help with robustness.  Maybe you can tell us a bit more about it at our next meeting?

src/krylov/HYPRE_krylov.h
```diff
@@ -177,6 +177,24 @@ HYPRE_Int HYPRE_PCGSetRecomputeResidual(HYPRE_Solver solver,
 HYPRE_Int HYPRE_PCGSetRecomputeResidualP(HYPRE_Solver solver,
                                          HYPRE_Int    recompute_residual_p);
 
+/**
+ * (Optional) Setting this to 1 allows use of Polak-Ribiere Method (flexible)
+ * this incrceases robustness, but adds an additional dot product per iteration
+ **/
+HYPRE_Int HYPRE_PCGSetFlex(HYPRE_Solver solver,
```
victorapm, line 184: Nice to have an option for flexible conjugate gradient now!

victorapm (approved):
Thanks!!

victorapm:
PS: should we add new regression tests in this PR or in a following one?

PR #927, opened 2023-06-15
L1 HSGS

liruipeng:
This PR provides a convergent l1-hybrid symmetric Gauss-Seidel (HSGS) method. It is named so since locally within the process it does l1-SGS, while across processes it does Jacobi-type updates, i.e., hybrid updates. The current l1-HSGS (of relaxation type 8) uses the same l1 term as in l1-hybrid Gauss-Seidel, which may not be convergent for general SPD matrices (e.g., the matrix included in PR). The new method, which uses the type number 88, has a new l1 weight that guarantees the convergence. Also, this PR provides the l1-symmetric hybrid Gauss-Seidel (SHGS), using the number 89, which is the symmetrized l1-HGS. The implementation is just simply calling the forward L1-HGS (13) followed by the backward one (14).
Based on hypre's regression tests, the convergence behavior of the new smoother is close to the old one, with up to 1 more iteration, for the problems that the old smoother can work.

- [x] tux
- [x] lassen
- [x] tioga
- [x] sunspot

PR #950, opened 2023-07-20
Add CUDA support to dense direct solver options

victorapm:
This PR adds CUDA support to the dense direct solver options (98, 99, 198, and 199) of BoomerAMG and MGR:

- Options 98 and 99 compute the LU factorization with pivoting.
- Options 198 and 199 compute the dense inverse matrix explicitly.

The difference between 98 and 99 or 198 and 199 lies on the strategy used for computing the dense local matrix from the distributed ParCSR matrix. In the first case (98 and 198), hypre's internal DataExchange is used , in the latter case (99 and 199), MPI collectives defined on a sub-communicator are used.

GPU support is achieved via the vendor math libraries (cuSOLVER in this case) or via MAGMA, when hypre is configured with MAGMA support. In the latter scenario, MAGMA takes precedence over cuSOLVER.

In addition to these changes, the `par_gauss_elim.c` file was majorly refactored.

TODOs for following PRs:

- Add regression tests for the new capabilities.
- Create a separate data structure for accessing dense direct solvers.
- Add GPU-aware MPI to the matrix/vector gathering phases.
- Add HIP and SYCL support.
- Add a runtime option to switch between solver implementations (vendor or MAGMA)


rfalgout (approved):
Looks good to me @victorapm!  Thanks!

ulrikeyang (approved):
this looks fine, as far as I can tell.

oseikuffuor1 (approved):
Thanks again Victor. Nice work.

src/utilities/device_utils.h
```diff
@@ -360,7 +360,6 @@ using hypre_DeviceItem = sycl::nd_item<3>;
    if (cudaSuccess != err) {                                                                 \
       printf("CUDA ERROR (code = %d, %s) at %s:%d\n", err, cudaGetErrorString(err),          \
                    __FILE__, __LINE__);                                                      \
-      hypre_assert(0); exit(1);                                                              \
```
liruipeng, line 363: Why remove? In the case of CUDA ERROR, do we want to exit right away?

victorapm: @rfalgout and I talked about this at some point. My understanding was that we shouldn't exit from hypre and the application that calls hypre would decide how to handle the error. If confirmed, this behavior needs to be replicated in the other macros (in a separate PR)

liruipeng: OK. Then, at least a `return hypre_error` should be called? IMHO, there's no reason to continue to when a CUDA error occurs.

victorapm: Agreed

liruipeng:
Looks good to me. I would like to merge #927 before this. Thanks!

victorapm:
OK! I can wait for you

PR #959, opened 2023-08-28
fix checkpar to pass on armhf, ppc64el and detect test failures and define hypre_FinalizeAllTimings()

### LINKED ISSUE #955: tests time out on ppc64el architecture
Hypre tests (in debian builds) have started timing out on ppc64el architecture, see https://buildd.debian.org/status/logs.php?pkg=hypre&arch=ppc64el.  The timeouts are happening for both hypre v2.28.0 and v2.29.0. The timeout is somewhat random, happening in a different test in different attempts.

Previously (April 2023) 2.28.0 passed tests.  Since then Debian has made a new stable release (using gcc-12, openmpi 4.1.4). The failing environment now builds with gcc-13 and openmpi 4.1.5. I'm not sure how relevant that is, but evidently the timeout is related to the build environment.

For reference, there is also a recent timeout on hppa, https://buildd.debian.org/status/logs.php?pkg=hypre&arch=hppa
But 2.29.0 recently built successfully on hppa, so I expect success with 2.28.0 after trying again.  The ppc64el failure has been more reproducible (4 recent failed attempts)

In general there's been a degree of flakiness in the hypre tests. Occasionally I've had timeouts locally on my own machine (amd64). I just run the tests again, they pass the next time. Is there any thing that can be done to make the tests more robust?

Possibly there's nothing hypre can do about it.  But if you have any ideas, I'll be glad to try them.

drew-parsons:
Enable tests to detect failure and to pass on less common architectures such as armhf, ppc64el.

Introduces Make variable HYPRE_CHECKPAR_TOL with default value 1e-6 suitable for "normal" architectures, allowing a relaxed tolerance to be set where needed. For instance.  [debian builds](https://buildd.debian.org/status/package.php?p=hypre) need HYPRE_CHECKPAR_TOL=1e-3 on arm64 i386 ppc64el s390x hppa powerpc ppc64.

Taken from debian patch [fix_make_check.patch](https://salsa.debian.org/science-team/hypre/-/blob/master/debian/patches/fix_make_check.patch).

Fixes: #955

Also applies debian patch [no_timings_finalizeAllTimings.patch](https://salsa.debian.org/science-team/hypre/-/blob/master/debian/patches/no_timings_finalizeAllTimings.patch) fixing definitions missed by PR #519.

Tests:

- [x] tux
- [x] lassen
- [x] tioga
- [x] sunspot

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -171,15 +171,15 @@ checkpar:
         set -e; \
         (cd test; $(MAKE) all); \
         echo "Testing IJ ..."; \
-        (cd test; ./runtest.sh -tol 1.e-06 -mpi "$(CHECKRUN)" TEST_ij/solvers.sh); \
+        (cd test; ./runtest.sh -atol 1.e-06 -mpi "$(CHECKRUN)" TEST_ij/solvers.sh); \
         (cd test; ./checktest.sh); \
         (cd test; ./cleantest.sh); \
         echo "Testing Struct ..."; \
-        (cd test; ./runtest.sh -tol 1.e-06 -mpi "$(CHECKRUN)" TEST_struct/solvers.sh); \
+        (cd test; ./runtest.sh -atol 1.e-06 -mpi "$(CHECKRUN)" TEST_struct/solvers.sh); \
         (cd test; ./checktest.sh); \
         (cd test; ./cleantest.sh); \
         echo "Testing SStruct ..."; \
-        (cd test; ./runtest.sh -tol 1.e-06 -mpi "$(CHECKRUN)" TEST_sstruct/solvers.sh); \
+        (cd test; ./runtest.sh -atol 1.e-06 -mpi "$(CHECKRUN)" TEST_sstruct/solvers.sh); \
         (cd test; ./checktest.sh); \
         (cd test; ./cleantest.sh); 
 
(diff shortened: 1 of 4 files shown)


src/Makefile
```diff
@@ -167,17 +172,18 @@ check:
 checkpar:
 	@ \
         echo "Checking the library ..."; \
+        set -e; \
         (cd test; $(MAKE) all); \
         echo "Testing IJ ..."; \
-        (cd test; ./runtest.sh -tol 1.e-06 -mpi "$(CHECKRUN)" TEST_ij/solvers.sh); \
+        (cd test; ./runtest.sh -atol $(HYPRE_CHECKPAR_TOL) -mpi "$(CHECKRUN)" TEST_ij/solvers.sh); \
```
victorapm, line 178: @rfalgout I'm wondering if it is more appropriate to use `rtol` here and in the similar occurrences below. What do you think?

rfalgout: It looks like we generally use atol in our AUTOTEST testing

victorapm: Thanks, Rob!

victorapm:
@waynemitchell , can you test this on sunspot? I'll run the other tests

rfalgout (approved):
This looks good to me.  I worry that we may accidentally break this in the future, since we generally think of these test scripts as being for internal use.

victorapm:
That's a good point, Rob. We could add `make check` and `make checkpar` to the regression tests

waynemitchell (approved):

victorapm (approved):

PR #969, opened 2023-09-22
Add HYPRE_GetExecutionPolicyName

victorapm:
This routine (to be used in a following PR) returns a string denoting the execution policy in use.


liruipeng (approved):
Looks good. Thank you!

PR #972, opened 2023-09-27
SYCL triangular solves, Chebyshev relaxation, etc.

waynemitchell:
This add triangular solves with the SYCL back end, which in turn enables ILU solves and Gauss-Seidel relaxation on Intel GPUs. Chebyshev relaxation is also added.


src/parcsr_ls/par_ilu_solve_device.c
```diff
@@ -1142,7 +1187,11 @@ hypre_ILUSolveRAPGMRESDevice(hypre_ParCSRMatrix   *A,
          /* perm back */
          if (perm)
          {
+#if defined(HYPRE_USING_SYCL)
+            hypreSycl_scatter(xtemp_data, xtemp_data + n, perm, ftemp_data);
```
victorapm, line 1191: Not for this PR, we could think of a way to wrap scatter for thrust and oneDPL in a single call. The same reasoning for gather as well

src/parcsr_mv/par_csr_triplemat_device.c
```diff
@@ -859,6 +856,40 @@ hypre_ParCSRMatrixRAPKTDevice( hypre_ParCSRMatrix *R,
...
+                                                          hypre_CSRMatrixNumNonzeros(zero_offd));
+
+      hypre_CSRMatrixDestroy(hypre_ParCSRMatrixDiag(zero));
+      hypre_ParCSRMatrixDiag(zero) = zero_diag;
+
+      hypre_CSRMatrixDestroy(hypre_ParCSRMatrixOffd(zero));
+      hypre_ParCSRMatrixOffd(zero) = zero_offd;
+
+      hypre_ParCSRMatrix *C_new;
+      hypre_ParCSRMatrixAdd(1.0, C, 1.0, zero, &C_new);
```
victorapm: Thinking again, would it be better to just use `hypre_CSRMatrixAdd` on the diagonal portion instead? This way we don't need `zero_offd` and the ParCSRMatrix.

src/parcsr_ls/par_relax_device.c
```diff
@@ -123,7 +123,8 @@ hypre_BoomerAMGRelaxTwoStageGaussSeidelDevice ( hypre_ParCSRMatrix *A,
    /* Sanity checks */
    hypre_assert(u_vecstride == num_rows);
    hypre_assert(r_vecstride == num_rows);
-   hypre_assert(z_vecstride == num_rows);
+   /* WM: TODO - commenting out the assert below, which fails */
```
victorapm: We can remove the TODO here and the assertion itself if you want, I have a branch fixing the assert failure coming soon

src/seq_mv/csr_matop_device.c
```diff
@@ -1634,12 +1633,96 @@ hypre_CSRMatrixCheckDiagFirstDevice( hypre_CSRMatrix *A )
...
+   HYPRE_Int  *A_i      = hypre_CSRMatrixI(A);
+   HYPRE_Int  *A_j      = hypre_CSRMatrixJ(A);
+   HYPRE_Int   num_rows = hypre_CSRMatrixNumRows(A);
+   HYPRE_Int  *result;
+   HYPRE_Int   ierr;
+
+   /* This test is only for square matrices */
+   if (hypre_CSRMatrixNumRows(A) != hypre_CSRMatrixNumCols(A))
+   {
+      return 0;
```
victorapm, line 1698: I think it's better to `return hypre_error_flag`

src/seq_mv/csr_matop_device.c
```diff
@@ -1634,12 +1633,96 @@ hypre_CSRMatrixCheckDiagFirstDevice( hypre_CSRMatrix *A )
...
+                             result + hypre_CSRMatrixNumRows(A) );
+#else
+   ierr = HYPRE_THRUST_CALL( reduce,
+                             result,
+                             result + hypre_CSRMatrixNumRows(A) );
+#endif
+
+   hypre_TFree(result, HYPRE_MEMORY_DEVICE);
+
+   return ierr;
```
victorapm, line 1721: ~~We could check if `ierr != 0` and give a `hypre_error_w_msg`, then we would `return hypre_error_flag`~~

Nevermind, now I see it's a check function meant to return true/false :)

victorapm (approved):
Looks great, thanks!! Just some minor comments/ideas...

src/parcsr_ls/par_relax_device.c
waynemitchell: Done. I also remove the `z_vecstride` variable to avoid a compiler warning about it non being used.

src/parcsr_ls/par_ilu_solve_device.c
waynemitchell: Yeah, I think this is part of a much larger question about whether/how we want to unify standard-library-ish calls implemented either in thrust or oneDPL. It might be nice to have our own `hypre_` wrappers for all this functionality, but it also might be too much work to develop/maintain.

src/parcsr_mv/par_csr_triplemat_device.c
waynemitchell: Yes, I agree this is the correct way to do it, particularly since the check for missing diagonals is local to each rank, so not all ranks may enter the code branch guarded by the if statement here.

src/parcsr_ls/par_ilu_solve_device.c
victorapm: Good! Seems like a topic for discussion in one of the SLS's

src/parcsr_ls/par_gauss_elim.c
```diff
@@ -338,7 +338,7 @@ hypre_GaussElimSetup(hypre_ParAMGData *amg_data,
     *  Factorization phase
     *-----------------------------------------------------------------*/
 
-#if defined(HYPRE_USING_GPU)
```
liruipeng, line 341: I guess no sycl support for this yet.

victorapm: Yes, sorry, my bad!

src/parcsr_ls/par_ilu_setup.c
```diff
@@ -1773,7 +1776,9 @@ hypre_ILUSetupLDUtoCusparse(hypre_ParCSRMatrix  *L,
    hypre_CSRMatrixData(LDU_diag) = LDU_diag_data;
 
    /* now sort */
+#if defined(HYPRE_USING_GPU)
    hypre_CSRMatrixSortRow(LDU_diag);
```
liruipeng, line 1780: I am a little surprised that we didn't have this #if guard. @oseikuffuor1 , we don't need sorted LDU on CPU, right?

liruipeng (approved):
Looks good. Just had a quick question about a sorting on CPUs for @oseikuffuor1 .

rfalgout (approved):
Looks good.  Thanks!

src/parcsr_ls/par_gauss_elim.c
waynemitchell: Correct, this functionality is brand new from a recent PR from Victor, so I'll look at that for a future PR.

src/parcsr_ls/par_ilu_setup.c
oseikuffuor1: Sorry I'm just seeing this. No we don't need it on CPU. Thanks Wayne.

PR #897, opened 2023-05-03
Add MGR statistics

victorapm:
This PR adds statistics info to MGR. An example for a 4-levels reduction strategy below:

```
 Num MPI tasks = 1
 Num OpenMP threads = 1


MGR SETUP PARAMETERS:

    MGR num levels = 5
    AMG num levels = 3
  Total num levels = 8

            Global           Fine           Coarse
lev     relaxation     relaxation      grid method    Interpolation      Restriction
====================================================================================
  0             --       User AMG       NG-BlkDiag         Diag Inv        Injection
  1     Blk-Jacobi      Dense Inv              RAP     Blk-Diag Inv        Injection
  2        BJ-ILU0         Jacobi              RAP         Diag Inv        Injection
  3         Blk-GS         Jacobi   NG-CPR-BlkDiag        Injection        Injection
  4     Blk-Jacobi         Jacobi               --               --               --



Full Operator Matrix Hierarchy Information:

                     nonzero   actual         |       entries/row        |              rowsums 
lev    rows    fine  entries  entries  sparse |  min  max     avg  stdev |      min      max      avg    stdev 
==============================================+==========================+=====================================
  0    2203     864    81704    81699  98.316 |    1  105    37.1  29.28 | -9.7e+11  8.9e+04 -5.6e+10  1.6e+11
  1    1339      16    68347    68342  96.188 |    1   87    51.0  30.86 | -9.7e+11  2.1e+05 -9.3e+10  2.0e+11
  2    1323     441    67960    67960  96.117 |    1   81    51.4  30.75 | -9.7e+11  1.8e-03 -9.4e+10  2.0e+11	( MGR )
  3     882     441   104304   104304  86.592 |    1  250   118.3  85.12 | -4.6e+11  1.3e-03 -1.0e+11  1.7e+11
  4     441     319    31431    26139  83.839 |    1  125    71.3  31.50 | -4.6e+11  9.0e-04 -1.0e+11  1.7e+11
----------------------------------------------+--------------------------+------------------------------------- MGR's coarsest level
  5     122      95     8514     8514  42.798 |   36  104    69.8  16.96 | -6.7e+11  4.5e+10 -2.1e+11  2.1e+11
  6      27      18      707      707   3.018 |   20   27    26.2   1.47 | -1.1e+12 -1.4e+11 -5.2e+11  2.3e+11	( AMG )
[...]

PR #976, opened 2023-10-02
Improve MGR data printing

victorapm:
This enhances what `print_level` in MGR can achieve. Particularly, now we can dump linear system info to files according to the `print_level` code. We also have the ability now of printing a sequence of linear systems to file (useful when hypre is used in time-stepping application).

See the new documentation for `HYPRE_MGRSetPrintLevel` below:

https://github.com/hypre-space/hypre/blob/45622494720a6d691518559c3db9173d352f8d1b/src/parcsr_ls/HYPRE_parcsr_ls.h#L4277-L4304

Depends on #897


src/parcsr_ls/par_mgr_setup.c
```diff
@@ -177,6 +177,16 @@ hypre_MGRSetup( void               *mgr_vdata,
    hypre_MPI_Comm_size(comm, &num_procs);
    hypre_MPI_Comm_rank(comm, &my_id);
 
+   /* Reset print_level codes. This is useful for printing
```
oseikuffuor1, line 180: @victorapm why do these codes need to be reset? Do they change their state?

src/parcsr_ls/par_mgr_solve.c
```diff
@@ -252,43 +233,32 @@ hypre_MGRSolve( void               *mgr_vdata,
...
+      {
+         hypre_printf("\n\n==============================================");
+         hypre_printf("\n NOTE: Convergence tolerance was not achieved\n");
+         hypre_printf("      within the allowed %d iterations\n", max_iter);
+         hypre_printf("==============================================");
+      }
    }
 
-   if (print_level > 1)
+   if ((my_id == 0) && (tol > 0) && (print_level & HYPRE_MGR_PRINT_INFO_SOLVE))
```
oseikuffuor1: The logic seems to have changed here. This will only print if not using MGR as a preconditioner. Is that intended, and if so, why?

src/parcsr_ls/par_mgr_setup.c
victorapm: This is useful for the following scenario:
1) User calls `HYPRE_MGRSetup`
2) Current LS data is printed
3) Current LS is solved
4) User *does not* call `HYPRE_MGRDestroy`
5) New LS is assembled and passed in to `HYPRE_MGRSetup`
6) Now we have a new LS to be dumped to file, the reserved bits will signal that we need to print the new data

src/parcsr_ls/par_mgr_solve.c
victorapm: Yes, logic has changed. New logic is: only print MGR's cycle data if the second bit of `print_level` is active (0x2). This is intended, because we have new print_level codes now (any number between 0 and 31) and we don't want to print MGR's cycle data always when `print_level > 1`. We want to print it only when the second bit is active, hence the bitwise AND with `HYPRE_MGR_PRINT_INFO_SOLVE`

src/parcsr_ls/par_mgr_setup.c
victorapm: > Do they change their state?

- `((print_level & HYPRE_MGR_PRINT_RESERVED_B) == HYPRE_MGR_PRINT_RESERVED_B)` checks if the result of a bitwise AND operation between print_level and `HYPRE_MGR_PRINT_RESERVED_B` is equal to `HYPRE_MGR_PRINT_RESERVED_B`. In other words, it checks if certain bits in print_level are set to the same values as in `HYPRE_MGR_PRINT_RESERVED_B`.
- If the condition above is true, then `HYPRE_MGR_PRINT_INFO_MATRIX` is bitwise OR'ed with `print_level`, i.e., certain bits in print_level will be set to 1 if they are active in `HYPRE_MGR_PRINT_INFO_MATRIX`, effectively turning on the print flag responsible for printing out the matrix.
- if the condition is false, 0 is bitwise OR'ed with `print_level`, effectively leaving `print_level` unchanged.

Same reasoning for the other checks

src/parcsr_ls/par_mgr_solve.c
victorapm: Sorry, Daniel! I read your msg too fast before... Now I see the point was the `tol > 0` check. I agree with you, it shouldn't be there. Fixed with [cef1f17](https://github.com/hypre-space/hypre/pull/976/commits/cef1f178dda6eba2b1378475d2e83d650f46eaf9)

src/parcsr_ls/par_mgr.c
```diff
@@ -6590,62 +6598,151 @@ hypre_MGRPrintCoarseSystem( void *mgr_vdata, HYPRE_Int print_flag)
...
+      else
+      {
+         hypre_IntArraySetInterleavedValues(dofmap, block_size);
+      }
+
+      /* Print Matrix */
+      hypre_ParPrintf(comm, "Writing matrix to path: %s\n", info_path);
+      filename = hypre_TAlloc(char, strlen(info_path) + 16, HYPRE_MEMORY_HOST);
+      hypre_sprintf(filename, "%s/IJ.out.A", info_path);
+      hypre_ParCSRMatrixPrintBinaryIJ(par_A, 0, 0, filename);
```
rfalgout: When printing to file, are you only supporting binary files?  Based on my past experience interacting with users via binary files, it can sometimes be problematic and is the reason we have used ASCII.  Can we support both, but have binary be the default?

victorapm: Yes, that's a good idea. What about:
1. Rename `HYPRE_MGR_PRINT_INFO_MATRIX` to `HYPRE_MGR_PRINT_INFO_MATRIX_BIN`
2. Add `HYPRE_MGR_PRINT_INFO_MATRIX_ASCII`

Same for RHS

victorapm: (This way we don't need a separate flag to signal binary or ascii and we can handle all cases with `print_level` alone)

rfalgout: Well, I was thinking of only using print_level anyway.  Wouldn't it be easier to just add a '32' to the print level to signify printing ASCII?  Otherwise, don't you need both a '32' and a '64'?

rfalgout (approved):
This looks good modulo the ASCII discussion we are having.  Nice work!

src/parcsr_ls/par_mgr.c
victorapm: Oh yes, now I see it. Agreed, will do this way

victorapm: Rob, I changed code "8" to signal ASCII printing and moved the matrix and RHS printout codes to "16" and "32". I also reserved "64" and "128" to print additional debugging information (to be implemented in another PR along with code "4"). See changes in [4562249](https://github.com/hypre-space/hypre/pull/976/commits/45622494720a6d691518559c3db9173d352f8d1b) and [9f1f672](https://github.com/hypre-space/hypre/pull/976/commits/9f1f672e317e2b14aaf5f4700b006f73c3479d85)

oseikuffuor1 (approved):
@victorapm, I like the idea quite a bit, but my main concern was in line with @rfalgout comment about the approach being crytpic for users (and potentially developers too). Your edit to the documentation makes things much clearer for users (than what was there before). I also agree it will be a good idea to discuss this with the larger team at some point, since this is likley to be adopted for other solvers and preconditioners in hypre. Nice job!

victorapm:
Thanks all! I'll merge this into master after #897 gets in

PR #983, opened 2023-10-09
Add `HYPRE_MGRSetFSolverAtLevel`

victorapm:
Add an option to set MGR's F-relaxation solver at any level

TODO:

- [ ] Remove `mgr_data -> fsolver_mode`?


src/parcsr_ls/par_mgr_setup.c
```diff
@@ -1527,7 +1524,7 @@ hypre_MGRSetup( void               *mgr_vdata,
                   (mgr_data -> fsolver_mode) = 1;
                }
             }
-            else
+            else if (aff_solver[lev])
```
oseikuffuor1, line 1530: @victorapm is this change necessary, since lev is 0?

src/parcsr_ls/par_mgr_setup.c
```diff
@@ -1536,6 +1533,19 @@ hypre_MGRSetup( void               *mgr_vdata,
                hypre_error_w_msg(0, msg);
             }
          }
+         else if (aff_solver[lev])
```
oseikuffuor1, line 1539: @victorapm This piece of code always assumes that we call the setup routines that users give us on MGR's internal data right?

src/parcsr_ls/par_mgr_setup.c
victorapm: This ensures that the warning message is displayed only when the F-relaxation solver at level `lev` is not NULL

src/parcsr_ls/par_mgr_setup.c
victorapm: 1. Not sure if I understood the question. but this `else if` provides the mechanism for running the F-relaxation solver created externally by the user.
2. In this code, they are not related. Do you want to create a new code for `Frelax_type` to denote "custom", i.e., "created externally by the user"?

src/parcsr_ls/par_mgr_solve.c
```diff
@@ -984,10 +985,23 @@ hypre_MGRCycle( void              *mgr_vdata,
             if (Frelax_type[level] == 2)
             {
                /* Do F-relaxation using AMG */
-               fine_grid_solver_solve((mgr_data -> aff_solver)[fine_grid],
-                                      A_ff_array[fine_grid],
-                                      F_fine_array[coarse_grid],
-                                      U_fine_array[coarse_grid]);
+               if (level == 0)
+               {
+                  /* TODO (VPM): unify with the next block */
```
oseikuffuor1, line 990: I agree these two should be merged at some point.

victorapm: Yes, the idea is to remove `fine_grid_solver_solve` and `fine_grid_solver_setup` from the MGR's struct. But first, we need to extend the idea of `hypre_Solver` to other preconditioners in hypre...

oseikuffuor1 (approved):
Thanks again @victorapm. This looks good.

PR #991, opened 2023-10-19
fix compile error

liruipeng:
This PR fixes compile errors/warnings on lassen.

- [x] tux
- [x] lassen
- [x] tioga

### DIFF
--- a/src/parcsr_ls/par_ge_device.c
+++ b/src/parcsr_ls/par_ge_device.c
@@ -105,9 +105,9 @@ hypre_GaussElimSetupDevice(hypre_ParAMGData *amg_data,
 
 #else
    /* Silence declared but never referenced warnings */
-   (A_piv  = NULL);
-   (A_mat  = NULL);
-   (A_work = NULL);
+   (A_piv  += 0);
+   (A_mat  += 0);
+   (A_work += 0);
    (buffer_size *= 1);
 
    hypre_error_w_msg(HYPRE_ERROR_GENERIC,
@@ -309,8 +309,8 @@ hypre_GaussElimSolveDevice(hypre_ParAMGData *amg_data,
 
 #else
    /* Silence declared but never referenced warnings */
-   (A_mat = NULL);
-   (A_piv = NULL);
+   (A_mat += 0);
+   (A_piv += 0);
    (i_one *= 1);
    (d_one *= 1.0);
    (zero  *= zero);
--- a/src/utilities/int_array_device.c
+++ b/src/utilities/int_array_device.c
@@ -152,6 +152,7 @@ hypre_IntArrayNegateDevice( hypre_IntArray *v )
  * hypre_IntArraySetInterleavedValuesDevice
  *--------------------------------------------------------------------------*/
 
+#if defined(HYPRE_USING_CUDA) || defined(HYPRE_USING_HIP)
 struct hypre_cycle_functor
 {
    HYPRE_Int cycle_length;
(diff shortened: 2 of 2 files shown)


victorapm (approved):
Thanks for the fixes!

src/parcsr_ls/par_ge_device.c
```diff
@@ -105,9 +105,9 @@ hypre_GaussElimSetupDevice(hypre_ParAMGData *amg_data,
 
 #else
    /* Silence declared but never referenced warnings */
-   (A_piv  = NULL);
-   (A_mat  = NULL);
-   (A_work = NULL);
```
waynemitchell, line 110: What was the issue with setting these to `NULL`? Just curious. I find it funny that we keep modifying this code that is really just here to get rid of some compiler warnings.

waynemitchell (approved):

src/parcsr_ls/par_ge_device.c
liruipeng: The previous code doesn’t work for the compilers used on lassen (still having the set-but-not-used warning). In general, I don't like to trick the compilers in a way to silence warnings. Warnings are there for reasons. I would rather write "the most accurate and robust" code, even if it's longer. This "fix" seems to work fine and a minimum change to what we have now. I would also like to hear suggestions from @rfalgout @ulrikeyang @oseikuffuor1 .  Thanks.

waynemitchell: Oh, weird. I am the one who put in the `= NULL` version of this, and I thought I had gotten a clean lassen regression with this change in. But this must have been a last minute change that I didn't test... thanks for fixing!

rfalgout: This seems reasonable to me.  The alternative would be to put the variables in an #if as well.  I doubt that's cleaner.

liruipeng: > Oh, weird. I am the one who put in the `= NULL` version of this, and I thought I had gotten a clean lassen regression with this change in. But this must have been a last minute change that I didn't test... thanks for fixing!

Strange... This was the warning:
```
/usr/tce/packages/cuda/cuda-10.1.243/bin/nvcc -ccbin=mpixlC -gencode arch=compute_70,code=sm_70  -g -O0 -lineinfo -expt-extended-lambda -std=c++11 --x cu -Xcompiler "-g "  -DHAVE_CONFIG_H -I.. -I../distributed_ls/Euclid -I. -I./.. -I./../blas -I./../lapack -I./../multivector -I./../utilities -I./../krylov -I./../seq_mv -I./../parcsr_mv -I./../distributed_matrix -I./../matrix_matrix -I./../IJ_mv -I./../parcsr_block_mv   -I/usr/tce/packages/cuda/cuda-10.1.243/include         -c par_ge_device.c -o par_ge_device.obj
par_ge_device.c(32): warning: variable "A_piv" was set but never used

par_ge_device.c(33): warning: variable "A_mat" was set but never used

par_ge_device.c(34): warning: variable "A_work" was set but never used

par_ge_device.c(226): warning: variable "A_mat" was set but never used

par_ge_device.c(227): warning: variable "A_piv" was set but never used
```
Do we use the same compiler, flags, etc?

waynemitchell: > Do we use the same compiler, flags, etc?

I just ran the `machine-lassen.sh` script to test, I didn't do any of my own configure/compile.

PR #993, opened 2023-10-20
Skipbreak

ulrikeyang:
This PR is my previous PR called 'skipcheck' but updated to the current master.
It contains options to avoid breaking for certain very small variables in CG and adds a flexible option.
I only tested this on tux box.

### DIFF
--- a/src/krylov/pcg.c
+++ b/src/krylov/pcg.c
@@ -99,6 +99,7 @@ hypre_PCGCreate( hypre_PCGFunctions *pcg_functions )
    (pcg_data -> recompute_residual) = 0;
    (pcg_data -> recompute_residual_p) = 0;
    (pcg_data -> stop_crit)    = 0;
+   (pcg_data -> skip_break)   = 0;
    (pcg_data -> converged)    = 0;
    (pcg_data -> hybrid)       = 0;
    (pcg_data -> owns_matvec_data ) = 1;
@@ -111,6 +112,8 @@ hypre_PCGCreate( hypre_PCGFunctions *pcg_functions )
    (pcg_data -> p)            = NULL;
    (pcg_data -> s)            = NULL;
    (pcg_data -> r)            = NULL;
+   (pcg_data -> r_old)        = NULL;
+   (pcg_data -> v)            = NULL;
 
    HYPRE_ANNOTATE_FUNC_END;
 
@@ -161,6 +164,16 @@ hypre_PCGDestroy( void *pcg_vdata )
          (*(pcg_functions->DestroyVector))(pcg_data -> r);
          pcg_data -> r = NULL;
       }
+      if ( pcg_data -> r_old != NULL )
+      {
+         (*(pcg_functions->DestroyVector))(pcg_data -> r_old);
+         pcg_data -> r_old = NULL;
+      }
+      if ( pcg_data -> v != NULL )
+      {
+         (*(pcg_functions->DestroyVector))(pcg_data -> v);
+         pcg_data -> v = NULL;
+      }
       hypre_TFreeF( pcg_data, pcg_functions );
       hypre_TFreeF( pcg_functions, pcg_functions );
    }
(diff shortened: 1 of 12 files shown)


waynemitchell (approved):
Looks good to me!

src/krylov/pcg.c
```diff
@@ -540,18 +585,59 @@ hypre_PCGSolve( void *pcg_vdata,
...
+      /* alpha = <r, Cr> / <p, Ap> */
+      /* for alpha close to 0 x and r will not change much unless x and/or r are close to 0 */
+      if (alpha <=0.0)
+      {
+         if (print_level > 1 && my_id == 0)
+	 {
+            hypre_printf("alpha %e", alpha);
+	 }
+         hypre_error_w_msg(HYPRE_ERROR_CONV, "Negative or zero alpha value in PCG");
+         if (skip_break == 0)
```
rfalgout: Maybe I'm not reading this right, but shouldn't this break for `skip_break` equal `0` or `1`?  I'm looking at the documentation in `HYPRE_krylov.h`.

src/krylov/pcg.c
```diff
@@ -540,18 +585,59 @@ hypre_PCGSolve( void *pcg_vdata,
...
+	 }
+      }
+      else if (! (alpha  >= HYPRE_REAL_TRUE_MIN))
+      {
+         hypre_error_w_msg(HYPRE_ERROR_CONV, "alpha value less than TRUE_MIN in PCG");
+         if (print_level > 1 && my_id == 0)
+	 {
+            hypre_printf("alpha %e", alpha);
+	 }
+         if (skip_break < 2)
```
rfalgout, line 611: This should break for values `0` or `2` (not `1`).  This block can only happen if
(0 < alpha < HYPRE_REAL_TRUE_MIN)

src/krylov/pcg.c
```diff
@@ -540,18 +585,59 @@ hypre_PCGSolve( void *pcg_vdata,
...
+      else if (! (alpha  >= HYPRE_REAL_MIN))
       {
          hypre_error_w_msg(HYPRE_ERROR_CONV, "Subnormal alpha value in PCG");
-         if (i == 1) { i_prod = i_prod_0; }
-         break;
+         if (print_level > 1 && my_id == 0)
+	 {
+            hypre_printf("alpha %e", alpha);
+	 }
+         if (skip_break < 3)
```
rfalgout: This should break for values `0` or `3` (not `1` or `2`).  This block can only happen if
(HYPRE_REAL_TRUE_MIN < alpha < HYPRE_REAL_MIN)

rfalgout:
Hi @ulrikeyang .  Thanks for getting this ready again.  I am confused about the break logic relative to the documentation.  Maybe I'm misreading this, and if so, my apologies.  I seemed to think this was correct the first time I reviewed it because it's exactly the same.  Let me know.  Thanks!

src/krylov/pcg.c
ulrikeyang: Here we only check for alpha <= 0, however you are right that there is one more branch missing, where skip_break should be 1, the one where 0 < alpha < HYPRE_REAL_TRUE_MIN, so I will add that.

ulrikeyang: Actually what I just said is not correct. I think the problem is just that the documentation is wrong. We need to shift the parameters by 1 down, and make the default 2, right now 0 and 3 are the same. If we do this things should be fine.

ulrikeyang:
I fixed the documentation which should resolve all the issues Rob mentioned.

rfalgout:
Hi @ulrikeyang .  This almost matches the documentation now, except that `2` doesn't break for all subnormal values, it only breaks when (HYPRE_REAL_TRUE_MIN <= alpha < HYPRE_REAL_MIN).  It looks like `0` breaks for all values.

Regarding the documentation, though, it seems to me that the way a user would want this to behave is as follows (and now that I stare at it more carefully, I think you just need to change the `if (skip_break)` lines to make this happen, i.e., put `< 3` in the first if block and `< 1` in the last if block):
- If set to 0: break if ( < HYPRE_REAL_MIN) or ( < HYPRE_REAL_TRUE_MIN) or  ( <=0)  <-- Default setting (all subnormal)
- If set to 1: break if ( < HYPRE_REAL_TRUE_MIN) or ( <=0)
- If set to 2: break if ( <=0)
- If set to 3: don't break at all

I can't imagine a user wanting to break if (0 < alpha < HYPRE_REAL_TRUE_MIN) but not when it's negative, for example.

ulrikeyang:
Ok. It should be all fixed now.

rfalgout:
> Ok. It should be all fixed now.

Looks great!  Thanks again for adding this feature!  Our users will surely appreciate having the ability to ignore these errors, get more informative messages, and have the flexibility to tune it any way they like.

rfalgout (approved):
Thanks!

liruipeng (approved):
Looks good to me. Thank you!

PR #995, opened 2023-10-25
Fix sycl shared build

waynemitchell:
This fixes the shared library build for hypre with the sycl backend.


src/configure
```diff
@@ -4582,30 +4577,6 @@ then :
...
-
-
-
-# Check whether --with-magma-lib was given.
-if test "${with_magma_lib+set}" = set; then :
-  withval=$with_magma_lib; for magma_lib in $withval; do
-    MAGMA_LIBS="$MAGMA_LIBS $magma_lib"
- done
-
-fi
```
waynemitchell, line 4608: @victorapm , when I reran `config/update.sh`, there were some deletions of magma-related stuff. This stuff mostly looks redundant, but I wanted to double check with you that this looks OK.

waynemitchell: Ah, these deletions actually break the build. I get the error:

icpx: error #10005: unable to open argument file 'MAGMA_LIBS@'

I notice that we have both `MAGMA_LIBS` and `HYPRE_MAGMA_LIB` listed in Makefile.config.in:

https://github.com/hypre-space/hypre/blob/master/src/config/Makefile.config.in#L177

Should one of these sets of magma-related variables be removed?

victorapm: Hi Wayne, yes, the ones without the hypre prefix should be removed.

I thought i had fixed this before... Sorry..

victorapm (approved):

ulrikeyang (approved):
This looks good, as far as I can tell. Thanks for taking care of the MAGMA inclusions also. We want those to work.

src/configure
waynemitchell: No problem. You can double check my final changes. I removed `MAGMA_LIBS` replacing with `HYPRE_MAGMA_LIB` and `HYPRE_MAGMA_LIB_DIR` where appropriate.

osborn9: I just tested a MAGMA enabled build and everything works as expected.

rfalgout (approved):
Looks good to me.  Thanks!

osborn9 (approved):
Looks good! 

Note - there are still some issues with the spack build, but are unrelated to this change (and maybe even hypre proper)

PR #996, opened 2023-10-26
use hipPointerAttribute_t.type as HIP is removing hipPointerAttribute_t.memoryType

rakesroy:
- In ROCm6.0 hipPointerAttribute_t.memoryType will be removed from HIP.
- Instead hipPointerAttribute_t.type to be used.
- This is causing build failure for AMG

### DIFF
--- a/src/utilities/memory.c
+++ b/src/utilities/memory.c
@@ -1142,11 +1142,19 @@ hypre_GetPointerLocation(const void *ptr, hypre_MemoryLocation *memory_location)
    {
       *memory_location = hypre_MEMORY_UNIFIED;
    }
+#if (HIP_VERSION_MAJOR >= 6)
+   else if (attr.type == hipMemoryTypeDevice)
+#else // (HIP_VERSION_MAJOR < 6)
    else if (attr.memoryType == hipMemoryTypeDevice)
+#endif // (HIP_VERSION_MAJOR >= 6)
    {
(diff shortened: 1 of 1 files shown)


pbauman:
@rakesroy Is this a backwards compatible change? Is HYPRE going to compile with previous versions of ROCm with this change? If not, then this PR is insufficient - guards will be needed based on the ROCm (or HIP?) version to support both pre- and post-ROCm6.0.

rakesroy:
> @rakesroy Is this a backwards compatible change? Is HYPRE going to compile with previous versions of ROCm with this change? If not, then this PR is insufficient - guards will be needed based on the ROCm (or HIP?) version to support both pre- and post-ROCm6.0.

rakesroy:
@pbauman,
The member hipPointerAttribute_t.type has been existing since ROCm5.5.
So this change will be backward compatible till ROCm5.5.

victorapm:
Thanks @rakesroy! We would also need the build with rocm 5.2.0 to work

rakesroy:
I have updated this PR to to use .type when (HIP_VERSION_MAJOR >= 6), otherwise to use .memoryType for backwards compatibility.

victorapm (approved):
Thank you!

rakesroy:
I am not authorized to merge this PR.
Can anyone please merge it?

PR #1003, opened 2023-11-03
CUDA compile flag check

### LINKED ISSUE #1000: CudaCompileFlagCheck for major and minor versions
Hello,
I am building hypre for multiple architectures, Quadro, Volta, and Ampere with    

`CUDA_HOME=/usr/local/cuda ./configure --with-cuda --with-gpu-arch='60 70 80' --prefix /usr/local --enable-shared --enable-device-memory-pool  --disable-static  --enable-unified-memory`

When  initializing hypre, device utils asserts that my actual architecture doesnt match the compiled architecture.  Shouldn't the device downgrade when running a lower architecture than what it states?

actual=610
compiled=600

Why assert in this case?  

Thanks
Joe

liruipeng:
To fix #1000

### DIFF
--- a/src/utilities/device_utils.c
+++ b/src/utilities/device_utils.c
@@ -2688,7 +2688,12 @@ hypre_CudaCompileFlagCheck()
 
    /* HYPRE_CUDA_CALL(cudaDeviceSynchronize()); */
 
-   if (cuda_arch_actual / 100 != cuda_arch_compile / 100)
+   const hypre_int cuda_arch_actual_major  = cuda_arch_actual  / 100;
+   const hypre_int cuda_arch_compile_major = cuda_arch_compile / 100;
+   const hypre_int cuda_arch_actual_minor  = cuda_arch_actual  % 100;
+   const hypre_int cuda_arch_compile_minor = cuda_arch_compile % 100;
+
+   if (cuda_arch_actual_major != cuda_arch_compile_major || cuda_arch_actual_minor < cuda_arch_compile_minor)
    {
       char msg[256];
 
--- a/src/utilities/device_utils.c
+++ b/src/utilities/device_utils.c
@@ -2688,7 +2688,7 @@ hypre_CudaCompileFlagCheck()
 
    /* HYPRE_CUDA_CALL(cudaDeviceSynchronize()); */
 
-   if (cuda_arch_actual != cuda_arch_compile)
+   if (cuda_arch_actual / 100 != cuda_arch_compile / 100)
    {
       char msg[256];
 


victorapm:
Looks good!! Should we do:
```
cuda_arch_actual < cuda_arch_compile
```

liruipeng:
> Code compiled for a major arch would work on the minor arches of that same generation.

So just need to match the major version.

victorapm:
Thanks! I'm thinking about the situation where someone compiles for `sm_70` and wants to run on a A100 (`sm_80`). CUDA provides backward compatibility and allow that, but should we allow this or not?

liruipeng:
> Thanks! I'm thinking about the situation where someone compiles for `sm_70` and wants to run on a A100 (`sm_80`). CUDA provides backward compatibility and allow that, but should we allow this or not?

Are you sure this works fine (compiling with 70 and running on 80)?

victorapm:
That seems the case to me. Just verified on lassen:
```
$ cat hello.cu
#include <stdio.h>
#include <cuda.h>
#include <cuda_runtime.h>

#define CUDA_CHECK_ERROR(err)                          \
  do {                                                 \
     cudaError_t err_ = (err);                         \
     if (err_ != cudaSuccess) {                        \
        fprintf(stderr, "CUDA error: %s at %s:%d\n",   \
        cudaGetErrorString(err_), __FILE__, __LINE__); \
        exit(EXIT_FAILURE);                            \
     }                                                 \
  } while (0)

__global__ void helloCUDA() {
  printf("Hello, CUDA World!\n");
}

int main() {
  helloCUDA<<<1, 1>>>();

  CUDA_CHECK_ERROR(cudaGetLastError());

  cudaDeviceSynchronize();

  return 0;
}

$ nvcc -arch=sm_60 hello.cu

$ lrun -n1 ./a.out 
Hello, CUDA World!

$ nvcc -arch=sm_50 hello.cu

$ lrun -n1 ./a.out 
Hello, CUDA World!
```

So, we would need to decide whether allowing this or not

liruipeng:
> That seems the case to me. Just verified on lassen:
> 
> ```
> $ cat hello.cu
> #include <stdio.h>
> #include <cuda.h>
> #include <cuda_runtime.h>
> 
> #define CUDA_CHECK_ERROR(err)                          \
>   do {                                                 \
>      cudaError_t err_ = (err);                         \
>      if (err_ != cudaSuccess) {                        \
>         fprintf(stderr, "CUDA error: %s at %s:%d\n",   \
>         cudaGetErrorString(err_), __FILE__, __LINE__); \
>         exit(EXIT_FAILURE);                            \
>      }                                                 \
>   } while (0)
> 
> __global__ void helloCUDA() {
>   printf("Hello, CUDA World!\n");
> }
> 
> int main() {
>   helloCUDA<<<1, 1>>>();
> 
>   CUDA_CHECK_ERROR(cudaGetLastError());
> 
>   cudaDeviceSynchronize();
> 
>   return 0;
> }
> 
> $ nvcc -arch=sm_60 hello.cu
> 
> $ lrun -n1 ./a.out 
> Hello, CUDA World!
> 
> $ nvcc -arch=sm_50 hello.cu
> 
> $ lrun -n1 ./a.out 
> Hello, CUDA World!
> ```
> 
> So, we would need to decide whether allowing this or not

Can you try hypre? I remember this didn't work before when we had machine ray.

victorapm:
Yep! It seems thrust doesn't allow:
```
terminate called after throwing an instance of 'thrust::system::system_error'
```

Not clear the reason why, as NVIDIA states that it should work, but I don't think it's worth digging into it...

As a matter of fact, I also tried compiling hypre with `--with-gpu-arch=72`, which should work on lassen with this PR, but got the same error above.

So, now, I'm questioning myself if we should even allow running on a device with the same major version as the code compiled (this PR)... Not even my "Hello world" code works with `sm_72` or `sm_75` on lassen

victorapm (approved):

PR #1011, opened 2023-11-15
Add iterative ILU0

victorapm:
This PR adds an interface to the iterative ILU preconditioner with zero fill-in implemented in rocSPARSE. The method is iterative in the context of the factorization phase.

This works only in device runs (AMD). It can be used as precondtitioner or complex smoother to BoomerAMG.

###  Iterative ILU0 as preconditioner to GMRES:

``` bash
$ srun -n 1 ./src/test/ij -n 128 128 128 -solver 81 -ilu_type 0 -ilu_reordering 0 -ilu_iter_setup_type 1 -ilu_iter_setup_option 10
```
### Iterative ILU0 as smoother to BoomerAMG-GMRES:

``` bash
$ srun -n 1 ./src/test/ij -n 128 128 128 -solver 3 -smlv 4 -smtype 5 -ilu_type 0 -ilu_reordering 0 -ilu_iter_setup_type 1 -ilu_iter_setup_option 10
```


victorapm:
Hi all, let me know your thoughts about this PR when you have time. Thanks!

src/parcsr_ls/HYPRE_parcsr_ls.h
```diff
@@ -4508,6 +4536,58 @@ HYPRE_ILUSolve( HYPRE_Solver solver,
...
+ *    -  3 : Synchronous with explicit storage splitting
+ *    -  4 : Semi-synchronous with explicit storage splitting
+ *
+ * Note: Iterative ILU is available only for zero fill-in and it depends on rocSPARSE.
+ **/
+HYPRE_Int
+HYPRE_ILUSetIterativeSetupType( HYPRE_Solver solver, HYPRE_Int iter_setup_type );
+
+/**
+ * (Optional) Set the compute option for iterative ILU in a bitwise fashion. Options are:
```
rfalgout: I would shy away from talking about bits.  Instead, maybe refer to these as "additive options" where several of them can be set at the same time by adding the individual numeric values.  I think you had a good user-level explanation of this somewhere else in the library (having to do with print options?).

victorapm: Yes! I can replicate what I did for `HYPRE_MGRSetPrintLevel` here

rfalgout (approved):
Looks good to me.  Thanks!

src/parcsr_ls/HYPRE_parcsr_ls.h
victorapm: Done in https://github.com/hypre-space/hypre/pull/1011/commits/39621a3ee3f00d74d343513db4bf4be4e46e94a5

src/parcsr_ls/HYPRE_parcsr_ls.h
```diff
@@ -4508,6 +4536,58 @@ HYPRE_ILUSolve( HYPRE_Solver solver,
 HYPRE_Int
 HYPRE_ILUSetMaxIter( HYPRE_Solver solver, HYPRE_Int max_iter );
 
+/**
+ * (Optional) Set the algorithm type for compute the ILU factorization. Options are:
```
oseikuffuor1: Typo. "to compute the iterative ILU ..."

victorapm: Fixed in https://github.com/hypre-space/hypre/pull/1011/commits/0ce97929ef0248becd67bf7551d2485937c701ed

src/parcsr_ls/HYPRE_parcsr_ls.h
```diff
@@ -4508,6 +4536,58 @@ HYPRE_ILUSolve( HYPRE_Solver solver,
...
+
+/**
+ * (Optional) Set the compute option for iterative ILU in an additive fashion, i.e.; multiple
+ * options can be turned on by summing their respective numeric codes as given below:
+ *
+ *    -  2: Use stopping tolerance to finish the algorithm
+ *    -  4: Compute correction norms
+ *    -  8: Compute residual norms
+ *    - 16: Save convergence history
+ *    - 32: Use rocSPARSE's internal COO format
```
oseikuffuor1, line 4563: Option 32 seems independent of the other options which determine the stopping criterion. I would assume one could set option 32 in addition to one of the other options?

victorapm: Yes! More about this on rocSPARSE's documentation

src/parcsr_ls/par_ilu_setup_device.c
```diff
@@ -146,7 +179,17 @@ hypre_ILUSetupILUDevice(HYPRE_Int               ilu_type,
          hypre_CSRMatrixPermute(hypre_ParCSRMatrixDiag(A), perm_data, rqperm_data, &A_diag);
 
          /* Compute ILU0 on the device */
-         hypre_CSRMatrixILU0(A_diag);
+         if (iter_setup_type)
```
oseikuffuor1, line 180: Since the check for using rocsparse is embedded in the hypre_ILUSetupIterativeILU0Device call, the assumption is that new support for iterative ILU factorization on other devices will have the same function prototype. Is that the idea?

victorapm: That's correct

src/parcsr_ls/par_ilu_setup_device.c
```diff
@@ -373,4 +416,190 @@ hypre_ILUSetupILUDevice(HYPRE_Int               ilu_type,
...
+ *   A - input matrix
+ *   type - algorithm for computing iterative ILU0
+ *   option - internal flags used by rocSPARSE
+ *   max_iter - max. number of iterations
+ *   tolerance - stopping criteria in iterative algorithm
+ *
+ * Output arguments:
+ *   num_iter_ptr - number of iterations
+ *   history_ptr - list of corrections and residual values for each iteration
+ *                 (computed only when residuals_ptr points to NULL at entrance)
```
oseikuffuor1: Do you mean if the history_ptr points to NULL here (instead of residuals_ptr? Also, the logic seems a bit confusing for me. We want to populate history information if the pointer is NULL. What if it is not? Would we not want updated info?

oseikuffuor1 (approved):
@victorapm this looks very good to me. I have left a few comments for your consideration. Thanks.

src/parcsr_ls/HYPRE_parcsr_ls.h
oseikuffuor1: So should we support two option arguments here then?

src/parcsr_ls/par_ilu_setup_device.c
victorapm: The comment is from an old version of the code and it's wrong now. I'll fix!

src/parcsr_ls/HYPRE_parcsr_ls.h
victorapm: We do. Users can add the codes to turn on/off different behaviors

src/parcsr_ls/par_ilu_setup_device.c
victorapm: Done in [951faeb](https://github.com/hypre-space/hypre/pull/1011/commits/951faeb683eef098643fae4b2501db948c84f021)

victorapm:
@liruipeng Can you review this PR?

liruipeng (approved):
Thanks!

PR #1013, opened 2023-11-18
Fix regression tests with-concurrent-hopscotch 

liruipeng:
Concurrent-hopscotch (OMP) cannot guarantee some of the "diff" tests. So leave them out from the main test scripts and don't "diff" with `HYPRE_NO_SAVED`.

- [x] tux
- [x] lassen

### DIFF
--- a/src/test/TEST_error/error.sh
+++ b/src/test/TEST_error/error.sh
@@ -37,3 +37,4 @@ egrep -v "hypre error" ${TNAME}.tmp        >  ${TNAME}.err
 #=============================================================================
 
 rm -f ${TNAME}.testdata*
+rm -f ${TNAME}.tmp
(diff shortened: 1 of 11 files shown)


victorapm (approved):

rfalgout:
It looks like this is the first time we use HYPRE_NO_SAVED inside the individual scripts.  I guess I don't see a problem with doing that, though in this particular case, it's not really turning off a "saved test" since all of the output being diffed is generated during the test.  Maybe it would be good to add a comment just before the `if` tests to make it clear what's happening here.  I was confused at first since HYPRE_NO_SAVED was already turned on in the tux openmp tests (and always has been).

rfalgout (approved):
Looks good modulo my one comment.  Thanks!

liruipeng:
> Looks good modulo my one comment. Thanks!

Thank you! The comment was added.

PR #1022, opened 2023-12-05
Add `hypre_DenseBlockMatrix`

victorapm:
This PR adds a data structure for storing dense block diagonal matrices in a single MPI rank. This is useful for preconditioners/smoothers employing batched linear solvers.

- Add new seq_block_mv folder
- Add `hypre_DenseBlockMatrix` struct


src/lib/Makefile
```diff
@@ -5,69 +5,53 @@
...
-$(PARCSRBLOCKMVFILES)\
-$(SEQMVFILES)\
-$(SSTRUCTLSFILES)\
-$(SSTRUCTMVFILES)\
-$(STRUCTLSFILES)\
-$(STRUCTMVFILES)\
-$(UTILITIESFILES)\
-$(BLASFILES)\
-$(LAPACKFILES)\
-$(IJMVOBJFILES)\
```
rfalgout, line 62: Are you sure you can delete these?  We used to have to do something special for the '.obj' files so I think these were needed at one time.

victorapm: Yes, this is fine. I moved the `.obj` files to their respective folders, e.g., the `.obj` files under `$(IJMVOBJFILES)` are now in `$(IJMVFILES)`. This allowed to reduce some lines of code... 

GPU compilation after this change looks fine (tested on lassen and tioga)

rfalgout: I see.  You might still want to check with Rui Peng just in case he had another reason for organizing things the way he did.  I remember he and I chatting about this at the time, but I don't recall the details.

victorapm: @liruipeng can you verify this particular change looks good?

src/seq_block_mv/dense_block_matrix.c
```diff
@@ -0,0 +1,334 @@
...
+                              HYPRE_Int  num_rows_block,
+                              HYPRE_Int  num_cols_block )
+{
+   hypre_DenseBlockMatrix  *A;
+   HYPRE_Int                num_blocks[2];
+
+   /* Compute number of blocks */
+   num_blocks[0] = 1 + ((num_rows - 1) / num_rows_block);
+   num_blocks[1] = 1 + ((num_cols - 1) / num_cols_block);
+   if (num_blocks[0] != num_blocks[1])
```
ulrikeyang, line 33: You want this matrix block structure to be square? But you allow non-square dense blocks?
Is this needed?

victorapm: Not necessarily square. I need block rectangular as well (at least for the purpose of the new MGR interpolation, but this might be useful in other places as well in the future (AIR?))

ulrikeyang: I just don't understand why you don't allow that the number of blocks in x-direction are not equal to the number of blocks in y-direction.

ulrikeyang (approved):
Overall this is fine except for the part I don't understand :)

src/seq_block_mv/dense_block_matrix.c
victorapm: Because it you think in 2D, there's actually just one block that spans both dimensions.

Suppose the full matrix is 8x8 and you are trying to build it with 2x4 local blocks. That shouldn't work because there isn't enough space in the second dimension. The check prevents a situation like this

rfalgout (approved):

src/seq_block_mv/dense_block_matrix.c
ulrikeyang: That is just a matter of definition. I don't see an issue there. I am thinking about a much more general matrix than you apparently. It seems that you want the original matrix to be square.

victorapm: No, the original matrix can be rectangular. But that implies the local blocks to be rectangular as well.

What I don't support is a mix of both, i.e., full matrix square while local block matrices rectangular (or vice-versa)

ulrikeyang: So, you don't want to deal with interpolation matrices that might have the same block sizes? Those would not be allowed with that condition.

src/lib/Makefile
liruipeng: I guess it's OK assuming it's tested. I don't see or remember there has to be a particular order for the .o/.objs.

src/seq_block_mv/dense_block_matrix.c
victorapm: Yes, that isn't supported here. I guess if we need this scenario to work, we could revisit.

Issue #1026, opened 2023-12-07
Change `--amdgpu-target` to `--offload-arch`


victorapm:
PS: we need to change `--amdgpu-target` to `--offload-arch`, but that seems a separate issue :)

_Originally posted by @victorapm in https://github.com/hypre-space/hypre/issues/1023#issuecomment-1843533142_

See the warning:
```
Warning: The --amdgpu-target option has been deprecated and will be removed in the future. Use --offload-arch instead.
```

victorapm:
Closed by #1027

PR #1027, opened 2023-12-07
Replace hipcc option `--amdgpu-target` with `--offload-arch`

victorapm:
Solves #1026.

Tested with rocm 5.2 and newer (tioga)

### DIFF
--- a/src/config/configure.in
+++ b/src/config/configure.in
@@ -2644,7 +2644,7 @@ AS_IF([test x"$hypre_using_hip" == x"yes"],
 
         if test "x${HYPRE_CUDA_SM}" != "x"
         then
-           HYPRE_CUDA_GENCODE="--amdgpu-target="
+           HYPRE_CUDA_GENCODE="--offload-arch="
            for sm in ${HYPRE_CUDA_SM}; do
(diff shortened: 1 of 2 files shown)


liruipeng:
Is it back compatible?

victorapm:
Tested with rocm 5.2, which is the minimum we support now

liruipeng:
> Tested with rocm 5.2, which is the minimum we support now

Do we state this somewhere? In the manual?

victorapm:
I didn't find it either. We should add to the manual/Wiki...

Right now, 5.2 is what we test with machine-tioga and hypre doesn't compile with versions older than 5.2 (I tried).

I don't think we should support rocm older than 5.2

liruipeng:
> I didn't find it either. We should add to the manual/Wiki...
> 
> Right now, 5.2 is what we test with machine-tioga and hypre doesn't compile with versions older than 5.2 (I tried).
> 
> I don't think we should support rocm older than 5.2

Agreed. I used to test 5.1 or even 5.0. Not sure if it's still worth it.

liruipeng (approved):

PR #1033, opened 2023-12-13
Update `hypre_ParVectorPrintIJ`

victorapm:
Add multi-component vector support to `hypre_ParVectorPrintIJ`

This feature is important for GEOS


src/parcsr_mv/par_vector.c
```diff
@@ -972,37 +974,59 @@ hypre_ParVectorPrintIJ( hypre_ParVector *vector,
...
 
    hypre_MPI_Comm_rank(comm, &myid);
    hypre_MPI_Comm_size(comm, &num_procs);
 
    hypre_sprintf(new_filename, "%s.%05d", filename, myid);
-
    if ((file = fopen(new_filename, "w")) == NULL)
    {
-      hypre_error_w_msg(HYPRE_ERROR_GENERIC, "Error: can't open output file %s\n");
+      hypre_error_w_msg(HYPRE_ERROR_GENERIC, "Error: can't open output file!");
```
oseikuffuor1: Should we actually include the filename here in the print statement to be clear?

src/parcsr_mv/par_vector.c
```diff
@@ -972,37 +974,59 @@ hypre_ParVectorPrintIJ( hypre_ParVector *vector,
...
+         hypre_fprintf(file, "\n");
+      }
+   }
+   else
+   {
+      /* Single-component (regular) vectors */
+      for (j = 0; j < local_size; j++)
+      {
+         hypre_fprintf(file, "%b %.14e\n",
+                       (HYPRE_BigInt) j + base_j,
```
oseikuffuor1: Does this need to be BigInt? I thought local sizes were always int?

src/parcsr_mv/par_vector.c
```diff
@@ -972,37 +974,59 @@ hypre_ParVectorPrintIJ( hypre_ParVector *vector,
...
+                       hypre_VectorEntryI(local_vector, j));
```
oseikuffuor1, line 1024: This change replaces local_data[(HYPRE_Int)(j - part0)]. Is part0 assumed to be zero here? If so, is that always true?

oseikuffuor1 (commented):
@victorapm I left you a few comments for your review.

src/parcsr_mv/par_vector.c
victorapm: Yes, good idea

src/parcsr_mv/par_vector.c
victorapm: These are global indices, so need to be BigInt

src/parcsr_mv/par_vector.c
victorapm: `part0` is not assumed to be zero. The new form is an alternative option to before which doesn't require the additional subtraction of indices

src/parcsr_mv/par_vector.c
victorapm: Done in [ba9687f](https://github.com/hypre-space/hypre/pull/1033/commits/ba9687f24dbd6a1f25da04ed6d073634bc5a1c2e)

src/parcsr_mv/par_vector.c
oseikuffuor1: @victorapm is base_j supposed to be BigInt then? Index j only goes from 0 to local_size, so it seems not a BigInt to me. base_j is passed as argument as HYPRE_Int.

src/parcsr_mv/par_vector.c
oseikuffuor1: Ok seems reasonable to me then. Is this the only place the "old" form was being used? Otherwise will other instances elsewhere need to be changed too?

victorapm: Yes, let me update this, thanks for catching!

src/parcsr_mv/par_vector.c
victorapm: Yes, let me update this, thanks for catching!

src/parcsr_mv/par_vector.c
victorapm: It's the only place, thanks for checking!

src/parcsr_mv/par_vector.c
victorapm: Done in [200bbf9](https://github.com/hypre-space/hypre/pull/1033/commits/200bbf9c7d5810878d8c884e1be832e806714cf5)

oseikuffuor1 (approved):
Thanks Victor!!

ulrikeyang (approved):
This looks fine. 
Should we extend the ReadIJ function to be able to read vectors that have been printed with this extended function?

victorapm:
Yes, I have a separate PR for that :)

PR #1034, opened 2023-12-13
Add `hypre_ParCSRMatrixBlockColSum`

victorapm:
Add a `ParCSRMatrix` function for computing its block column sum.

This function will be used by MGR (GEOS).

Depends on #1022


src/parcsr_mv/par_csr_matop.c
```diff
@@ -6218,26 +6219,20 @@ hypre_ParCSRMatrixDropSmallEntries( hypre_ParCSRMatrix *A,
       return hypre_error_flag;
    }
 
-   hypre_GpuProfilingPushRange("ParCSRMatrixDropSmallEntries");
```
oseikuffuor1, line 6221: @victorapm why is this no longer needed? Was it a bug to be here and not in device code? Similar question in the DiagScaleVector function.

oseikuffuor1 (approved):
@victorapm I left you one comment. Otherwise this looks good to me. Thanks.

src/parcsr_mv/par_csr_matop.c
victorapm: It's already inside the `Device` function and no longer needed here

victorapm:
Hi all, I updated the new function in this PR to use `hypre_DenseBlockMatrix` type as output. This change is more appropriate for the purpose of the new function (to create new block-interpolation options for MGR).

ulrikeyang (approved):
This looks fine.

PR #1035, opened 2023-12-18
Refactor MGR interpolation

victorapm:
Move MGR interpolation functions to a new and dedicated file.


src/parcsr_ls/par_mgr_setup.c
```diff
@@ -1180,19 +1180,17 @@ hypre_MGRSetup( void               *mgr_vdata,
             hypre_MGRBuildBlockJacobiWp(A_FF, A_FC, block_jacobi_bsize,
                                         CF_marker, coarse_pnts_global, &Wp);
          }
-         hypre_MGRBuildInterp(A_array[lev], A_FF, A_FC, CF_marker, Wp,
-                              coarse_pnts_global, 1, dof_func_buff_data,
-                              debug_flag, trunc_factor, P_max_elmts[lev],
-                              block_jacobi_bsize, &P, interp_type[lev],
-                              num_interp_sweeps);
+         hypre_MGRBuildProlongation(A_array[lev], A_FF, A_FC, CF_marker, Wp,
```
rfalgout: Why the name change?

rfalgout (approved):
Looks good modulo the name change from Interp to Prolongation.  I don't understand the reason for it, and it seems to make the code less consistent wrt naming.

src/parcsr_ls/par_mgr_setup.c
victorapm: I think "Prolongation" is more specific (we are building an operator that goes from coarse to fine), even though the two terms are used quite interchangeably in multigrid methods.

If we keep the new name, I agree that some other variable names should change as well, e.g., `interp_type` -> `prol_type`.

Let me know what to do!

victorapm: PS: I'm putting both "Restriction" and "Prolongation" routines in the same file suffixed with "interp". So, I wanted to have a BuildProlongation and a BuildRestriction serving as entry points to all possible MGR "interpolations"

rfalgout: I don't think we should use "prolongation".  We nearly uniformly use "interpolation" in hypre (with just a few exceptions that probably could be fixed), so I would prefer to be consistent.  The two names are interchangeably in the multigrid community, even though "prolongation" is a more general term in a broader context.  I don't see a problem with putting restriction in the "interp" file.

victorapm: Ok, thanks! I'll revert this back

victorapm: Done in [72a07a8](https://github.com/hypre-space/hypre/pull/1035/commits/72a07a8fbd08cc975384a70440334acc53b398aa)

oseikuffuor1 (approved):
Nice and clean. Thanks Victor!

PR #1037, opened 2023-12-19
Add MGR restriction helper

victorapm:
Add `hypre_MGRBuildRFromW` for building classical restriction operators `R = [-Wr  I]` from an input matrix `Wr`

Depends on #1035


rfalgout (approved):

src/parcsr_ls/par_mgr_interp.c
```diff
@@ -2375,3 +2375,139 @@ hypre_MGRTruncateAcfCPR(hypre_ParCSRMatrix  *A_CF,
...
+
+#if defined (HYPRE_USING_GPU)
+   HYPRE_ExecutionPolicy exec = hypre_GetExecPolicy1(memory_location_W);
+
+   if (exec == HYPRE_EXEC_DEVICE)
+   {
+      /* TODO (VPM): Implement hypre_MGRBuildRFromWDevice */
+      hypre_ParCSRMatrixMigrate(W, HYPRE_MEMORY_HOST);
+      hypre_ParCSRMatrixMigrate(R, HYPRE_MEMORY_HOST);
+      hypre_MGRBuildRFromWHost(C_map, F_map, W, R);
```
oseikuffuor1, line 2496: Do you think we need a warning here? Not sure what we do in similar cases there there's no device option yet and we revert to running on host?

oseikuffuor1 (approved):
Looks good @victorapm . Just left a minor comment/ question for your consideration.

src/parcsr_ls/par_mgr_interp.c
victorapm: I think it isn't necessary as the code still works despite being less efficient than running fully on the device. hypre warnings are mostly for situations where the code doesn't work as they generate non-zero error codes. 

Also, I'm planning to implement `hypre_MGRBuildRFromWDevice` in the next couple of months, so this is temporary

oseikuffuor1: Ok sounds good.

PR #1040, opened 2024-01-02
Improve MGR robustness

victorapm:
Improve MGR robustness when Gaussian Elimination is used as F-relaxation.

A new logic has been added: If GE fails, then fall back to Jacobi.

In addition, this PR improves MGR's memory usage by reducing the size of certain allocations.

### DIFF
--- a/src/parcsr_ls/par_mgr_device.c
+++ b/src/parcsr_ls/par_mgr_device.c
@@ -579,7 +579,7 @@ hypre_ParCSRMatrixExtractBlockDiagDevice( hypre_ParCSRMatrix   *A,
 #else
    HYPRE_Int            *pivots;
    HYPRE_Complex       **tmpdiag_aop;
-   HYPRE_Int            *infos;
+   HYPRE_Int            *info;
 #endif
    HYPRE_Int            *blk_row_indices;
    HYPRE_Complex        *tmpdiag;
@@ -592,7 +592,7 @@ hypre_ParCSRMatrixExtractBlockDiagDevice( hypre_ParCSRMatrix   *A,
 
    /* Additional variables for debugging */
 #if HYPRE_DEBUG
-   HYPRE_Int            *h_infos;
+   HYPRE_Int            *h_info;
    HYPRE_Int             k, myid;
 
    hypre_MPI_Comm_rank(hypre_ParCSRMatrixComm(A), &myid);
@@ -664,7 +664,7 @@ hypre_ParCSRMatrixExtractBlockDiagDevice( hypre_ParCSRMatrix   *A,
     *-----------------------------------------------------------------*/
    {
       dim3 bDim = hypre_GetDefaultDeviceBlockDimension();
-      dim3 gDim = hypre_GetDefaultDeviceGridDimension(num_rows / blk_size, "warp", bDim);
+      dim3 gDim = hypre_GetDefaultDeviceGridDimension(num_blocks, "warp", bDim);
 
       if (CF_marker)
       {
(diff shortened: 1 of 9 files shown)


src/utilities/error.c
```diff
@@ -85,6 +85,66 @@ hypre_error_handler(const char *filename, HYPRE_Int line, HYPRE_Int ierr, const
    hypre__global_error = err;
 }
 
+/*--------------------------------------------------------------------------
+ *--------------------------------------------------------------------------*/
+
+void
+hypre_error_handler_rank_0(HYPRE_Int   myid,
```
rfalgout: Remove this function.  I don't understand its purpose and you aren't using it anyway.

src/utilities/error.h
```diff
@@ -26,16 +27,23 @@ typedef struct
...
 extern hypre_Error hypre__global_error;
 #define hypre_error_flag  hypre__global_error.error_flag
+#define hypre_error_temp_flag  hypre__global_error.temp_error_flag
 
 /*--------------------------------------------------------------------------
  * HYPRE error macros
  *--------------------------------------------------------------------------*/
 
 void hypre_error_handler(const char *filename, HYPRE_Int line, HYPRE_Int ierr, const char *msg);
+void hypre_error_handler_rank_0(HYPRE_Int myid, const char *filename, HYPRE_Int line,
```
rfalgout: Remove this and the '_0' functions below.

src/utilities/error.c
victorapm: Done in [8ed50af](https://github.com/hypre-space/hypre/pull/1040/commits/8ed50af2b09c7125308badda38a3f57b4c1171ba)

src/utilities/error.h
victorapm: Done in [8ed50af](https://github.com/hypre-space/hypre/pull/1040/commits/8ed50af2b09c7125308badda38a3f57b4c1171ba)

src/utilities/error.c
```diff
@@ -85,6 +85,66 @@ hypre_error_handler(const char *filename, HYPRE_Int line, HYPRE_Int ierr, const
...
+
+   /* Reset temporary error code */
+   hypre_error_temp_flag = 0;
+}
+
+/*--------------------------------------------------------------------------
+ *--------------------------------------------------------------------------*/
+
+HYPRE_Int
+HYPRE_GetMaxGlobalError(MPI_Comm comm)
```
rfalgout: What we really should do here is provide a "union" and just call it `GetGlobalError`.  Using max can miss errors in "smaller bits".

victorapm: Right! How about using MPI_BOR (bitwise OR) instead of MPI_MAX?

rfalgout (approved):
This looks good to me, modulo the comment about MaxError.  Thanks!

src/utilities/error.c
rfalgout: Perfect.  Thanks!

victorapm: As a side question, since we are dealing with this...

Should we allow negative error codes? 

If not, then I propose changing `HYPRE_Int error_flag` to `hypre_uint error_flag` in the `hypre_Error` struct

victorapm: Bringing this up also because bitwise OR on signed data types might produce unexpected results...

rfalgout: This might be something to do, but let's save it for another PR.  Functionally in hypre, we only ever have non-negative error codes, so bitwise OR shouldn't be a problem.

victorapm: Done in [80e65b4](https://github.com/hypre-space/hypre/pull/1040/commits/80e65b40b05de97b1f44871491092298b4b4f03a)

src/parcsr_ls/par_mgr_setup.c
```diff
@@ -1830,16 +1830,15 @@ hypre_MGRSetup( void               *mgr_vdata,
       TODO (VPM): move this block closer to global smoother setup */
    for (j = 0; j < num_c_levels; j++)
    {
-      if (((mgr_data -> num_relax_sweeps)[j] > 0) && (l1_norms[j] == NULL))
+      if (!l1_norms[j])
```
oseikuffuor1, line 1833: Why is the check for num_relax_sweeps > 0 removed?

src/parcsr_ls/par_mgr_device.c
```diff
@@ -664,7 +664,7 @@ hypre_ParCSRMatrixExtractBlockDiagDevice( hypre_ParCSRMatrix   *A,
     *-----------------------------------------------------------------*/
    {
       dim3 bDim = hypre_GetDefaultDeviceBlockDimension();
-      dim3 gDim = hypre_GetDefaultDeviceGridDimension(num_rows / blk_size, "warp", bDim);
+      dim3 gDim = hypre_GetDefaultDeviceGridDimension(num_blocks, "warp", bDim);
```
oseikuffuor1, line 667: Is num_blocks = num_rows / blk_size? Or has the argument here changed? num_blocks is set to num_points above.

src/parcsr_ls/par_mgr_device.c
```diff
@@ -692,27 +692,27 @@ hypre_ParCSRMatrixExtractBlockDiagDevice( hypre_ParCSRMatrix   *A,
 
       /* Memory allocation */
       tmpdiag     = hypre_TAlloc(HYPRE_Complex, bdiag_size, HYPRE_MEMORY_DEVICE);
-      diag_aop    = hypre_TAlloc(HYPRE_Complex *, num_rows, HYPRE_MEMORY_DEVICE);
+      diag_aop    = hypre_TAlloc(HYPRE_Complex *, num_blocks, HYPRE_MEMORY_DEVICE);
```
oseikuffuor1, line 695: num_blocks is used to replace num_rows in various places. Was this a bug before, or were we allocating and using more memory than was needed?

oseikuffuor1 (commented):
Hi Victor, I made a few comments/ questions for your review. Thanks.

src/parcsr_ls/par_mgr_device.c
victorapm: I was allocating more info than needed

src/parcsr_ls/par_mgr_setup.c
victorapm: Because we need to compute l1_norms since global relaxation might need it. 

To be more precise, there are situations where l1_norms is not required. So, I could have been more strict here by adding checks instead of removing the one we have.

I plan to improve this in a future PR

src/parcsr_ls/par_mgr_device.c
victorapm: This was a bug :)

Correct is `num_blocks`

victorapm:
Thanks for reviewing! Answered!

oseikuffuor1:
@victorapm thanks. Will these changes work for the CPU case as well? In particular the reversion to use Jacobi if GE fails?

victorapm:
Yes, this works for CPU as well

oseikuffuor1 (approved):
Thanks Victor!

PR #1043, opened 2024-01-04
New print level codes in MGR

victorapm:
This PR adds two print level codes for MGR intended for developers:

- Code 64: Print the coarsest level matrix to NP files.
- Code 128: Print the full MGR hierarchy (operator, interpolation, and restriction).

This PR also:
- Cleans MGR's setup implementation by removing old debugging macros
- Completes the work started with #976


rfalgout (approved):
Looks good to me!

PR #1045, opened 2024-01-05
Add `hypre_DenseBlockMatrixMultiply`

victorapm:
Add function for multiplying two dense block matrices.

To be used in MGR.

PR #1044, opened 2024-01-05
Bug fix in MGR

victorapm:
Fix RAP computation for options 2 and 3. The logic for each option was flipped.

PR #1036, opened 2023-12-19
New IntArray capabilities

victorapm:
This PR adds:
- `hypre_IntArraySeparateByValue`: function for creating new arrays after segregating indices of an input array by value. (to be used in MGR)
- New `hypre_IntArrayArray` type to denote an array of `hypre_IntArray`, just like `hypre_BoxArrayArray` does for `hypre_BoxArray`.

PR #1046, opened 2024-01-05
New MGR restriction based on column sum

victorapm:
This PR adds a new option for defining MGR's restriction via (block) column sums.

Depends on

- #1036 
- #1040
- #1037 
- #1035 
- #1044 
- #1045 
- #1048


rfalgout (approved):
This looks good to me.  It looks like some of the changes are replicated from other outstanding PRs, correct?  Thanks!

victorapm:
Yes, some changes are new from the point of view of master because the other PRs haven't been merged yet. Thanks!

src/parcsr_mv/par_csr_matop.c
```diff
@@ -6695,39 +6695,39 @@ hypre_ParCSRMatrixBlockColSumHost( hypre_ParCSRMatrix     *A,
...
+   HYPRE_Int              *send_map_elmts;
+   HYPRE_Int              *send_map_starts;
 #if defined(HYPRE_USING_PERSISTENT_COMM)
    hypre_ParCSRPersistentCommHandle *comm_handle;
 #else
    hypre_ParCSRCommHandle           *comm_handle;
 #endif
 
    /* Update commpkg offsets */
-   hypre_ParCSRCommPkgUpdateVecStarts(comm_pkg, num_cols_block_B, 1, num_cols_block_B);
```
oseikuffuor1, line 6707: Why is this no longer needed?

src/parcsr_mv/par_csr_matop.c
```diff
@@ -6695,39 +6695,39 @@ hypre_ParCSRMatrixBlockColSumHost( hypre_ParCSRMatrix     *A,
...
+   send_map_elmts  = hypre_ParCSRCommPkgSendMapElmts(comm_pkg);
+   send_map_starts = hypre_ParCSRCommPkgSendMapStarts(comm_pkg);
 
    /* Allocate the recv and send buffers  */
 #if defined(HYPRE_USING_PERSISTENT_COMM)
    comm_handle = hypre_ParCSRCommPkgGetPersistentCommHandle(HYPRE_COMM_PKG_JOB_COMPLEX, comm_pkg);
    recv_data = (HYPRE_Complex *) hypre_ParCSRCommHandleRecvDataBuffer(comm_handle);
    send_data = (HYPRE_Complex *) hypre_ParCSRCommHandleSendDataBuffer(comm_handle);
    send_data = hypre_Memset((void *) send_data, 0,
-                            (size_t) (num_cols_offd_A * num_cols_block_B) * sizeof(HYPRE_Complex),
```
oseikuffuor1, line 6715: Were these allocations a bug before? Looks like the new allocation is smaller?

src/seq_mv/vector.c
```diff
@@ -900,15 +900,9 @@ hypre_SeqVectorElmdivpyMarked( hypre_Vector *x,
...
 
    /* Sanity checks */
-   if (hypre_VectorSize(y) != hypre_VectorSize(b))
+   if (hypre_VectorSize(x) != hypre_VectorSize(b))
    {
-      hypre_error_w_msg(HYPRE_ERROR_GENERIC, "sizes of y and b do not match!\n");
-      return hypre_error_flag;
-   }
-
-   if (hypre_VectorSize(x) < hypre_VectorSize(y))
```
oseikuffuor1, line 970: Why don't we no longer check for the sizes of x and y?

src/parcsr_ls/par_mgr.c
```diff
@@ -1042,216 +1046,190 @@ hypre_ParCSRMatrixLeftScale(HYPRE_Real *vector,
...
+ * Methods 1-4 assume that restriction is the injection operator.
+ * Method 5 assumes that interpolation is the injection operator.
  *
  * TODO (VPM): Can we have a single function that works for host and device?
- *             RT is not being used.
  *--------------------------------------------------------------------------*/
 
 HYPRE_Int
-hypre_MGRComputeNonGalerkinCoarseGrid(hypre_ParCSRMatrix    *A,
+hypre_MGRComputeNonGalerkinCoarseGrid(hypre_ParCSRMatrix    *A_FF,
```
oseikuffuor1, line 1054: Why is the function signature changed? Is there a reason one would want to pass in the A_{XX} matrices instead of what was done previously?

src/parcsr_ls/par_mgr_interp.c
```diff
@@ -2096,167 +2121,6 @@ hypre_MGRBuildInterpApproximateInverse(hypre_ParCSRMatrix   *A,
    return hypre_error_flag;
 }
 
-/*--------------------------------------------------------------------------
- * hypre_MGRGetAcfCPR
- *--------------------------------------------------------------------------*/
-
-HYPRE_Int
-hypre_MGRGetAcfCPR(hypre_ParCSRMatrix     *A,
```
oseikuffuor1, line 2104: Has this been consolidated into the MGRBuildR function?

src/parcsr_ls/par_mgr_interp.c
```diff
@@ -170,109 +187,113 @@ hypre_MGRBuildRestrict( hypre_ParCSRMatrix    *A,
...
+      hypre_MGRBuildInterpApproximateInverse(AT, CF_marker_data, num_cpts_global, &RT);
+      hypre_BoomerAMGInterpTruncation(RT, trunc_factor, max_elmts);
    }
    else if (restrict_type == 12)
    {
-      hypre_MGRBuildPBlockJacobi(AT, A_FFT, A_FCT, NULL, blk_size, CF_marker, num_cpts_global,
-                                 &R);
+      hypre_MGRBuildPBlockJacobi(AT, A_FFT, A_FCT, NULL, blk_size, CF_marker_data, &RT);
    }
    else if (restrict_type == 13) // CPR-like restriction operator
```
oseikuffuor1, line 222: Does this new implementation require A_CF to be present always? Since it is only needed for this restrict_type, it may be NULL, right? If so, should we check here to make sure it is not NULL?

src/parcsr_ls/par_mgr_setup.c
```diff
@@ -1108,19 +1119,14 @@ hypre_MGRSetup( void               *mgr_vdata,
...
-          (Frelax_type[lev] == 99)  ||
-          (Frelax_type[lev] == 199) ||
-          (interp_type[lev] == 12)  ||
-          (mgr_coarse_grid_method[lev] != 0))
-#endif
-      {
-         hypre_ParCSRMatrixGenerateFFFC(A_array[lev], CF_marker, coarse_pnts_global,
-                                        NULL, &A_FC, &A_FF);
-      }
+      /* Compute C/F splitting (needed by RAP computation and other operations) */
```
oseikuffuor1, line 1122: It looks like we are doing to the work of generating A_CF even if it might not be needed?

src/parcsr_ls/par_mgr_setup.c
```diff
@@ -1351,23 +1349,40 @@ hypre_MGRSetup( void               *mgr_vdata,
...
             hypre_GpuProfilingPushRange(region_name);
             HYPRE_ANNOTATE_REGION_BEGIN("%s", region_name);
-            RAP_ptr = hypre_ParCSRMatrixRAPKT(RT, A_array[lev], P, 1);
+            if (RT)
+            {
+               RAP_ptr = hypre_ParCSRMatrixRAPKT(RT, A_array[lev], P, 1);
+            }
+            else if (R)
+            {
+               AP      = hypre_ParCSRMatMat(A_array[lev], P);
```
oseikuffuor1, line 1370: I believe the triple matrix product is faster (not sure if only on GPU). Could this be made to use that?

oseikuffuor1 (approved):
Hi @victorapm overall, it looks good to me. Quite a few function signatures have changed so I left some comments for you. Please take a look to make sure these changes are for improved refactoring and not a change to just support currently added functionalities. Otherwise I fear we may constantly change them as new strategies are implemented. Otherwise I am happy with it. Thanks :-)

src/parcsr_mv/par_csr_matop.c
victorapm: I simplified the previous implementation of this function and this line became:
`hypre_ParCSRCommPkgUpdateVecStarts(comm_pkg, 1, 0, 1);`

src/parcsr_mv/par_csr_matop.c
victorapm: Not a bug, but as for my previous comment, this change was due to a simplification of the previous implementation of this function

src/seq_mv/vector.c
victorapm: This check (introduced by me last year) was too strict. The function still works if y_size > x_size. So I decided to remove the check

src/parcsr_ls/par_mgr.c
victorapm: The signature has changed for performance purposes, i.e., the A_{XX} matrices were being computed inside this function (on master). However, when calling this function, we already know some of them since they are used to build restriction/interpolation. Therefore, we can just pass that information in and we don't need to recompute it inside `hypre_MGRComputeNonGalerkinCoarseGrid`

src/parcsr_ls/par_mgr_interp.c
victorapm: This function does two things:
1. Computes A_CF
2. Truncates A_CF to block (rectangular) diagonal.

We already have other code that does 1 and 2 separately. In addition, when calling this function in MGR, we already have knowledge about A_CF (it's formed right at the beginning of each level)

So, I removed this function and replaced it with a call to `hypre_MGRTruncateAcfCPR`, which does 2.

The final code is better performant (we do less extra computation) and yields the same result

src/parcsr_ls/par_mgr_interp.c
victorapm: Yes, good idea, I'll add the check

src/parcsr_ls/par_mgr_setup.c
victorapm: Correct. This also happens on master as you had already noted in the comment below: 
https://github.com/hypre-space/hypre/blame/6fc944427268ad18b4e3cd29ab2fc7f1b0854658/src/parcsr_ls/par_mgr.c#L1118-L1124

The issue is that `hypre_ParCSRMatrixGenerateFFFC` always generates both matrices and we always need A_CC. So, A_CF is a byproduct not always used.

There's room for improvement in a separate PR

liruipeng: you may give NULL pointers to what you don't need. need to double check, not 100% sure.

src/parcsr_ls/par_mgr_setup.c
victorapm: Maybe yes. @liruipeng which is preferable? Two MatMats or a single call to `hypre_ParCSRMatrixRAPKT`?

src/parcsr_ls/par_mgr_setup.c
victorapm: It's not possible:

https://github.com/hypre-space/hypre/blob/6fc944427268ad18b4e3cd29ab2fc7f1b0854658/src/parcsr_mv/gen_fffc.c#L445

https://github.com/hypre-space/hypre/blob/6fc944427268ad18b4e3cd29ab2fc7f1b0854658/src/parcsr_mv/gen_fffc.c#L484

liruipeng: OK. The device version can do this.

src/parcsr_ls/par_mgr_setup.c
liruipeng: It depends. It may be useful to leave an option for this.

victorapm:
> Please take a look to make sure these changes are for improved refactoring and not a change to just support currently added functionalities

It's mostly the latter case unfortunately

> Otherwise I fear we may constantly change them as new strategies are implemented

True, but with the current structure, I don't see how to get away with this.

One idea is to pass MGR's data object around instead of all the variables you might need. So, for example:

Instead of:
```
hypre_MGRBuildInterp(hypre_ParCSRMatrix   *A,
                     hypre_ParCSRMatrix   *A_FF,
                     hypre_ParCSRMatrix   *A_FC,
                     HYPRE_Int            *CF_marker,
                     hypre_ParCSRMatrix   *aux_mat,
                     HYPRE_BigInt         *num_cpts_global,
                     HYPRE_Real            trunc_factor,
                     HYPRE_Int             max_elmts,
                     HYPRE_Int             blk_size,
                     hypre_ParCSRMatrix  **P_ptr,
                     HYPRE_Int             interp_type,
                     HYPRE_Int             num_sweeps_post)
```
We could have something like:
```
hypre_MGRBuildInterp(hypre_MGRData       *mgr_data,
                     HYPRE_Int.           level,
                     hypre_ParCSRMatrix  **P_ptr)
```

Let me know your thoughts @oseikuffuor1 @rfalgout 

This would be for a separate PR...

src/parcsr_ls/par_mgr_setup.c
victorapm: Good idea, when is it better to use one or the other?

What should be the default?

liruipeng: can you use the AMG parameter, rap2, here?

victorapm: We would need to implement it for MGR as we don't have. But yes, it's possible

liruipeng: I suggest you have an option that mirrors rap2 in the AMG parameters.

victorapm: Yes, that's what I'm thinking

when is it better to use one or the other?

What should be the default?

liruipeng: rap2 = 0 is the default for AMG.

victorapm: You said it depends above. In which situation would it be better to do two MatMats (rap2 = 1) instead of the the triple-mat (rap = 0)?

liruipeng: It is hard to say which situation. I don't have an exact answer for this question. We have rap = 0 as default for AMG based on results on Laplacians.

PR #1048, opened 2024-01-08
Add `hypre_ceildiv`

victorapm:
Add a macro for computing division where the result is always rounded up to the nearest integer.

This fixes a bug in master for the case when the dividend is zero.

### DIFF
--- a/src/parcsr_ls/par_mgr.c
+++ b/src/parcsr_ls/par_mgr.c
@@ -3108,8 +3108,7 @@ hypre_MGRComputeNonGalerkinCoarseGrid(hypre_ParCSRMatrix    *A,
       }
       else
       {
-         /* TODO (VPM): Shouldn't blk_inv_size = bsize for method == 3? Check with DOK */
-         HYPRE_Int blk_inv_size = method == 2 ? bsize : 1;
+         HYPRE_Int blk_inv_size = method == 2 ? 1 : bsize;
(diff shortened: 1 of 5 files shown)


rfalgout:
What are your assumptions about the arguments for this macro?  Integer?  Non-negative?

victorapm:
I assume non-negative dividend and positive divisor. It works for integers (non-negative) and floating point numbers

victorapm:
Hi Rob, do the assumptions above sounds good? Another idea would be creating a `static inline` function

rfalgout:
The assumptions sound good, but you may want to add a comment in front of the macro that state them.  Also, this will only do a 'ceil' if the result is cast to an integer (either implicitly or explicitly).  Maybe you should do the cast in the macro (to HYPRE_Int, for example).

rfalgout (approved):

victorapm:
Thanks, I'll add the comment about assumptions

It's better to do the cast outside as this might be used for BigInts as well.

Thanks!

rfalgout:
Yes, I thought about the BigInt question, and that's a good point.  The reason I brought it up was because of the macro name.  You might also want to say something about casting in the comment.  If all the numbers are real, and the result is real, then a ceil isn't done.

victorapm:
Yes, good points! I'll add something about casting to the comment. Thanks again!

PR #1054, opened 2024-01-17
Add strided copy function for ParVectors

victorapm:
Add `hypre_ParVectorStridedCopy` for performing a strided copy operation from a source data array to the data array of a hypre_Vector structure (x->data). 

The source and destination arrays can be accessed with different strides. 

A strided copy means that elements are copied from the source to the destination at regular stride intervals.

To be used in MGR

### DIFF
--- a/src/seq_mv/vector_device.c
+++ b/src/seq_mv/vector_device.c
@@ -345,18 +345,22 @@ hypre_SeqVectorSumEltsDevice( hypre_Vector *vector )
 
 HYPRE_Int
 hypre_SeqVectorStridedCopyDevice( hypre_Vector  *vector,
-                                  HYPRE_Int      stride,
+                                  HYPRE_Int      istride,
+                                  HYPRE_Int      ostride,
                                   HYPRE_Int      size,
                                   HYPRE_Complex *data)
 {
    HYPRE_Complex  *v_data = hypre_VectorData(vector);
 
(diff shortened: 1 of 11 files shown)


src/seq_mv/vector_device.c
```diff
@@ -339,6 +339,32 @@ hypre_SeqVectorSumEltsDevice( hypre_Vector *vector )
...
+                                  HYPRE_Int      stride,
+                                  HYPRE_Int      size,
+                                  HYPRE_Complex *data)
+{
+   HYPRE_Complex  *v_data = hypre_VectorData(vector);
+
+#if defined(HYPRE_USING_CUDA) || defined(HYPRE_USING_HIP)
+   auto counting_iter = thrust::make_counting_iterator(0);
+
+   HYPRE_THRUST_CALL( transform_if, counting_iter, counting_iter + size, v_data, v_data,
```
liruipeng: This call has the complexity of O(size). There should be a more efficient implementation of O(size/stride), I assume.

victorapm: Yes, done in https://github.com/hypre-space/hypre/pull/1054/commits/b507d9cc0e54939e4104772eae57ff1225bd23d8

I've also generalized the new code to accept different strides for the source and destination arrays

liruipeng (approved):
Looks good to me. Thanks!

rfalgout (approved):
Looks good to me too

PR #1056, opened 2024-01-23
add the option of IJ base

liruipeng:
This PR fixes the issue of the IJ interface on GPU when index base $\neq 0$

- [x] tux
- [x] lassen
- [x] tioga
- [x] sunspot


rfalgout:
@liruipeng .  This looks good to me so far.  Let me know when you move it out of draft status.

One question/comment.  If the `ij_assembly.c` routine is useful for testing, should we add a test to the regression suite?  Same question for `ij_device.c`.

liruipeng:
> @liruipeng . This looks good to me so far. Let me know when you move it out of draft status.
> 
> One question/comment. If the `ij_assembly.c` routine is useful for testing, should we add a test to the regression suite? Same question for `ij_device.c`.

Thanks for the suggestion. Regression for `ij_assembly.c` was added. Not sure about `ij_device.c`, which was useful when we developed AMG on device but not so much right now. Maybe it can be removed.

rfalgout:
> Thanks for the suggestion. Regression for `ij_assembly.c` was added. Not sure about `ij_device.c`, which was useful when we developed AMG on device but not so much right now. Maybe it can be removed.

Awesome, thanks!  Why are some of the files empty, though (e.g., `fsai.jobs`)?

victorapm:
I think GitHub wanted to say that there aren't changes in any lines of the files. What has changed here are the file permissions, from 644 to 755

liruipeng:
@rfalgout LANL tested this fix and it worked for them. This PR should be good to go. Thanks!

rfalgout (approved):

PR #1060, opened 2024-01-25
Fix check-double autotest script

rfalgout:
This fixes a recent false positive in the check-double autotest script.  The updated script ignores '-' before and after the string "double", so it will not detect "long-double" in a comment, for example.  This should still find all valid instances of the `double` keyword.


victorapm (approved):
Thanks, Rob!

PR #1062, opened 2024-01-26
CMake: add CUDA dependencies for downstream packages

streeve:
Added `find_dependency` calls which mirror `find_package` in `cmake/HYPRE_SetupCUDAToolkit.cmake`. Without this the downstream app has to add the same dependency paths as the hypre build


victorapm (approved):
Thank you for your contribution!

rfalgout (approved):
Looks good to me.  Thanks!

Issue #960, opened 2023-08-29
Compiler warnings


prj-:
With `gcc` 12.3.0, I get multiple warnings such as:
```
arch-linux-c-opt/include/_hypre_utilities.h(2714): warning #3012-D: a volatile destination type for a compound assignment expression is deprecated
              move_bucket->hopInfo |= (1U << move_free_dist);
arch-linux-c-opt/include/_hypre_utilities.h:2742:91: warning: unused parameter ‘m’ [-Wunused-parameter]
 2742 | hypre_UnorderedBigIntMapFindCloserFreeBucket( hypre_UnorderedBigIntMap   *m,
      |                                                                 ~~~~~~~~~~~~              ^
```
Could you please fix this?

victorapm:
Hi Pierre, I'm working on a branch to get rid of those warnings: #896

prj-:
Hi Victor, any idea when this will land in `master`?

victorapm:
Hi Pierre, I have a few other tasks ahead of this. I would say about a month to get this into master. Would this update be critical for PETSc?

prj-:
No, it's only critical for lunatics like me who use all imaginable compiler warnings.

victorapm:
hahaha I understand you... I'll let you know once this is ready...

PR #1064, opened 2024-01-31
Thoroughly address #960

prj-:
See also: https://github.com/hypre-space/hypre/pull/896

### DIFF
--- a/src/utilities/_hypre_utilities.h
+++ b/src/utilities/_hypre_utilities.h
@@ -2969,8 +2969,8 @@ hypre_UnorderedIntMapFindCloserFreeBucket( hypre_UnorderedIntMap  *m,
             #pragma omp flush
 #endif
 
-            move_bucket->hopInfo |= (1U << move_free_dist);
-            move_bucket->hopInfo &= ~(1U << move_new_free_dist);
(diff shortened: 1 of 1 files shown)


prj-:
I don't know why I never took the time to fix this @victorapm, but here you go (see referenced MR and/or issue for the current warnings I get without this patch).

victorapm (approved):
Looks great, thanks @prj-

victorapm:
@rfalgout , Pierre fixed some compilation warnings like:
```
: warning #3012-D: a volatile destination type for a compound assignment expression is deprecated
              move_bucket->hopInfo &= ~(1U << move_new_free_dist);
```

Which happens when compiling with `-std=c++20`

rfalgout (approved):
Looks good.  Thanks!

liruipeng:
What exactly is the problem? Is it `|=`, `&=`, etc, or `startBucket->hopInfo`? Thanks

rfalgout:
> What exactly is the problem? Is it `|=`, `&=`, etc, or `startBucket->hopInfo`? Thanks

Here is something I found explaining the issue:

https://stackoverflow.com/questions/65528802/incrementing-a-volatile-in-c20-with-operations-on-volatile-being-deprecated

Issue #1067, opened 2024-02-13
New hypre release 2.31.0


rfalgout:
Hello @hypre-space/developers  ,

We are doing a new release, so please hold off on making changes to the master branch until it is done.

Thanks!

-Rob

rfalgout:
- [x] Create a PR with title "Release 2.31.0" on branch `release-2.31.0`.
  - [x] Create branch
  - [x] Update `CHANGELOG` file with information regarding contents of this release.
  - [x] Update `config/version.sh` with release number and date, then run `config/update.sh` and commit changes.
  - [x] Create PR, reference this issue in the description, and add reviewers
- [x] [MAJOR.MINOR] Update user and reference manuals.
- [x] Check regression tests.
  - [x] machine-tux
  - [x] machine-tux-spack (same errors as the recent tests on master)
  - [x] machine-mac
  - [x] machine-rztopaz
  - [x] machine-lassen (small benchmark difference in one run)
  - [x] machine-tioga
- [x] Squash merge the PR.
- [x] Tag repository with release number: `git tag -a v2.31.0 -m "Release v2.31.0"`
- [x] Push tags to central repository: `git push --tags`
- [x] Create a GitHub Release.
- [x] On the CASC cluster, copy the GitHub release tar file to `/usr/casc/hypre`.
- [x] On the RZ, copy the GitHub tar file to `/usr/gapps/hypre` (for backup reasons).
- [x] Update Spack Package and create Spack pull request.
- [x] Close the new release GitHub issue.

rfalgout:
Would someone mind doing the tioga runs (branch release-2.31.0)?

victorapm:
I can do it :)

rfalgout:
> I can do it :)

Awesome, thanks!

rfalgout:
As usual, rztopaz is taking forever.  I'll check it again tomorrow.  Otherwise, all of the tests are clean.

rfalgout:
Hi All.  The release is done.  Thanks!

PR #1068, opened 2024-02-13
Release 2.31.0

rfalgout:
See issue #1067 .


rfalgout:
@victorapm I am having issues with the `tail +7` command in `seq_block_mv/headers`.  The `_hypre*.h` files are supposed to basically behave as if we did `#include file.h` directly in the source, but we do it this way to minimize the number of header files to install.  Why do you not want to `cat` the entire two files?

victorapm:
Hi Rob, I do `tail +7` to exclude the headers (didn't want to include the copyright header multiple times in the `_hypre` file)

What was the issue exactly?

If it turns out time-consuming, I would say let's just do `cat`

rfalgout:
I get these error messages when I run `headers` on my tux box:
```
tail: cannot open ‘+7’ for reading: No such file or directory
tail: cannot open ‘+7’ for reading: No such file or directory
```
This came up because the `config/astyle-apply.sh` script also runs the `headers` scripts.  I noticed differences between my tux box and my mac (which doesn't complain).

victorapm (approved):
Thanks!

liruipeng (approved):

oseikuffuor1 (approved):
Thanks!

rfalgout:
@waynemitchell Does this release look good to you?

waynemitchell (approved):
Looks good to me. Thanks!

PR #1077, opened 2024-02-29
skip transpose test with OMP

liruipeng:
This PR removes the test of assembling $A^T$ using the IJ interface on CPUs with OpenMP, since the test violates an assumption made by the IJ interface.


rfalgout (approved):
Thanks, Rui Peng!

PR #1079, opened 2024-03-04
Asynchronous device to device hipMemcpy

tomstitt:
Switch `hipMemcpy` to `hipMemcpyAsync` for device to device copies to avoid unneeded host-side syncs; matches cuda's performant behavior.

See https://github.com/mfem/mfem/pull/2780


victorapm (approved):
Sorry for the delay. Thank you for your contribution!

rfalgout (approved):

PR #1081, opened 2024-03-11
Sycl cleanup

### LINKED ISSUE #1162: Building with SYCL Fails
Build with Intel OneAPI 2024.2 fails with the following error:

par_csr_filter_device.c:182:25: error: use of undeclared identifier 'hypre_popc'; did you mean 'hypre_Log2'?
  182 |          offd_offset += hypre_popc(ballot);
      |                         ^~~~~~~~~~
      |                         hypre_Log2

My configure command is:

./configure --with-sycl --enable-onemklsparse --enable-onemklblas --enable-onemklrand --enable-unified-memory

waynemitchell:
Fixes the (currently broken) SYCL build for master. Provides some minor updates and removes several workarounds that are no longer necessary for the SYCL back end:

1. oneDPL reverse iterator
2. oneDPL exclusive scan in place
3. dpct constant iterator
4. Replace generic_space with local_space when setting up atomics
5. Optimization of atomic function calls in spgemm
6. Some syntax/API updates for oneMKL
7. Add CPU launch option for SYCL for debugging purposes

This also removes some old notes and updates the regressions to run on Aurora with the latest default software environment there.


src/parcsr_mv/par_csr_filter_device.c
```diff
@@ -171,12 +172,20 @@ hypreGPUKernel_ParCSRMatrixBlkFilterFill(hypre_DeviceItem &item,
...
+#if defined(HYPRE_USING_SYCL)
+               /* WM: question - why is the atomic necessary here? */
+               col_map_marker[col] = 1;
+               /* auto atomic_key = sycl::atomic_ref < */
+               /*                   HYPRE_Int, sycl::memory_order::relaxed, */
+               /*                   sycl::memory_scope::device, */
+               /*                   sycl::access::address_space::local_space > (col_map_marker[col]); */
+               /* atomic_key.fetch_or(1); */
+#else
                atomicOr(col_map_marker + col, 1);
```
waynemitchell: @victorapm , I think this is code you recently added. I was wondering why you put an atomic operation here, since it seems like you are just writing 1 to the memory location when the if condition is satisfied. Even if you have multiple threads trying to write the same location at once, I don't see why the atomic is necessary?

victorapm: Thanks for catching, Wayne. You are right, we can remove `atomicOr` here

waynemitchell:
@rfalgout , @liruipeng , @victorapm , could y'all look through this PR when you get a chance? I'd like to get it merged soon since it fixes the currently broken SYCL build. Regressions are passing on Aurora, and I'm currently running lassen, tioga, and tux regressions.

src/test/ams_driver.c
```diff
@@ -163,6 +163,9 @@ main (hypre_int argc,
 #endif
    /* use vendor implementation for SpGEMM */
    HYPRE_SetSpGemmUseVendor(0);
+#if defined(HYPRE_USING_SYCL)
+   HYPRE_SetSpGemmUseVendor(1);
```
victorapm, line 167: Should we do this for the other drivers as well?

victorapm (approved):
Nice changes and thanks for the fixes

victorapm:
All, we might also want to merge #1141 soon since it fixes the build with cuda >= 12.5

rfalgout (approved):
Looks good.  Thanks!

liruipeng (approved):
Looks good. Thanks!

PR #1091, opened 2024-04-08
Add `hypre_MGRSetGlobalSmootherAtLevel` to MGR

victorapm:
Add `hypre_MGRSetGlobalSmootherAtLevel` which allows users to set "external" hypre solvers as a global relaxation method in MGR.

Tests:

- [x] tux
- [x] lassen
- [x] tioga


oseikuffuor1 (approved):
@victorapm this looks good to me, if it has been tested to work. It will be good to have a regression test to cover this case and make sure things are being destroyed correctly. I am fine with it if that is done in a separate PR. Thanks!!

rfalgout (approved):
Looks good.  Thanks!

src/parcsr_ls/HYPRE_parcsr_ls.h
```diff
@@ -4436,20 +4436,54 @@ HYPRE_MGRSetGlobalSmoothType( HYPRE_Solver solver,
...
- * @param solver [IN] MGR solver/preconditioner object
- * @param fsolver [IN] Global smoother solver object
- **/
+ * This function configures a global relaxation (smoother) method for a specific level within the
+ * multigrid reduction (MGR) scheme.
+ *
+ * @note Unlike some other setup functions that might require an array to set options across multiple
+ *       levels, this function focuses on a single level, identified by the \e level parameter.
+ *
+ * @warning If `hypre_MGRSetLevelSmoothType` has previously been called, setting a global smoother at
```
oseikuffuor1: Maybe we should rephrase this statement to say that this call takes precedence for setting the smoother at that level or overrides all other options for the smoother at that level (irrespective of when they are called).

src/parcsr_ls/HYPRE_parcsr_ls.h
```diff
@@ -4436,20 +4436,54 @@ HYPRE_MGRSetGlobalSmoothType( HYPRE_Solver solver,
...
+ *          a specific level using this function will override the smoother type for that level to an
+ *          inactive code, indicating the use of an user-defined smoother instead.
+ *
+ * @param[in] \e level The level identifier for which the global relaxation method is to be set.
+ *            Must be within the range of the configured levels in the MGR solver.
```
oseikuffuor1: change configured to "number of"

src/parcsr_ls/HYPRE_parcsr_ls.h
victorapm: Sounds good! Do you want to make the change?

src/parcsr_ls/HYPRE_parcsr_ls.h
victorapm: Fixed

src/parcsr_ls/par_mgr.c
```diff
@@ -3487,43 +3465,74 @@ hypre_MGRSetGlobalSmoothType( void *mgr_vdata, HYPRE_Int gsmooth_type )
...
-   HYPRE_Int *level_smooth_type = hypre_CTAlloc(HYPRE_Int, max_num_coarse_levels, HYPRE_MEMORY_HOST);
+   /* Set level_smooth_type array */
+   level_smooth_type = hypre_CTAlloc(HYPRE_Int, max_num_coarse_levels, HYPRE_MEMORY_HOST);
+   hypre_TFree((mgr_data -> level_smooth_type), HYPRE_MEMORY_HOST);
    if (gsmooth_type != NULL)
    {
       for (i = 0; i < max_num_coarse_levels; i++)
       {
-         level_smooth_type[i] = gsmooth_type[i];
+         /* For meaningful values of global smoothing type, the option set via
```
oseikuffuor1, line 3490: Should this check be made once in mgr_setup, since we would know then whether level_smoother and level_smooth_type are both set then set precedence and invalidate the level_smooth_type entry as appropriate? Perhaps before this line: https://github.com/hypre-space/hypre/blob/4f5d6ad50639d6ba8f1250d569221e1f13ebbe1e/src/parcsr_ls/par_mgr_setup.c#L958

src/parcsr_ls/par_mgr.c
```diff
@@ -3573,15 +3588,35 @@ hypre_MGRSetGlobalSmootherAtLevel( HYPRE_Int     level,
...
-      hypre_error_in_arg(2);
-      return hypre_error_flag;
+      (mgr_data -> level_smooth_type) = hypre_CTAlloc(HYPRE_Int,
+                                                      max_num_coarse_levels,
+                                                      HYPRE_MEMORY_HOST);
    }
 
    (mgr_data -> level_smoother)[level] = smoother;
 
+   /* Obtain corresponding smoother type */
```
oseikuffuor1, line 3601: See previous comment above about doing this check once. Level_smooth_type should be 'unset' for this level (and in the previous call) since the mgr_solve routine checks both level_smooth_type  (first) and then level_smoother to call the smoother. This smoother set here may not be called then.

src/parcsr_ls/par_mgr_solve.c
```diff
@@ -744,6 +744,16 @@ hypre_MGRCycle( void              *mgr_vdata,
                               U_array[fine_grid]);
                hypre_ParVectorAllZeros(U_array[fine_grid]) = 0;
             }
+            else if ((mgr_data -> level_smoother)[fine_grid])
```
oseikuffuor1, line 738: See comment above about possible issue if level_smooth_type is not 'unset' if user set specific smoother at this level.

src/parcsr_ls/par_mgr_solve.c
```diff
@@ -1214,6 +1224,16 @@ hypre_MGRCycle( void              *mgr_vdata,
                               A_array[fine_grid], F_array[fine_grid],
                               U_array[fine_grid]);
             }
+            else if ((mgr_data -> level_smoother)[fine_grid])
```
oseikuffuor1, line 1234: Similar comment as above

oseikuffuor1 (commented):
@victorapm, I made a few comments for your review. Otherwise, the overall idea looks good to me. Thanks.

src/parcsr_ls/par_mgr.c
victorapm: Resolved offline

src/parcsr_ls/par_mgr.c
victorapm: Resolved offline

src/parcsr_ls/par_mgr_solve.c
victorapm: Resolved offline

src/parcsr_ls/par_mgr_solve.c
victorapm: Resolved offline

src/parcsr_ls/HYPRE_parcsr_ls.h
victorapm: Done in [eeb6ce9](https://github.com/hypre-space/hypre/pull/1091/commits/eeb6ce9cee13d01e2800484601e6269f4ab848e5)

PR #1094, opened 2024-04-24
Lassen regression test update

liruipeng:
This PR updates the regression test on lassen because the default CUDA version has been upgraded.

- [x] lassen
- [x] tioga


src/test/TEST_bench/benchmark_spgemm.jobs
```diff
@@ -26,7 +26,7 @@ mpirun -np 1 ./ij_mm -n 1024 1024 1  -5pt -verify 1 -spgemm_binned 0 -spgemmalg
 mpirun -np 1 ./ij_mm -n 1024 1024 1  -9pt -verify 1 -spgemm_binned 0 -spgemmalg 3 -rep 10 > benchmark_spgemm.out.12
 
 mpirun -np 1 ./ij_mm -n 128 128 128  -7pt -verify 1 -vendor 1                     -rep 10 > benchmark_spgemm.out.13
-mpirun -np 1 ./ij_mm -n 128 128 128 -27pt -verify 1 -vendor 1                     -rep 10 > benchmark_spgemm.out.14
+mpirun -np 1 ./ij_mm -n 80   80  80 -27pt -verify 1 -vendor 1                     -rep 10 > benchmark_spgemm.out.14
```
ulrikeyang, line 29: Why did you change the size of the problem?

src/test/TEST_bench/benchmark_spgemm.perf.saved.lassen
```diff
@@ -23,23 +23,23 @@ Device Parcsr Matrix-by-Matrix wall clock time = 0.008427 seconds
 # Output file: benchmark_spgemm.out.12
 Device Parcsr Matrix-by-Matrix wall clock time = 0.011918 seconds
 # Output file: benchmark_spgemm.out.13
-Device Parcsr Matrix-by-Matrix wall clock time = 0.122758 seconds
+Device Parcsr Matrix-by-Matrix wall clock time = 0.027048 seconds
```
ulrikeyang, line 26: Wow. this is quite a bit faster.

src/test/TEST_bench/benchmark_spgemm.jobs
liruipeng: It was because cuda 11 now fails on the older size, out of memory :( . Note that this is with cusparse SpGEMM.

ulrikeyang (commented):
Looks good to me. Thanks

src/test/TEST_bench/benchmark_spgemm.perf.saved.lassen
liruipeng: ~~Oh. This is because the problem size is smaller.~~
Actually, cusparse version 11 is much faster for this case compared with cuda10.

ulrikeyang:
There were several faster ones. Was this all the same problem?

liruipeng:
> There were several faster ones. Was this all the same problem? 

Sorry I replied too fast. SpGEMM from cusparse version 11 is significantly faster than CUDA 10, as shown in many cases. But it seems to be more memory demanding, so there is one case we used before that failed due to OOM error. I changed the problem size to a smaller one, and it shows a more faster speed obviously.

rfalgout (approved):
Looks good.  Thanks!

Issue #1070, opened 2024-02-17
ParCSRAAt for nonsquare matrices


jesusbonilla:
I am trying to use ParCSRAAt with a rectangular matrix. Is there a reason for the following check or does it simply come from the standard matmul? Everything seems to be working on my end if I just remove it. Thanks!
https://github.com/hypre-space/hypre/blob/ee74c20e7a84e4e48eec142c6bb6ff2a75db72f1/src/parcsr_mv/par_csr_aat.c#L392

liruipeng:
Good question. I think it is fine to remove the check.

PR #1098, opened 2024-06-14
Bugs fixed related to transposed operations

jesusbonilla:
I noticed a few minor bugs related to transposed operations (one related to issue #1070 ). 

I have fixed all the ones I knew how to do it, but I believe there is still one more bug that I am no sure how to fix. In /src/parcsr_mv/communicationT.c col_starts is used as if it was an array of length num_procs+1, but this is no longer true. 
https://github.com/hypre-space/hypre/blob/ee74c20e7a84e4e48eec142c6bb6ff2a75db72f1/src/parcsr_mv/communicationT.c#L262

Is a possible fix the communication of column start positions and locally constructing a different col_starts_ array with all the column starts for all process, or is there a better alternative with communication involved?

### DIFF
--- a/src/parcsr_mv/communicationT.c
+++ b/src/parcsr_mv/communicationT.c
@@ -360,7 +360,7 @@ hypre_MatTCommPkgCreate_core (
                            }
                         }
             */
-            for ( kc = row_starts[my_id]; kc < row_starts[my_id + 1]; kc++ )
+            for ( kc = row_starts[0]; kc < row_starts[1]; kc++ )
             {
                if ( kc == col && i != my_id )
                {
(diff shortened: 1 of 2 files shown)


victorapm:
Hi @jesusbonilla, thank you for the proposed fix!

It seems that the routine `hypre_ParCSRAAt` and related ones such as `hypre_MatTCommPkgCreate` became quite outdated and additional fixes to them need to be implemented. I will bring this up in our next hypre team meeting and will update you after we make a plan.

Assuming that your goal is to multiply two rectangular matrices (are they tall-and-skinny?), I would suggest, for now, to compute the transpose explicitly and perform the multiplication:
```
hypre_ParCSRMatrixTranspose(A, At, 1);
AAt = hypre_ParCSRMatMat(A, At);
```
This has the advantage of being GPU-enabled wrt `hypre_ParCSRAAt`, but the disadvantage of using more memory.

jesusbonilla:
Hi @victorapm thanks for taking a look to it and for your suggestions! 

Yes, indeed the matrix is tall-and-skinny. I'll follow your suggestion of explicitly computing the transpose. For the moment, performance shouldn't be an issue, and I'll just update my code once `hypre_ParCSRAAt` is updated. Thanks!

PR #1099, opened 2024-06-17
IJ assembly memory

liruipeng:
This PR adds early assembly option to optimize the memory usage in the IJ interface to assemble matrices on GPUs.

User level APIs are
```
HYPRE_IJMatrixSetEarlyAssemble
HYPRE_SStructMatrixSetEarlyAssembly
```


src/sstruct_mv/HYPRE_sstruct_mv.h
```diff
@@ -511,6 +511,10 @@ HYPRE_SStructMatrixDestroy(HYPRE_SStructMatrix matrix);
 HYPRE_Int
 HYPRE_SStructMatrixInitialize(HYPRE_SStructMatrix matrix);
 
+HYPRE_Int
+HYPRE_SStructMatrixSetEarlyAssembly( HYPRE_SStructMatrix matrix,
```
rfalgout: Here you use the name "Assembly" but in IJ you use "Assemble".

liruipeng: Ah, they need to be consistent. Maybe both use "Assemble"?

rfalgout: Either is fine with me, but maybe "Assemble" is better because it connects more explicitly to the assemble function.

rfalgout (approved):
This looks good to me.  Thank!

src/IJ_mv/HYPRE_IJ_mv.h
```diff
@@ -344,6 +344,15 @@ HYPRE_Int HYPRE_IJMatrixSetDiagOffdSizes(HYPRE_IJMatrix   matrix,
 HYPRE_Int HYPRE_IJMatrixSetMaxOffProcElmts(HYPRE_IJMatrix matrix,
                                            HYPRE_Int      max_off_proc_elmts);
 
+HYPRE_Int HYPRE_IJMatrixSetInitAllocation(HYPRE_IJMatrix matrix,
```
victorapm, line 352: Should we document these functions in this PR?

src/IJ_mv/IJMatrix_parcsr.c
```diff
@@ -242,6 +242,90 @@ hypre_IJMatrixSetMaxOffProcElmtsParCSR(hypre_IJMatrix *matrix,
...
+
+   if (!aux_matrix)
+   {
+      HYPRE_Int local_num_rows = (HYPRE_Int)(row_partitioning[1] - row_partitioning[0]);
+      HYPRE_Int local_num_cols = (HYPRE_Int)(col_partitioning[1] - col_partitioning[0]);
+      hypre_AuxParCSRMatrixCreate(&aux_matrix, local_num_rows, local_num_cols, NULL);
+      hypre_IJMatrixTranslator(matrix) = aux_matrix;
+   }
+   hypre_AuxParCSRMatrixInitAllocFactor(aux_matrix) = factor;
+#endif
```
victorapm, line 271:
```suggestion
#else
   HYPRE_UNUSED_VAR(matrix);
   HYPRE_UNUSED_VAR(factor);
#endif
```

src/IJ_mv/IJMatrix_parcsr.c
```diff
@@ -242,6 +242,90 @@ hypre_IJMatrixSetMaxOffProcElmtsParCSR(hypre_IJMatrix *matrix,
...
+
+   if (!aux_matrix)
+   {
+      HYPRE_Int local_num_rows = (HYPRE_Int)(row_partitioning[1] - row_partitioning[0]);
+      HYPRE_Int local_num_cols = (HYPRE_Int)(col_partitioning[1] - col_partitioning[0]);
+      hypre_AuxParCSRMatrixCreate(&aux_matrix, local_num_rows, local_num_cols, NULL);
+      hypre_IJMatrixTranslator(matrix) = aux_matrix;
+   }
+   hypre_AuxParCSRMatrixEarlyAssemble(aux_matrix) = early_assemble;
+#endif
```
victorapm, line 302:
```suggestion
#else
   HYPRE_UNUSED_VAR(matrix);
   HYPRE_UNUSED_VAR(early_assemble);
#endif
```

src/IJ_mv/IJMatrix_parcsr.c
```diff
@@ -242,6 +242,90 @@ hypre_IJMatrixSetMaxOffProcElmtsParCSR(hypre_IJMatrix *matrix,
...
+
+   if (!aux_matrix)
+   {
+      HYPRE_Int local_num_rows = (HYPRE_Int)(row_partitioning[1] - row_partitioning[0]);
+      HYPRE_Int local_num_cols = (HYPRE_Int)(col_partitioning[1] - col_partitioning[0]);
+      hypre_AuxParCSRMatrixCreate(&aux_matrix, local_num_rows, local_num_cols, NULL);
+      hypre_IJMatrixTranslator(matrix) = aux_matrix;
+   }
+   hypre_AuxParCSRMatrixGrowFactor(aux_matrix) = factor;
+#endif
```
victorapm, line 333:
```suggestion
#else
   HYPRE_UNUSED_VAR(matrix);
   HYPRE_UNUSED_VAR(factor);
#endif
```

victorapm (approved):
Thanks, Rui Peng!

src/IJ_mv/IJMatrix_parcsr_device.c
```diff
@@ -108,32 +145,47 @@ hypre_IJMatrixSetAddValuesParCSRDevice( hypre_IJMatrix       *matrix,
...
-                                                                                  char, stack_elmts_max,          char, stack_elmts_max_new, HYPRE_MEMORY_DEVICE);
-      hypre_AuxParCSRMatrixMaxStackElmts(aux_matrix) = stack_elmts_max_new;
+      else
+      {
+         if (early_assemble)
+         {
+            stack_elmts_max_new = stack_elmts_required;
+            if (early_assemble == 1)
+            {
+               early_assemble_flag = 1;
```
waynemitchell: I don't quite follow everything here. Why do you have both `early_assemble` and `early_assemble_flag`? Also, why do you check `early_assemble` in the outer `if` statement and then check again if `early_assemble == 1` inside? Does `early_assemble` take other non-zero values besides 1? I had thought it was just a flag to turn early assemble on/off.

waynemitchell (approved):
I had one question above, but this all looks good. Also, I'll give a +1 to Victor's suggestion to document the new user-level functions. Thanks, Rui Peng!

src/IJ_mv/IJMatrix_parcsr_device.c
liruipeng: Hi @waynemitchell Thanks for catching this. In the old code before we fixed the code hang issue, we had a situation that `early_assemble = 2`. This was used when we still did a communication step in early assemble that calls `Addvalues` which possible triggers another early assemble, and another communication, and even possibly another early assembly and so on ... recursively. So we hacked it by setting it to `2` by temporally disable early assemble when communicating for other processes' entries.  Now that the whole communication step in early assemble is gone, the `=2` thing is no longer there which leaves the silly code in place
```
          if (early_assemble) // 0 or 1
          {
             ...
             if (early_assemble == 1)
```
I have updated the code slightly, removing the inner `if`. The logic should be simple and clear now: `early_assemble` is a global flag that turns on/off this feature and `early_assemble_flag` is locally per each call of `Addvalues`. When out of memory it triggers the assemble action at the end of `hypre_IJMatrixSetAddValuesParCSRDevice`.

Hope this explains. 

Thanks

PR #1101, opened 2024-06-25
Set HYPRE_HAVE_MPI_COMM_F2C in CMake build

Yannicked:
In the CMake build, the `HYPRE_HAVE_MPI_COMM_F2C` variable was not set, which causes invalid communicators in some MPI implementations when HYPRE was called from Fortran.

This pull request addresses the issue by using CMake's `check_symbol_exists` to determine if the function or macro is available, and sets the variable accordingly.


victorapm:
@Yannicked Thank you for your contribution!

I apologize for the delay. I'm changing the base to another development branch so I can make a few adjustments.

PR #1103, opened 2024-07-03
row_indexes for IJMatrixGetValues

waynemitchell:
Add option to pass row_indexes to the IJMatrixGetValues routines. This aligns with existing implementation for the SetValues routines and is necessary for upcoming functionality in the recmat branch.

### DIFF
--- a/src/IJ_mv/HYPRE_IJMatrix.c
+++ b/src/IJ_mv/HYPRE_IJMatrix.c
@@ -118,6 +118,59 @@ HYPRE_IJMatrixCreate( MPI_Comm        comm,
 }
 
 /*--------------------------------------------------------------------------
+ * HYPRE_IJMatrixPartialClone
+ *
+ * Creates a new IJMatrix with data copied from an existing matrix except
+ * for the members:
+ *    1) hypre_IJMatrixObject
+ *    2) hypre_IJMatrixTranslator
+ *    3) hypre_IJMatrixAssumedPart
+ *--------------------------------------------------------------------------*/
+
+HYPRE_Int
+HYPRE_IJMatrixPartialClone( HYPRE_IJMatrix  matrix_in,
+                            HYPRE_IJMatrix *matrix_out )
+{
+   hypre_IJMatrix *ijmatrix_in = (hypre_IJMatrix *) matrix_in;
+   hypre_IJMatrix *ijmatrix_out;
+   HYPRE_BigInt    ilower;
+   HYPRE_BigInt    iupper;
+   HYPRE_BigInt    jlower;
+   HYPRE_BigInt    jupper;
+
+   if (!ijmatrix_in)
+   {
+      hypre_error_in_arg(1);
+      return hypre_error_flag;
+   }
+
+   HYPRE_IJMatrixGetLocalRange(ijmatrix_in, &ilower, &iupper, &jlower, &jupper);
+
+   ijmatrix_out = hypre_CTAlloc(hypre_IJMatrix, 1, HYPRE_MEMORY_HOST);
+
+   hypre_IJMatrixComm(ijmatrix_out)               = hypre_IJMatrixComm(ijmatrix_in);
+   hypre_IJMatrixObject(ijmatrix_out)             = NULL;
+   hypre_IJMatrixTranslator(ijmatrix_out)         = NULL;
+   hypre_IJMatrixAssumedPart(ijmatrix_out)        = NULL;
+   hypre_IJMatrixObjectType(ijmatrix_out)         = hypre_IJMatrixObjectType(ijmatrix_in);
+   hypre_IJMatrixAssembleFlag(ijmatrix_out)       = 0;
+   hypre_IJMatrixPrintLevel(ijmatrix_out)         = hypre_IJMatrixPrintLevel(ijmatrix_in);
+   hypre_IJMatrixOMPFlag(ijmatrix_out)            = hypre_IJMatrixOMPFlag(ijmatrix_in);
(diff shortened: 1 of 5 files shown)


src/IJ_mv/HYPRE_IJMatrix.c
```diff
@@ -814,7 +867,133 @@ HYPRE_IJMatrixGetValues( HYPRE_IJMatrix matrix,
...
+      return hypre_error_flag;
+   }
+
+   if (!ijmatrix)
+   {
+      hypre_error_in_arg(1);
+      return hypre_error_flag;
+   }
+
+   if (!ncols)
```
victorapm, line 906: In `HYPRE_IJMatrixSetValues2` we allow this pointer to be NULL. I'm wondering if it should be the same here

src/IJ_mv/HYPRE_IJMatrix.c
```diff
@@ -814,7 +867,133 @@ HYPRE_IJMatrixGetValues( HYPRE_IJMatrix matrix,
...
```
victorapm, line 980: Same comment as for `HYPRE_IJMatrixGetValues2`

src/IJ_mv/HYPRE_IJ_mv.h
```diff
@@ -222,6 +222,37 @@ HYPRE_Int HYPRE_IJMatrixGetValues(HYPRE_IJMatrix  matrix,
...
+                                   HYPRE_Complex  *values);
+
+/**
+ * Gets values for \e nrows rows or partial rows of the matrix
+ * and zeros out those entries in the matrix.
+ *
+ * Same as IJMatrixGetValues2, but zeros out the entries after getting them.
+ *
+ **/
+HYPRE_Int HYPRE_IJMatrixGetValuesAndZero(HYPRE_IJMatrix  matrix,
```
victorapm: Minor point: should we call this `HYPRE_IJMatrixGetValuesAndZeroOut`?

src/IJ_mv/IJMatrix_parcsr.c
```diff
@@ -473,6 +475,7 @@ hypre_IJMatrixGetValuesParCSR( hypre_IJMatrix *matrix,
       col_map_offd = hypre_ParCSRMatrixColMapOffd(par_matrix);
    }
 
+   /* WM: question - what does rows < 0 signify? Do I need to add row_indexes here? */
```
victorapm: @ulrikeyang I'm curious about this as well. What is the use case for `nrows < 0` here?

src/IJ_mv/_hypre_IJ_mv.h
```diff
@@ -440,8 +440,8 @@ HYPRE_Int hypre_IJMatrixSetMaxOffProcElmtsParCSR ( hypre_IJMatrix *matrix,
 HYPRE_Int hypre_IJMatrixInitializeParCSR ( hypre_IJMatrix *matrix );
 HYPRE_Int hypre_IJMatrixGetRowCountsParCSR ( hypre_IJMatrix *matrix, HYPRE_Int nrows,
                                              HYPRE_BigInt *rows, HYPRE_Int *ncols );
-HYPRE_Int hypre_IJMatrixGetValuesParCSR ( hypre_IJMatrix *matrix, HYPRE_Int nrows, HYPRE_Int *ncols,
-                                          HYPRE_BigInt *rows, HYPRE_BigInt *cols, HYPRE_Complex *values );
+HYPRE_Int hypre_IJMatrixGetValuesParCSR ( hypre_IJMatrix *matrix, HYPRE_Int nrows, HYPRE_Int *ncols, HYPRE_BigInt *rows,
```
victorapm, line 443: Should we add a warning that this does not support GPUs for now?

victorapm (approved):
Thank you, Wayne! I left some minor comments

src/IJ_mv/IJMatrix_parcsr.c
rfalgout: I'm pretty sure this means "get all rows".  I think we've also used something similar to "get all cols" for a given row.  Looking through the code here, that's what seems to be happening.

rfalgout (approved):
Looks good to me.  Thanks!

src/IJ_mv/IJMatrix_parcsr.c
victorapm: Thanks, Rob! In that case, it seems `row_indexes` is not needed here

src/IJ_mv/HYPRE_IJMatrix.c
waynemitchell: `ncols` is used in `hypre_IJMatrixGetValuesParCSR()` and seems required to me in this case.

src/IJ_mv/HYPRE_IJMatrix.c
waynemitchell: Same as above, I think `ncols` is necessary here.

src/IJ_mv/HYPRE_IJ_mv.h
waynemitchell: Sure, I like that name better.

src/IJ_mv/_hypre_IJ_mv.h
waynemitchell: Added some "not implemented" error messages. Take a look at the additions and see whether that is what you had in mind.

src/IJ_mv/HYPRE_IJMatrix.c
victorapm: Sounds good, thanks for clarifying!

src/IJ_mv/_hypre_IJ_mv.h
victorapm: Great, thanks!

PR #1111, opened 2024-08-04
Add Non-UVM support to preconditioner options

victorapm:
Removes the dependency on Unified Virtual Memory (UVM) for executing HMIS coarsening in BoomerAMG and ILU-threshold in ILU when Hypre is configured for GPUs. Although these algorithms continue to execute on the CPU, they can now run without requiring UVM support.

### DIFF
--- a/src/parcsr_ls/par_ilu_setup.c
+++ b/src/parcsr_ls/par_ilu_setup.c
@@ -330,7 +330,6 @@ hypre_ILUSetup( void               *ilu_vdata,
          else
 #endif
          {
-
             hypre_ILUSetupILUT(matA, max_row_elmts, droptol, perm, perm, nLU, nLU,
                                &matL, &matD, &matU, &matS, &u_end);
          }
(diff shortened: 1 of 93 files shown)


rfalgout (approved):
Looks good to me.  Thanks!

src/parcsr_ls/par_ilu_setup.c
```diff
@@ -1775,42 +1764,31 @@ hypre_ILUSetupLDUtoCusparse(hypre_ParCSRMatrix  *L,
...
-      LDU_diag_data[pos++] = 1.0 / D[i - 1];
-      for (j = U_diag_i[i - 1]; j < U_diag_i[i]; j++)
+      for (j = U_diag_i[i]; j < U_diag_i[i + 1]; j++)
       {
-         LDU_diag_j[pos] = U_diag_j[j];
-         LDU_diag_data[pos++] = U_diag_data[j];
+         LDU_diag_j[pos]   = U_diag_j[j];
+         LDU_diag_a[pos++] = U_diag_a[j];
       }
    }
```
oseikuffuor1: This seems to be resetting the diagonal to the first entry. Is there a reason for this other than to be consistent with some of the other situations in hypre? We may need to make sure it doesn't break any algorithmic assumptions downstream.

oseikuffuor1 (commented):
@victorapm I have a question about the implications to doing solves on device if the UVM support is removed. Presumably, we would explicitly move data to device for the solve then. Let's discuss this at the meeting this afternoon

liruipeng (approved):
Looks good. Thank you @victorapm

oseikuffuor1 (approved):
Thanks Victor!

PR #1114, opened 2024-08-15
Add filtering option to AMG

victorapm:
Adds a filtering option to AMG targeting systems of PDEs.

Filtering is done in the input matrix to BoomerAMG to remove inter-variable couplings. This only affects the matrix used for preconditioning purposes.

The new option leads to smaller operator complexities at the expense of more linear solver iterations (in general). This can be turned on via `HYPRE_BoomerAMGSetFilterFunctions`

A few results obtained with linear elasticity matrices, using a single core and `-solver 1 -nf 3 -th 0.6`:

setup
solve

### DIFF
--- a/src/parcsr_ls/HYPRE_parcsr_ls.h
+++ b/src/parcsr_ls/HYPRE_parcsr_ls.h
@@ -170,7 +170,19 @@ HYPRE_Int HYPRE_BoomerAMGGetFinalRelativeResidualNorm(HYPRE_Solver  solver,
  * The default is 1, i.e. a scalar system.
  **/
 HYPRE_Int HYPRE_BoomerAMGSetNumFunctions(HYPRE_Solver solver,
-                                         HYPRE_Int          num_functions);
+                                         HYPRE_Int    num_functions);
+
+/**
+ * (Optional) Sets filtering for system of PDEs (\e num_functions > 1).
+ *
+ * \param filter_functions An integer flag to enable or disable filtering of inter-variable
+ * connections in the input matrix used for preconditioning.
+ *   - A value of 0 (default) indicates no filtering, preserving all inter-variable connections.
+ *   - A value of 1 enables filtering, removing inter-variable connections to lower
+ *     operator and memory complexities.
+ **/
+HYPRE_Int HYPRE_BoomerAMGSetFilterFunctions(HYPRE_Solver solver,
+                                            HYPRE_Int    filter_functions);
 
 /**
  * (Optional) Sets the mapping that assigns the function to each variable,
--- a/src/parcsr_ls/HYPRE_parcsr_ls.h
+++ b/src/parcsr_ls/HYPRE_parcsr_ls.h
@@ -180,6 +180,11 @@ HYPRE_Int HYPRE_BoomerAMGSetNumFunctions(HYPRE_Solver solver,
  *   - A value of 0 (default) indicates no filtering, preserving all inter-variable connections.
  *   - A value of 1 enables filtering, removing inter-variable connections to lower
  *     operator and memory complexities.
+ *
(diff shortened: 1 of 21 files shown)


src/parcsr_mv/par_csr_filter.c
ulrikeyang, line 1: It looks like you assume a certain ordering of the variables, namely that each function is represented at each grid point and ordered accordingly. Is that correct? So, if that is not the case, this doesn't make sense.

src/parcsr_ls/par_amg_setup.c
ulrikeyang, line 1: Are you saving A somewhere, or are you replacing it with A-tilde?

src/parcsr_mv/par_csr_filter.c
victorapm: Correct, I do make that assumption in this implementation

src/parcsr_ls/par_amg_setup.c
victorapm: The A passed as input is not changed. A_tilde is simply a pointer to A if the new option is off, otherwise it's a pointer to the new filtered matrix if the option is on. In the latter case, A_tilde is destroyed before completing AMG's setup

ulrikeyang: So to save memory, you need more memory? Did you see any time improvements using this new option?

victorapm: More memory at the finest level, but less memory in the following ones, overall we use less memory.

I've seen about up to 2x improvement in execution times for loosely coupled problems (see plots above)

ulrikeyang (approved):
Looks good.

src/parcsr_ls/par_amg_setup.c
victorapm: I've added results to the initial comment in this PR

rfalgout (approved):
Looks good.  Thanks!

src/parcsr_ls/HYPRE_parcsr_ls.h
```diff
@@ -170,7 +170,19 @@ HYPRE_Int HYPRE_BoomerAMGGetFinalRelativeResidualNorm(HYPRE_Solver  solver,
...
+/**
+ * (Optional) Sets filtering for system of PDEs (\e num_functions > 1).
+ *
+ * \param filter_functions An integer flag to enable or disable filtering of inter-variable
+ * connections in the input matrix used for preconditioning.
+ *   - A value of 0 (default) indicates no filtering, preserving all inter-variable connections.
+ *   - A value of 1 enables filtering, removing inter-variable connections to lower
+ *     operator and memory complexities.
+ **/
+HYPRE_Int HYPRE_BoomerAMGSetFilterFunctions(HYPRE_Solver solver,
```
oseikuffuor1, line 189: Victor, can you add a note here about the assumption of the ordering (interleaved, I believe, right)?

oseikuffuor1 (approved):
Looks good @victorapm . I left a comment about the documentation. Also, would regression testing this option be appropriate?

victorapm:
Thanks, yes! I will add a regression test to this PR as well

src/parcsr_ls/HYPRE_parcsr_ls.h
victorapm: Done in [dbc513f](https://github.com/hypre-space/hypre/pull/1114/commits/dbc513f6940ef515170b06b1452d192de118bb0f)

victorapm:
Hi all, I've added 4 new tests involving the new option. They use systems of laplacians and test for sizes 2 and 3, sequential and parallel

oseikuffuor1:
[like]  Osei-Kuffuor, Daniel reacted to your message:

PR #1117, opened 2024-08-27
Update struct matrix print routines and TEST_structmat saved files

rfalgout:
I changed all of the index printing for struct matrices to use parentheses with commas.  Let me know what you think.

Note that the matrix read routine doesn't work yet.  I need to decide how to deal with symmetric matrices and then do some simple debugging.


victorapm (approved):
Thanks, Rob!

rfalgout:
I made a bunch of changes and also implemented the read routines (symmetric, constant, and variable).  I tested one symmetric matrix that had both constant and variable coefficients, and it seemed to be read in correctly (printed it out and diffed with the original).  However, the struct driver dies and I think it's probably an unrelated issue (with the driver) that I will have to debug.

The matrix/vector file format is slightly changed: I added a "num_values" line just before both the constant and variable data; and I changed the names in the matrix file to "Constant Data" and "Variable Data".

victorapm:
Thank you for finishing up some of the TODOs I had left :)

rfalgout:
I am ready to merge this whenever it's okay with you @victorapm and @waynemitchell.  Let me know.
* I cleaned up a memory leak, then added a couple of structmat tests that produce a new leak. :(  The new tests multiply RA and AP (where these are the usual RAP matrices) just to test the print routines.  I'll track this down later since it's a leak that arises in a scenario we don't typically have.
* The `-fromfile` feature of the `struct` driver is still broken, but that feature really requires a complete driver overhaul that also can wait.

victorapm:
Hi Rob, I reviewed the branch again and it looks ready to be merged. Thanks for fixing the leak!

waynemitchell (approved):
Looks good to me. Thanks, Rob!

PR #1121, opened 2024-09-03
Add `hypre_ParCSRMatMatDiag`

victorapm:
This PR introduces a new function, used in MGR, for computing the diagonal entries of the product of two ParCSRMatrix instances, A and B.

The new function is aimed at optimizing scenarios where only the diagonal of the resulting matrix C is needed, avoiding the computational and memory overhead associated with full matrix multiplication.


src/parcsr_ls/par_mgr_interp.c
```diff
@@ -731,171 +733,159 @@ hypre_ExtendWtoPHost(HYPRE_Int      P_nr_of_rows,
...
-   hypre_ParCSRMatrixDestroy(A_FF);
-   hypre_ParCSRMatrixDestroy(A_FC);
-   if (method <= 0)
-   {
-      hypre_CSRMatrixDestroy(W_diag);
-      hypre_CSRMatrixDestroy(W_offd);
-   }
+   *Wp_ptr = Wp;
+   *P_ptr  = P;
+   *Wp_ptr = NULL;
```
rfalgout: Why is this set to NULL right after it's set to Wp?

src/parcsr_ls/par_mgr_interp.c
```diff
@@ -731,171 +733,159 @@ hypre_ExtendWtoPHost(HYPRE_Int      P_nr_of_rows,
 
 HYPRE_Int
 hypre_MGRBuildPHost( hypre_ParCSRMatrix   *A,
+                     hypre_ParCSRMatrix   *A_FF,
```
rfalgout, line 739: It would be good to add some comments about the return values of this function and what it means when they are NULL or not NULL, etc.

rfalgout (approved):
This looks good to me, modulo the questions/comments I had.  Thanks!

src/parcsr_ls/par_mgr_interp.c
victorapm: Thanks for catching! This was the result of a bad merge

src/parcsr_ls/par_mgr_interp.c
victorapm: Yes, I will add some comments!

src/parcsr_ls/par_mgr_interp.c
victorapm: Fixed in [e81a8fb](https://github.com/hypre-space/hypre/pull/1121/commits/e81a8fb110c5924b8805613f97a42762b6f8861c)

src/parcsr_ls/par_mgr_interp.c
victorapm: Done in [a746eae](https://github.com/hypre-space/hypre/pull/1121/commits/a746eaec1828562c240f639a1d44334285df301b)

victorapm:
@oseikuffuor1 @liruipeng could you take a look at this? Thanks!

src/parcsr_ls/par_mgr_interp.c
```diff
@@ -714,188 +718,182 @@ hypre_ExtendWtoPHost(HYPRE_Int      P_nr_of_rows,
...
+ *
+ * This function computes:
+ *   - *Wp_ptr: Pointer to the constructed auxiliary matrix Wp
+ *              (NULL if method == 0, meaning Wp is an empty matrix)
+ *   - *P_ptr: Pointer to the constructed interpolation matrix P
+ *             (Always non-NULL if successful)
  *--------------------------------------------------------------------------*/
-
 HYPRE_Int
 hypre_MGRBuildPHost( hypre_ParCSRMatrix   *A,
```
oseikuffuor1, line 738: Is there a reason why we are changing the argument signature here? AFF and AFC are not always needed right?

src/parcsr_ls/par_mgr_setup.c
```diff
@@ -1188,7 +1189,10 @@ hypre_MGRSetup( void               *mgr_vdata,
                                                    -1, CF_marker, inv_size, 1, diag_inv);
             frelax_diaginv[lev] = diag_inv;
             blk_size[lev] = block_jacobi_bsize;
-            hypre_MGRBuildAff(A_array[lev], CF_marker, debug_flag, &A_FF);
+            if (!A_FF)
```
oseikuffuor1, line 1192: Was this a bug before? Could AFF have been build previously?

src/parcsr_mv/par_csr_matmat.c
```diff
@@ -0,0 +1,159 @@
+/******************************************************************************
```
oseikuffuor1, line 1: Maybe these should be put together with hypre_ParCSRMatMat<Host, Device> codes? Not sure how much additional work that will be to move those into this file or these into par_csr_triplemat.c.

oseikuffuor1 (commented):
@victorapm it looks mostly good to me. I left a few minor comments for your review.

src/parcsr_ls/par_mgr_interp.c
victorapm: They are not always needed, but when they are this avoids extra work inside `MGR_BuildPHost` to recompute them

src/parcsr_ls/par_mgr_setup.c
victorapm: Right, this was a bug before that our tests didn't catch

src/parcsr_mv/par_csr_matmat.c
victorapm: My idea is to move the `ParCSRMatMat` codes to this new file actually, since `triplemat` became quite big

oseikuffuor1 (approved):

liruipeng (approved):

PR #1124, opened 2024-09-06
Fix build with umpire host pool

victorapm:
This PR fixes the build when using the `--with-umpire-host` option.

The build was broken due to a call to `hypre_CTAlloc` happening prior to umpire's initialization. The solution here is to use `calloc` instead in a couple of places.

This PR also updates memory usage reporting with umpire's host pool info.

### DIFF
--- a/AUTOTEST/check-mem.sh
+++ b/AUTOTEST/check-mem.sh
@@ -38,7 +38,9 @@ find . -type f -print | egrep '[.]*[.](c|cc|cpp|cxx|C|h|hpp|hxx|H)$' |
   egrep -v '/FEI_mv' |
   egrep -v '/hypre/include' |
   egrep -v '/utilities/memory_tracker.c' |
-  egrep -v '/utilities/memory.c' > check-mem.files
+  egrep -v '/utilities/memory.c' |
+  egrep -v '/utilities/general.c' |
+  egrep -v '/utilities/device_utils.c' > check-mem.files
 
 egrep '(^|[^[:alnum:]_]+)malloc[[:space:]]*\('  `cat check-mem.files` >&2
 egrep '(^|[^[:alnum:]_]+)calloc[[:space:]]*\('  `cat check-mem.files` >&2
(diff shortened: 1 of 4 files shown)


rfalgout:
Can you explain what breaks with hypre_CTAlloc?  That should ultimately call calloc so I don't understand the issue.  We should be able to use our own wrapper routines for memory allocation.  Either we need to fix something in our wrappers or maybe this is something that needs to be taken up with the Umpire folks.  Let me know.  Thanks!

victorapm:
Sure, Rob!

The issue arises because hypre_CTAlloc calls hypre_umpire_host_pooled_allocate when hypre is built with the Umpire host pool enabled, as seen in this code segment:

https://github.com/hypre-space/hypre/blob/3caa81955eb8d1b4e35d9b450e27cf6d07b50f6e/src/utilities/memory.c#L253-L268

For hypre_umpire_host_pooled_allocate to function correctly, Umpire needs to be initialized, which happens at:

https://github.com/hypre-space/hypre/blob/3caa81955eb8d1b4e35d9b450e27cf6d07b50f6e/src/utilities/general.c#L384-L386

However, this initialization occurs after the hypre_HandleCreate() call, which internally calls hypre_CTAlloc. Since hypre_CTAlloc attempts to allocate memory using Umpire (before Umpire is initialized), the allocation fails at this stage.

I considered moving the Umpire initialization to occur prior to hypre_HandleCreate(), but this creates another issue. Umpire initialization itself depends on the hypre handle being available, as shown here:

https://github.com/hypre-space/hypre/blob/3caa81955eb8d1b4e35d9b450e27cf6d07b50f6e/src/utilities/memory.c#L1976-L1979

This leads to a circular dependency: hypre_HandleCreate() requires memory allocation, but the Umpire pool (needed for the allocation) cannot be initialized until after the handle is created.

To resolve this, the cleanest solution in my opinion, is to use calloc directly in a few special cases (specifically in hypre_Handle and hypre_DeviceData), bypassing hypre_CTAlloc and avoiding the dependency issue altogether.

Let me know if you think we should explore other options or engage with the Umpire team, but this solution seems to avoid the circular dependency without major changes to the memory allocation logic.

liruipeng (approved):
Looks good to me. Thanks!

rfalgout (approved):
This looks good for now, but let's continue to think about possible alternative approaches.  The global variable approach might be good since we don't support thread safety in hypre and already use global variables for other things.  Maybe it would be good to add a note in the code about considering alternatives (we'll likely forget about it otherwise).  Thanks!

PR #1135, opened 2024-09-19
Tux regression test fixes for rhel9 system

rfalgout:
The RHEL9 upgrade resulted in compiler version changes, etc. that broke some of the regression tests.  Most of the changes are because of new compiler warnings and saved file differences.  There are still one or two things to fix, but this is mostly ready to merge.

### DIFF
--- a/AUTOTEST/machine-tux-spack.sh
+++ b/AUTOTEST/machine-tux-spack.sh
@@ -37,7 +37,9 @@ src_dir=`cd $1; pwd`
 shift
 
 # OpenMPI limits the number of processes available by default - override
+# RDF: The first environment variable didn't work for me
 export OMPI_MCA_rmaps_base_oversubscribe=1
+export PRTE_MCA_rmaps_default_mapping_policy=:oversubscribe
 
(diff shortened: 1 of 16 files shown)


victorapm (approved):
Thanks, Rob!

What are the mpi library and gcc versions being used in the new tuxbox?

rfalgout:
> Thanks, Rob!
> 
> What are the mpi library and gcc versions being used in the new tuxbox?

gcc (GCC) 11.4.1 20231218 (Red Hat 11.4.1-3)

For MPI, I installed my own version of MPICH: 4.2.2

victorapm:
Thanks, Rob! I will need to update gcc on my box after this :)

rfalgout:
I am getting the following output from the `examples/ex12f` run:
```
Note: The following floating-point exceptions are signalling: IEEE_DENORMAL
```
I was unable to figure out exactly what is triggering it and I think we can probably ignore it anyway.  Any opinions on this?  If we just filter the message out, should that be done in `runtest.sh` or in a new `machine-tux.filters` file?

PR #1136, opened 2024-09-20
Fix rhel9 compiler differences in sludist saved file

rfalgout:
This should be the final fix for getting the spack tests to run cleanly again

### DIFF
--- a/src/test/TEST_superlu/sludist.saved
+++ b/src/test/TEST_superlu/sludist.saved
@@ -14,5 +14,5 @@ Final Relative Residual Norm = 3.172597e-08
 
 
 GMRES Iterations = 8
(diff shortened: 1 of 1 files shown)


victorapm (approved):
Very nice! So the OpenMPI issue was sorted out I guess. Thank you!

rfalgout:
> Very nice! So the OpenMPI issue was sorted out I guess. Thank you!

Yes.  Turns out I needed to set a different environment variable from the one you told me about to get it to work.  It took some googling to find the answer, because it isn't really documented well on any of the OpenMPI sites (I didn't see it at all actually).  They usually suggest using a command-line option, but that's not as easy for us to integrate into our testing.

PR #1141, opened 2024-09-26
Fix GPU build error and warnings

victorapm:
This PR contains fixes for hypre's GPU build. More precisely:

- Fix build error with cuda-12.5 due to the lack of using `thrust` namespace in a few places.
- Add definition to disable deprecated warnings from cuSPARSE-12 involving usage of ILU functions.
- Fix several warnings related to the CUDA build, most of them due to unused variables.
- Rename a few Cuda prefixed functions to a more generic name since they make sense for all GPU backends.


rfalgout (approved):
Looks good to me.  Thanks!

liruipeng (approved):
Looks good. Thanks!

waynemitchell:
@victorapm , there are some issues here for the sycl build. Let me take a closer look, and I may push some chages.

victorapm:
Thanks Wayne!

waynemitchell:
Looks like the issue wasn't actually sycl-related. I think there was just one call to hypre_SyncComputeStream() that didn't get updated and still had the handle arg. So the compile failed. I went ahead and merged the most recent master and fixed that one function call. Running the sycl regressions now.

waynemitchell:
Out of curiosity, what's the strategy behind using `HYPRE_UNUSED_VAR` vs. just removing the unused variable from the function? I guess that actually removing the unused variables is probably more changes/effort. But is there any other reason?

liruipeng:
> Out of curiosity, what's the strategy behind using `HYPRE_UNUSED_VAR` vs. just removing the unused variable from the function? I guess that actually removing the unused variables is probably more changes/effort. But is there any other reason?

Sometimes where we have #ifdef, variables are not used in some branch but used in others. We can probably refactor the code but the HYPRE_UNUSED_VAR approach is probably easier. Just to silence compilers.

waynemitchell (approved):

victorapm:
Just adding to Rui Peng's response, that is particularly true with `#if defined HYPRE_USING_GPU`. While, in other places such as BoomerAMG functions, refactoring would be easier to do.

PR #1142, opened 2024-09-26
Recmat structmat

rfalgout:
This fixes some regression test issues with structmat and sstructmat.  It's difficult to explain what the problem was, but it had to do with calling BoxManAssemble more than once.  There is an effort in the code to minimize the calls to this routine, but it leads to rather confusing code.  I ultimately had to add an argument to hypre_StructMatmultSetup to tell it when it should and should not assemble the grid for the product matrix.  I still think there is a potential issue with this that we will have to work through in the future.

### DIFF
--- a/src/sstruct_ls/maxwell_semi_interp.c
+++ b/src/sstruct_ls/maxwell_semi_interp.c
@@ -18,6 +18,26 @@
 #include "_hypre_sstruct_ls.h"
 
 /*--------------------------------------------------------------------------
+ * Adding a local FineToCoarse function to fix --with-print-errors issues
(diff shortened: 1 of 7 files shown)


victorapm (approved):
Thanks, Rob! I agree the old code was a bit confusing regarding the boxmanager calls. This solution is good!

rfalgout:
I also just pushed fixes for the single and longdouble tests.  We only needed to update the saved files.

rfalgout:
I added a fix for the maxwell issues.  With all of these changes, the nightly regression tests should be clean.

waynemitchell (approved):
All changes look good to me. Thanks, Rob!

PR #1143, opened 2024-09-27
Handle hipMemoryTypeUnregistered introduced in rocm 6

helloworld922:
rocm 6 seems to return this type for host memory allocated outside of rocm


victorapm (approved):
Thank you for your contribution!

rfalgout (approved):

PR #1150, opened 2024-10-02
Fix complex build

victorapm:
Fix compilation errors for build with complex numbers support: `./configure --enable-complex --enable-maxdim=4`

Example `ex18comp` now runs successfully


rfalgout (approved):
That does the trick.  Thanks!

victorapm:
I found other compilation errors with gcc, shall I open another PR for them?

PR #1154, opened 2024-10-04
Updating the nonmixedint tests to use the new emptyProc input file

rfalgout:
I didn't realize that some of the emptyProc tests were in the nonmixedint jobs file so they were still using the old input files.


victorapm (approved):
Thanks, Rob!! Curious to know why these are here. Do you know what is missing to get them to work in the mixedint build?

rfalgout:
It looks like these are all Euclid and Parasails runs, and we are moving away from supporting these in hypre.

victorapm:
Oh yes, right! Thanks!

PR #1156, opened 2024-10-06
cmake: add so version to libhypre

junghans:


src/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ endif ()
 
 # The version number.
 set(HYPRE_VERSION 2.31.0)
+set(HYPRE_SOVERSION 2)
```
junghans: On a second thought the `soversion` should maybe be `31` instead.

junghans:
ping @waynemitchell

victorapm:
Hi @junghans, thank you for your contribution! I talked to @streeve last week and pointed out this development branch: https://github.com/hypre-space/hypre/tree/cmake-hip

This branch will include the soversion change that your proposed here. I appreciate if you can test the branch and provide feedback. It's almost complete, except for documentation changes. Thank you!

junghans:
Perfect, I will look!

junghans:
@streeve @victorapm The `cmake-hip` branch would work for me, but given https://github.com/hypre-space/hypre/blob/cmake-hip/INSTALL.md#abi-compatibility, we might not want to use `2` as the so version.

victorapm:
@junghans Thanks for bringing this up. I updated the branch to use `major.minor` as the SOVERSION

junghans:
I have to check but I believe the so version needs to be an integer.

junghans:
Ok, I think CMake actually allows it, but i generated a non-standard symlink hierarchy: 
```
libhypre.so -> libhypre.so.2.32 -> libhypre.so.2.32.0
```

As a normal user I would more expect something like:
```
libhypre.so -> libhypre.so.2 -> libhypre.so.2.32
```
(see https://tldp.org/HOWTO/Program-Library-HOWTO/shared-libraries.html).

But of course as there is no [abi-compatibility](https://github.com/hypre-space/hypre/blob/cmake-hip/INSTALL.md#abi-compatibility), maybe:
```
libhypre.so -> libhypre.so.32
```
or
```
libhypre.so -> libhypre.so.232
```

would be better.

victorapm:
@junghans sounds great. Thanks for digging into this!

PR #1167, opened 2024-10-18
CMake build improvements

helloworld922:
Made it easier and less error prone to change configurations in CMake.

- Get rid of HYPRE_INSTALL_PREFIX, use CMAKE_INSTALL_PREFIX only. This
follows cmake common practices and allows changing the install prefix
in a single place.
- Get rid of HYPRE_BUILD_TYPE, use CMAKE_BUILD_TYPE only. This
follows cmake common practices and allows changing the build type
in a single place.
- Ensure that certain internal cache flags are toggled on/off as
needed. These probably shouldn't be cached at all, however I left them
cached. Specifically this ensures that HYPRE_USING_DEVICE_MEMORY and
HYPRE_USING_DEVICE_MEMORY cannot be simultaneously enabled.


ZedThree:
There's also the conventional (and built in!) `BUILD_SHARED_LIBS` which is clobbered by `HYPRE_ENABLE_SHARED`

victorapm:
Hi Andrew, thank you for the contribution! I included this on #1181

victorapm:
The contents of this PR have been included in #1181, which has been merged into master

PR #1170, opened 2024-10-22
Recmat pmatmult multi-variable fixes

rfalgout:
This implements a partial extension of PMatmult() to multiple variables.  Some comments on the implementation:
* It assumes there is only one struct-matrix product term for each M_ij block of pmatrix M
* It requires having only one variable type (probably cell-centered only)
* The struct matmult API was restructured to compute a collection of matmults
* The main files that changed were `struct_matmult.c` and `sstruct_matmult.c` (and related header files)
* With this, all of the regression tests work except for `TEST_sstruct/io.sh`


rfalgout:
FYI, everything is basically running now.  There is a small residual-norm difference between PFMG and SysPFMG that I want to look into a little bit more (on master, these produce the same results when they can both be run on the same problem).  After that, I plan to:
* Try to reintroduce commented-out runs in the regression tests (I was already able to do this in periodic.jobs)
* Update the comments in the code

rfalgout:
@victorapm and @waynemitchell.  This PR is ready now.  I am running the full autotest runs just to be sure, but it's close enough for you start giving it a look/review.  The main changes are in the two `matmult.c` files.  I will need to merge in the `master` branch before merging the PR, and we should coordinate on how to best merge this with the work you are doing on other branches.  Let me know.  Thanks!

rfalgout:
> @victorapm and @waynemitchell. This PR is ready now.

I just realized I hadn't formally requested your review on this.  Oops.  Thanks again!

src/sstruct_ls/sys_pfmg_solve.c
```diff
@@ -47,14 +47,19 @@ hypre_SysPFMGSolve( void                 *sys_pfmg_vdata,
...
-   HYPRE_Real            b_dot_b, r_dot_r, eps = 0;
-   HYPRE_Real            e_dot_e = 0, x_dot_x = 1;
+   HYPRE_Real              b_dot_b, r_dot_r, eps = 0;
+   HYPRE_Real              e_dot_e = 0, x_dot_x = 1;
 
-   HYPRE_Int             i, l;
-   char                  filename[255];
+   HYPRE_Int               i, l;
+#if 0
+   HYPRE_Int               print_level     = (sys_pfmg_data -> print_level);
```
victorapm, line 56: What is the reason for turning this off?

src/sstruct_mv/sstruct_matmult.c
```diff
@@ -11,7 +11,25 @@
...
 
 /*==========================================================================
- * SStructPMatrix routines
+ * SStructPMatrix matrix-multiply routines
+ *
+ * The pmatrix (SStructPMatrix) contains an nvars x nvars array of struct
+ * (StructMatrix) matrices.  The multiply is then a block-matrix multiply
+ * involving these struct matrices.
+ *
+ * NOTE: This only works for cell-centered variable types (see below comment).
```
victorapm, line 20: Should we add a runtime check for the variable type? If it's not cell-centered, then return an error

src/sstruct_mv/sstruct_matmult.c
```diff
@@ -75,37 +99,114 @@ hypre_SStructPMatmultCreate(HYPRE_Int                   nmatrices_input,
...
+         /* Run through the nested for-loop */
+         for (ii = 0; ii < nn; ii++)
          {
-            (pmmdata -> smmdata)[vi][vj] = NULL;
+            /* M_ij += A1_{i,k1} * A2_{k1,k2} * ... * AN_{km,j} */
+
+            zero_product = 0;
+            for (k = 0; k < nterms; k++)
+            {
+               if (trans[k])
```
victorapm: How about the one-liner?

```
smatrices[k] = hypre_SStructPMatrixSMatrix(pmatrices[terms[k]],
                                           trans[k] ? i[k + 1] : i[k],
                                           trans[k] ? i[k] : i[k + 1]);
```

src/sstruct_mv/sstruct_matmult.c
```diff
@@ -217,79 +319,110 @@ hypre_SStructPMatmultSetup( hypre_SStructPMatmultData  *pmmdata,
...
+   hypre_SStructPGridCellSGrid(pgrid) = NULL;
 
    /* Create part matrix data structure */
    hypre_SStructPMatrixCreate(comm, pgrid, pstencils, &pM);
    smaps = hypre_SStructPMatrixSMaps(pM);
 
+   /* Initialize the struct matmults for this part */
+   /* RDF NOTE: This does not assemble the struct grids.  They are assembled
+    * below or in HYPRE_SStructGridAssemble() to reduce box manager overhead.
+    * It's not clear if this is useful anymore. */
```
victorapm: This should still be useful to reduce communication overhead

victorapm (approved):
Thank you for all the work!

In my opinion, we should merge this to `recmat` first and I can solve the conflicts in my branch locally.

src/sstruct_ls/sys_pfmg_solve.c
rfalgout: It creates a compile warning because line 338 is turned off now.  The reason for the latter is to match the behavior of PFMG (it also eliminates a residual computation, which we can debate the pros and cons of), which is one of the regression tests.

src/sstruct_mv/sstruct_matmult.c
rfalgout: Yes, probably.  I haven't verified that it fails for other variable types, but it looks like it probably would.  I'll go ahead and add a check.  Thanks!

src/sstruct_mv/sstruct_matmult.c
rfalgout: I'm not a huge fan of this shorthand notation, but in this case it would be okay accompanied with a descriptive comment.  I'll make this change.

src/sstruct_mv/sstruct_matmult.c
rfalgout: What I am referring to here is the `assemble_grid` feature of the `MatmultInit()` function.  I think we achieve the same communication reduction by way of the restructuring of the struct matmult interface, but I need to look more carefully (hence the RDF NOTE; you'll see that I left a number of these lying around :) ).

src/sstruct_ls/sys_pfmg_solve.c
victorapm: I see now, thanks Rob!

src/sstruct_mv/sstruct_matmult.c
victorapm: Sounds good, thanks!

src/sstruct_mv/sstruct_matmult.c
victorapm: Sounds good! Thanks!

src/sstruct_mv/sstruct_matmult.c
victorapm: Oh I see now, we could compare before vs after with caliper. I guess we need a couple hundreds of ranks at least to see a difference

rfalgout: I haven't removed the feature yet, but caliper tests would still be good to do as part of our performance testing.  For now, I've just updated this comment to make it more clear.  Thanks!

rfalgout:
> In my opinion, we should merge this to `recmat` first and I can solve the conflicts in my branch locally.

I am good with that plan, though I think it might be good for us to resolve the current recmat regression issues first.  I will then need to merge those updates as well as master into this branch.

victorapm:
Yes, that sounds better!

waynemitchell (approved):
I can't say I've absorbed all the details here, but everything looks fine to me at a high level. Thanks, Rob!

PR #1171, opened 2024-10-29
Error verbosity

rfalgout:
This adds the function `HYPRE_SetPrintErrorVerbosity()` to control which error messages to record for printing.


rfalgout:
If any of you have ideas about how we can regression test this and the other error interface routines, let me know.

victorapm (approved):
Thanks, Rob!

Related to this topic: when the krylov solver doesn't converge, should we write the error msg in all ranks or just rank 0?

rfalgout:
> Related to this topic: when the krylov solver doesn't converge, should we write the error msg in all ranks or just rank 0?

Maybe we can chat about this at our next meeting.  I think this may be something we do *not* want to do, but that's because I typically think of error codes as a debugging tool, and when debugging errors, I'm not sure we can assume that all ranks are in the same state.

rfalgout:
Another note about printing error messages...  We could modify the routine `HYPRE_PrintErrorMessages()` to be a non-collective routine callable from any rank (to do that, we only need to remove an MPI_Barrier call from the current implementation).  Then the user can decide if they only want to print messages on rank zero.  Messages generated on non-zero ranks (and not also on rank zero) would be missed, of course.

victorapm:
This sounds a good idea to me

rfalgout:
I changed `HYPRE_PrintErrorMessages()` to be non-collective

ulrikeyang (approved):
This looks good. Thank you.

oseikuffuor1 (approved):

rfalgout:
Hi @liruipeng and @waynemitchell .  Does this look okay to you?

waynemitchell (approved):
Looks good. Thanks, Rob!

liruipeng (approved):
Looks good to me. Thanks!

PR #1174, opened 2024-10-31
Fix for hypre_BoxArrayCreateFromIndices()

waynemitchell:
There are cases where the algorithm implemented in hypre_BoxArrayCreateFromIndices() does not find a change in sign in the computed Laplacian. In these cases, the cut dimension defaults to 0, which may not be a valid direction (e.g. if all points lie in the same plane in that dim). This PR addresses this case by selecting the cut dimension to be the box's longest dimension.

### DIFF
--- a/src/struct_mv/box.c
+++ b/src/struct_mv/box.c
@@ -909,7 +909,9 @@ hypre_BoxArrayCreateFromIndices( HYPRE_Int         ndim,
    }
 
    /* Set defaults */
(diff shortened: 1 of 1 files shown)


victorapm (approved):
Thanks, Wayne!

The reason for box_minvol=1 is to forbid a cut when the box volume is equal to 1 (single cell): https://github.com/hypre-space/hypre/blob/3aac68acbb35f6e6e2d36dd24416817f40735fbf/src/struct_mv/box.c#L1016

PR #1180, opened 2024-11-05
Fix broken regressions

waynemitchell:
Fixing the regression tests broken by https://github.com/hypre-space/hypre/pull/1174

### DIFF
--- a/src/struct_mv/box.c
+++ b/src/struct_mv/box.c
@@ -893,7 +893,7 @@ hypre_BoxArrayCreateFromIndices( HYPRE_Int         ndim,
    HYPRE_Int         *indices[HYPRE_MAXDIM];
    HYPRE_Int         *lbox_indices[HYPRE_MAXDIM];
    HYPRE_Int         *rbox_indices[HYPRE_MAXDIM];
-   HYPRE_Int          splitdir, dir, d, i;
+   HYPRE_Int          splitdir, dir, sign_change, d, i;
    HYPRE_Int          index, size, capacity, change;
    HYPRE_Int          num_indices;
    HYPRE_Int          num_lbox_indices, num_rbox_indices;
@@ -1095,12 +1095,14 @@ hypre_BoxArrayCreateFromIndices( HYPRE_Int         ndim,
             }
 
             /* Look for largest sign change among all directions */
-            dir = -1;
+            dir = 0;
+            sign_change = 0;
             for (d = 1; d < ndim; d++)
             {
                if (hypre_IndexD(sign, dir) < hypre_IndexD(sign, d))
                {
                   dir = d;
+                  sign_change = 1;
                }
             }
 
@@ -1109,7 +1111,7 @@ hypre_BoxArrayCreateFromIndices( HYPRE_Int         ndim,
             splitdir = dir;
 
             /* If no change of sign in the Laplacian was found, just cut the longest dim in half */
-            if (dir < 0)
+            if (!sign_change)
             {
                change = 0;
                for (d = 0; d < ndim; d++)
--- a/src/test/TEST_sstruct/amr2d.saved
+++ b/src/test/TEST_sstruct/amr2d.saved
@@ -51,20 +51,20 @@ Iterations = 18
 Final Relative Residual Norm = 5.252407e-07
 
 # Output file: amr2d.out.14
-Iterations = 28
-Final Relative Residual Norm = 8.819880e-07
+Iterations = 35
+Final Relative Residual Norm = 9.882763e-07
 
 # Output file: amr2d.out.15
-Iterations = 28
-Final Relative Residual Norm = 8.819880e-07
+Iterations = 35
+Final Relative Residual Norm = 9.882763e-07
 
 # Output file: amr2d.out.16
-Iterations = 10
-Final Relative Residual Norm = 7.820153e-07
+Iterations = 11
+Final Relative Residual Norm = 8.466675e-07
 
 # Output file: amr2d.out.17
-Iterations = 10
-Final Relative Residual Norm = 7.820153e-07
+Iterations = 11
+Final Relative Residual Norm = 8.466675e-07
 
 # Output file: amr2d.out.18
 Iterations = 6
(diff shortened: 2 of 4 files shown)


src/struct_mv/box.c
```diff
@@ -1095,12 +1095,14 @@ hypre_BoxArrayCreateFromIndices( HYPRE_Int         ndim,
             }
 
             /* Look for largest sign change among all directions */
-            dir = -1;
+            dir = 0;
```
rfalgout, line 1098: Is the issue that `dir` needs to be a valid direction (set to `0` by default) even when no sign change is detected?

waynemitchell: Yeah, you can see that `dir` is used in line 1102 below. I was trying to just reuse this variable for convenience, but I wasn't careful enough about it.

This should fix the valgrind issues in the regressions, but there were also some iteration count differences reported in some error files. Some of these are expected: namely any place that the unstructured interpolation for SSAMG is used, the change here may change some of the numerics. But I think there are still some tests where iteration counts change that I don't quite understand.

waynemitchell:
OK, I found the culprit for the regression changes in the amr2d and amr3d regressions (and a general reason why behavior will change here). The `hypre_BoxArrayCreateFromIndices()` function is used here in assembling the sstruct graph:
https://github.com/hypre-space/hypre/blob/recmat/src/sstruct_mv/HYPRE_sstruct_graph.c#L763
My initial change affects the behavior of how the boxes are generated where there is no sign change in the Laplacian when running the Berger-Rigoutsos algorithm. I don't know this algorithm well enough to understand exactly when this happens... But I do see that this occurs somewhat frequently. The previous behavior of the `hypre_BoxArrayCreateFromIndices()` code was to split the box in dim 0 by default. This can cause issues if the box already has no thickness in dim 0. My change avoids this issue by splitting along the dimension where the box is largest. Again, I don't know the underlying algorithm well here, so I'm not sure if there are other implications here that I'm not aware of, but I think my change is at least an improvement in that it prevents the code from trying to split a box along dim 0 when it cannot (which causes it to error out). If there are no additional concerns from @victorapm or @rfalgout , I think we should go ahead and merge this fix and then updated the .saved files to account for the changes in convergence due to the change in behavior here.

victorapm:
﻿Thanks for the update, Wayne!

Your idea sounds good to me, but I think we should return to this at some point since convergence has degraded for a few cases with the new changes

Victor

From: Wayne Mitchell ***@***.***>
Date: Friday, November 8, 2024 at 1:34 PM
To: hypre-space/hypre ***@***.***>
Cc: Paludetto Magri, Victor ***@***.***>, Mention ***@***.***>
Subject: Re: [hypre-space/hypre] Fix broken regressions (PR #1180)

OK, I found the culprit for the regression changes in the amr2d and amr3d regressions (and a general reason why behavior will change here). The hypre_BoxArrayCreateFromIndices() function is used here in assembling the sstruct graph:
https://github.com/hypre-space/hypre/blob/recmat/src/sstruct_mv/HYPRE_sstruct_graph.c#L763<https://urldefense.us/v3/__https:/github.com/hypre-space/hypre/blob/recmat/src/sstruct_mv/HYPRE_sstruct_graph.c*L763__;Iw!!G2kpM7uM-TzIFchu!1sPoyOHjaC3kunl1LWB2EBW-PpoqQNb8-gdZXyl4vzS3LfiZzJ67Hy_b23DD3AzgjsMWQcuNwBdv1m-epQddEDZ_jLs_gRc$>
My initial change affects the behavior of how the boxes are generated where there is no sign change in the Laplacian when running the Berger-Rigoutsos algorithm. I don't know this algorithm well enough to understand exactly when this happens... But I do see that this occurs somewhat frequently. The previous behavior of the hypre_BoxArrayCreateFromIndices() code was to split the box in dim 0 by default. This can cause issues if the box already has no thickness in dim 0. My change avoids this issue by splitting along the dimension where the box is largest. Again, I don't know the underlying algorithm well here, so I'm not sure if there are other implications here that I'm not aware of, but I think my change is at least an improvement in that it prevents the code from trying to split a box along dim 0 when it cannot (which causes it to error out). If there are no additional concerns from @victorapm<https://urldefense.us/v3/__https:/github.com/victorapm__;!!G2kpM7uM-TzIFchu!1sPoyOHjaC3kunl1LWB2EBW-PpoqQNb8-gdZXyl4vzS3LfiZzJ67Hy_b23DD3AzgjsMWQcuNwBdv1m-epQddEDZ_aS53o7w$> or @rfalgout<https://urldefense.us/v3/__https:/github.com/rfalgout__;!!G2kpM7uM-TzIFchu!1sPoyOHjaC3kunl1LWB2EBW-PpoqQNb8-gdZXyl4vzS3LfiZzJ67Hy_b23DD3AzgjsMWQcuNwBdv1m-epQddEDZ_B6YfpZM$> , I think we should go ahead and merge this fix and then updated the .saved files to account for the changes in convergence due to the change in behavior here.

rfalgout:
> I think we should return to this at some point since convergence has degraded for a few cases with the new changes

Which regression tests are affected?  Do you need one of use to generate the saved file changes on a tux box?  I would like to see what the impact is, whether we address it now or later.

waynemitchell:
> > I think we should return to this at some point since convergence has degraded for a few cases with the new changes
> 
> Which regression tests are affected? Do you need one of use to generate the saved file changes on a tux box? I would like to see what the impact is, whether we address it now or later.

@rfalgout , yes, can you generate updated .saved files (since I'm still on RHEL 8 and may not get exactly the right results on mine)?

rfalgout:
Okay, the saved files are there now.

victorapm (approved):

rfalgout:
@waynemitchell  Do you want me to run the tux tests on this?  Let me know if you think it's ready.  Thanks!

rfalgout (approved):
I agree with @victorapm that we should look into the convergence changes at some point later.  This looks good otherwise.  Thanks!

rfalgout:
The sstruct regression tests on my tux box are clean.  I think this is ready to merge (it's up to date with master too).  Thanks!

PR #1183, opened 2024-11-13
Recmat io regression test fixes

rfalgout:
This PR fixes the sstruct `io` regression tests and adds similar tests to struct.  It mainly involved fixing a few problems in the struct read routines.

FYI, most of the commits below can be ignored; they are from the pmatmult PR that was just merged.


rfalgout:
I had to comment out one test in `TEST_sstruct/io.jobs` to get things running.  We need to extend a certain conversion routine that we use for printing an IJ matrix in the `sstruct.c` driver.  I left a comment in the jobs file.  Otherwise, I think this is all clean.  I will run the full regression suite to be sure, but the PR is ready for review.

src/struct_mv/struct_vector.c
```diff
@@ -1681,9 +1681,11 @@ hypre_StructVectorRead( MPI_Comm    comm,
    hypre_StructVectorAssemble(vector);
 
    /*----------------------------------------
-    * Close file
+    * Clean up and close file
     *----------------------------------------*/
 
+   HYPRE_StructGridDestroy(grid);
```
waynemitchell, line 1687: Why do you destroy the grid here? Don't you want to keep it around in the vector data structure? Similar question for the read matrix routine also.

Also, answering your comment above in line 1669, it looks like the num ghost is set inside `HYPRE_StructVectorCreate()`, so no need to call `hypre_StructVectorSetNumGhost()` here.

rfalgout: I believe a reference to the grid is kept in the vector, so destroying it here really only decrements a reference counter.  I needed to add this to fix a memory leak.

Regarding the num_ghost question, we do pass this into the routine, and there was another issue I had to fix related to passing in an incorrectly sized num_ghost array from the struct driver.  I guess I was wondering if there was a case of reading a vector where we would want to have a differently-defined num_ghost argument (from the default set in Create).

waynemitchell: Ah, I see the ref counter thing now (I didn't realize it was incremented both on grid creation and vector creation). Thanks!

waynemitchell (approved):

PR #1195, opened 2024-12-10
Updated axpy, copy, and innerprod to work with different base grids

rfalgout:
This PR will do the following (when complete):
* Allow for different base grids and non-unitary strides in vector routines
* Implement logic in matvec to support future optimizations using the vector `Forget()` functionality


rfalgout:
FYI, this now works with the changes to grow the data space and minimize copies.  I have run all of the tests manually, and everything is clean.  I am running the autotest script just to make sure something wasn't forgotten.  This is also up-to-date with the 'recmat' branch.

I added a MemoryMode to vectors, and added a hardcoded switch in PFMG to adjust it for all of the matvecs.  I only did a couple of quick PFMG tests.  It is faster than the previous recmat version overall.  The difference between MemoryMode 0 (always resize) and 2 (minimize resize) is negligible, so I'm not sure where the savings came in versus the previous recmat branch (I touched a lot of code).  I did verify that many fewer resizes are done with MemoryMode 2 than with 0.

So, this is ready for Wayne to take over.  Let me know when you are well and ready to go over this.  In the meantime, I will try to update and expand on the comments in the code.

rfalgout:
Hi @waynemitchell and @victorapm .  This is ready for you to start reviewing.  Thanks!

src/struct_mv/struct_copy.c
```diff
@@ -9,69 +9,91 @@
  *
  * Structured copy routine
  *
+ * RDF TODO: The names for the vector class functions needs to be revisited.
```
victorapm, line 12: I think we should rename it to `hypre_StructVectorCopy` and move it to `struct_vector.c` or perhaps a `struct_vector_op.c`

src/struct_ls/HYPRE_struct_pcg.c
```diff
@@ -210,43 +210,49 @@ HYPRE_StructDiagScale( HYPRE_StructSolver solver,
...
+   loop_box = hypre_BoxCreate(ndim);
+   hypre_SetIndex(ustride, 1);
+
+   for (i = 0; i < nboxes; i++)
    {
-      box = hypre_BoxArrayBox(boxes, i);
+      hypre_StructVectorGridBoxCopy(x, i, loop_box);
+      start = hypre_BoxIMin(loop_box);
 
       A_data_box = hypre_BoxArrayBox(hypre_StructMatrixDataSpace(A), i);
```
victorapm, line 232: I'm wondering if we should have a `hypre_StructMatrixGridDataBox` as we have for vectors now

src/struct_ls/smg_residual_unrolled.c
```diff
@@ -81,12 +83,12 @@ hypre_SMGResidualSetup( void               *residual_vdata,
    grid    = hypre_StructMatrixGrid(A);
    stencil = hypre_StructMatrixStencil(A);
 
-   hypre_SetIndex3(unit_stride, 1, 1, 1);
+   hypre_SetIndex3(ustride, 1, 1, 1);
```
victorapm, line 86:
```suggestion
```

victorapm: Suggesting to delete this since we have `hypre_SetIndex(ustride, 1)` above

src/struct_mv/protos.h
```diff
@@ -72,6 +72,14 @@ HYPRE_Int hypre_AddIndexes ( hypre_Index index1, hypre_Index index2, HYPRE_Int n
 HYPRE_Int hypre_SubtractIndexes ( hypre_Index index1, hypre_Index index2, HYPRE_Int ndim,
                                   hypre_Index result );
 HYPRE_Int hypre_IndexesEqual ( hypre_Index index1, hypre_Index index2, HYPRE_Int ndim );
+HYPRE_Int
+hypre_IndexesGreater( hypre_Index  index1,
```
victorapm, line 76: Very minor comment: I was wondering if we should have the return data type (`HYPRE_Int`) in the same line as the function name as we have in the other prototypes

src/struct_ls/smg_residual_unrolled.c
rfalgout: I agree with your suggestions, but SMG needs more of an overhaul first.  This file, for example, will probably disappear altogether in favor of a matvec.  I've generally made changes to SMG where absolutely necessary, though I can't promise that's always been the case. :)

src/struct_ls/HYPRE_struct_pcg.c
rfalgout: Yes, it should.  I was being somewhat lazy.  This is still a work in progress and I would prefer to make this change more globally and all at once.  Let me see if that involves too many changes to justify doing in this PR.  Thanks!

src/struct_mv/protos.h
rfalgout: Actually, I was wondering if we should make our lives easier and just use copy-paste from now on in protos.h. :)  The reason they were one-liners before is historical - we used to have a script that automatically generated the prototypes from the C-files.  Now that we are doing it by hand (which we've been doing for a while, of course), it probably makes more sense to make it simple for developers.  These are internal to hypre, so the format choice we make only affects developers.

src/struct_mv/struct_copy.c
rfalgout: Probably.  We had a discussion about this in our Dec 11 SLS meeting and came up with a plan of action.  I think we should do that first, then make all of the changes in one PR.

src/struct_mv/struct_matrix.c
```diff
@@ -15,22 +15,50 @@
 #include "_hypre_struct_mv.hpp"
 
 /*--------------------------------------------------------------------------
- * Matrix data is currently stored relative to the largest matrix stride
  *--------------------------------------------------------------------------*/
 
 HYPRE_Int
-hypre_StructMatrixGetDataMapStride( hypre_StructMatrix *matrix,
-                                    hypre_IndexRef     *stride )
+hypre_StructMatrixGetFStride( hypre_StructMatrix *matrix,
```
victorapm, line 21: Very minor: I'm also wondering if we should spell this name out for clarity: `hypre_StructMatrixGetFineStride` Same for coarse

rfalgout: Thanks again for all the suggestions, and sorry to push back on so many of them, this one included. :(  The issue here is that I'm not sure yet if the fstride/cstride terminology is general enough to support what we want to do in the future.  It may be that this should be something like "stencil stride" (the stride defining the index space where the stencil is applied) and "data stride" (the stride defining the index space of the data).  Or something else.  This will require a little more thought and maybe some more code to flesh things out.  So, let's leave it for now.  It's only used in two files at the moment (ignoring prototype files).

src/struct_mv/struct_vector.c
```diff
@@ -314,19 +306,86 @@ hypre_StructVectorComputeDataSpace( hypre_StructVector *vector,
...
+   if (cdata_space == NULL)
+   {
+      /* resize if no current data space (cdata_space) */
+      need_resize = 1;
+   }
+   else if ( !hypre_IndexesEqual(stride, sstride, ndim) )
+   {
+      /* resize if stride and saved stride are different */
+      need_resize = 1;
+   }
```
waynemitchell, line 379: Why do we need a resize if the stride and saved strides differ?

rfalgout: Maybe we don't need it, since the next else clause will catch the case where the number of boxes is different.  Maybe I should give that a try.

rfalgout: The regression tests don't run correctly without this check.  I don't remember why it's needed, but it looks like it was there for a reason.  I'll let you know if it comes to me.

rfalgout:
Just wanted to point out that I did a quick update of the checktest.sh script.  Somehow the -summary feature was dropped and I couldn't even find a hash for that code in the git logs.  Weird.  It's a simple fix, fortunately.

waynemitchell (approved):
I think this all looks good. I'm curious about my one question above, but otherwise I think I understand most of what's going on here. Thanks for figuring this all out, Rob!

src/struct_mv/struct_vector.c
rfalgout: Okay, I remember.  The rebase was originally there to ensure that the base grid for x was the same as the one for A.  Now we ensure that it is "at least as fine" as the one for A (possibly finer).  There is logic in the code that assumes this is true, most importantly in the routine for computing the communication package information.  So, when the strides differ (because of the rebase), the resize is needed to finalize the rebase.  Otherwise, it just gets undone, and the above assumption no longer holds, and other stuff breaks.  It may be possible to remove this requirement, but it would require additional code changes at a minimum.

waynemitchell: I'm still not sure I'm seeing the case that gets missed if you don't do this check. In such a case, the strides differ, but the `BoxArrayInBoxArray` check is true below (so the code doesn't flag a need for resize). So there has been a rebase, but the vector data space is still has the same number of boxes and is contained within the `data_space` argument passed in? When does this happen? 

In other words, I don't see why the else clause below isn't sufficient (as you hinted at above).

rfalgout: Interpolation is the case to consider.  Here, the vector x may have a coarse base grid and a data space with enough ghost zones to support interpolation.  The rebase changes the base grid to a finer grid, so now the strides differ.  If the number of boxes is the same in both base grids and we ignore the stride differences as being a reason for doing a resize, then the BoxArrayInBoxArray check is okay and the resize never gets done.  The result is that x will have a base grid that is not at least as fine as the base grid for A, and that's an assumption that other parts of the code depend on.  Does this make sense, or am I misunderstanding what you are saying?

Note that currently, if we don't do a resize, then the vector is restored and that reverses the rebase.  We probably could rework the code to allow the rebase to stick but still avoid the copy in the resize.  That's something to keep in mind and think about a bit more.  I had considered the idea of doing some of this checking to avoid copies inside the resize itself, but decided that we wanted to have more explicit control over what resize does.  So, maybe we just need to extend its functionality (e.g., resize-without-copy), or expand on the else statement here to do a Forget() in this case, or something like that:

https://github.com/hypre-space/hypre/blob/e985f89ccd1f49d5dd8417c6136b0e01bef931fb/src/struct_mv/struct_matvec.c#L168-L172

victorapm (approved):
Thank you for all the work, Rob, and sorry it took me some time to go over everything. Thanks!

src/struct_mv/protos.h
victorapm: That makes sense! I guess bringing back the script would be even better :)

PR #1200, opened 2025-01-01
expose HYPRE_FMANGLE etc. to CMake

JakeOShannessy:
`HYPRE_FMANGLE` controls how symbols for Fortran are mangled. This can be changed via the configure script, but could not be configured during a CMake build. This change adds 3 options to CMakeLists.txt exposing `HYPRE_FMANGLE`, `HYPRE_FMANGLE_BLAS`, and `HYPRE_FMANGLE_LAPACK` such that they can be configured during a CMake build.


victorapm:
@JakeOShannessy thanks for your contribution. I will make a few other changes and include this in PR #1181, which involves several cmake updates to hypre.

PR #1202, opened 2025-01-07
Mixed Int Sstruct

waynemitchell:
Some fixes for the sstruct interface using the mixed int build.


victorapm (approved):
Thanks for the fixes!

rfalgout (approved):
Looks great!  Thanks @waynemitchell !

PR #1207, opened 2025-01-17
Check for HIP min. required version

victorapm:
Adds a check to cmake build system for the minimum HIP version we currently support (5.2.0).


rfalgout (approved):
Looks good to me

PR #1234, opened 2025-02-10
Strength matrix

rfalgout:
This branch implements a new strength measure idea.


rfalgout:
Hi @oseikuffuor1 and @waynemitchell .  This branch is now up to date with master.  I removed the cogent proxy from the Makefile (it doesn't build currently, and we probably don't want it in hypre anyway) and I added a one line description of the strength option to `ij.c` (it will need to be updated at some point).

The code doesn't work correctly for the matrices I am testing, but I'm pretty sure it's because the matrices are negative definite and the current code looks for positive off-diagonal entries instead of same-sign-as-diagonal.  I'll try to fix that.

oseikuffuor1:
Thanks Rob!

-Daniel





Daniel Osei-Kuffuor

Center for Applied Scientific Computing

Lawrence Livermore National Laboratory

Livermore, CA 94550

Mail Code: L-561

Phone: 925-422-7683

Email: ***@***.******@***.***>

rfalgout:
Okay, this seems to fix the negative diagonal case.  The strength measure doesn't really help with the matrices I am testing though. :(  This deserves a much closer investigation, of course.

PR #1236, opened 2025-02-19
Fix C++ dependency lib when compiling with IBM XL

victorapm:
This PR adds a condition to handle linking libraries for the IBM XL C++ compiler.
This allows using a C linker with hypre for GPU builds.


oseikuffuor1 (approved):
This works. Thanks @victorapm. I see stdc++ is already there so I'm not sure why gnu did not work before. Thanks again.

PR #1238, opened 2025-02-21
Fix unregistered mem path for HIP

### LINKED ISSUE #1138: Inconsistent behavior of hypre_GetPointerLocation
The function `hypre_GetPointerLocation` in `src/utilities/memory.c` behaves differently on CUDA and HIP builds. 

The functions `cudaPointerGetAttributes` and `hipPointerGetAttributes` are unreliable at identifying host pointers that weren't obtained from the CUDA/HIP host allocators respectively. For CUDA builds there are these lines in the above mentioned function:
```
if (attr.type == cudaMemoryTypeUnregistered)
{
   *memory_location = hypre_MEMORY_HOST;
}
```
A similar set of lines should be present for HIP builds, but are not.

Both CUDA and HIP builds will return `hypre_MEMORY_HOST` in the event of an error. However, given an arbitrary host pointer `hipPointerGetAttributes` will typically not error. Instead it will return a location of unregistered, and then `hypre_GetPointerLocation` will return `hypre_MEMORY_UNDEFINED` instead of `hypre_MEMORY_HOST`.

On debug builds with HIP this can result in a failed assertion in `hypre_CheckMemoryLocation`. I observed the following sequence of events recently:

1. Called `HYPRE_IJVectorAssemble`
2. That called `hypre_IJVectorAssembleParDevice`
3. That called `hypre_AuxParVectorDestroy(aux_vector)`
4. That deallocates the data fields inside the `aux_vector` struct correctly by specifying `HYPRE_MEMORY_DEVICE`
5. Deallocating the struct itself fails since it is of course on host, and the above assertion fails when `hypre_GetPointerLocation` can not figure out that the struct is indeed on host.

victorapm:
Closes #1138

### DIFF
--- a/src/utilities/memory.c
+++ b/src/utilities/memory.c
@@ -1193,10 +1193,12 @@ hypre_GetPointerLocation(const void *ptr, hypre_MemoryLocation *memory_location)
    }
 #if (HIP_VERSION_MAJOR >= 6)
    else if (attr.type == hipMemoryTypeUnregistered)
+#else
+   else
+#endif
    {
       *memory_location = hypre_MEMORY_HOST;
    }
-#endif
 #endif // defined(HYPRE_USING_HIP)
 
 #if defined(HYPRE_USING_SYCL)


liruipeng (approved):
Looks good. Thanks @victorapm

PR #1242, opened 2025-03-04
Thrust usage refactor for cuda 11+

victorapm:
This PR includes several changes to improve compatibility with different versions of the Thrust library and to replace deprecated Thrust functions as of cuda 12.8.

The most important changes include modifying the usage of `thrust::not1` to `HYPRE_THRUST_NOT`, and updating functor structures to handle different Thrust versions.

These changes remains compatible with older versions of thrust, including the one distributed with cuda 10.

### DIFF
--- a/src/parcsr_ls/par_mod_multi_interp_device.c
+++ b/src/parcsr_ls/par_mod_multi_interp_device.c
@@ -86,7 +86,11 @@ struct local_equal_plus_constant : public
 };
 
 /* transform from local C index to global C index */
+#if (defined(THRUST_VERSION) && THRUST_VERSION < 101000)
 struct globalC_functor : public thrust::unary_function<HYPRE_Int, HYPRE_BigInt>
+#else
+struct globalC_functor
+#endif
 {
    HYPRE_BigInt C_first;
 
@@ -330,7 +334,7 @@ hypre_BoomerAMGBuildModMultipassDevice( hypre_ParCSRMatrix  *A,
                                               thrust::make_counting_iterator(n_fine),
                                               CF_marker,
                                               points_left,
-                                              thrust::not1(equal<HYPRE_Int>(1)) );
+                                              HYPRE_THRUST_NOT(equal<HYPRE_Int>(1)) );
    remaining = points_end - points_left;
 
    /* Cpts; number of C pts */
@@ -544,7 +548,7 @@ hypre_BoomerAMGBuildModMultipassDevice( hypre_ParCSRMatrix  *A,
                                       points_left_old + remaining,
                                       diag_shifts,
                                       points_left,
-                                      thrust::not1(thrust::identity<HYPRE_Int>()) );
+                                      HYPRE_THRUST_NOT(thrust::identity<HYPRE_Int>()) );
 #endif
 
          hypre_assert(new_end - points_left == cnt_rem);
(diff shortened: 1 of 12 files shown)


src/parcsr_ls/par_mod_multi_interp_device.c
```diff
@@ -86,7 +86,11 @@ struct local_equal_plus_constant : public
 };
 
 /* transform from local C index to global C index */
+#if (defined(THRUST_VERSION) && THRUST_VERSION < 101000)
 struct globalC_functor : public thrust::unary_function<HYPRE_Int, HYPRE_BigInt>
+#else
+struct globalC_functor
```
liruipeng, line 92: The thrust in cuda 11+ doesn't support `public thrust::unary_function` anymore?

src/parcsr_mv/par_csr_fffc_device.c
```diff
@@ -21,10 +21,10 @@ typedef thrust::tuple<HYPRE_Int, HYPRE_Int> Tuple;
 /* transform from local F/C index to global F/C index,
  * where F index "x" are saved as "-x-1"
  */
-#if defined(HYPRE_USING_SYCL)
-struct FFFC_functor
-#else
+#if (defined(THRUST_VERSION) && THRUST_VERSION < 101000)
 struct FFFC_functor : public thrust::unary_function<Tuple, HYPRE_BigInt>
+#else
+struct FFFC_functor
```
liruipeng, line 27: the same question here.

src/parcsr_mv/par_csr_fffc_device.c
```diff
@@ -48,10 +48,10 @@ struct FFFC_functor : public thrust::unary_function<Tuple, HYPRE_BigInt>
 
 /* this predicate selects A^s_{FF} */
 template<typename T>
-#if defined(HYPRE_USING_SYCL)
-struct FF_pred
-#else
+#if (defined(THRUST_VERSION) && THRUST_VERSION < 101000)
 struct FF_pred : public thrust::unary_function<Tuple, bool>
+#else
+struct FF_pred
```
liruipeng, line 54: the same q.

rfalgout (approved):
Looks good to me, but I have no expertise on thrust.

liruipeng (approved):
Looks good to me. Thanks

src/parcsr_ls/par_mod_multi_interp_device.c
victorapm: Hi Rui Peng, It's been deprecated:

```
path-to-hypre/src/utilities/_hypre_utilities.hpp:1538:57: warning: ‘template<class Argument, class Result> struct thrust::THRUST_200700_860_NS::unary_function’ is deprecated [-Wdeprecated-declarations]
 1538 | struct type_cast : public thrust::unary_function<T1, T2>
      |                                                         ^             
path-to-spack/cuda-12.8.0-hd4fbybxbgvypwjgbekpfccmbevzr623/targets/x86_64-linux/include/thrust/functional.h:85:23: note: declared here
   85 | struct THRUST_DEPRECATED unary_function
      |                       ^~~~~~~~~~~~~~
```

liruipeng: OK. Thanks. It is frustrated that they keep changing things like this.

I "remembered" in the old version `: public thrust::unary_function<HYPRE_Int, HYPRE_BigInt>` is not actually necessary after `struct functor` but I could be wrong. If it is the case, we can remove it totally, to simplify the code. But I am happy with the way it is now. Thanks for the work.

victorapm: Sounds good! If we decide to drop cuda 10, then we can remove support to `unary_function` in hypre

PR #1243, opened 2025-03-07
Fix memory leak affecting ext+e interpolation in AMG

victorapm:
This PR includes a memory leak fix to the `hypre_BoomerAMGBuildModExtPEInterpHost` function. The change adds a call to free the `dof_func_offd` memory.

### DIFF
--- a/AUTOTEST/machine-aurora.sh
+++ b/AUTOTEST/machine-aurora.sh
@@ -57,14 +57,14 @@ save="aurora"
 
 # SYCL with UM in debug mode [ij, struct]
 # WM: I suppress all warnings for sycl files for now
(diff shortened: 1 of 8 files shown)


rfalgout (approved):
Thanks, Victor!  Do we need to add a run to our regression tests to ensure this gets tested with valgrind?  I assume we weren't doing any runs before that used this particular code path.

victorapm:
Hi Rob, yes, that's a good idea. Let me do it

victorapm:
@rfalgout I've added a regression test for case showing the leak (`interptype 18`) and also for other interpolation types just for completeness. 

@waynemitchell could you generate the baselines in aurora? I don't have access

victorapm:
@waynemitchell could you run this on aurora and add the results for the new tests? Thank you!

waynemitchell (approved):
I went ahead and ran the full regression suite on Aurora, and everything looks good. Thanks, Victor!

PR #1247, opened 2025-03-12
Add special case for RAP in MGR

victorapm:
This PR introduces a new condition in MGR's coarse matrix computation to handle cases where the restriction operator is equal to the injection operator.

Should be merged after #1243

### DIFF
--- a/src/parcsr_ls/par_mgr_rap.c
+++ b/src/parcsr_ls/par_mgr_rap.c
@@ -233,9 +233,9 @@ hypre_MGRBuildNonGalerkinCoarseOperatorHost(hypre_ParCSRMatrix    *A_FF,
    hypre_ParCSRMatrix    *minus_Wp = NULL;
 
    HYPRE_Int              blk_inv_size;
-   HYPRE_Real             neg_one = -1.0;
-   HYPRE_Real             one = 1.0;
-   HYPRE_Real             beta = neg_one;
+   HYPRE_Complex          neg_one = -1.0;
+   HYPRE_Complex          one = 1.0;
+   HYPRE_Complex          beta = 0.0;
 
    if (Wp != NULL && max_elmts > 0)
    {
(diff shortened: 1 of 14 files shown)


src/parcsr_ls/par_mgr_rap.c
```diff
@@ -587,6 +587,14 @@ hypre_MGRBuildCoarseOperator(void                *mgr_vdata,
          hypre_ParCSRMatrixAdd(1.0, A_CC, -1.0, RAP_c, &RAP);
          hypre_ParCSRMatrixDestroy(RAP_c);
       }
+      else if (!Wr && Wp)
+      {
+         /* Prolongation is not the injection operator (Wp != NULL) and
+            Restriction is the injection operator (Wr == NULL) */
+         RAP_c = hypre_ParCSRMatMat(A_CF, Wp);
```
oseikuffuor1, line 611: @victorapm Wp here is not negated already right?

oseikuffuor1 (approved):

oseikuffuor1 (approved):
Thanks @victorapm

src/parcsr_ls/par_mgr_rap.c
victorapm: Good point, right!

victorapm: For reference: we found different definitions of Wp and Wr throughout MGR's code (without the negative sign and with the negative sign). We decided to use a single convention that considers the negative sign already in Wp and Wr, so that prolongation is computed as `[Wp I]^T` and restriction as `[Wr I]`. The unification work was done in [89c2d0e](https://github.com/hypre-space/hypre/pull/1247/commits/89c2d0ef0e2112a644aef2842c1649f08d9dbc8e)

rfalgout (approved):
Looks good to me.  Thanks!

PR #1249, opened 2025-03-21
Recmat - add support for symmetric matrices in Matmult

rfalgout:
This adds support for symmetric matrices in the StructMatmult routine and adds optimizations to PFMG.

- Adjusts communication for symmetric storage
- Automatically determines if the product matrix in matmult is symmetric
- Computes only the stored symmetric coefficients in matmult (the "lower triangle" so to speak)
- Modifies the PFMG setup to reduce memory copies in the matmult calls
- Cleans up the kernel optimization code in matmult.  Using a mask (not a bit mask) simplifies things.
- Makes a better decision about when a mask is needed in matmult.  Now, the most common RAP scenario in PFMG does not need a mask, while it did before because of the constant diagonal in P.  This speeds up the kernel a little bit.
- Computes a better data space for the matrices involved in the matrix product.
- Tries to avoid unnecessary matrix resizes in matmult.  Now, PFMG often does not require a resize-with-copy.
- Adds loop unrolling optimizations to matvec for rectangular matrices (previously only square matrices were unrolled).

Some regression tests needed to be adjusted:
- `emptyProc` and `solvers` changed because of having no mask in matmult.  The input files for these tests were not defined correctly because they didn't zero out coefficients along boundaries.  The mask was originally hiding this fact.  With the boundary condition fixes, any solver involving PFMG still had small differences because the interpolation changes near the boundaries.
- `amr2d` changed because there was apparently a bug somewhere in the original `recmat` code.  If you use the `generic` matmult routine on `recmat` you get the same answer as here on `recmat-symmetric`.
- `nonmixedint` was not broken, but it was updated to use `sstruct.in.cube` instead of `sstruct.in.emptyProc` which was removed because it was redundant.


rfalgout:
I am going to do a few more updates to the code and run a few timing tests on tux, but this is now ready for review.

rfalgout:
Hi @victorapm and @waynemitchell .  The work here is done except for having to sync up with recmat when the time comes.  So feel free to start looking through this.  Thanks!

src/struct_ls/pfmg_setup.c
```diff
@@ -188,33 +201,39 @@ hypre_PFMGSetup( void               *pfmg_vdata,
...
-         hypre_zPFMGSetupRestrictOp(RT_l[l], A_l[l], cdir);
-      }
-#endif
-      HYPRE_ANNOTATE_REGION_END("%s", "Interpolation");
 
-      HYPRE_ANNOTATE_REGION_BEGIN("%s", "RAP");
       if (rap_type == 0)
       {
+         hypre_StructMatrixInitializeShell(P_l[l]);
+#if 0 /* TODO: Allow RT != P */
```
waynemitchell, line 196: Just curious: what else needs to be done to allow `RT != P`? Also, is there a distinction between the `nonsymmetric_cycle` condition here and the `(RT_l[l] != P_l[l])` condition below?

src/struct_mv/struct_matmult.c
```diff
@@ -1302,19 +1409,16 @@ hypre_StructMatmultCompute( hypre_StructMatmultData  *mmdata,
       switch (nterms)
       {
          case 2:
```
waynemitchell, line 1414: Add `HYPRE_FALLTHROUGH` here after merging the most recent `recmat` changes from Victor?

waynemitchell (approved):
Looks good to me. Thanks for all the work on this, Rob!

src/struct_ls/pfmg_setup.c
rfalgout: This is all placeholder code, though the `(RT_l[l] != P_l[l])` code is runnable now if this condition were ever to be true.  It's currently all intended to support a future nonsymmetric option for PFMG.  The main thing that would need to happen would be to add the user option flag and implement the `Create/SetupRestrictOp()` routines which would be similar to the ones for interpolation but based on A^T.

src/struct_mv/struct_matmult.c
rfalgout: Will do.  Good catch!

src/struct_mv/struct_matvec.c
```diff
@@ -336,24 +334,18 @@ hypre_StructMatvecCompute( void               *matvec_vdata,
    hypre_IndexRef           start;
 
    HYPRE_Complex            temp;
-   HYPRE_Int                compute_i, i, j, k, si, ssi;
-   HYPRE_Int                csi[UNROLL_MAXDEPTH], vsi[UNROLL_MAXDEPTH];
+   HYPRE_Int                compute_i, i, j, si, se;
+   HYPRE_Int                cnentries, vnentries;
+   HYPRE_Int                centries[nentries], ventries[nentries];
```
victorapm, line 342: Should we consider allocating `centries` and `ventries` on the heap instead of using VLAs?

src/struct_mv/struct_matvec_core.c
```diff
@@ -0,0 +1,634 @@
...
+   for (si = 0; si < nentries; si += UNROLL_MAXDEPTH)
+   {
+      depth = hypre_min(UNROLL_MAXDEPTH, (nentries - si));
+
+      if (!transpose)
+      {
+         switch (depth)
+         {
+            case 9:
+               HYPRE_SET_CAX(Ap8, xoff8, entries[si + 8]);
```
victorapm, line 121:
```suggestion
               HYPRE_SET_CAX(Ap8, xoff8, entries[si + 8]);
               HYPRE_FALLTHROUGH;
```

Should we do the same for the other occurrences in this file as well?

rfalgout: I can't implement FALLTHROUGH until I pull in the recmat changes, but I wanted to wait until the GPU work was all integrated (or mostly integrated).  If you think it's okay to do the merge now, let me know.

victorapm (approved):
Thank you, Rob! I liked the comments you added throughout the code

src/struct_mv/struct_matvec.c
rfalgout: The only issue I can see from googling is that VLAs (as written here) are allocated on the stack which is easier to overflow.  Otherwise, VLAs are much faster and produce more readable code, IMO.  Let me do more research on this first before deciding.  Thanks!

src/struct_mv/struct_matvec_core.c
rfalgout: This is done now

rfalgout:
@victorapm and @waynemitchell .  I have merged recmat into this branch and I am running the regression tests.  I forgot to comment out the code that causes some of the sstruct tests to fail, but I can do that later.

@victorapm .  Most of the conflicts were in `sstruct_matmult.c`, `struct_matmult.c`, and `struct_matvec.c`.  Now that the main kernels are in separate `core` files, you may have to fix some GPU-related stuff that got left out.  Let me know if you want to do that now, on this branch, or wait until after I merge into recmat.

victorapm:
Thanks, Rob! I can do it on another branch if you agree. I plan to include also the remaining changes involving UVM dependency. Thanks!

src/struct_mv/struct_matvec.c
rfalgout: I will think about this more and consider changing later.  I think it's more important to merge this into recmat right now.  Thanks again!

PR #1251, opened 2025-03-25
Recmat partial device support

victorapm:
This PR adds device support to PFMG in recmat. This requires hypre to be configured with UVM support


src/IJ_mv/HYPRE_IJMatrix.c
```diff
@@ -1011,7 +1011,14 @@ HYPRE_IJMatrixGetValuesAndZeroOut( HYPRE_IJMatrix matrix,
 
    if (exec == HYPRE_EXEC_DEVICE)
    {
-      hypre_error_w_msg(HYPRE_ERROR_GENERIC, "HYPRE_IJMatrixGetValues not implemented for GPUs!");
+      /* TODO: add device implementation */
```
rfalgout: Do the non-sstruct changes here (IJ, parcsr, etc.) match what's on master?  Just curious.

src/sstruct_mv/sstruct_matrix.c
```diff
@@ -823,10 +823,12 @@ hypre_SStructPMatrixGetDiagonal( hypre_SStructPMatrix  *pmatrix,
  *==========================================================================*/
 
 /*--------------------------------------------------------------------------
+ * hypre_SStructUMatrixInitialize
  *--------------------------------------------------------------------------*/
 
 HYPRE_Int
-hypre_SStructUMatrixInitialize( hypre_SStructMatrix *matrix )
+hypre_SStructUMatrixInitialize( hypre_SStructMatrix  *matrix,
```
rfalgout, line 830: Just wondering if memory_location should be set through a 'Set' function instead.  We should parallel what is done in parcsr, and maybe that's what you've already done here.

src/struct_ls/pfmg_setup.c
```diff
@@ -187,7 +188,7 @@ hypre_PFMGSetup( void               *pfmg_vdata,
          RT_l[l] = hypre_zPFMGCreateRestrictOp(A_l[l], cdir, stride);
       }
 #endif
-      HYPRE_StructMatrixSetTranspose(RT_l[l], 1);
+      hypre_StructMatrixSetTranspose(RT_l[l], 1, &resize);
```
rfalgout: This isn't necessary here because there is no data allocated yet, i.e., there is nothing to resize.

src/struct_ls/red_black_constantcoef_gs.c
```diff
@@ -316,7 +316,7 @@ hypre_RedBlackConstantCoefGS( void               *relax_vdata,
                                 x_dbox, stencil_shape[offd[5]]);
                      xoff4 = hypre_BoxOffsetDistance(
                                 x_dbox, stencil_shape[offd[4]]);
-                  /* fall through */
+                     HYPRE_FALLTHROUGH;
```
rfalgout, line 319: What is this?  And why?

rfalgout (approved):
Looks good to me.  Just minor comments/questions.

src/struct_ls/red_black_constantcoef_gs.c
victorapm: This informs the compiler that it's safe to fall through the next case statement. By default, modern compilers assume that's not safe, thus producing a compilation warning. So with this macro, we don't get the warning

rfalgout: Okay, so it's just to avoid compiler complaints.  Thanks!

src/IJ_mv/HYPRE_IJMatrix.c
victorapm: They don't match, I can create a separate PR with the non-struct changes to be merged on master

src/sstruct_mv/sstruct_matrix.c
victorapm: In parcsr, we have the "_v2" functions that take the memory location as input argument. I think having a separate SetMemoryLocation is a good idea

src/sstruct_ls/ssamg_interp.c
```diff
@@ -264,239 +261,138 @@ hypre_SSAMGSetupInterpOp( hypre_SStructMatrix  *A,
...
-                  }
-                  else
-                  {
-                     ventries[nventries++] = si;
-                  }
-               }
-
-               /* Get stencil space on base grid for entry 1 of P (valid also for entry 2) */
-               hypre_StructMatrixGetStencilSpace(P_s, 1, 0, origin, stride);
+            continue;
```
waynemitchell, line 284: Is there a reason you reorganized this in this way? I am not crazy about having the `Assemble` and `ClearBoundary` stuff repeated here and below and then using `continue` to skip over the loop. For me, it's less clear and causes unnecessary code replication.

src/sstruct_mv/HYPRE_sstruct_matrix.c
```diff
@@ -425,7 +425,7 @@ HYPRE_SStructMatrixInitialize( HYPRE_SStructMatrix matrix )
    iupper--; jupper--;
    HYPRE_IJMatrixCreate(comm, ilower, iupper, jlower, jupper,
                         &hypre_SStructMatrixIJMatrix(matrix));
-   hypre_SStructUMatrixInitialize(matrix);
+   hypre_SStructUMatrixInitialize(matrix, HYPRE_MEMORY_DEVICE);
```
waynemitchell, line 428: Should the sstruct matrix data structure have a `MemoryLocation` similar to what we do in parcsr or struct? Then we should initialize all the struct/parcsr data structures underneath a sstruct matrix in that memory location (rather than hard-coding to device). I guess this will require a lot more code changes, so maybe that's best left to a future PR.

src/sstruct_ls/ssamg_interp.c
```diff
@@ -264,239 +261,138 @@ hypre_SSAMGSetupInterpOp( hypre_SStructMatrix  *A,
...
+         }
 
-               hypre_CopyToIndex(stride, ndim, Astride);
-               hypre_StructMatrixMapDataStride(A_s, Astride);
-               hypre_CopyToIndex(stride, ndim, Pstride);
-               hypre_StructMatrixMapDataStride(P_s, Pstride);
+         if (hypre_StructMatrixConstEntry(P_s, 1))
+         {
+            /* Off-diagonal entries are constant */
+            hypre_TMemcpy(hypre_StructMatrixConstData(P_s, 1), &half, HYPRE_Complex, 2,
```
waynemitchell: Is this correct? `half` is just a single value, but this is trying to copy a memory block of size 2?

victorapm: Thanks for catching!

src/sstruct_ls/ssamg_interp.c
victorapm: The reason was to remove one level of indentation. But I'm happy to revert

src/sstruct_mv/HYPRE_sstruct_matrix.c
victorapm: I think that's a good idea. @rfalgout do you agree?

rfalgout: Off the top of my head, yes, I agree.  I also agree that it might be left for later since I have been making lots of changes on recmat for matmult that will have to be merged with your work.

rfalgout: If, however, the change is easy, then feel free.  Thanks!

src/struct_ls/pfmg_setup.c
victorapm: The reason for this change is to use the private `hypre_` API instead of the public one `HYPRE_`. Should I revert?

src/sstruct_mv/HYPRE_sstruct_matrix.c
victorapm: I was checking and it involves changing some files that we both touched. Let's plan to do it after both our PRs are merged

src/struct_ls/pfmg_setup.c
rfalgout: Yes, I would revert.  The internal API provides the extra resize feedback (rarely needed), but you don't need it here and it confused me. :)  I realize one just calls the other but...

victorapm: Done, thanks!

src/sstruct_ls/ssamg_interp.c
waynemitchell: OK, well I guess I don't have a strong opinion either way. Up to you.

src/IJ_mv/HYPRE_IJMatrix.c
victorapm: Created PR #1263 for these changes

PR #1252, opened 2025-03-25
Fix parallel implementation of `hypre_ParCSRMatrixBlkFilterHost`

victorapm:
Bug fix involving the calculation of `col_map_offd`

TODO: another PR will be opened to fix the device implementation


rfalgout (approved):
You should ask someone else who is familiar with this part of the code to review.  Thanks!

victorapm:
Hi @waynemitchell could you take a look at this PR?

src/parcsr_mv/par_csr_filter.c
```diff
@@ -102,11 +104,14 @@ hypre_ParCSRMatrixBlkFilterHost( hypre_ParCSRMatrix  *A,
          }
       }
 
-      for (j = A_offd_i[i]; j < A_offd_i[i + 1]; j++)
+      if (A_offd_i[num_rows] - A_offd_i[0] > 0)
```
waynemitchell, line 107: This check seems redundant to me? What's the reason for it?

src/parcsr_mv/par_csr_filter.c
```diff
@@ -157,27 +162,48 @@ hypre_ParCSRMatrixBlkFilterHost( hypre_ParCSRMatrix  *A,
...
-            marker[A_offd_j[j]] = 1;
          }
       }
    }
 
-   /* Update col_map array */
-   num_cols_offd_B = 0;
-   for (i = 0; i < num_cols_offd_A; i++)
+   /* Allocate work array */
+   work = hypre_TAlloc(HYPRE_BigInt, B_offd_nnz, HYPRE_MEMORY_HOST);
```
waynemitchell, line 176: Why do you need an extra work array? Can you just copy directly into `col_map_offd_B` and work with that array below?

src/parcsr_mv/par_csr_filter.c
victorapm: This is important for skipping the for-loop in case we don't have external connections in a given rank

src/parcsr_mv/par_csr_filter.c
victorapm: I need the extra work array because the binary search below works on `B_offd_bj`

victorapm: just adding to the previous response: the work array is sorting out `B_offd_bj` in order to build col_map_offd. Unfortunately, we can't copy directly into col_map_offd without sorting first

waynemitchell: Sorry, I'm probably just missing something here... Why do you have to sort before copying into `col_map_offd_B`? It looks like `col_map_offd_B` is not used until after you copy the `work` array into it. So can't you just copy `B_offd_bj` into `col_map_offd_B`, then do the sorting and removal of duplicates in place in `col_map_offd_B`?

src/parcsr_mv/par_csr_filter.c
```diff
@@ -157,27 +162,48 @@ hypre_ParCSRMatrixBlkFilterHost( hypre_ParCSRMatrix  *A,
...
       {
-         col_map_offd_B[num_cols_offd_B++] = col_map_offd_A[i];
+         work[num_cols_offd_B++] = work[i + 1];
       }
    }
+
+   /* Build B's col_map array */
+   for (i = 0; i < num_cols_offd_B; i++)
+   {
+      col_map_offd_B[i] = work[i];
```
waynemitchell, line 194: Why a for loop instead of just a memcpy?

src/parcsr_mv/par_csr_filter.c
victorapm: No worries! I need to sort *and remove duplicate columns* before copying because the col_map_offd array gives a mapping for the compressed offd matrix. I can't copy `B_offd_bj` into `col_map_offd` because the first generally has more entries than the second one.

There's a similar code in the matrix assembly function: https://github.com/hypre-space/hypre/blob/79ab05654eacd34acd7c07e0a2d58f46008e0b43/src/IJ_mv/IJMatrix_petsc.c#L885-L914

src/parcsr_mv/par_csr_filter.c
victorapm: That would work as well! I picked the for loop option because `num_cols_offd_B` is generally small and the compiler can do efficient loop unrolling strategies that might be faster than memcpy. But the difference is very minimal to be honest...

src/parcsr_mv/par_csr_filter.c
victorapm: @waynemitchell not sure if my previous answer was clear. Please, let me know if I'm missing something

waynemitchell: @victorapm , I see it now. I had it in my head that `col_map_offd` was the same allocation size as these other arrays. Thanks for explaining!

src/parcsr_mv/par_csr_filter.c
waynemitchell: I didn't realize a for loop might actually be faster than memcpy! Anyway, doesn't matter to me either way, I guess, you can do either here.

waynemitchell (approved):

src/parcsr_mv/par_csr_filter.c
victorapm: You're welcome, thanks for looking carefully at this!

PR #1256, opened 2025-03-29
Extract raw pointer from accessor in SYCL build

victorapm:
This PR modifies the `HYPRE_GPU_LAUNCH2` macro to explicitly extract the raw pointer from the SYCL local memory accessor before passing it to the kernel. This change prevents the following deprecated conversion warning in SYCL 2020:

```
warning: 'operator __local char *' is deprecated: Conversion to pointer type is deprecated since SYCL 2020. Please use get() instead.
```

### DIFF
--- a/src/utilities/_hypre_utilities.hpp
+++ b/src/utilities/_hypre_utilities.hpp
@@ -522,10 +522,11 @@ using hypre_DeviceItem = sycl::nd_item<3>;
       hypre_HandleComputeStream(hypre_handle())->submit([&] (sycl::handler& cgh) {           \
          sycl::range<1> shmem_range(shmem_size);                                             \
          sycl::local_accessor<char, 1> shmem_accessor(shmem_range, cgh);                     \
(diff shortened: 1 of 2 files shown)


rfalgout (approved):

waynemitchell:
@victorapm , I'm getting some errors trying to run this branch on Aurora. Let me do some investigation and get back to you.

waynemitchell:
Yeah, apparently you cannot call `get_multi_ptr` from the host for a local accessor. I get the following error message:

terminate called after throwing an instance of 'sycl::_V1::exception'
  what():  get_multi_ptr must not be called on the host for a local accessorterminate called after throwing an instance of 'sycl::_V1::exception'
  what():  get_multi_ptr must not be called on the host for a local accessor

I modified the code accordingly. I think it should now work and also avoid the warning (I'm testing one more time).

waynemitchell (approved):

PR #1263, opened 2025-04-08
Non-struct code refactoring and fixes

victorapm:
Non-struct code changes from #1251 

- Fix implementation of hypre_ParCSRMatrixSetNumRownnz
- Add copy assignment operator to ReduceSum struct
- Fix fallthrough compiler warnings
- Add temporary implementation for `HYPRE_IJMatrixGetValuesAndZeroOut`

### DIFF
--- a/src/IJ_mv/HYPRE_IJMatrix.c
+++ b/src/IJ_mv/HYPRE_IJMatrix.c
@@ -1012,6 +1012,8 @@ HYPRE_IJMatrixGetValuesAndZeroOut( HYPRE_IJMatrix matrix,
    if (exec == HYPRE_EXEC_DEVICE)
    {
       /* TODO: add device implementation */
+      hypre_error_w_msg(HYPRE_ERROR_GENERIC, "HYPRE_IJMatrixGetValuesAndZeroOut not implemented for GPUs!");
+
       if ( hypre_IJMatrixObjectType(ijmatrix) == HYPRE_PARCSR )
       {
          hypre_IJMatrixMigrateParCSR(ijmatrix, HYPRE_MEMORY_HOST);
--- a/src/IJ_mv/HYPRE_IJMatrix.c
+++ b/src/IJ_mv/HYPRE_IJMatrix.c
@@ -1011,7 +1011,14 @@ HYPRE_IJMatrixGetValuesAndZeroOut( HYPRE_IJMatrix matrix,
 
    if (exec == HYPRE_EXEC_DEVICE)
    {
-      hypre_error_w_msg(HYPRE_ERROR_GENERIC, "HYPRE_IJMatrixGetValues not implemented for GPUs!");
+      /* TODO: add device implementation */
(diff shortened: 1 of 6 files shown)


rfalgout (approved):
Looks good to me.  Maybe ask Rui Peng or Wayne to take a look also.  Thanks!

src/IJ_mv/HYPRE_IJMatrix.c
```diff
@@ -1011,7 +1011,14 @@ HYPRE_IJMatrixGetValuesAndZeroOut( HYPRE_IJMatrix matrix,
...
    {
-      hypre_error_w_msg(HYPRE_ERROR_GENERIC, "HYPRE_IJMatrixGetValues not implemented for GPUs!");
+      /* TODO: add device implementation */
+      if ( hypre_IJMatrixObjectType(ijmatrix) == HYPRE_PARCSR )
+      {
+         hypre_IJMatrixMigrateParCSR(ijmatrix, HYPRE_MEMORY_HOST);
+         hypre_IJMatrixGetValuesParCSR(ijmatrix, nrows, ncols,
+                                       rows, row_indexes, cols, values, 1);
+         hypre_IJMatrixMigrateParCSR(ijmatrix, HYPRE_MEMORY_DEVICE);
+      }
```
waynemitchell, line 1023: I think this is fine as a temporary implementation, but we might want to throw a warning or something here. Particularly, since this is calling down to the CPU version of `hypre_IJMatrixGetValuesParCSR()`, if `values` is an array in device memory, this will cause an error.

src/parcsr_mv/par_csr_matrix.c
```diff
@@ -423,25 +423,34 @@ hypre_ParCSRMatrixSetNumRownnz( hypre_ParCSRMatrix *matrix )
...
+   local_num_rownnz = 0;
+   if (rownnz_diag && !rownnz_offd)
    {
-      local_num_rownnz++;
-      if (rownnz_diag[i] < rownnz_offd[j])
-      {
-         i++;
-      }
-      else
+      local_num_rownnz = num_rownnz_diag;
```
waynemitchell, line 438: `local_num_rownnz` is overwritten below (see line 457). Particularly, with your initialization of `i` and `j` above, I don't think it will end up with the right value?

src/IJ_mv/HYPRE_IJMatrix.c
victorapm: Good idea! Added a warning

src/parcsr_mv/par_csr_matrix.c
victorapm: The idea here is that `(num_rownnz_diag - i) + (num_rownnz_offd - j)` is zero, so line 457 doesn't change `local_num_rownnz`

waynemitchell: Ah, I see the `+=` on line 457 now... I had read it as `=`. Thanks!

waynemitchell (approved):

PR #1272, opened 2025-04-22
Device porting for struct and sstruct 

victorapm:
- Remove UVM dependency from PFMG's interpolation code
- Update struct's communication package with device support (on-going)
- Create dedicate file for PFMG coarsening and remove UVM dependency
- Rename `hypre_PFMGZeroDiagonal` to `hypre_StructMatrixZeroDiagonal`
- Rename `hypre_InitializeCommunication` to `hypre_StructCommunicationInitialize`
- Rename `hypre_FinalizeCommunication` to `hypre_StructCommunicationFinalize`
- Add `hypre_StructMatrixSetMemoryLocation` routine

Comes after #1251


src/sstruct_ls/fac_restrict2.c
rfalgout, line 1: Remember that we are planning to remove FAC, so if you find yourself needing more extensive changes than these, it's probably not worth it.

src/struct_mv/struct_communication.c
```diff
@@ -659,7 +665,9 @@ hypre_CommPkgCreate( hypre_CommInfo   *comm_info,
 }
 
 /*--------------------------------------------------------------------------
- * Note that this routine uses an identity coordinate transform
+ * hypre_CommBlockSetEntries
```
rfalgout: Note that I have been removing the extra function name in the comments because it's redundant and I've found it to be incorrect at times.  So, we are working against each other with this. :)  Not a big deal, but I thought I'd mention it.

rfalgout (approved):
Looks good to me.  Thanks!

src/sstruct_ls/fac_restrict2.c
victorapm: That's a good point, thanks for bringing it up!

src/struct_mv/struct_communication.c
victorapm: Thanks for mentioning! I wasn't aware of your effort. I will refrain from adding comment entries like this

src/struct_ls/pfmg.h
```diff
@@ -17,6 +17,18 @@
 //#define DEBUG_SETUP
 //#define DEBUG_SOLVE
 
+#define hypre_PFMGSetCIndex(cdir, cindex)       \
+   {                                            \
+      hypre_SetIndex(cindex, 0);                \
+      hypre_IndexD(cindex, cdir) = 0;           \
+   }
```
waynemitchell, line 24: I see this is just copy/paste (so no code changes), but I'm kind of curious about this function. It looks like it just sets the index to 0?

waynemitchell (approved):
I didn't go through all of this in great detail, but this looks good to me, and I don't want to hold you up. Thanks, Victor!

PR #1279, opened 2025-05-04
Make HYPRE_StructVectorSetValues signature consistent

gartavanis:
HYPRE_StructVectorSetValues takes a "values" argument, but similar functions like HYPRE_StructVectorSetBoxValues, HYPRE_StructMatrixSetValues etc., take a pointer argument instead (*values). This becomes a problem when using HYPRE_StructVectorSetValues on GPUs, as internally, the function expects a pointer to device memory (otherwise an assert is triggered). We've encountered this problem when trying to integrate HYPRE with the GPU backend for ParFlow.

This PR fixes this by making the signature consistent with the other functions. Alternatively, if no changes to the API are desired in order to maintain backwards compatibility, we could have different function, e.g. HYPRE_StructVectorSetValuesPtr, with the new pointer argument.

### DIFF
--- a/src/struct_mv/HYPRE_struct_mv.h
+++ b/src/struct_mv/HYPRE_struct_mv.h
@@ -420,7 +420,7 @@ HYPRE_Int HYPRE_StructVectorInitialize(HYPRE_StructVector vector);
  **/
 HYPRE_Int HYPRE_StructVectorSetValues(HYPRE_StructVector  vector,
                                       HYPRE_Int          *index,
-                                      HYPRE_Complex       value);
+                                      HYPRE_Complex      *values);
 
 /**
  * Add to vector coefficients index by index.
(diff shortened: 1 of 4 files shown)


victorapm:
@gartavanis thank you for your contribution. I'll merge this into a development branch so that we can make a couple of changes.

As a note, for GPUs we recommend using the SetBoxValues functions in order to reduce the overhead (wrt to SetValues) associated with memory transfers in all D2H, H2D, and D2D cases

liruipeng:
Forgot to talk about this PR in today's meeting. This changes the user level API of hypre, we need to discuss it within the team.

rfalgout:
> Forgot to talk about this PR in today's meeting. This changes the user level API of hypre, we need to discuss it within the team.

I agree, we should discuss.  However, after looking at the code, it does appear to be inconsistent with `SStructVectorSetValues()`.  I don't remember why we did that.  In both cases, only one value is set at a time.  Since we are encouraging people to use `SStruct` anyway, and since we also encourage people to use `SetBoxValues` instead, my feeling is that we won't break many user codes.  We could also give a heads up to the code groups here in case this will affect them.  Anyway, let's chat at the next meeting.

liruipeng:
The user who needs this at LLNL is ParFlow team. Steve Smith,  @oseikuffuor1

PR #1281, opened 2025-05-06
Fix reg tests

victorapm:
Extends `hypre_CSRMatrixPrint` and `hypre_SeqVectorPrint` to work correctly according to the memory location of the input matrix and vector, respectively.

This fixes the regression tests on lassen


rfalgout (approved):
Looks good to me.  Thanks!

liruipeng (approved):

PR #1285, opened 2025-05-20
[recmat]: GPU porting for StructMatMult routines

victorapm:
* Fix CMake build
* Fix GPU build
* Clean up unused functions and rename "z"-prefixed routines
* Finalize porting of StructMatMult routines (core and fuse)
* Add profiling markers throughout the struct code 
* Extend matmult fuse routines to work for `nterms=2` and `nterms=3`


src/struct_mv/CMakeLists.txt
```diff
@@ -61,9 +61,12 @@ if (HYPRE_USING_GPU)
     struct_data.c
     struct_innerprod.c
     struct_matmult.c
+    struct_matmult_core.c
```
rfalgout, line 64: This is already correct on recmat.  Do you need to do a merge maybe?

victorapm: Interesting... This shows up as a new change to me. Also, recmat-dev-3 seems up-to-date with recmat

Screenshot from 2025-05-20 19-26-17

src/struct_mv/struct_matmult.c
```diff
@@ -1220,43 +1224,80 @@ hypre_StructMatmultCompute( hypre_StructMatmultData  *mmdata,
...
+#if defined(HYPRE_USING_GPU)
+   HYPRE_MemoryLocation       memory_location = hypre_StructMatrixMemoryLocation(matrices[0]);
+   HYPRE_ExecutionPolicy      exec_policy     = hypre_GetExecPolicy1(memory_location);
+#endif
 
    HYPRE_ANNOTATE_FUNC_BEGIN;
+   hypre_GpuProfilingPushRange("StructMatmultCompute");
+
+   /* Create work arrays on host memory for storing constant coefficients */
+   A_const = hypre_TAlloc(HYPRE_Complex*, nmatrices + 1, HYPRE_MEMORY_HOST);
```
rfalgout, line 1319: This is for future discussion, but I'm wondering if we should have a constant host array pointer in every matrix that would be a copy of the device data (when using GPUs) and add a HostConstData() macro to access it.  For CPU runs, the host data would just be a pointer to the device data, so no copying needed.  For GPU runs, we would need a SyncHostConstData() function (or something) that we would call first to ensure that the host data is correct.  Just a thought.  It might make for slightly cleaner code and maybe some reuse in other parts of the code.

src/struct_mv/struct_matrix.c
```diff
@@ -1285,7 +1285,9 @@ hypre_StructMatrixInitializeData( hypre_StructMatrix *matrix,
 {
    if (data == NULL)
    {
-      data = hypre_CTAlloc(HYPRE_Complex, hypre_StructMatrixDataSize(matrix),
+      /* TODO (VPM): we should be able to just do hypre_TAlloc, but that worsens PFMG convergence */
```
rfalgout, line 1290: This may be because of ghost zones.  We need to initialize them to zero and this is the cheapest way to ensure that will be the case.  This might also be something to think about wrt the other CTAlloc() calls changed in this file.

src/struct_mv/struct_matmult_core.c
```diff
@@ -18,56 +18,467 @@
...
+   cprod_##m     = cprod[(k + m)];    \
+   tptrs_##m##_0 = tptrs[(k + m)][0]
+
+#define HYPRE_SMMLOAD_3VARS(k, m)     \
+   cprod_##m     = cprod[(k + m)];    \
+   tptrs_##m##_0 = tptrs[(k + m)][0]; \
+   tptrs_##m##_1 = tptrs[(k + m)][1]
+
+#define HYPRE_SMMLOAD_4VARS(k, m)     \
+   cprod_##m     = cprod[(k + m)];    \
```
rfalgout, line 135: Why do we need to lose the k-array structure here?  Any idea if this affects the CPU runs?

src/struct_mv/struct_matmult.c
victorapm: I like this idea, specially since it will make things cleaner.

src/struct_mv/struct_matrix.c
victorapm: Oh yes! The reason I'm replacing CTAlloc with TAllocs is for performance. Memset on the GPU can be costly. We could gain at least 10% in exec. time on the setup phase if we get rid of them. Should we think of another way of initializing the ghost zones?

src/struct_mv/struct_matmult_core.c
victorapm: This is because we need to pass device pointers to the BoxLoops in the case of GPU runs, and the double pointers like tptrs can be an issue because if any of the inner pointers reside on the host (or were set up on the host), they become invalid when dereferenced on the device. So, here the idea is to reduce the level of indirection and create variables to point directly to device addresses.

I don't expect this change to affect the performance on CPU runs, but I will double check with experiments.

src/struct_mv/CMakeLists.txt
rfalgout: The answer is probably due to the fact that github PRs use a triple-dot diff, which does some funny things wrt common commit ancestors.  So, if the common ancestor between your branch and recmat is before the cmake file commit on recmat, then your cmake changes will show up even if there is no actual difference (double-dot diff). This is not important, but it's good to be aware of the triple-dot stuff because it has bitten us before.

src/struct_mv/struct_matmult.c
rfalgout: Okay, let's keep it in mind.

src/struct_mv/struct_matmult_core.c
rfalgout: Thanks for the explanation, and just do a quick performance check for a single rank run.  No need to do anything extensive because it's probably all okay.

src/struct_mv/struct_matrix.c
rfalgout: After thinking about this more, I think the ghost zones shouldn't matter.  We needed to be more careful about them on the master branch, but I can't think of why we need to worry here.  It would be good to understand why PFMG suffers by making this change.  That may be a symptom of a possibly bigger issue we need to fix.

victorapm: Thanks, Rob! I will investigate this. I realized the issue happens also with a single rank, so it's probably something else other than the ghost zones

src/struct_mv/CMakeLists.txt
victorapm: Oh ok! That makes sense, something to keep in mind...

rfalgout (approved):

src/struct_mv/struct_matmult.c
```diff
@@ -13,6 +13,7 @@
 
 #include "_hypre_struct_mv.h"
 #include "_hypre_struct_mv.hpp"
+//#define USE_FUSED_MATMULT 1
```
waynemitchell: The results you showed in our last meeting showed that the fused version was a bit faster, yes? At least on the GPU? Should we enable this by default, or is there a reason not to?

waynemitchell (approved):

src/struct_mv/struct_matmult.c
victorapm: I think we should enable it by default. I left it disabled just to make my tests easier :)

victorapm:
@rfalgout @waynemitchell I added a function for selecting the implementation used for computing struct mat/mat multiplication in https://github.com/hypre-space/hypre/pull/1285/commits/f031de879afea69ee294b3f6081ea40c040ee692

Could you double check and let me know if you have any suggestions (e.g., name changes...)? After this, I'm ready to merge this PR. Thanks!

rfalgout:
> Could you double check and let me know if you have any suggestions (e.g., name changes...)? After this, I'm ready to merge this PR. Thanks!

Hi @victorapm .  Overall, the interface looks good to me, but I do think you should use different names:
- Use `Matmult` instead of `MatMat` to be consistent with the function name where this feature is implemented. Do something similar for the variable names as well.
- Use the word "kernel" instead of "algorithm".  So, set the "kernel type" or just "kernel" if you want to shorten it.
- In PFMG, it would be good if we could just pull the matmult kernel selection into the `SetRAPType` function somehow since this only applies to RAP in that context anyway.  I guess this would have to be done by adding a "tens" digit in front of the usual RAPType numbers.  I'm not sure if I like this idea though, so we probably should stick with `PFMGSetMatmultKernel` as you have it (with suggested name change).

What do you think?  Thanks!

victorapm:
Thanks for the suggestions, Rob! Do the changes in [fb7795f](https://github.com/hypre-space/hypre/pull/1285/commits/fb7795f8086f1e89d36d9b514cb154a0c262dbca) look good? 

I used `PFMGSetMatmultType` to follow the others we have: `PFMGSetRAPType`, `PFMGSetRelaxType`

Please, let me know if this looks good. Thanks again!

src/struct_mv/protos.h
```diff
@@ -440,6 +440,9 @@ hypre_StructMatmultCreate( HYPRE_Int                  max_matmults,
 HYPRE_Int
 hypre_StructMatmultDestroy( hypre_StructMatmultData *mmdata );
 HYPRE_Int
+hypre_StructMatmultSetAlgoType( hypre_StructMatmultData  *mmdata,
```
rfalgout: This name needs to be changed, right?

src/struct_mv/struct_matmult.c
```diff
@@ -1005,7 +1049,7 @@ hypre_StructMatmultInitialize( hypre_StructMatmultData  *mmdata,
       if ( hypre_StructMatrixNeedResize(matrices[m], data_spaces[m]) )
       {
          hypre_StructMatrixResize(matrices[m], data_spaces[m]);
-         hypre_StructMatrixForget(matrix); // RDF: Add MemoryMode() to Resize() then remove this
+         hypre_StructMatrixForget(matrices[m]); // RDF: Add MemoryMode() to Resize() then remove this
```
rfalgout, line 1052: Was this breaking something?

rfalgout:
> Please, let me know if this looks good. Thanks again!

Looks good.  Thanks!

src/struct_mv/protos.h
victorapm: Thanks for catching! Fixed in [7fdb175](https://github.com/hypre-space/hypre/pull/1285/commits/7fdb175bcdf3e85f631e5705292ce3acf12e73a1)

src/struct_mv/struct_matmult.c
victorapm: This is a bug that I caught after reading the code. In this loop, we're not updating the `matrix` pointer as we do in the previous loop where `matrix` is used (lines 963 - 973). Hence, the change `matrix` -> `matrices[m]`.

rfalgout: What's odd is that it works correctly either way, right?  I guess we don't really stress the matrix resize much though, so maybe that's why.  Thanks!

victorapm: Yes, right! Sorry I didn't answer your question properly before. This wasn't breaking anything. Will need to think how to exercise this part of the code

PR #1286, opened 2025-05-22
 Add support for computing error norms in GMRES 

victorapm:
This pull request introduces enhancements to GMRES, including support for reference solutions and improvements to logging and error reporting. Additionally, it updates outdated URLs across multiple files. Below is a summary of the most important changes:

* Added new functions `HYPRE_GMRESSetRefSolution` and `HYPRE_GMRESGetRefSolution` to allow setting and retrieving a reference solution vector for error computation. 
* Introduced a new member `xref` in the GMRES data structure to store the reference solution. Updated `hypre_GMRESCreate`, `hypre_GMRESSetup`, and `hypre_GMRESSolve` to handle computations involving `xref`.
* Enhanced logging options to include error norms and convergence history when a reference solution is provided. New print levels (6-9) were added for debugging purposes. 
* Updated outdated URLs in comments across multiple files to point to the current location of the IEEE 754 status document. 
* Add regression tests for new tagged residual/error norms feature

Extra changes:
* Enable FPE exceptions in debug mode on linux + Fix some exceptions
* Fix memory leaks when calling `MGRSetup` multiple times without a matching `MGRDestroy` call

Example usage:

```
$ mpirun -np 1 test/ij -fromfile ../src/test/TEST_ij/data/beam_tet_dof459_np1/A.IJ -fromrhsfile ../src/test/TEST_ij/data/beam_tet_dof459_np1/b.IJ -nf 3 -solver 3 -k 50 -xstarfromfile ../src/test/TEST_ij/data/beam_tet_dof459_np1/xstar.IJ -funcsinterleaved -iout 7
Hypre init times:
Hypre init:
  wall clock time = 0.000001 seconds
  wall MFLOPS     = 0.000000
  cpu clock time  = 0.000001 seconds
  cpu MFLOPS      = 0.000000


Using HYPRE_DEVELOP_STRING: v2.33.0-15-g58eb704fc (branch tagged-error; not the develop branch)

  solver ID    = 3

Generate Matrix:
Spatial Operator:
  wall clock time = 0.002656 seconds
  wall MFLOPS     = 0.000000
  cpu clock time  = 0.002656 seconds
  cpu MFLOPS      = 0.000000

  Number of vector components: 1
  RHS vector has unit coefficients
  Initial guess is 0
  Reference solution vector read from file ../src/test/TEST_ij/data/beam_tet_dof459_np1/xstar.IJ
  Calling BuildFuncTagsInterleaved with num_functions = 3
IJ Vector Setup:
RHS and Initial Guess:
  wall clock time = 0.000077 seconds
  wall MFLOPS     = 0.000000
  cpu clock time  = 0.000076 seconds
  cpu MFLOPS      = 0.000000

Solver: AMG-GMRES
HYPRE_GMRESGetPrecond got good precond


 Num MPI tasks = 1

 Num OpenMP threads = 1


BoomerAMG SETUP PARAMETERS:

 Max levels = 25
 Num levels = 5

 Strength Threshold = 0.250000
 Interpolation Truncation Factor = 0.000000
 Maximum Row Sum Threshold for Dependency Weakening = 1.000000

 Coarsening Type = HMIS 
 measures are determined locally


 No global partition option chosen.

 Interpolation = extended+i interpolation
 Number of functions = 3
 Functions filtering is off

Operator Matrix Information:

             nonzero            entries/row          row sums
lev    rows  entries sparse   min  max     avg      min         max
======================================================================
  0     459    11277  0.054    12   37    24.6  -4.167e+00   7.500e+01
  1     216     8652  0.185    20   59    40.1  -9.241e-01   6.583e+01
  2      33      715  0.657    13   31    21.7  -2.815e-02   5.056e+01
  3      11      109  0.901     8   11     9.9  -2.146e-05   3.279e+01
  4       5       25  1.000     5    5     5.0   1.132e-02   3.863e+01


Interpolation Matrix Information:
                    entries/row        min        max            row sums
lev  rows x cols  min  max  avgW     weight      weight       min         max
================================================================================
  0   459 x 216     0    4   2.6   1.111e-01   7.537e-01   0.000e+00   1.000e+00
  1   216 x 33      1    4   2.6   2.888e-02   1.000e+00   4.435e-01   1.000e+00
  2    33 x 11      1    2   1.9   1.177e-01   8.823e-01   3.328e-01   1.000e+00
  3    11 x 5       1    2   1.3   1.724e-01   9.681e-01   1.724e-01   1.000e+00


     Complexity:   grid = 1.577342
               operator = 1.842511
                 memory = 1.968077




BoomerAMG SOLVER PARAMETERS:

  Maximum number of cycles:         1 
  Stopping Tolerance:               0.000000e+00 
  Cycle type (1 = V, 2 = W, etc.):  1

  Relaxation Parameters:
   Visiting Grid:                     down   up  coarse
            Number of sweeps:            1    1     1 
   Type 0=Jac, 3=hGS, 6=hSGS, 9=GE:     13   14     9 
   Point types, partial sweeps (1=C, -1=F):
                  Pre-CG relaxation (down):   0
                   Post-CG relaxation (up):   0
                             Coarsest grid:   0

Setup phase times:
GMRES Setup:
  wall clock time = 0.001154 seconds
  wall MFLOPS     = 0.000000
  cpu clock time  = 0.001154 seconds
  cpu MFLOPS      = 0.000000

L2 norm of b: 2.142429e+01
L2 norm of b0: 1.236932e+01
L2 norm of b1: 1.236932e+01
L2 norm of b2: 1.236932e+01
Initial L2 norm of residual: 2.142429e+01
Initial L2 norm of r0: 1.236932e+01
Initial L2 norm of r1: 1.236932e+01
Initial L2 norm of r2: 1.236932e+01
Initial L2 norm of error: 1.471136e+04
Initial L2 norm of e0: 2.314426e+03
Initial L2 norm of e1: 1.027296e+04
Initial L2 norm of e2: 1.027296e+04

  Iters             |e|_2           |e0|_2           |e1|_2           |e2|_2
 ------     -------------    -------------    -------------    ------------- 
      1      1.466438e+04     2.308647e+03     1.024096e+04     1.023898e+04 
      2      1.463625e+04     2.304914e+03     1.022218e+04     1.021832e+04 
      3      1.442954e+04     2.275311e+03     1.008754e+04     1.006360e+04 
      4      1.221520e+04     1.931907e+03     8.671306e+03     8.383748e+03 
      5      5.094395e+03     8.085714e+02     4.088546e+03     2.929652e+03 
      6      2.080738e+03     3.422215e+02     1.981676e+03     5.341493e+02 
      7      1.391606e+03     2.420118e+02     1.353518e+03     2.144464e+02 
      8      9.986826e+02     1.771530e+02     9.511145e+02     2.477196e+02 
      9      7.989485e+02     1.364089e+02     7.829958e+02     8.141819e+01 
     10      6.065046e+02     1.019223e+02     5.654708e+02     1.941710e+02 
     11      5.650393e+02     9.612819e+01     4.770110e+02     2.872095e+02 
     12      3.116541e+02     5.510059e+01     2.413861e+02     1.892749e+02 
     13      2.012461e+01     2.395113e+00     1.036004e+01     1.708604e+01 
     14      5.868421e+01     1.118560e+01     3.604868e+01     4.493564e+01 
     15      5.336111e+01     1.039796e+01     3.089835e+01     4.224432e+01 
     16      4.119659e+01     7.979591e+00     2.296989e+01     3.325462e+01 
     17      7.368214e+00     1.605038e+00     4.094696e+00     5.911674e+00 
     18      2.885672e+00     5.753945e-01     9.771622e-01     2.653521e+00 
     19      3.307042e-01     4.718693e-02     2.664438e-01     1.901219e-01 
     20      2.339605e-01     2.944370e-02     1.891974e-01     1.344429e-01 
     21      2.273797e-01     2.941917e-02     1.805359e-01     1.350661e-01 
     22      1.401470e-01     1.427303e-02     8.323082e-02     1.118485e-01 
     23      1.828833e-01     3.282356e-02     1.165284e-01     1.370768e-01 
     24      1.903516e-01     3.635983e-02     1.444598e-01     1.185034e-01 
     25      1.789139e-01     3.468979e-02     1.343629e-01     1.129311e-01 
     26      1.609038e-01     3.110288e-02     1.201061e-01     1.024557e-01 
     27      6.165855e-02     1.156958e-02     4.707162e-02     3.810754e-02 
     28      7.766007e-03     1.219527e-03     6.039698e-03     4.727120e-03 
     29      2.365765e-03     3.501504e-04     1.419066e-03     1.860239e-03 
     30      3.314674e-04     5.140338e-05     1.849251e-04     2.702426e-04 
     31      4.492168e-05     7.230694e-06     2.545445e-05     3.630077e-05 
     32      1.105975e-06     2.410587e-07     1.070490e-06     1.382844e-07 


Final L2 norm of residual: 9.300487e-08
Final L2 norm of r0: 4.020169e-08
Final L2 norm of r1: 5.205879e-08
Final L2 norm of r2: 6.575418e-08
Final L2 norm of error: 1.105975e-06
Final L2 norm of e0: 2.410584e-07
Final L2 norm of e1: 1.070490e-06
Final L2 norm of e2: 1.382853e-07

Solve phase times:
GMRES Solve:
  wall clock time = 0.011715 seconds
  wall MFLOPS     = 0.000000
  cpu clock time  = 0.011692 seconds
  cpu MFLOPS      = 0.000000


GMRES Iterations = 32
Final GMRES Relative Residual Norm = 4.341096e-09

```


src/test/ij.c
```diff
@@ -119,6 +115,8 @@ hypre_int
 main( hypre_int argc,
       char *argv[] )
 {
+   MPI_Comm            comm = hypre_MPI_COMM_WORLD;
```
rfalgout, line 118: This change is fine, but I don't understand why it's needed.  When will the communicator in our drivers ever be anything other than MPI_COMM_WORLD?

victorapm: Hi Rob, I agree. Let me know if you prefer to revert this

rfalgout: Either way is fine with me.  I was mostly wondering if you had some other reason for making the change.  Thanks!

victorapm: Sounds good, Rob! I did this just for organization: some places we were using `comm` while others we were using `hypre_MPI_Comm_World`

rfalgout:
This all looks good to me, @victorapm .  I want to make a general comment here about the Krylov methods in hypre, though maybe this is something to bring up in our biweekly meeting.  Over the years these routines have gotten more and more difficult to follow, yet Krylov methods are really simple from a pseudocode perspective.  I think it would be good to think about ways to make the code more readable while continuing to support the new features we add.  The debugging-related stuff (print statements, especially) could probably be reorganized into separate functions, for example.  Just something to think about for the future (not to address in this PR).  Thanks!

rfalgout (approved):

victorapm:
Thanks, Rob! I agree with the idea of making the code simpler, we could definitely do some refactoring to improve readability

src/parcsr_ls/ams.c
```diff
@@ -1260,8 +1260,9 @@ HYPRE_Int hypre_AMSSetAlphaPoissonMatrix(void *solver,
    hypre_AMSData *ams_data = (hypre_AMSData *) solver;
    ams_data -> A_Pi = A_Pi;
 
-   /* Penalize the eliminated degrees of freedom */
-   hypre_ParCSRMatrixSetDiagRows(A_Pi, HYPRE_REAL_MAX);
+   /* Penalize the eliminated degrees of freedom.
+      We divide HYPRE_REAL_MAX by 1000 to avoid FPE errors. */
+   hypre_ParCSRMatrixSetDiagRows(A_Pi, HYPRE_REAL_MAX / 1000);
```
oseikuffuor1, line 1265: @victorapm can you tell why HYPRE_REAL_MAX is generating the FPE errors? Why is the division by 1000 helping here? This is essentially setting the entry in diagonal rows to a large number right? I suppose this has little impact on the algorithm downstream, but I am curious why FPE errors are generated

src/parcsr_ls/par_mod_lr_interp.c
```diff
@@ -979,7 +979,10 @@ hypre_BoomerAMGBuildModExtPIInterpHost(hypre_ParCSRMatrix  *A,
                   break;
                }
             }
-            As_FF_diag_data[j] /= value;
+            if (value != 0.0)
+            {
```
oseikuffuor1, line 983: @victorapm Should we be using HYPRE_REAL_MIN for instances of comparison to 0.0?

src/krylov/gmres.c
```diff
@@ -373,13 +375,22 @@ hypre_GMRESSolve(void  *gmres_vdata,
...
    real_r_norm_old = b_norm;
 
+   /* Compute initial error when a reference solution is provided */
+   if (xref)
+   {
+      (*(gmres_functions->CopyVector))(x, w_3);
+      (*(gmres_functions->Axpy))(-1.0, xref, w_3);
+      (*(gmres_functions->InnerProd))(w_3, w_3, &num_tags, &xiprod);
+      e_norm = hypre_sqrt(xiprod[0]);
+   }
```
oseikuffuor1, line 392: Should this also check for print_level > 6 before performing the computations (and printing the initial error info)?

src/krylov/gmres.c
```diff
@@ -441,14 +452,44 @@ hypre_GMRESSolve(void  *gmres_vdata,
...
+         if (num_tags > 1)
+         {
+            for (tag = 0; tag < num_tags; tag++)
+            {
+               hypre_printf("Initial L2 norm of r%*d: %e\n", hypre_ndigits(num_tags),
+                            tag, hypre_sqrt(iprod[tag + 1]));
+            }
+         }
+
+         if (xref)
```
oseikuffuor1, line 481: Should we check print_level > 6 here?

src/parcsr_ls/par_mgr.c
```diff
@@ -247,6 +247,20 @@ hypre_MGRDestroy( void *data )
...
+            (mgr_data -> R_array)[i - 1] = NULL;
+         }
+
+         if ((mgr_data -> P_array)[i - 1] && (mgr_data -> RT_array)[i - 1] &&
+             ((mgr_data -> P_array)[i - 1] == (mgr_data -> RT_array)[i - 1]))
+         {
+            hypre_ParCSRMatrixDestroy((mgr_data -> P_array)[i - 1]);
+            (mgr_data -> P_array)[i - 1] = NULL;
+         }
+
```
oseikuffuor1, line 263: @victorapm a few comments here.
* Could we get away with just checking for "=="? Are the first two checks in the if-statement needed?
* Do we have currently options where we set RT_array[j] = R_array[j]?
* Similarly for RT_array[j] = P_arrayj], do we currently have this in master? (I have support for this in the sMGR branch based on our previous discussions, but it has not been merged into master yet)
* If the above comment are inconsequential, should we set (mgr_data->RT_array)[i-1] = NULL; as well?

src/parcsr_ls/par_mgr_setup.c
```diff
@@ -404,6 +404,20 @@ hypre_MGRSetup( void               *mgr_vdata,
...
+            RT_array[j] = NULL;
+         }
+
+         if (P_array[j] && RT_array[j] && (P_array[j] == RT_array[j]))
+         {
+            hypre_ParCSRMatrixDestroy(P_array[j]);
+            P_array[j]  = NULL;
+            RT_array[j] = NULL;
+         }
+
```
oseikuffuor1, line 420: Same comments as bullets 1,2 and 3 in preceding comment above.

src/parcsr_ls/par_mgr_setup.c
```diff
@@ -74,8 +74,8 @@ hypre_MGRSetup( void               *mgr_vdata,
 #if defined(HYPRE_USING_GPU)
    hypre_ParCSRMatrix  **P_FF_array = (mgr_data -> P_FF_array);
 #endif
-   hypre_ParCSRMatrix  **P_array = (mgr_data -> P_array);
-   hypre_ParCSRMatrix  **R_array = (mgr_data -> RT_array);
+   hypre_ParCSRMatrix  **P_array  = (mgr_data -> P_array);
+   hypre_ParCSRMatrix  **R_array  = (mgr_data -> R_array);
```
oseikuffuor1, line 78: Thanks for correcting this here.

oseikuffuor1 (commented):
Hi @victorapm , this looks good to me, but I left a few comments for your review.

src/parcsr_ls/ams.c
victorapm: Hi Daniel, Using `HYPRE_REAL_MAX` here leads to a SIGFPE when debugging hypre (built in debug mode) on gdb.  So this modification resolves the issue with SIGFPE. The new factor didn't cause any regressions in our tests

src/krylov/gmres.c
victorapm: The idea here is to report initial error norms in case the use passes a reference solution. `print_level > 6` would apply if the user wants to see convergence history of the error norms

src/parcsr_ls/par_mod_lr_interp.c
victorapm: In this particular case, I think it's fine to compare against `0.0`

`value` comes from the `D_q` array, which is initialized to zeros. The situation we are trying to guard against is when `D_q` doesn't get updated in the loop

src/krylov/gmres.c
victorapm: I think checking if `xref` is passed in should be enough here as we still want to see initial/final error norms when `print_level > 1`

src/parcsr_ls/par_mgr.c
victorapm: 1. We can have situations where both `RT_array[i-1]` or `R_array[i-1]` are NULL, one at a time. For example: one asks for injection - `R` is null - or for column-lumped - `RT` is null. (Edited from my previous reply, which was incorrect)
2. It doesn't seem we have. I can remove this
3. We don't have this in master. I added this check in anticipation to upcoming features
4. I think this doesn't apply, but correct me if I'm wrong

src/parcsr_ls/par_mgr_setup.c
victorapm: Does the previous answer address this one?

src/parcsr_ls/ams.c
oseikuffuor1: Oh ok. I was more curious about whether you knew why this was triggering SIGFPE error. It seems like it is not clear why.

src/krylov/gmres.c
oseikuffuor1: ok

src/parcsr_ls/ams.c
victorapm: It's because the matrix data array (coefficients) are later used on another place (can't remember which) that tries to do a multiplication with that value

src/parcsr_ls/par_mgr.c
oseikuffuor1: 1. I suppose the goal is to avoid potential double free error when RT == R (is this correct?). Just to clarify, is the first two checks meant to address the case when both could be NULL? Is that what you meant by your example?
2. Yes if you can't think of a case where we would need that, I agree we should remove that.
3. Sure, that is good with me
4. This is mainly for completeness since the data it points to has been deleted. So similar to what you have in (line 418 of par_mgr_setup.c (or see code snippet in comment below)

victorapm: 1. Right, goal is to avoid double free. However, for the second question, the real reason for the first two checks is to test for existence of both R and RT
2. OK!
3. OK!
4. What action point do you suggest here?

oseikuffuor1: Oh for 4. I was just pointing out that we should probably add:
(mgr_data -> RT_array)[i - 1] = NULL;
after
(mgr_data -> P_array)[i - 1] = NULL;

oseikuffuor1 (approved):
Thanks Victor!

src/parcsr_ls/par_mgr.c
victorapm: Sure, I will do that

src/utilities/general.c
```diff
@@ -405,6 +410,10 @@ HYPRE_Initialize(void)
    hypre_MagmaInitialize();
 #endif
 
+#if defined(__linux__) && defined(HYPRE_DEBUG)
+   feenableexcept(FE_INVALID | FE_DIVBYZERO | FE_OVERFLOW);
+#endif
+
```
waynemitchell, line 416: @victorapm , for some reason the changes in `general.c` here are causing a couple of the regressions to fail on Aurora. I'm not sure what's going on here... took me a while to track it down because this seems unrelated. But if I back up to the commit before this, the regressions are clean. Then if I only apply the changes in this file, I get seg faults in a couple of the elasticity regressions. Any idea what's going on here?

victorapm: I added this for debug builds a couple of months ago. It's for triggering SIGFPE in case we hit floating-point exceptions. If we have tests that are failing, we should understand why. Which ones are failing?

waynemitchell: It's the `TEST_ij/elast.jobs` 17 and 18 jobs.

victorapm: Ok! How about the build options for hypre? Is this the one? https://github.com/hypre-space/hypre/blob/435e042d5355a55ebb460fea6d02846fd0274d1e/AUTOTEST/machine-aurora.sh#L60

waynemitchell: I'm just doing:
./configure --with-sycl --enable-debug --enable-unified-memory --with-print-errors
Also, for the environment on Aurora, I'm using the default oneAPI (oneapi/release/2025.0.5) but a custom oneDPL (cloned from their git hub and at hash 716ab2fdb6ba8eefd17a730eeb003ca6eb419f53). You can point hypre to the custom oneDPL location by setting the environment variable DPLROOT=<path/to/top/level/oneDPL/directory>.

waynemitchell: Sorry, I think I said I was getting seg faults before, but that's not right. I'm seeing "signal 8," which refers to this SIGFPE thing. Interestingly, the code runs fine (the job converges) if I only comment out the line:
`feenableexcept(FE_INVALID | FE_DIVBYZERO | FE_OVERFLOW);`
So whatever arithmetic error it is encountering seems pretty benign.

victorapm: Thanks Wayne, I'll take a look... It's interesting I can't reproduce this on other machines... Might be a sycl/icx issue... Let's see!

waynemitchell: Yes, it seems quite strange. I'll add that I tried this with a CPU build on Aurora also and was not able to reproduce the problem. You get a different hierarchy with the CPU code, so I'm guessing the GPU code path for this particular run specifically leads to these arithmetic errors (they occur somewhere deep in the hierarchy in the setup).

PR #1287, opened 2025-05-23
Recmat - fix for part boundary box coarsening and SSAMG convergence

rfalgout:
This fixes an issue with the coarsening of the part boundary boxes.  Some notes:

- This PR is intended to first illustrate the changes needed and to support further discussion towards a final fix.
- The name of the function `hypre_CoarsenBoxArrayArrayNeg` is no longer appropriate and should be changed and possibly moved elsewhere since it is only needed in two places in the `sstruct_mv` directory.  It is pretty specific to the needs of the two calling routines: `hypre_SStructGridCoarsen` and `hypre_SStructPMatmultInitialize`.
- The function `hypre_CoarsenBoxNeg` is no longer used and can be removed.  It is only called by the function in the previous bullet (so maybe not generally useful) and it either has a bug in it or is not appropriate for coarsening the part boundary boxes.
- The ssamg regression tests in `amr2d` slow down slightly with these changes, `amr3d` stays the same, and the other ssamg tests (without a Krylov method) speed up considerably.
- The convergence of the `amr2d` problems is pretty slow overall (as a preconditioner it's fast), which is puzzling.

### DIFF
--- a/src/struct_mv/coarsen.c
+++ b/src/struct_mv/coarsen.c
@@ -336,10 +336,17 @@ hypre_CoarsenBoxArrayArrayNeg( hypre_BoxArrayArray   *boxaa,
    hypre_BoxArray        *new_boxa;
    hypre_BoxArrayArray   *new_boxaa;
 
+   hypre_Index            unit_index;
+   hypre_Index            grow_index;
+
    HYPRE_Int              count_box;
    HYPRE_Int              count_boxa;
    HYPRE_Int              box_id;
    HYPRE_Int              i, ii, j;
+
+   /* Set grow_index = stride - unit_index */
+   hypre_SetIndex(unit_index, 1);
+   hypre_SubtractIndexes(stride, unit_index, ndim, grow_index);
 
    /* Allocate box */
    box = hypre_BoxCreate(ndim);
@@ -358,7 +365,10 @@ hypre_CoarsenBoxArrayArrayNeg( hypre_BoxArrayArray   *boxaa,
       hypre_ForBoxI(j, boxa)
       {
          hypre_CopyBox(hypre_BoxArrayBox(boxa, j), box);
-         hypre_CoarsenBoxNeg(box, refbox, origin, stride);
+         //hypre_CoarsenBoxNeg(box, refbox, origin, stride);
+         hypre_BoxGrowByIndex(box, grow_index);
+         hypre_CoarsenBox(box, origin, stride);
+         hypre_IntersectBoxes(box, refbox, box);
          if (hypre_BoxVolume(box))
          {
             count_boxa++;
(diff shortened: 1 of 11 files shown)


waynemitchell:
@rfalgout, OK, I'm mostly documenting this for my own understanding (I think you know all this already). I did some testing to verify this. The fix here in coarsen.c affects the correctness of the specialized matmult routine AND has an impact on the ssamg hierarchy. That is why even if you are using the general/slower matmult code path (which always yields correct matmults), the fix can still yield a difference in convergence. Out of curiosity, do you know exactly how the incorrect pbnd boxes were affecting the ssamg hierarcy? I guess that they will affect how the interpolation operator is setup, since you do some adjustment of weights at the part boundaries defined by these boxes. Anything else?

rfalgout:
> @rfalgout, OK, I'm mostly documenting this for my own understanding (I think you know all this already). I did some testing to verify this. The fix here in coarsen.c affects the correctness of the specialized matmult routine AND has an impact on the ssamg hierarchy. That is why even if you are using the general/slower matmult code path (which always yields correct matmults), the fix can still yield a difference in convergence. Out of curiosity, do you know exactly how the incorrect pbnd boxes were affecting the ssamg hierarcy? I guess that they will affect how the interpolation operator is setup, since you do some adjustment of weights at the part boundaries defined by these boxes. Anything else?

Hi @waynemitchell .  I meant to say something more about the connection between the specialized matmult and general matmult, specifically with respect to the Frobenius norm differences.  They are all zero to machine precision.

I was not sure why the affect on `SStructGridCoarsen` changed convergence, but intended to investigate that.  Your guess may be a good one.  @victorapm, do you know (or have a guess)?

BTW, I am wondering if there is a better way to do some of this long term.  When I have some time, I will organize my thoughts in more detail so we can all discuss it.

victorapm:
Hi all, the part boundary array affects how interpolation is computed as Wayne brought up. This should be the only reason for the differences

victorapm (approved):
Thank you, Rob! I think it makes sense to remove `hypre_CoarsenBoxNeg` and `hypre_CoarsenBoxArrayArrayNeg` as it does not seem they are working properly for some cases and your new solution involving subtraction of boxes sounds appropriate

rfalgout:
@waynemitchell and @victorapm .  I have done the renaming, etc.  Let me know if you want me to change anything else.  Otherwise, I am ready to merge.  Thanks!

victorapm:
Hi Rob, thank you for the changes. I think it looks good and this can be merged

src/struct_mv/coarsen.c
```diff
@@ -358,7 +308,10 @@ hypre_CoarsenBoxArrayArrayNeg( hypre_BoxArrayArray   *boxaa,
       hypre_ForBoxI(j, boxa)
       {
          hypre_CopyBox(hypre_BoxArrayBox(boxa, j), box);
-         hypre_CoarsenBoxNeg(box, refbox, origin, stride);
+         //hypre_CoarsenBoxNeg(box, refbox, origin, stride);
```
waynemitchell: Remove the commented line here? Particularly since `hypre_CoarsenBoxNeg()` was removed. Same thing below.

waynemitchell (approved):
Thanks, Rob! Glad you were able to track this down.

src/struct_mv/coarsen.c
rfalgout: Yes, done.  Thanks!

PR #1289, opened 2025-05-27
Add matrix scaling and element-wise vector functions

victorapm:
This PR introduces new matrix functions and element-wise vector operations for both host and device (GPU) runs.

1. New element-wise operations for `hypre_Vector`:
- `hypre_SeqVectorPointwiseProduct`: z[i] = x[i] * y[i]
- `hypre_SeqVectorPointwiseDivision`: z[i] = y[i] / x[i]
- `hypre_SeqVectorPointwiseInverse`: y[i] = 1.0 / x[i]

2. Corresponding `hypre_ParVector` and public APIs:
- `(hypre|HYPRE)_ParVectorPointwiseProduct`
- `(hypre|HYPRE)_ParVectorPointwiseDivision`
- `(hypre|HYPRE)_ParVectorPointwiseInverse`

3. Placeholder for the corresponding `hypre_StructVecor` and `hypre_SStructVector` implementations 
5. New private function and public API for computing scaling weights for a `ParCSRMatrix`


src/parcsr_ls/par_amg_solve.c
```diff
@@ -119,18 +120,18 @@ hypre_BoomerAMGSolve( void               *amg_vdata,
    }
 
    /* Update work vectors */
-   hypre_ParVectorResize(Vtemp, num_vectors);
-   hypre_ParVectorResize(Rtemp, num_vectors);
-   hypre_ParVectorResize(Ptemp, num_vectors);
-   hypre_ParVectorResize(Ztemp, num_vectors);
+   hypre_ParVectorResize(Vtemp, num_rows_fine, num_vectors);
```
rfalgout, line 123: It looks like this is the only place ParVectorResize() is called.  Do we really need to introduce the `size` feature?  Do we envision this being used elsewhere?  Just wondering.

src/parcsr_mv/HYPRE_parcsr_matrix.c
```diff
@@ -408,3 +408,43 @@ HYPRE_ParCSRMatrixMatvecT( HYPRE_Complex      alpha,
...
+   return ( hypre_ParCSRMatrixDiagScale((hypre_ParCSRMatrix *) A,
+                                        (hypre_ParVector *) left,
+                                        (hypre_ParVector *) right) );
+}
+
+/*--------------------------------------------------------------------------
+ *--------------------------------------------------------------------------*/
+
+HYPRE_Int
+HYPRE_ParCSRMatrixCompScalingTagged(HYPRE_ParCSRMatrix   A,
```
rfalgout: Where is this used?  Where would it be used?

src/struct_mv/struct_vector.c
```diff
@@ -1320,3 +1320,69 @@ hypre_StructVectorClone(hypre_StructVector *x)
 
    return y;
 }
+
+/*--------------------------------------------------------------------------
+ *--------------------------------------------------------------------------*/
+
+HYPRE_Int
+hypre_StructVectorElmdivpy( hypre_StructVector *x,
```
rfalgout: I didn't see this one in ParVector.  What does it do?

rfalgout:
Hi @victorapm .  This generally looks good.  I have a few questions (asked already).  The only other thing I might suggest is to get feedback from the team on the names, maybe at our next meeting.

src/parcsr_ls/par_amg_solve.c
victorapm: The new element-wise functions added in this PR also make use of `hypre_ParVectorResize`. The `size` field is important because the user code might pass a non-empty `z_ptr` that points to a vector with size smaller than what we actually need here. This could happen in a scenario where the vector might have been created before in a transient simulation, but the number of unknowns have increased during the simulation

https://github.com/hypre-space/hypre/blob/bc07da4bbfa55e4ff2c49e9312c5705d463e0772/src/parcsr_mv/par_vector.c#L731-L750

src/parcsr_mv/HYPRE_parcsr_matrix.c
victorapm: Currently this is used only in geos, let me write an example usage in the `ij` driver

src/struct_mv/struct_vector.c
victorapm: `Elmdivpy` was added in the past for `hypre_ParVector`. It computes `y = y + x ./ b`

https://github.com/hypre-space/hypre/blob/bc07da4bbfa55e4ff2c49e9312c5705d463e0772/src/parcsr_mv/par_vector.c#L688-L704

rfalgout (approved):

src/parcsr_mv/HYPRE_parcsr_matrix.c
victorapm: Hi Rob, I added an example usage and regression test for this feature in [830736a](https://github.com/hypre-space/hypre/pull/1289/commits/830736a065952a041b876ff076bd0b120c7d1193)

victorapm:
@oseikuffuor1 @rfalgout I updated the `Elm` function names to `Pointwise`. Please, let me know if this looks good

src/parcsr_mv/par_csr_matop.c
```diff
@@ -7053,3 +7053,122 @@ hypre_ParCSRMatrixColSum( hypre_ParCSRMatrix   *A,
...
+ * Example usage:
+ *   hypre_ParCSRMatrixCompScalingTagged(A, 1, num_blocks, block_tags, &scaling);
+ *   hypre_ParCSRMatrixDiagScale(A, scaling, scaling);
+ *
+ * Notes:
+ *   - If type=0 or tags=NULL, the function returns without scaling.
+ *--------------------------------------------------------------------------*/
+
+HYPRE_Int
+hypre_ParCSRMatrixCompScalingTagged(hypre_ParCSRMatrix  *A,
```
oseikuffuor1: Is the 'type' argument here a placeholder for different types of scaling in the future? Otherwise I am not sure why it would be needed, since if it is 0, one would just not call the function

src/seq_mv/vector.c
```diff
@@ -1466,74 +1469,233 @@ hypre_SeqVectorSumElts( hypre_Vector *v )
...
+                                      hypre_Vector *y,
+                                      hypre_Vector *z )
+{
+   HYPRE_Int      size   = hypre_VectorSize(x);
+   HYPRE_Complex *x_data = hypre_VectorData(x);
+   HYPRE_Complex *y_data = hypre_VectorData(y);
+   HYPRE_Complex *z_data = hypre_VectorData(z);
+   HYPRE_Int      i;
 
+   /* Element-wise division y[i] = y[i] / x[i] */
```
oseikuffuor1: Can you correct this comment? LHS should be z[i].

src/parcsr_mv/par_csr_matop.c
victorapm: Yes, this is for an upcoming branch with different methods for computing the norms

src/seq_mv/vector.c
victorapm: Great, thanks for catching

src/parcsr_mv/par_csr_matop.c
oseikuffuor1: We should probably add in the comments what type of scaling this is, but I suppose that will be included in the upcoming branch when there's more than one option?

oseikuffuor1 (approved):
Thanks Victor, this looks good to me.

src/parcsr_mv/par_csr_matop.c
victorapm: Yes, good point. I can do it in this PR. Thanks!

PR #1290, opened 2025-06-03
oneAPI 2025.0, 2025.1 workarounds

waynemitchell:
There are some workarounds needed in order to avoid issues with the oneDPL versions packaged with oneAPI 2025.0 and 2025.1.


victorapm (approved):
Thank you! This seems a subset of #1314, right?

PR #1296, opened 2025-06-16
Multiprecision functions

rfalgout:
This PR adds runtime multiprecision function support to hypre.  Some comments:

* This is a proposal and work-in-progress intended for discussion
* Currently only the `krylov` directory is implemented for illustrative purposes.
* The code compiles cleanly when configured normally and also with `--enable-mixed-precision` (note that the `struct_mp.c` driver is commented out in `test/Makefile` because I didn't feel like fixing the compile warnings).
* All of the default regression tests (ams, ij, sstruct, struct, lobpcg) run cleanly with `--enable-mixed-precision`.
* I verified that the runtime-switching code is being called (i.e., `Foo` --> `Foo_pre`) by walking through the debugger.
* The `sstruct.c` driver has been extended to exercise the runtime precision capability via the `-precision` flag.

Some notes on the implementation:

* All of the runtime multiprecision code is generated automatically by way of the script `krylov/mup_code`.
* The underlying code generation scripts (`config/gen_*.sh`) are modifications of Daniel's original scripts (`config/gen*.sh`).
* All of the multiprecision-related files begin with `mup` (or `mp`) with the exception of the header files which follow the `HYPRE_` and `_hypre_` naming conventions and end with `_mup*.h`.  This change was made to keep the multiprecision files together in the file listing (for sanity).
* The original `krylov_functions.saved` file is now split into three files: `mup.functions`, `mup.methods`, and `mup.fixed`.  Functions should only appear in one of the lists.  The latter contains functions that do not have runtime mup support.
* [There is no longer a fourth default compile] There are four compiles done (as before) but the last one is now named `MP_BUILD_DEFAULT` to distinguish it from the actual default build in the `test` directory (which is what a user would do).  This was needed to manage the tricky symbol naming and prototyping, especially with regard to the standard symbol names `HYPRE_Foo` and `hypre_Foo` (the symbols with no suffix).  The user prototypes for these need to use `hypre_long_double` and `void *` for `HYPRE_Real` and `HYPRE_Real *` to support runtime precision switching.  For the `MP_BUILD_DEFAULT` library compile, these same prototypes are needed, but only in the `mup_functions.c` file where these symbols are implemented.  For all the other files, macro renaming is needed (we don't want to do any renaming in user code).
* The scripts should work for any function prototypes that don't have parentheses in either the return type or argument types.  As a result, many other changes were needed in the `krylov` directory because it was a bit of a mess.  This should make the other directories much easier to convert.
* The `HYPRE_ZZZZZPRECISION_FUNC` macro in `_hypre_krylov_mup_def.h` was introduced to allow us to transition each of the other directories one at a time.  Once they are all migrated over, this should be renamed to the original macro name `HYPRE_MULTIPRECISION_FUNC` and the duplicate deleted.

Additional comments:

* [This approach is now used throughout] Consider modifying the build system to use different object-file extensions instead of directories.  For example, use `.o`, `.o_flt`, `.o_dbl`, `.o_long_dbl`, `.obj`, `.obj_flt`, etc.  This would mimic the `.obj` approach we already use for the GPU compilation and hence should have a natural cmake approach that can be used.  Also, it should allow most of the "Build Rules" section in each of the makefiles to be implemented only once in `Makefile.config` because there would be no directory-specific naming to do.  It would be easy to test this idea in the `krylov` directory first. 
* [This approach is now used throughout] Consider switching the meaning of `FILES` and `MuP_FILES` in the makefiles.  It seems more natural to me to let `FILES` represent the standard files in hypre (as currently) and use `MP_FILES` to mean files that only pertain to the `--enable-mixed-precision` configuration.


src/krylov/pcg.c
```diff
@@ -690,7 +690,7 @@ hypre_PCGSolve( void *pcg_vdata,
                HYPRE_Real r2ob2;
                /* v = C*s = C*(r_old-r_new) */
                (*(pcg_functions->ClearVector))(v);
-               precond(precond_data, A, s, v);
+               precond(precond_data, A, s, v);      // RDF: Should this use precond_Mat?
```
ulrikeyang: Yes, that should be precond_Mat instead of A. Apparently, we missed this since it is in a spot that is not used that often.

rfalgout: Okay, fixed.  Thanks!

src/krylov/HYPRE_krylov.h
```diff
@@ -1289,4 +1310,12 @@ HYPRE_Int HYPRE_CGNRGetPrecond(HYPRE_Solver  solver,
 }
 #endif
 
+#ifdef HYPRE_MIXED_PRECISION
+#include "_hypre_krylov_mup_undef.h"
+#include "HYPRE_krylov_mup.h"
+#ifdef BUILD_MP_FUNC
+#include "_hypre_krylov_mup_def.h"
+#endif
+#endif
```
oseikuffuor1: @rfalgout can you explain a bit what is going on here? _hypre_krylov_mup_def.h is included earlier on in this file (between the HYPRE_MIXED_PRECISION guards), then undef'd and included again. Could there be multiple definition for this second inclusion since HYPRE_krylov_mup.h may also define the same function name? I suppose this works, but I think I'm missing something.

rfalgout: The include at the top of the file for `mup_def.h` affects everything in the header file, changes names, etc.  We don't want to do that to user code when they include this file, so at the bottom the name change macros are undefined.  The extra redefine is when we use this header file to build code in hypre where we do want the name changes to affect the code.  I suppose this could be changed to say `#ifndef BUILD_MP_FUNC` then include the `undef.h` and `krylov_mup.h` files.  Maybe I'll experiment with that, but hopefully this clears things up a bit.

rfalgout: One more thing I remembered here.  The file `HYPRE_krylov_mup.h` contains all the prototypes needed whether compiling hypre code or user code.  It needs to be included here without getting modified by the `mup_def.h` macros.  I think one way to do that would be to include it near the top of this file, then use the `#ifndef` idea in the previous comment at the bottom.  I would have to test this to be sure I'm not forgetting something.

Also, now I see that we would potentially have two mismatched `Foo` prototypes the way this is right now if we try to drop the `_def` rename (e.g., `SetAbsoluteTol`).  So, if we implement the idea we talked about for getting rid of the `_def` symbols in the library, we would still have to use a rename like this in the header file to prevent the conflict (we just wouldn't compile the normal hypre source code with this rename).

src/utilities/multiprecision.h
```diff
@@ -23,68 +23,126 @@
...
 #ifndef HYPRE_LONG_DOUBLE
 #define HYPRE_LONG_DOUBLE 1
 #endif
-#elif defined(MP_BUILD_DOUBLE)
-#define HYPRE_OBJECT_PRECISION HYPRE_REAL_DOUBLE
+
+#elif defined(MP_BUILD_DEFAULT)
+
+#define HYPRE_MULTIPRECISION_FUNC(a) a
+#define HYPRE_ZZZZZPRECISION_FUNC(a) CONCAT_(a, def)  /* RDF: Temporary */
```
oseikuffuor1: What will this macro return now, in the case when we get rid of Foo_def? Could it be a no-op as before? I suppose this could have implications on the way the name changes are handled in the header files (eg. HYPRE_Krylov.h)?

rfalgout: In the case where we don't build the symbols for Foo_def (they will not be in the library file), I think we may still need to have this macro redefinition to Foo_def to avoid prototype conflicts in the header files.  This is what I was trying to say in the comments for the HYPRE_krylov.h file.

rfalgout:
Hi @oseikuffuor1 .  I made minor changes to `HYPRE_krylov.h` and `HYPRE_lobpcg.h` to simplify the def/undef stuff a little bit and hopefully make it clearer with some comments.  Let me know what you think.  If it looks okay, then I will also incorporate the same approach into the internal header files.  Thanks!

rfalgout:
@oseikuffuor1 .  One more thing to point out, if you haven't noticed it already.  When user files are compiled (e.g., the test drivers in hypre), `HYPRE_MP_FUNC` is not defined.  As a result, the mixed-precision prototypes are included and available in `HYPRE_krylov.h` but the macro-based prototype renaming does not extend outside of the header file and into user code.  Note also that the extra mixed precision files in hypre (e.g., `mup_pre.c`, `mup_functions.c`) are also compiled as if they were user files.  In summary, the standard files in hypre are compiled with macro renaming turned on and mixed precision prototypes unavailable, while the situation is reversed for mixed-precision files and users.

rfalgout:
Hi @oseikuffuor1  and @ulrikeyang  .  I've done a bunch of mostly cosmetic things today that required touching a lot of files.  Here is a synopsis:

- Renamed `krylov.h` to `_hypre_krylov.h` to conform to header naming conventions.
- Renamed `lobpcg.h` to `_hypre_lobpcg.h` to conform to header naming conventions.
- Moved `utilities/multiprecision.h` to `_hypre_mup_def.h` and `HYPRE_utilities.h`.  This helps us to better conform to header file practices in hypre (we don't want to install headers without a hypre prefix of some sort).  The file `_hypre_mup_def.h` is only every included in the header files `_hypre_<directory>_mup_def.h`, so the name is consistent with them.  This also freed up the use of `multiprecision.h` to implement the runtime precision variable.
- Added a top-level `mup_code` script in the src directory that runs each of the `mup_code` scripts in the subdirectories to write all of the multiprecision code.  It currently only calls `krylov/mup_code` but that will change as we update the rest of hypre.
- Implemented a runtime precision global variable.  See `multiprecision.c` and `multiprecision.h` in the utilities directory.  Also look at `krylov/mup_functions.c` to see how it is used.

rfalgout:
Hi @oseikuffuor1 @ulrikeyang .  I just pushed changes that incorporate the `sstruct_ls` directory.  This ended up being more of a headache than I had hoped for because of two things:
- I had issues with FAC and Maxwell prototypes, so I decided to pull the trigger and remove them from hypre as we've talked about doing anyway.
- A weird bug came up with building the library file that I eventually figured out.  I will send an email to the whole team to get some feedback.

victorapm:
@rfalgout @oseikuffuor1 Just sharing an idea... Would it make sense to use `f64`, `f32`, ... instead of `dbl`, `flt`, ...? I'm thinking about a future GPU implementation where we would like to have f16 or even lower precisions... Maybe having the precision number spelled out would be nicer.

rfalgout:
> @rfalgout @oseikuffuor1 Just sharing an idea... Would it make sense to use `f64`, `f32`, ... instead of `dbl`, `flt`, ...? I'm thinking about a future GPU implementation where we would like to have f16 or even lower precisions... Maybe having the precision number spelled out would be nicer.

Let's chat about this at our meeting.  I don't really know what's best from a naming perspective.  Fortunately, with the updates we are making now, these hard-coded names will almost always be called by auto-generated code, so it would be easy to change in the future without affecting users.

rfalgout:
Hi @oseikuffuor1 and @ulrikeyang .  I still have 6 directories to convert (they will be straightforward and basically follow exactly what Daniel did before), but it would be good if you could complete your reviews as soon as you are ready.  Thanks!

ulrikeyang (approved):
Overall, I am fine with these changes that were done in the original PR. I don't have enough time right now to go through the additional 300+ file changes with my upcoming travel in detail, but I assume that regression tests will catch bugs, and I will be able to look more into this when I get back, but I don't want to hold up the hypre-3.0 branch. So, thank you for all this hard work. It  looks good to me.

rfalgout:
Hi @oseikuffuor1 and @ulrikeyang .  This is all done now.  The only thing we don't have runtime multiprecision support for is LOBPCG.  There are some 'mv_MultiVector' symbols referenced by our test drivers and I didn't write the functions to support these because the API doesn't conform to using 'HYPRE_' functions.  Maybe we can try to update this later, but I'd like to fix the API if we do.  We still have '_flt', etc. symbols for LOBPCG that can be called directly.

oseikuffuor1 (approved):
Thanks a lot Rob!!

PR #1303, opened 2025-06-30
Allow major compile compute capability less than actual

### LINKED ISSUE #1302: Assertion of equivalent major compute capability for cuda
What are the reasons for the assertion? We have an apptainer container that we distribute to users and we compile with cuda arch 70 for maximum compatibility (recognizing performance may not be as good). We may just start passing multiple arch's to configure, but I was curious about the rationale behind the assertion

lindsayad:
Closes #1302

### DIFF
--- a/src/utilities/device_utils.c
+++ b/src/utilities/device_utils.c
@@ -2739,13 +2739,7 @@ hypre_CudaCompileFlagCheck()
 
    /* HYPRE_CUDA_CALL(cudaDeviceSynchronize()); */
 
-   const hypre_int cuda_arch_actual_major  = cuda_arch_actual  / 100;
-   const hypre_int cuda_arch_compile_major = cuda_arch_compile / 100;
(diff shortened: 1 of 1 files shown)


lindsayad:
I'd also be fine just removing the `hypre_assert(0)` line

prj-:
@victorapm, is it worth getting this merged before the 3.0 release?

victorapm:
Sorry I missed this one @lindsayad. I'll check with the team this afternoon, but it looks it's worth merging on hypre-3. Thanks for bringing it up

lindsayad:
No problem, thanks for looking into it!

victorapm:
Moving this to #1363 
Fixed also our CMake build so it allows something like `-DCMAKE_CUDA_ARCHITECTURES=80;90;100`

lindsayad:
Closing in favor of #1363

PR #1307, opened 2025-07-18
GPU additions to IJ interface

victorapm:
This PR contains mainly improvements to GPU support in hypre. Details are given below:

- `hypre_IJMatrixGetValuesParCSRDevice`: Complete GPU implementation for retrieving values from ParCSR matrix
- `HYPRE_IJMatrixGetValues2()`: Now calls device implementation when execution policy is GPU
- `HYPRE_IJMatrixGetValuesAndZeroOut()`: Replaced error message with proper device implementation call
- Moved `hypre_PrefixSumInt()` implementation to `utilities`
- `hypre_CSRMatrixSetRownnzDevice()`: New device implementation for row non-zero counting
- Added GPU profiling range markers to key computational kernels
- Added flag to control addition of structural zeros in triple matrix multiplication algorithms for GPU execution
- Added `log_level_saved` in `hypre_handle` field for temporary log level storage
- Reorganized function declarations with clear separation between host and device implementations
- Added proper device function prototypes with complete parameter specifications
- Enhanced documentation comments for device-specific functions
- Enhanced test drivers with additional BoomerAMG configuration options:

PR #1308, opened 2025-07-18
[recmat]: GPU support for SStruct interface and SSAMG

victorapm:
This PR contains fixes and additions enabling GPU support throughout the SStruct interface, including SSAMG.

Non-struct changes in this PR can be found in #1307


src/test/TEST_sstruct/amr2d.saved
```diff
@@ -52,19 +52,19 @@ Final Relative Residual Norm = 5.252407e-07
 
 # Output file: amr2d.out.14
 Iterations = 37
-Final Relative Residual Norm = 8.428671e-07
+Final Relative Residual Norm = 8.012871e-07
```
rfalgout, line 55: I can't remember if you had an explanation for these differences.  Do you?  I have found these tests to be super sensitive to changes in the code, which is a bit worrying.

rfalgout (approved):
This looks good to me @victorapm , modulo my question about the saved file differences.  Thanks!

src/test/TEST_sstruct/amr2d.saved
victorapm: Hi Rob, this is mainly due to changes in the defaults in `sstruct` driver, which now uses 6 levels of pure SSAMG before switching to BoomerAMG. I found this to be robust and efficient in general so I changed the defaults, but I can revert if you prefer

rfalgout: This is a perfectly good explanation, so let's keep it as is.  Thanks!

src/sstruct_ls/ssamg_csolver.c
```diff
@@ -104,6 +109,7 @@ hypre_SSAMGCoarseSolverSetup( void *ssamg_vdata )
       HYPRE_BoomerAMGSetLogging(csolver, 1);
       //HYPRE_BoomerAMGSetAggNumLevels(csolver, 1);
       HYPRE_BoomerAMGSetup(csolver, par_Ac, par_b, par_x);
+      hypre_SetLogLevel(0);
```
waynemitchell, line 112: Why set the log level here?

src/sstruct_ls/ssamg_setup.c
```diff
@@ -69,39 +69,44 @@ hypre_SSAMGSetup( void                 *ssamg_vdata,
...
       hypre_error_w_msg(HYPRE_ERROR_GENERIC, "RHS is not HYPRE_SSTRUCT");
-      HYPRE_ANNOTATE_FUNC_END;
 
       return hypre_error_flag;
    }
 
+   HYPRE_ANNOTATE_FUNC_BEGIN;
+   hypre_GpuProfilingPushRange("SSAMG-Setup");
+   hypre_MemoryPrintUsage(comm, hypre_HandleLogLevel(hypre_handle()), "SSAMG setup begin", 0);
+   hypre_SetLogLevel(0);
```
waynemitchell, line 101: Same question here on the log level.

src/sstruct_mv/sstruct_matmult.c
```diff
@@ -1097,8 +1099,18 @@ hypre_SStructMatmultComputeU( hypre_SStructMatmultData *mmdata,
...
+      if (!urap2)
+      {
+         /* Note: this call does NOT add structural zeros to the diagonal */
+         parcsr_uM = hypre_ParCSRMatrixRAPKT(parcsr_sP, parcsr_sA, parcsr_sP, 1, 0);
+      }
+      else
+      {
+         hypre_ParCSRMatrix *Q = hypre_ParCSRMatMat(parcsr_sA, parcsr_sP);
+         parcsr_uM = hypre_ParCSRTMatMatKT(parcsr_sP, Q, 1);
+         hypre_MatvecCommPkgCreate(parcsr_uM);
```
waynemitchell, line 1112: I'm curious what the reason for the two different options here is (toggled by `urap2`). Also, why do you create the matvec comm pkg here but not above?

waynemitchell (approved):
This looks good. There are some sycl changes that are needed, but the sycl side needs work in general, I think, so I'm happy to leave that to a future PR from me. I also didn't look too closely through the ij/parcsr changes since I think those changes will be reviewed separately as part of https://github.com/hypre-space/hypre/pull/1307, correct?

src/sstruct_ls/ssamg_csolver.c
victorapm: This is for resetting the log level so we don't get extra info about memory usage from the IJ calls. The goal was to get memory info only at the beginning and end of SSAMG and BoomerAMG setup (4 calls in total)

src/sstruct_ls/ssamg_setup.c
victorapm: The previous answer also applies here

src/sstruct_mv/sstruct_matmult.c
victorapm: I was experimenting with both ways of computing the unstructured RAP on GPUs and how that affect performance. It turns out they are pretty similar, but the current default is slightly faster. I left the `urap2` check there in case we want to revisit this or expose this to users (as we do for BoomerAMG). 

Regarding the commpkg, it turns out that the RAPKT function calls it under the hood, while the MatMatKT doesn't. We should probably unify the behavior here, but I didn't do it in this PR

PR #1310, opened 2025-07-21
Make HYPRE_StructVectorSetValues signature consistent

liruipeng:
copied from #1279


rfalgout:
Can this wait until hypre-3.0?  We could make the change there instead, unless this is urgent.

PR #1312, opened 2025-07-22
Add Chebyshev solver

victorapm:
This PR exposes the existing Chebyshev relaxation from the `IJ` interface as a solver/preconditioner so that it can be used by MGR or other applications. Additional notes:

- Created getter/setter interface for all Chebyshev parameters
- Restructured the monolithic Chebyshev solver into separate setup and solve phases for improved maintainability
- Added Chebyshev as a preconditioner to krylov solvers (Chebyshev-PCG and Chebyshev-GMRES) with full integration into the main test driver and configuration parameters control
- Added regression tests with 20 different configurations covering sequential/parallel execution, and specific Chebyshev parameters
- Preserved all existing function signatures and interfaces to ensure backward compatibility


rfalgout:
Consider doing this as a general polynomial method similar to the Krylov methods.  If the API in ParCSR would not change, then we could continue with this in the interim.

victorapm:
Thanks, Rob! I checked the ParCSR APIs again and it doesn't seem they would change with the more general polynomial implementation.

@waynemitchell @oseikuffuor1  @rfalgout Therefore, I propose we work on reviewing and merging this for now keeping in mind that the internal Chebyshev implementation will be moved to a more general (interface agnostic) one as future work

rfalgout (approved):
Looks good.  Thanks!

oseikuffuor1 (approved):
Thanks Victor!

waynemitchell (approved):
Looks good. Thanks!

PR #1314, opened 2025-07-24
SYCL cmake fix

waynemitchell:
The cmake build with the sycl backend was broken due to `HYPRE_USING_CUDA_STREAMS` no longer being set. In addition the cmake commands to find oneDPL had been removed somewhere along the way.


victorapm (approved):
Thanks, Wayne!

PR #1315, opened 2025-07-25
Merge recmat into hypre-3.0

rfalgout:
This will merge the recmat branch into hypre-3.0 where develop will continue.  Since recmat is already up-to-date with master (and so is hypre-3.0), this should basically be a no-op.  So the review can be quick.


victorapm (approved):
Thanks, Rob!

waynemitchell (approved):
Yup, looks good to me! Thanks!

PR #1317, opened 2025-07-28
Merge mp-build-all into hypre-3.0

rfalgout:
This will merge the mp-build-all branch into hypre-3.0 where develop will continue.


rfalgout:
Hi @oseikuffuor1 and @victorapm .  This is the multiprecision branch merged with recmat.  I ran the `machine-tux-mixed.sh` and `machine-tux-recmat.sh` tests on it, and they should both be clean now (I initially had to fix a few things).

I would like to get this merged into hypre-3.0 sometime in the next couple of days.  Since we will have time to do further reviews and changes on the hypre-3.0 branch before final release, I don't think it needs an extensive review.  But, this is a good opportunity to see how the multiprecision work differs from recmat.

I would like @victorapm to focus mainly on the struct and semi-structured files that have changed.  Most of the changes overall are due to new files that you can largely ignore since they are auto-generated (files that have 'mup' in the name).

@oseikuffuor1 , this is largely what you've already seen, but the symbols in the struct and sstruct part had to change quite a bit as expected.

Let me know if you see anything that doesn't look right.  I've gone over it again myself, but it's a lot of code!  Thanks!

src/sstruct_ls/sstruct_ls_functions.saved
```diff
@@ -0,0 +1,489 @@
...
+hypre_sstructsyspfmgsolve
+hypre_AMR_FCoarsen
+hypre_AMR_RAP
+hypre_ZeroAMRMatrixData
+hypre_ZeroAMRVectorData
+hypre_AMR_CFCoarsen
+hypre_CFInterfaceExtents
+hypre_CFInterfaceExtents2
+hypre_CF_StenBox
+hypre_FAC_IdentityInterp2
```
victorapm: Should we re-generate this file in this PR? It doesn't seem we need some of the functions listed here anymore

src/sstruct_ls/sstruct_ls_mup_func.h
```diff
@@ -0,0 +1,508 @@
...
+#define hypre_ZeroAMRVectorData HYPRE_MULTIPRECISION_FUNC ( hypre_ZeroAMRVectorData )
+#define hypre_AMR_CFCoarsen HYPRE_MULTIPRECISION_FUNC ( hypre_AMR_CFCoarsen )
+#define hypre_CFInterfaceExtents HYPRE_MULTIPRECISION_FUNC ( hypre_CFInterfaceExtents )
+#define hypre_CFInterfaceExtents2 HYPRE_MULTIPRECISION_FUNC ( hypre_CFInterfaceExtents2 )
+#define hypre_CF_StenBox HYPRE_MULTIPRECISION_FUNC ( hypre_CF_StenBox )
+#define hypre_FAC_IdentityInterp2 HYPRE_MULTIPRECISION_FUNC ( hypre_FAC_IdentityInterp2 )
+#define hypre_FacSemiInterpCreate2 HYPRE_MULTIPRECISION_FUNC ( hypre_FacSemiInterpCreate2 )
+#define hypre_FacSemiInterpDestroy2 HYPRE_MULTIPRECISION_FUNC ( hypre_FacSemiInterpDestroy2 )
+#define hypre_FacSemiInterpSetup2 HYPRE_MULTIPRECISION_FUNC ( hypre_FacSemiInterpSetup2 )
+#define hypre_FAC_WeightedInterp2 HYPRE_MULTIPRECISION_FUNC ( hypre_FAC_WeightedInterp2 )
```
victorapm: Same comment as above (regenerate file)

rfalgout: I forgot to say that the `*.saved` files are no longer relevant.  I need to delete them.  Same holds for the `*func.h` headers and the `*func_header` scripts.  I suppose I should do that now.

@oseikuffuor1 Any objections to removing these now?  The `config/genMuP*.sh` scripts are not needed either.  Let me know.

victorapm: Thanks for the info, Rob!

src/struct_mv/struct_mv_mp.c
```diff
@@ -0,0 +1,243 @@
...
+ *****************************************************************************/
+
+/*--------------------------------------------------------------------------
+ * Mixed precision hypre_StructVectorCopy
+ * copies data from x to y
+ * if size of x is larger than y only the first size_y elements of x are
+ * copied to y
+ *--------------------------------------------------------------------------*/
+HYPRE_Int
+hypre_StructVectorCopy_mp( hypre_StructVector_mp *x,
```
victorapm, line 31: I'd like to revisit later whether we need a new struct vector type (`hypre_StructVector_mp`) or if we could augment the one we already have

src/struct_mv/struct_matrix.h
```diff
@@ -98,8 +98,56 @@ typedef struct hypre_StructMatrix_struct
...
+
+   hypre_CommPkg        *comm_pkg;                  /* Info on how to update ghost data */
+
+   HYPRE_Int             ref_count;
+
+#if defined(HYPRE_MIXED_PRECISION)   
+   HYPRE_Precision matrix_precision;
+#endif
+
+} hypre_StructMatrix_mp;
```
victorapm, line 149: Similar comment than the one for `hypre_StructVector_mp`

src/struct_mv/struct_mv_mp.c
rfalgout: I agree we should revisit this.  I don't think this is used right now, but I could be wrong.  We don't want the code replication if we can help it.  @oseikuffuor1 , we should chat about this at some point, maybe when we start thinking about bringing in the AMG work, if not before.

src/struct_ls/setrandom.c
```diff
@@ -0,0 +1,79 @@
+/******************************************************************************
```
victorapm: We might not need this file since `hypre_StructVectorSetRandomValues` is implemented in `struct_vector.c`

src/sstruct_ls/protos.h
```diff
@@ -5,398 +5,10 @@
...
-HYPRE_Int HYPRE_SStructSysPFMGSetDxyz ( HYPRE_SStructSolver solver, HYPRE_Real *dxyz );
-HYPRE_Int HYPRE_SStructSysPFMGSetLogging ( HYPRE_SStructSolver solver, HYPRE_Int logging );
-HYPRE_Int HYPRE_SStructSysPFMGSetPrintLevel ( HYPRE_SStructSolver solver, HYPRE_Int print_level );
-HYPRE_Int HYPRE_SStructSysPFMGGetNumIterations ( HYPRE_SStructSolver solver,
-                                                 HYPRE_Int *num_iterations );
-HYPRE_Int HYPRE_SStructSysPFMGGetFinalRelativeResidualNorm ( HYPRE_SStructSolver solver,
-                                                             HYPRE_Real *norm );
+HYPRE_Int hypre_SStructPVectorSetRandomValues ( hypre_SStructPVector *pvector, HYPRE_Int seed );
+HYPRE_Int hypre_SStructVectorSetRandomValues ( hypre_SStructVector *vector, HYPRE_Int seed );
+HYPRE_Int hypre_SStructSetRandomValues ( void *v, HYPRE_Int seed );
```
victorapm: These 3 should be in `sstruct_mv/protos.h`

src/struct_ls/setrandom.c
rfalgout: Good catch.  Let me look into that.

src/sstruct_ls/protos.h
rfalgout: I'll check

src/parcsr_ls/par_lr_restr.c
```diff
@@ -16,9 +16,12 @@
 //       hypre_dense_topo_sort(HYPRE_Real *L, HYPRE_Int *ordering, HYPRE_Int n)
 //    to get ordering for triangular solve. Can provide
 
-
+/* Global variable for error handling */
+/* guard definition of global variables to avoid linker errors for multiprecision build */
+#if defined hypre_DEFINE_GLOBAL
 HYPRE_Int AIR_TOT_SOL_SIZE = 0;
 HYPRE_Int AIR_MAX_SOL_SIZE = 0;
+#endif
```
victorapm: Minor comment: should these be static? I agree the macro also works. Just asking to understand what would be the preferred solution for this kind of situation

src/parcsr_ls/par_amg_setup.c
```diff
@@ -8,7 +8,7 @@
 #include "_hypre_onedpl.hpp"
 #include "_hypre_parcsr_ls.h"
 #include "par_amg.h"
-#include "../parcsr_block_mv/par_csr_block_matrix.h"
+#include "../parcsr_block_mv/_hypre_parcsr_block_mv.h"
```
victorapm:
```suggestion
#include "_hypre_parcsr_block_mv.h"
```

Minor suggestion... Works since we have `-I$(src_dir)/../parcsr_block_mv` in the Makefile

src/parcsr_ls/par_cycle.c
```diff
@@ -13,7 +13,7 @@
 
 #include "_hypre_parcsr_ls.h"
 #include "par_amg.h"
-#include "../parcsr_block_mv/par_csr_block_matrix.h"
+#include "../parcsr_block_mv/_hypre_parcsr_block_mv.h"
```
victorapm:
```suggestion
#include "_hypre_parcsr_block_mv.h"
```

src/sstruct_ls/protos.h
rfalgout: I removed two of them.  The third one was already in recmat with a to-do comment above it.  It will take a tad more work to fix it so let's wait.

src/parcsr_ls/par_cycle.c
rfalgout: Yes, this one is okay.  Fixed.  Note that there are some other directory-dependent includes that I want to remain, namely directories such as Euclid, ParaSails, etc. that don't really conform to our header file approach (and I don't see a reason to change the approach in these cases).  Thanks!

src/parcsr_ls/par_amg_setup.c
rfalgout: Fixed.  See other comment for par_cycle.c.

src/parcsr_ls/par_lr_restr.c
rfalgout: These aren't even used anywhere in hypre, so I removed them.

I think we need two approaches for file scope variables.  If they are truly global variables, then we need the hypre_DEFINE_GLOBAL so that only one copy is referred to across a multiprecision run.  If it is only used as file scope, declaring it static eliminates the link issue of having it multiply defined when using mixed precision and the code will function correctly as well.

src/struct_ls/setrandom.c
rfalgout: I had removed this from the Makefile but forgot to remove it from hypre.  Fixed.

victorapm (approved):
Thank you, Rob! I just added a few minor fixes for the CMake build (mixed precision option still needs work). I noticed this needs an astyle pass. Should we leave it to the hypre 3.0 branch? Thanks again!

rfalgout:
Thanks for the CMake fix, @victorapm !  I checked it at one point, but should have done more testing later.

rfalgout:
> Thank you, Rob! I just added a few minor fixes for the CMake build (mixed precision option still needs work). I noticed this needs an astyle pass. Should we leave it to the hypre 3.0 branch? Thanks again!

Let's do the astyle pass in hypre 3.0.  There may be some manual formatting fixes we'd like to do as part of that, and a separate PR could be useful for that.

src/sstruct_ls/sstruct_ls_mup_func.h
rfalgout: The original lists, scripts, and headers have been removed now.  @oseikuffuor1 stopped by my office.

src/config/writeHeader.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
```
oseikuffuor1: I think we can remove this file right?

rfalgout: Good catch!  Just did that.

oseikuffuor1 (approved):
@rfalgout this merge looks good to me. I added 1 minor comment for your review. I also have a few more comments/ questions that are related to the multiprecision build and should not impact this merge. It is probably easier to stop by your office to chat about those.

src/struct_mv/struct_mv_mp.c
oseikuffuor1: Happy to chat about this so see what other alternatives there are.

src/parcsr_ls/par_lr_restr.c
oseikuffuor1: True, we do not need to do anything special for static functions since they are contained (and only accessible) within a unique compilation unit for each precision build. We use this a fair bit in some of the folders already (utilities, krylov, etc) and we should take advantage of it in other places as well if applicable.

rfalgout:
Thanks @oseikuffuor1 and @victorapm for taking a look at this!  I'll merge it soon.

PR #1321, opened 2025-08-01
[hypre 3.0]: update mup scripts

victorapm:
- Change script shebang to use bash (default in hypre)
- Cosmetic changes to color output messages (OK, UPDATE...)
- Write down full path of `mup_check.err` file (in case of update), so we can just double-click and open it more quickly
- Fix current issue pointed out by `mup_check`: 
```
 ./mup_check > /dev/null 2>&1 && cat utilities/mup_check.err 
*** mup_check.old	2025-07-31 18:24:40.065460669 -0700
--- mup_check.new	2025-07-31 18:24:40.088460944 -0700
***************
*** 32,39 ****
  HYPRE_SetUseGpuRand
  HYPRE_Version
  HYPRE_VersionNumber
- _hypre_Free
- _hypre_MAlloc
  hypre_BeginTiming_fcn
  hypre_BigBinarySearch
  hypre_BigLowerBound
--- 32,37 ----
```


src/config/generate_function_list.sh
```diff
@@ -1,15 +1,16 @@
...
+# The script uses 'nm' and searches for functions labeled with 'T', i.e.,
+# symbol is in the text (code) section and is globally visible.
 
 # This prevents unmatched patterns from expanding (e.g., when there are no .obj files)
 shopt -s nullglob
 
-nm -P *.o *.obj | grep ' T ' | awk '{print $1}' | sed 's/^_//' | sed 's/_$//'
-
+# Use awk to avoid issues with spacing
+nm -P *.o *.obj | awk '$2 == "T" {print $1}' | sed -e 's/_$//'
```
rfalgout: Why are you doing this?  The original scripts that Daniel wrote tried to use the position, but it wasn't portable because the position is not always the same ($2, $3, and $4 appeared to all be possible).  This was done intentionally.

src/config/mup_check_dir.sh
```diff
@@ -29,9 +42,8 @@ fi
...
 
 SZ=`ls -l mup_check.err | awk '{print $5}'`
-if [ $SZ != 0 ]
-then
-   echo "   UPDATE - see mup_check.err"
+if [ "$SZ" != 0 ]; then
+    echo -ne "${RED}UPDATE${NC} - see $(pwd)/mup_check.err\n"
 else
-   echo "   OK"
+    echo -ne "${GREEN}OK${NC}\n"
```
rfalgout, line 48: I think this is fancier than it needs to be, but okay.

src/config/generate_function_list.sh
rfalgout: I just tried it on my mac, and I get failure (red UPDATE) for every directory.  I think it's the portability issue.

src/config/mup_check_dir.sh
rfalgout: Also, I don't know why the full path is needed.  Did you have a reason for that?  The check is done from the 'src' directory, so isn't it clear where to look?

src/config/generate_function_list.sh
rfalgout: Out of curiosity, did the script not work for you?  It works on my mac and my tux box, and it also works on Daniel's mac.  It's important to make this as portable as possible, but using position does not seem like a viable solution.

rfalgout:
What was the issue with hypre_Free and hypre_Malloc mentioned in your description above?

src/config/generate_function_list.sh
victorapm: Good point! This was in fact not portable. I just pushed a fix and verified it works on MacOS (Darwin) and linux

src/config/mup_check_dir.sh
victorapm: This was just so I could double-click on the file name and open it quickly. But I agree there's no strong reason for the change. Happy to revert!

victorapm:
> What was the issue with hypre_Free and hypre_Malloc mentioned in your description above?

I ran the following on my tuxbox using the hypre-3.0 branch

```
$ ./mup_check > /dev/null 2>&1 && cat utilities/mup_check.err 
```

and got the following error: 
```
*** mup_check.old	2025-07-31 18:24:40.065460669 -0700
--- mup_check.new	2025-07-31 18:24:40.088460944 -0700
***************
*** 32,39 ****
  HYPRE_SetUseGpuRand
  HYPRE_Version
  HYPRE_VersionNumber
- _hypre_Free
- _hypre_MAlloc
  hypre_BeginTiming_fcn
  hypre_BigBinarySearch
  hypre_BigLowerBound
--- 32,37 ----
```

This branch should fix this issue

src/config/generate_function_list.sh
```diff
@@ -1,15 +1,33 @@
...
+# The script uses 'nm' and searches for functions labeled with 'T', i.e.,
+# symbol is in the text (code) section and is globally visible.
 
 # This prevents unmatched patterns from expanding (e.g., when there are no .obj files)
 shopt -s nullglob
 
-nm -P *.o *.obj | grep ' T ' | awk '{print $1}' | sed 's/^_//' | sed 's/_$//'
+# Set arguments to "sed" based on the operating system for portability purposes
+OS_TYPE="$(uname -s)"
+case ${OS_TYPE} in
```
rfalgout: I still don't really like this fix (I'm not a fan of case statements if they can be avoided).  What was the original problem with the output from `nm` (what does the nm line look like before you pass it to sed)?  I assume this was prompted by the issue with hypre_Free and hypre_Malloc.  Also, I'm not sure the '$2' usage is going to be reliable, because I was having issues with this to begin with.

rfalgout: After a standard compile, what do you get if you type the following in `utilities`?
```
nm -P *.o *.obj | grep hypre_MAlloc
```

victorapm: Thank you for sharing these concerns, Rob!. I agree with the point you raised about the case statement, and I've given it more thought. I believe the case statement is necessary in this situation, for a specific reason that is tied to symbol mangling conventions on macOS and Linux, and a single `sed` command would fail on one of the platforms.

The problem lies with how the `sed` command handles function names that start with an underscore. On Mac, all global C functions have a leading underscore added by the compiler. We need to strip this to get the canonical function name. However, on Linux, a leading underscore is not a standard compiler convention, but we have it in a couple of private functions in hypre (e.g., `_hypre_Free`). In this case, stripping the leading underscore would incorrectly change the function's name (`_hypre_Free` -> `hypre_Free`).

The case statement is designed to handle this exact difference: On Darwin, we strip the leading and trailing underscores. While, on Linux, we only strip the trailing underscore, preserving the intentional leading underscore.

While I agree that a case statement adds complexity, I think it is necessary to correctly handle these platform-specific differences without losing important information.

Now regarding the change to `awk '$2 == "T"'`, this is a more robust and portable way to filter the `nm` output, as it relies on the POSIX-compliant output format of `nm -P` and guarantees that we have an output like `name type value size`, so `T` will always be in the second column and thus `awk '$2 == "T"'` should always work correctly.

Hope this is clear, please let me know if I'm missing something!

victorapm: > what do you get if you type the following in utilities?

I'm getting the following on Linux:
```
hypre_MAlloc U         
hypre_MAlloc U         
hypre_MAlloc U         
hypre_MAlloc U         
hypre_MAlloc U         
hypre_MAlloc U         
hypre_MAlloc U         
hypre_MAlloc U         
hypre_MAlloc U         
hypre_MAlloc U         
_hypre_MAlloc T 100 23
hypre_MAlloc T 5d0 32
hypre_MAlloc_core t 130 e5
```

and on Mac:
```
_hypre_MAlloc U 0 0
_hypre_MAlloc U 0 0
_hypre_MAlloc U 0 0
_hypre_MAlloc U 0 0
_hypre_MAlloc U 0 0
_hypre_MAlloc U 0 0
_hypre_MAlloc U 0 0
_hypre_MAlloc U 0 0
_hypre_MAlloc U 0 0
_hypre_MAlloc U 0 0
__hypre_MAlloc T 178 0
_hypre_MAlloc T 70c 0
_hypre_MAlloc_core t 1a8 0
```

rfalgout: I understand the issue better now.  Thanks!  Let's chat tomorrow, though (if you have time).  I am not 100% sold yet, and I think there are other options that may (or may not) be better.

victorapm: Sounds good, Rob! Another solution is to rename the underscore functions in hypre. I'm happy with this as well!

rfalgout: Renaming the functions is what I was going to suggest, but let's chat anyway.  I'll email you first.

.gitignore
```diff
@@ -71,6 +73,9 @@ src/test/structmat
 src/test/structmat.exe
 src/test/struct_migrate
 src/test/struct_migrate.exe
+src/test/test_mp
```
rfalgout, line 76: Note that Daniel is updating some things related to these new drivers.  Not all of them will remain.

victorapm: Sounds good! I wanted to get a clean "git status" with this change. Let me know if I should remove some of the drivers from gitignore

rfalgout: This can stay.  I just wanted to warn you that you may need to change it again.  Also, you both should be aware that you are working on multiprecision code at the same time.

rfalgout:
This looks good, but I would like to wait until it's finished before approving, so let me know when you think it's all done.  Thanks!

victorapm:
Hi Rob, this is ready for another review pass. Thank you!

rfalgout (approved):
This looks good.  Thanks!

victorapm:
@oseikuffuor1 I updated the mup scripts here. Hope this doesn't interfere with your work or cause any conflicts. Let me know!

oseikuffuor1:
Sure, Thanks @victorapm

PR #1322, opened 2025-08-01
[hypre 3.0]: Improve `mup_code`

victorapm:
Adds multi-threading to `mup_code` so it sweeps several folders in parallel.

Old version:
```
$ time ./mup_code > /dev/null

real	0m17.289s
user	0m15.059s
sys	    0m19.251s
```
New version:
```
$ time ./mup_code > /dev/null

real	0m7.021s
user	0m15.079s
sys	    0m25.013s
```


src/mup_code
```diff
@@ -35,8 +35,14 @@ DIRS="\
...
+process_directory() {
+    local dir="$1"
+    echo "Writing multiprecision code for $dir ..."
+    (cd "$dir" && ./mup_code)
+}
+
+export -f process_directory
+
+# Use xargs for parallel execution with number of CPU cores
+echo "$DIRS" | tr ' ' '\n' | xargs -P "$(nproc)" -I {} bash -c 'process_directory "$1"' _ {}
```
rfalgout: Is the underscore an end-of-line marker?

victorapm: This is actually a dummy var for `$0` when calling `bash -c`. It’s required so that `{}` becomes `$1` in the script and the folders get parsed correctly by `process_directory`

rfalgout: Ah, now I see.  It could be any string, not only underscore.  I was thinking this had a special meaning for some reason.  Thanks!

How portable do you think this will be?  It seems okay, but I don't have much experience with nproc.

rfalgout: After a little research on nproc, maybe it would be better to just hardcode a max proc number, say 4 or 6.  I'm not sure we get much benefit with higher numbers here anyway.

victorapm: Yes, it seems MacOS doesn't get it. Let me update

victorapm: I updated to `"${NPROC:-4}"` so it uses 4 cores by default and allow the user to overwrite it if needed

rfalgout: Sounds good.  On my mac, this takes about 1m20 with 10 procs, 1m10 with 4 procs, and 2m12 with 1 proc.  So, it seems that 4 procs is a reasonable default (and even the fastest for these runs).  Thanks!

rfalgout (approved):

PR #1311, opened 2025-07-22
GPU fixes

victorapm:
This PR modernizes hypre's GPU compatibility by addressing deprecated API and eliminating compilation warnings with newer CUDA versions while maintaining backward compatibility. A detailed list of changes is given below:

- Removed deprecated `thrust::binary_function` inheritance from 6+ functor classes and introduced `HYPRE_THRUST_IDENTITY` macro for cross-version compatibility
- Improved CUDA detection with `${CUDA_DIR}` parameter and simplified NVTX setup using modern CMake handling
- Migrated from deprecated `nvToolsExt` to `nvtx3` library and renamed `nvtx.c` to `device_markers.c` for better semantics
- Added conditional compilation for ROCSPARSE 3.4.0+ features and enhanced iterative ILU0 with `num_free_iter` parameter support

PR #1323, opened 2025-08-04
[hypre 3.0]: Struct optimizations

victorapm:
This PR contains several optimizations for PFMG and struct kernels in general.

The main idea is to fuse BoxLoop operations whenever possible and thus reduce the overhead associated with launching boxloops. These changes mainly impact device execution, but also benefit host-only execution. A summary of the main changes is given below:

- Add a fully fused kernel for `StructMatMult` that condenses the most common multiplication patterns (involving `FFC`, `FC`, and `F` terms) in a single box loop
- Add a third vector argument to `StructMatvec` such that we can compute `z = beta * y + alpha * A * x`. Allows computing residual vector in a single boxloop.
- Refactor Jacobi and PFMG Solve implementations.
- Increase the loop unrolling factor in `StructMatvec`  to 27, thus covering all stencil combinations present in PFMG.
- Add a `StructMatvec` kernel for variable coefficients plus a single constant coefficient (used by restriction)
- Add option to initialize struct matrix/vector entries to zeros. In GPU runs, we want to avoid device `memset` calls.
- Add special cases to `hypre_PFMGComputeCxyz` 
- Run `src/mup_code`

Extra:

- Fix usage of CMake build options: `-DHYPRE_ENABLE_THRUST_NOSYNC=ON` and `-DHYPRE_ENABLE_CUDA_STREAMS=ON`
- Add make target `make tagsnomup` to generate a `TAGS` file without references to auto-generated `mup` files
- CUDA 12.9 compilation fixes (PR #1311)

TODO:
- [x] Re-generate performance comparisons for PFMG on `master`, `hypre-3.0`, and this branch
- [x] Resolve regressions in `TEST_sstruct`


src/struct_mv/struct_matvec.c
```diff
@@ -435,50 +414,6 @@ hypre_StructMatvecCompute( void               *matvec_vdata,
             xp = hypre_StructVectorData(x);
             hypre_InitializeIndtComputations(compute_pkg, xp, &comm_handle);
             compute_box_aa = hypre_ComputePkgIndtBoxes(compute_pkg);
-
-            /*--------------------------------------------------------------
-             * Initialize y = -(beta/alpha)*y and multiply everything by -alpha
-             * at the end.  This optimizes the matvec for residual computations
```
rfalgout, line 441: Did you get improvements on the CPU as well for matvec?  This previous change (different from master) sped up the solve phase enough to be faster than master, simply by prioritizing residual computations, which make up our most common use case.

src/struct_mv/struct_matvec_core.c
```diff
@@ -14,12 +14,16 @@
 #include "_hypre_struct_mv.h"
 #include "_hypre_struct_mv.hpp"
 
-#define UNROLL_MAXDEPTH 7
+#define HYPRE_UNROLL_MAXDEPTH 27
```
rfalgout, line 20: Do you think there could ever be a downside to going "too deep" with unrolling?  I am thinking not (or at least hoping compilers are smart enough to not slow things down).

src/struct_mv/struct_matvec_core.c
```diff
@@ -107,262 +607,141 @@ hypre_StructMatvecCompute_core_CC( hypre_StructMatrix *A,
...
-#define DEVICE_VAR is_device_ptr(yp,xp)
-   for (si = 0; si < nentries; si += UNROLL_MAXDEPTH)
+   hypre_CopyToIndex(start, ndim, zdstart);
+   hypre_MapToCoarseIndex(zdstart, NULL, ran_stride, ndim);
+
+   /* Initialize output vector in the first pass */
+#ifdef HYPRE_CORE_CASE
+#undef HYPRE_CORE_CASE
+#endif
+#define HYPRE_CORE_CASE(n)                                                     \
```
rfalgout, line 624: Clever use of macros

rfalgout (approved):
This looks good, modulo comments and questions.  I am eager to see the results as well.  Thanks for all the hard work on this!

src/struct_mv/struct_matvec_core.c
victorapm: This reduced a fair amount of code and reduced the risk of mistakes with code repetition

src/struct_mv/struct_matvec.c
victorapm: Yes, it's about 20-30% faster on CPUs as well for large problem sizes. Timings in seconds for 100 residual computations (b - A*x) using 7-pt stencil on a single core on Dane:
| Cube | recmat-dev-5 | hypre-3.0 | master   | Recmat-dev-5/master |
|------|---------------|-----------|----------|----------------------|
| 32   | 0.010974      | 0.010927  | 0.009495 | 86.52%               |
| 48   | 0.033406      | 0.032544  | 0.032761 | 98.07%               |
| 64   | 0.082546      | 0.081822  | 0.081748 | 99.03%               |
| 96   | 0.236781      | 0.344268  | 0.317361 | 134.03%              |
| 128  | 0.897015      | 0.880943  | 0.871767 | 97.19%               |
| 160  | 1.477560      | 1.817620  | 1.744341 | 118.06%              |
| 192  | 3.042691      | 3.196230  | 3.207345 | 105.41%              |
| 256  | 6.497967      | 7.946542  | 7.978162 | 122.78%              |

I think the main advantage in this branch is that we can compute the residual in a single BoxLoop. Previously we had it split into 3 BoxLoops. First, `y = -beta/alpha * y`; second, `y -= A*x`. Third, `y = -alpha*y`

src/struct_mv/struct_matvec_core.c
victorapm: I did some testing for CPUs and the answer is yes. 27 is leading to register pressure and spilling. It seems 18-21 is the optimal spot, but let me do more testing. This likely depends on the CPU as well. We might want to have a good default and let the user overwrite it during config time if needed. For GPUs though, it seems even the matmult kernels, which involve more data, do not use all registers available. So we are safe with the high values I have in this branch. We will probably want different defaults according to CPU/GPU build. Let me do more testing

src/struct_mv/struct_matvec.c
rfalgout: Nice!

src/struct_mv/struct_matvec_core.c
rfalgout: Perfect, this sounds good.  Just thought I would ask.

waynemitchell:
I pushed a couple minor changes to get the sycl build compiling again. At runtime, I'm still having some issues, however. For some struct runs, I'm getting:

error: Total size of kernel arguments exceeds limit! Total arguments size: 2356, limit: 2048
in kernel: 'typeinfo name for hypre_StructMatmultCompute_fuse_fcc_fc_f(int, double*, double* (*) [3], double**, int*, int, double*, double* (*) [3], double**, int*, int, double*, double* (*) [3], double**, int*, int, int*, hypre_Box_struct*, int*, int*, hypre_Box_struct*, int*, int*, hypre_Box_struct*, int*, int*, int, double**, int*)::'lambda1'(sycl::_V1::nd_item<3>)'
error: backend compiler failed build.

I haven't had a chance to look closely at this yet (mostly just flagging this here as a reminder for the next time I take a look). Not sure whether it's easy/feasible to reduce the arguments here.

victorapm:
Thanks for the fixes, Wayne! I'd say for now we could macro-out the `fcc_fc_f` call in case of `HYPRE_USING_SYCL`. So, it will fall back to the other fuse implementations

src/IJ_mv/IJMatrix_parcsr_device.c
```diff
@@ -297,30 +297,26 @@ hypre_IJMatrixSetAddValuesParCSRDevice( hypre_IJMatrix       *matrix,
...
+   __device__ Tuple operator()(const Tuple& x, const Tuple& y ) const
    {
-      return thrust::make_tuple( hypre_max(thrust::get<0>(x), thrust::get<0>(y)),
-                                 thrust::get<1>(x) + thrust::get<1>(y) );
+#if defined(HYPRE_USING_SYCL)
+      using namespace std;
+#else
+      using namespace thrust;
+#endif
+      return make_tuple( hypre_max(get<0>(x), get<0>(y)),
```
liruipeng, line 316: Does `make_tuple` also work for sycl?

src/parcsr_ls/ams.c
```diff
@@ -495,26 +495,16 @@ HYPRE_Int hypre_ParCSRMatrixFixZeroRows(hypre_ParCSRMatrix *A)
  *--------------------------------------------------------------------------*/
 
 #if defined(HYPRE_USING_GPU)
-#if defined(HYPRE_USING_SYCL)
 struct l1_norm_op1
-#else
-struct l1_norm_op1 : public thrust::binary_function<HYPRE_Complex, HYPRE_Complex, HYPRE_Complex>
```
liruipeng, line 501: I guess this is not needed for CUDA anymore?

src/parcsr_ls/par_mod_multi_interp_device.c
```diff
@@ -51,8 +51,7 @@ struct globalC_functor
 };
 #else
 template<typename T>
-struct tuple_plus : public
-   thrust::binary_function<thrust::tuple<T, T>, thrust::tuple<T, T>, thrust::tuple<T, T> >
```
liruipeng, line 55: Same q. as above

liruipeng (approved):

src/IJ_mv/IJMatrix_parcsr_device.c
victorapm: Yes, sycl supports `std::make_tuple`

src/parcsr_ls/ams.c
victorapm: Right, `binary_function` was deprecated in cuda 11.1 (5y+ ago)

src/parcsr_ls/par_mod_multi_interp_device.c
victorapm: https://github.com/hypre-space/hypre/pull/1323#discussion_r2257440529

src/struct_mv/struct_matvec_core.c
victorapm: I added a 27pt option to the struct driver and compared 100 matvecs with different loop unrolling depths on a single core. Here are the results:
```
| Cube |    7     |    14    |    18    |    21    |    24    |    27    |
|------|----------|----------|----------|----------|----------|----------|
|  32  | 0.025239 | 0.020624 | 0.020939 | 0.020969 | 0.021992 | 0.031626 |
|  48  | 0.078284 | 0.068046 | 0.068340 | 0.068330 | 0.074114 | 0.098346 |
|  64  | 0.161471 | 0.167139 | 0.164641 | 0.164509 | 0.174526 | 0.187288 |
|  96  | 0.598661 | 0.620033 | 0.585222 | 0.574362 | 0.620007 | 0.656750 |
| 128  | 1.629558 | 1.549541 | 1.410765 | 1.376458 | 1.532878 | 1.687267 |
| 160  | 3.371172 | 2.973456 | 2.753991 | 2.728814 | 2.913064 | 3.267841 |
| 192  | 5.908839 | 5.172203 | 4.699686 | 4.763094 | 4.999501 | 5.629243 |
| 256  |14.064792 |12.096380 |11.548962 |11.157585 |12.216588 |13.089007 |
```

It seems 21 is a good choice, so I'm setting this as the new default to CPU builds. Users can also change it by passing `-DHYPRE_UNROLL_MAXDEPTH=N` as CFLAGS.

rfalgout: Makes sense.  Thanks!  BTW, you can use the sstruct driver easily to generate a 27-pt problem and run PFMG, so there is no need to add 27pt to the struct driver.

src/struct_mv/struct_matmult_fuse.c
```diff
@@ -3681,7 +4661,63 @@ hypre_StructMatmultCompute_fuse( HYPRE_Int nterms,
...
+                                            Mdbox, Mdstart, Mdstride,
+                                            Mnum_values, mmptrs,
+                                            &combined_fcc_fc_f);
+#endif
+
+   /* Call individual fuse functions */
+   if (!combined_fcc_fc_f)
+   {
+      /* First, we need to initialize Mptrs to zeroes */
+      hypre_Memset((void *) hypre_StructMatrixVData(M), 0.0,
```
waynemitchell, line 4697: I'm getting an error on this line with the sycl build on Aurora:

SYCL ERROR (code = Native API failed. Native API returns: 45 (UR_RESULT_ERROR_INVALID_ARGUMENT)) at memory.c:174
struct: memory.c:174: void hypre_UnifiedMemset(void *, HYPRE_Int, size_t): Assertion `0' failed.

Not sure what the issue is here. I'm just doing a simple single rank `./struct` run.

victorapm: Hmm, it says invalid arg... Let's try printing out the pointer address of `hypre_StructMatrixVData(M)` and the value of  `sizeof(HYPRE_Complex) * (Mdata_size - Mnum_values)`

src/struct_mv/_hypre_struct_mv_mup.h
```diff
@@ -713,6 +713,13 @@ hypre_BoxVolume_dbl( hypre_Box *box );
 HYPRE_Int
 hypre_BoxVolume_long_dbl( hypre_Box *box );
 
+HYPRE_Int
+hypre_BoxesEqual_flt( hypre_Box *box1, hypre_Box *box2 );
+HYPRE_Int
+hypre_BoxesEqual_dbl( hypre_Box *box1, hypre_Box *box2 );
+HYPRE_Int
+hypre_BoxesEqual_long_dbl( hypre_Box *box1, hypre_Box *box2 );
```
waynemitchell, line 721: Why do we need this function in different precisions?

victorapm: @oseikuffuor1 @rfalgout can you help with this?

waynemitchell (approved):
This all looks good to me, modulo the issues I'm having with the sycl build. I don't want to hold up this PR just for the sycl build because we are having various issues with that anyway...

src/struct_mv/struct_matmult_fuse.c
waynemitchell: All the arguments to memset and the memory locations all look fine... very strange. I'll ask my Intel contact and keep poking at this.

waynemitchell: OK, looking closer, I do have some questions about what's happening here. The matrix data is allocated at `hypre_StructMatrixData(M)`, and then the first few elements are reserved for the constant stencil entries, and the rest is for the variable entries, correct? `hypre_StructMatrixVData(M)` points to the data location plus the `hypre_StructMatrixVDataOffset(M)`. But what is `Mnum_values` here? It looks to be different than the offset, and in fact, it looks to me (from printing things out) that `Mnum_values` is smaller than the offset. This in turn causes the size of the memset to be longer than the allocation size of the matrix data, I think. Am I misreading what's going on here? Seems like this will cause some memory out of bounds issues.

victorapm: Thanks for looking at this carefully, Wayne. Right, `Mnum_values` is not quite right here. It refers to the number of variable stencil entries associated with the matrix. However, `hypre_StructMatrixVDataOffset(M)` might be different than it, e.g., in the case of symmetric matrices where we just store the triangular part. Using `hypre_StructMatrixVDataOffset(M)` is the right thing to do in this case of `memset`. Let me update this!

ulrikeyang (approved):
I have looked through the files. Since I am not familiar with the GPU stuff I can't really comment. I am relying on the rest of you. It looks like this was a lot of work! Thanks for all this effort. I approve this from my end and leave the rest to you guys.

victorapm:
Thanks Ulrike and all, there are a few more regressions errors on `TEST_sstruct` involving residual norm differences that I'm investigating. After I figure this out, we should be able to merge this PR

src/struct_mv/struct_matmult_fuse.c
victorapm: This is ready for another sycl test when you have time, Wayne. Thanks!

PR #1325, opened 2025-08-06
[hypre 3.0]: CMake build - mixed precision

victorapm:
This PR adds a CMake option: `-DHYPRE_ENABLE_MIXED_PRECISION` for turning on mixed precision


victorapm:
@oseikuffuor1, Profiling the autotools and cmake builds on a 16-cores machine:

autotools **with** mixed precision:
```
$ time (cd src && make clean && ./configure --enable-mixed-precision && make -j && cd test && make clean && make -j && cd ../..) >/dev/null

real	0m25.922s
user	5m9.313s
sys	    1m11.863s
```

autotools **without** mixed precision
```
$ time (cd src && make clean && ./configure && make clean && make -j && cd test && make clean && make -j && cd ../..) >/dev/null

real	0m16.439s
user	1m23.667s
sys	    0m21.459s
```

cmake **with** mixed precision:
```
$ time (cd build && cmake -DHYPRE_BUILD_TESTS=ON -DHYPRE_ENABLE_MIXED_PRECISION=ON ../src && make -j && cd ..) >/dev/null

real	0m41.496s
user	12m13.727s
sys	    1m15.090s
```

cmake **without** mixed precision:
```
$ time (cd build && cmake -DHYPRE_BUILD_TESTS=ON ../src && make -j && cd ..) >/dev/null

real	0m19.414s
user	2m36.957s
sys	    0m17.174s
```

CMake build is a bit slower in both cases (with and without mixed precision). We could probably investigate this in the future

victorapm:
@oseikuffuor1 Following up on this topic... The reason is CMake's build defaults to "O3" while autotools does "O2". So, cmake uses higher opt level and takes longer... If we force cmake to use "O2", it looks better:

cmake **without** mixed precision and **with** `O2`
```
$ time (cd build && cmake -DHYPRE_BUILD_TESTS=ON -DCMAKE_C_FLAGS_RELEASE="-O2" ../src && make -j && cd ..) >/dev/null

real	0m12.462s
user	1m58.861s
sys	    0m16.447s
```

cmake **with** mixed precision and `O2`:
```
$ time (cd build && cmake -DHYPRE_BUILD_TESTS=ON -DHYPRE_ENABLE_MIXED_PRECISION=ON -DCMAKE_C_FLAGS_RELEASE="-O2" ../src && make -j && cd ..) >/dev/null

real	0m29.537s
user	9m14.819s
sys	    1m9.675s
```

src/config/cmake/HYPRE_CMakeUtilities.cmake
```diff
@@ -592,3 +592,65 @@ function(print_option_status)
...
+
+  if(NOT REGULAR_SRCS)
+    message(FATAL_ERROR "SRCS argument is required for setup_mixed_precision_compilation")
+  endif()
+
+  # Create object libraries for each precision
+  add_library(${module_name}_flt  OBJECT ${REGULAR_SRCS})
+  add_library(${module_name}_dbl  OBJECT ${REGULAR_SRCS})
+  add_library(${module_name}_ldbl OBJECT ${REGULAR_SRCS})
+  add_library(${module_name}_def  OBJECT ${REGULAR_SRCS})
```
rfalgout: This '_def' compile no longer happens in the autotools build, but I forgot to remove it from Makefile.config (there are also a few remnants in 'struct_mv/Makefile' in the 'clean' target.  If you are just mimicking autotools here, then you can remove both.

src/test/test_mp_pcg.c
```diff
@@ -24,9 +24,13 @@
 #include "HYPRE_parcsr_mv.h"
 
 #include "HYPRE_IJ_mv.h"
+#include "HYPRE_IJ_mv_mup.h"
```
rfalgout: This shouldn't be here.  It is supposed to be included automatically by way of HYPRE_IJ_mv.h if you have HYPRE_MIXED_PRECISION defined and do *not* define hypre_MP_BUILD (see the bottom of the header file).  The same comment holds for the other '*_mup.h' includes here.  The user compile should have no additional mixed precision variables defined except for the first one.

rfalgout (approved):
Looks great, Victor!  Thanks!

src/test/test_mp_pcg.c
oseikuffuor1: @victorapm you can 'ignore' this file if you want, as I am working on cleaning up the tests.

oseikuffuor1 (approved):
@victorapm this looks good to me also. There seems to be an issue with .gitignore being deleted from the build directory when 'make distclean' is called, but that doesn't seem to be specific to this PR.

victorapm:
Thanks for the comments, Rob and Daniel. Everything should be addressed in 5f867bc

For the record, these are the new timings for the various builds (using the 16-cores machine):

| Build                        | Time [s] |
|------------------------------|----------|
| autotools (O2)               |  16.4        |
| autotools (mixed prec. + O2) | 25.9         |
| cmake (O2)                   |  12.4       |
| cmake (O3)                   |  19.4        |
| cmake (mixed prec. + O2)     |  22.7       |
| cmake (mixed prec. + O3)     |  32.1        |

PR #1334, opened 2025-08-18
[hypre 3.0] TEST_struct GPU regression fixes

waynemitchell:
Aiming to fix some failing regression tests on GPUs. I'm focusing here on the `TEST_struct` jobs. Remaining failing jobs:

- [x] cgpfmg3d.err.1
- [x] emptyproc.err.01a.b
- [x] emptyproc.err.04a.b
- [x] parflow.err.2
- [x] parflow.err.5
- [x] pfmgbase1d.err.1
- [x] pfmgbase2d.err.2
- [x] pfmgbase3d.err.5
- [x] pfmgbase3d.err.6
- [x] pfmgbase3d.err.7
- [x] pfmgtest1d.err.1


src/struct_mv/struct_matrix.c
```diff
@@ -2354,7 +2356,26 @@ hypre_StructMatrixReadData( FILE               *file,
...
+
+   /* Move values to the device memory if necessary and free host values */
+   if (hypre_GetActualMemLocation(memory_location) != hypre_MEMORY_HOST)
+   {
+      vi = 0;
+      hypre_ForBoxI(i, boxes)
+      {
+         box = hypre_BoxArrayBox(boxes, i);
+         vi += num_values * hypre_BoxVolume(box);
+      }
```
victorapm:
```suggestion
      vi = hypre_BoxArrayVolume(boxes) * num_values;
```

waynemitchell:
@victorapm , @liruipeng , I think this is getting close enough that y'all can take a look. Still tracking down a bug related to struct matmult in PFMG on multiple boxes, hoping that's the last one.

src/struct_ls/pfmg_setup.c
```diff
@@ -96,7 +96,9 @@ hypre_PFMGSetup( void               *pfmg_vdata,
 
    if (exec == HYPRE_EXEC_DEVICE)
    {
-      zero_init = 0;
+      /* WM: debug - this is screwing things up on lassen? */
+      /* zero_init = 0; */
+      zero_init = 1;
```
waynemitchell: @victorapm , this is a bit of a weird one to me. If I don't enforce a zero initialization, I get errors in some tests. Digging into it, it seems that there are some non-zero (uninitialized?) values in the data array for the vector, x, on the PFMG coarse grids that should be zero. But I'm also unsure about why these zero values are getting stored in the first place and where exactly in the grid they lie? In short, this needs to be revisited, but for now, simply turning off the zero_init optimization fixes some things.

victorapm: Hi Wayne, the temporary fix makes sense, thanks! The uninitialized ones are probably ghost zones. Ideally we would like to avoid memset for better performance, but let's leave it for now to get everything working

victorapm (approved):

PR #1339, opened 2025-08-21
[hypre 3.0]: Struct regression fixes

victorapm:


rfalgout:
Hi @victorapm .  Out of curiosity, do you have any idea about the performance hit we get reverting to CTAlloc here compared to TAlloc?  I guess I am wondering how important this was for you to get the really nice performance results you got for PFMG versus master.  Thanks!

victorapm:
Hi Rob, setup is about 20% slower on single GPU runs. In the case of multiple GPUs, we have MPI overheads so the relative importance of this goes down to 5-10%

src/struct_mv/struct_matmult_fuse.c
```diff
@@ -4693,11 +4693,6 @@ hypre_StructMatmultCompute_fuse( HYPRE_Int nterms,
    /* Call individual fuse functions */
    if (!combined_fcc_fc_f)
    {
-      /* First, we need to initialize Mptrs to zeroes */
-      hypre_Memset((void *) hypre_StructMatrixVData(M), 0.0,
-                   sizeof(HYPRE_Complex) * (Mdata_size - Mvdata_offset),
-                   hypre_StructMatrixMemoryLocation(M));
-
```
waynemitchell, line 4700: Why was this memset in here in the first place? Is it OK to just remove it?

src/struct_mv/struct_matmult.c
```diff
@@ -1314,7 +1314,7 @@ hypre_StructMatmultCompute( hypre_StructMatmultData  *mmdata,
    hypre_GpuProfilingPushRange("StructMatmultCompute");
 
    /* Set flag for initializing matrix/vector entries to zeroes */
-#if defined(HYPRE_USING_GPU)
+#if defined(HYPRE_USING_GPU) && defined(HYPRE_FUSE_FCC_FC_F)
```
waynemitchell, line 1317: What is the `HYPRE_FUSE_FCC_FC_F` flag? I think this is not defined, so this code path will no longer be taken, correct? I guess the code block here only controls the behavior of `zero_init`, which we disabled for now anyways...

waynemitchell (commented):
Thanks for the fixes, Victor. Just had a couple questions below.

src/struct_mv/struct_matmult_fuse.c
victorapm: This one is fine removing because now we are doing `CTAlloc` always regardless of the value of `zero_init`. It was here in the first place because the kernels below assume that `hypre_StructMatrixVData(M)` is initialized

src/struct_mv/struct_matmult.c
victorapm: It turns on the specialized fuse kernel `fuse_fcc_fc_f`. It's currently not being defined anywhere in the library, so yes this code path will no longer be taken. I left it here anyway so we can go back to this later and re-enable the `fuse_fcc_fc_f` kernel

src/struct_mv/struct_matmult_fuse.c
waynemitchell: OK. Do you want to comment it out instead and mark it with a TODO for when we revisit the zero_init feature? Up to you.

victorapm: I prefer to remove it altogether. We might need to do something different to get `zero_init` working properly...

PR #1342, opened 2025-08-23
Fix Wextra compilation warnings on CUDA/HIP builds

victorapm:


rfalgout (approved):
Looks good.  Thanks!  At some point, we should consider adding a compile regression test somewhere to ensure this is checked.  Not urgent.

victorapm:
Yes, that's a good idea. Thanks!

PR #1316, opened 2025-07-25
HYPRE 3.0

rfalgout:
This (draft) PR helped us track differences between the hypre-3.0 branch and master, and provided a mechanism for discussion as we continued to develop hypre-3.0.  It is now the final PR for the new release.

See issue #1370 

Some general guidance since this is such a big update with a target release deadline:
- Use separate PRs to make changes to the hypre-3.0 branch
- Make fast review turnarounds a priority (suggest alternate reviewers if you don't have time)
- Communicate - let people know what you are working on and consider potential merge conflicts
- Not all items on our big to-do list are required for this release - prioritize

See https://hypre.readthedocs.io/en/hypre-3.0/ for readthedocs user manual

See https://github.com/hypre-space/hypre/blob/hypre-3.0/src/docs/wiki-dev/Mixed-Precision.md for developer wiki

PR #1344, opened 2025-08-25
Remove files from library

victorapm:
See https://github.com/hypre-space/hypre/pull/1316#discussion_r2291428337


ulrikeyang (approved):
Thank you. This looks good to me.

rfalgout:
Hi @victorapm .  Thanks for setting this up!  I need to think about it a bit more before approving.  I'm not crazy about moving the multiprecision machinery into the test directory, but maybe that's how it needs to be.  I'll look this week when I get back.

victorapm:
Thanks, Rob! I wasn't sure about that either, but couldn't find a better solution for now. Thanks for looking! We can reevaluate this

oseikuffuor1:
@victorapm Thanks for this work! Just to add to the discussion, I also would prefer the multiprecision machinery to be outside of the test/ examples directories. I just think it is cleaner and perhaps makes things easier to maintain that way. That said, you have demonstrated how one can use the tools to enable the capabilities in the test directory, though there might be a lighter-weight option to do that. I think in the future, when the multiprecision build becomes the more standard build/ use of hypre, we may revisit this to streamline our regression testing. For now, my work on mixed-precision regression testing is keeping things local to the driver file. If we decide to build the helper functions in multiple precisions, it should be straightforward to adapt to use them.

rfalgout:
It looks like the only test driver that needed the different precisions for some of these Generate routines is `ij_mp.c`.

@oseikuffuor1 : Is the `ij_mp.c` driver something you plan on keeping around?

Since it doesn't look like runtime precision is needed, at a minimum I would change the `mup_code` script to use the `onlydef` option as we do in Euclid and a few of the other directories.  This will minimize the number of files generated and only make the fixed precision interface symbols available (i.e., `_flt`, etc.).  It will not generate `mup_*.c` files.

But, if the driver mentioned above is not needed, then we don't need the multi-precision machinery and we can just move the helper functions in the standard way.

ulrikeyang:
The ij_mp driver is needed for testing of single precision preconditioners with double precision solvers. It generates a single precision and a double precision matrix. However, one could just generate double precision matrix and convert it to a single precision matrix. That conversion routine is a true hypre routine which is located outside of the test directory. That would require providing these conversion routines in hypre-3.0, which we hadn't planned yet. Right now we just copy the vector back and forth for the mixed precision preconditioner/solver.

rfalgout:
Hi @victorapm .  I removed myself from the review request.  Add me back if you want to revisit this.  I've got some outstanding requests that seem stale so I'm trying the "remove" route to clean up the list. :)

PR #1345, opened 2025-08-26
[hypre 3.0] GPU sstruct regression fixes

waynemitchell:
Fixing sstruct regressions on lassen.

### DIFF
--- a/src/sstruct_mv/sstruct_matrix.c
+++ b/src/sstruct_mv/sstruct_matrix.c
@@ -875,15 +875,64 @@ hypre_SStructUMatrixInitialize( hypre_SStructMatrix  *matrix,
       nrows    = hypre_SStructGridLocalSize(grid);
    }
 
-   /* Set row sizes */
 #if defined(HYPRE_USING_GPU)
    if (exec == HYPRE_EXEC_DEVICE)
    {
+      /* Set max_size (row_sizes not needed for GPU execution) */
       row_sizes = NULL;
+      for (part = 0; part < nparts; part++)
+      {
+         pgrid = hypre_SStructGridPGrid(grid, part);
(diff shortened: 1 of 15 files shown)


src/sstruct_mv/sstruct_matrix.c
```diff
@@ -875,15 +875,64 @@ hypre_SStructUMatrixInitialize( hypre_SStructMatrix  *matrix,
       nrows    = hypre_SStructGridLocalSize(grid);
    }
 
-   /* Set row sizes */
 #if defined(HYPRE_USING_GPU)
    if (exec == HYPRE_EXEC_DEVICE)
    {
+      /* Set max_size (row_sizes not needed for GPU execution) */
       row_sizes = NULL;
+      for (part = 0; part < nparts; part++)
```
rfalgout: Hi @waynemitchell .  Not sure if you are ready yet with this PR, but I wanted to give you a quick reaction.  There is a lot of code duplication here, so if it's possible to list the replicated code once outside of the if/else construct, that would be preferable to me.  That is, minimize the specialized GPU code as much as possible.  Thanks!

waynemitchell: @rfalgout , yes, I agree. I was just trying to get something together quick to make sure it worked, and there are a few things entangled here. I'll try to do something more elegant for the final version.

waynemitchell:
I think the only remaining sstruct regression tests that are failing on lassen are all SSAMG tests, which I will leave for another PR.

rfalgout (approved):
Looks good, thanks!

victorapm (approved):
Thank you, Wayne!

liruipeng (approved):
Thanks

PR #1347, opened 2025-08-27
[hypre 3.0] struct regression fixes on Tuolumne

waynemitchell:
Fix regressions on Tuolumne. All the struct and sstruct regressions (except for ssamg) should be passing on both lassen and tuolumne now. Note that we need to generate some .saved files for Tuolumne, but I'm going to leave that for another PR focused on a larger revamp of what tests we do on that machine.


victorapm (approved):
Thanks, Wayne!

rfalgout (approved):
Thanks!

PR #1348, opened 2025-08-29
[hypre 3.0]: GPU build fixes

victorapm:
- Fix corner case during warp size detection when building hypre via `./configure --with-hip`
- Update `cudaMemPrefetchAsync` API to be compatible with CUDA 13
- Fix compilation warnings

### DIFF
--- a/src/config/update.sh
+++ b/src/config/update.sh
@@ -60,4 +60,4 @@ mv CMakeLists.txt.tmp CMakeLists.txt
 
 ##### Update release information in documentation
 
-#(cd docs; ./update-release.sh)
+(cd docs; ./update-release.sh)
--- a/src/config/update.sh
+++ b/src/config/update.sh
@@ -60,4 +60,4 @@ mv CMakeLists.txt.tmp CMakeLists.txt
 
 ##### Update release information in documentation
 
-(cd docs; ./update-release.sh)
+#(cd docs; ./update-release.sh)
--- a/src/configure
+++ b/src/configure
@@ -7735,11 +7735,11 @@ if test x$ac_prog_cxx_stdcxx = xno
 then :
   { printf "%s\n" "$as_me:${as_lineno-$LINENO}: checking for $CXX option to enable C++11 features" >&5
 printf %s "checking for $CXX option to enable C++11 features... " >&6; }
-if test ${ac_cv_prog_cxx_11+y}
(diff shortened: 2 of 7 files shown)


src/config/update.sh
```diff
@@ -60,4 +60,4 @@ mv CMakeLists.txt.tmp CMakeLists.txt
 
 ##### Update release information in documentation
 
-(cd docs; ./update-release.sh)
+#(cd docs; ./update-release.sh)
```
victorapm:
```suggestion
(cd docs; ./update-release.sh)
```

rfalgout (approved):

src/configure
```diff
@@ -7735,11 +7735,11 @@ if test x$ac_prog_cxx_stdcxx = xno
 then :
   { printf "%s\n" "$as_me:${as_lineno-$LINENO}: checking for $CXX option to enable C++11 features" >&5
 printf %s "checking for $CXX option to enable C++11 features... " >&6; }
-if test ${ac_cv_prog_cxx_11+y}
+if test ${ac_cv_prog_cxx_cxx11+y}
```
waynemitchell, line 7738: Looks like maybe there were some unintentional changes here? The build on lassen in breaking for me after this PR with:

nvcc fatal   : Value 'gnu++11' is not defined for option 'std'

Here's a sample bad compile command:

/usr/tce/packages/cuda/cuda-11.2.0/bin/nvcc -ccbin=mpixlC -std=gnu++11 -gencode arch=compute_70,code=sm_70  -g -O0 -lineinfo -expt-extended-lambda -std=c++17 --x cu -Xcompiler "-g "  -DHAVE_CONFIG_H -I.. -I./.. -I./../struct_mv -I.  -I/usr/tce/packages/cuda/cuda-11.2.0/include          -c device_markers.c -o device_markers.obj

victorapm: Let me check why `-std=gnu++11` is getting added. This shouldn't happen

@rfalgout this is probably related to your other issue

waynemitchell: @victorapm , yeah, I'm not sure why these lines got changed in the first place, but rerunning `config/update.sh` fixes these lines.

victorapm: We probably used different versions of autotools

Now I'm wondering, we shouldn't be checking for C++11 to begin with... 

Our default is 17 now for GPU builds

PR #1350, opened 2025-08-30
HYPRE 3.0 - Fixing compiler errors when complex is enabled

rfalgout:
Most solvers in hypre don't support complex numbers, and they would need a lot of work to do that (if even possible at all).  So, most solvers in hypre just use `HYPRE_Real`, which produces compiler warnings but not errors.

This PR:
* Changes `HYPRE_Complex` to `HYPRE_Real` in a number of places where errors were previously occurring.
* Includes header file `complex.h` in a few places where it was possible to support complex numbers.


src/struct_mv/struct_matrix.c
```diff
@@ -11,6 +11,7 @@
  *
  *****************************************************************************/
 
+#include <complex.h>
```
victorapm: Should we include this only when `HYPRE_COMPLEX` is defined in `HYPRE_config.h`?

victorapm (approved):
Thanks, Rob!

src/struct_mv/struct_matrix.c
rfalgout: Probably. Maybe the best way to deal with this is to put the if/include in utilities.h. I'll give that a go. Thanks!

rfalgout: The utilities idea created issues in the blas directory that wasn't worth the effort to try to fix, since our support for complex numbers is so basic right now.  So, I just added the include guard.

PR #1324, opened 2025-08-05
[hypre 3.0]: Minor changes for CUDA 13 support

nmnobre:
Hey, I've tried to follow the existing code structure, but I'm happy to change things as required. :)

PR #1353, opened 2025-08-31
[hypre 3.0]: CUDA 13 support

### LINKED ISSUE #1326: hypre with cuda 13+
Nvidia made a bunch of breaking changes in cuda 13. Notably, C++17 or newer is now required, and they changed the namespace of a lot of items. There also appear to be a number of miscellaneous API changes.

[Guide for CCCL 3.0 migration](https://nvidia.github.io/cccl/cccl/3.0_migration_guide.html)

victorapm:
Fixes a couple of issues involving CUDA 13 support. See https://github.com/hypre-space/hypre/pull/1324#issuecomment-3239392950

CUDA 13 and rocm 7.0 both require C++17. This PR bumps our minimum C++ standard to 17

Closes #1326

### DIFF
--- a/src/config/cmake/HYPRE_SetupGPUToolkit.cmake
+++ b/src/config/cmake/HYPRE_SetupGPUToolkit.cmake
@@ -6,21 +6,10 @@
 # Enable CXX language
 enable_language(CXX)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
-if(HYPRE_ENABLE_SYCL)
-  # We enforce the use of Intel's oneAPI DPC++/C++ Compiler
-  if(NOT CMAKE_CXX_COMPILER MATCHES "dpcpp|icpx")
-    message(FATAL_ERROR "SYCL requires DPC++ or Intel C++ compiler")
-  endif()
 
-  # Enforce C++17 at least for SYCL
-  if(NOT DEFINED CMAKE_CXX_STANDARD OR CMAKE_CXX_STANDARD LESS 17)
-    set(CMAKE_CXX_STANDARD 17)
-  endif()
-else()
-  # Enforce C++14 at least for CUDA and HIP
-  if(NOT DEFINED CMAKE_CXX_STANDARD OR CMAKE_CXX_STANDARD LESS 14)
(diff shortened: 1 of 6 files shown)


rfalgout (approved):

waynemitchell (approved):

liruipeng (approved):

src/config/cmake/HYPRE_SetupGPUToolkit.cmake
```diff
@@ -6,21 +6,10 @@
...
-  if(NOT DEFINED CMAKE_CXX_STANDARD OR CMAKE_CXX_STANDARD LESS 17)
-    set(CMAKE_CXX_STANDARD 17)
-  endif()
-else()
-  # Enforce C++14 at least for CUDA and HIP
-  if(NOT DEFINED CMAKE_CXX_STANDARD OR CMAKE_CXX_STANDARD LESS 14)
-    set(CMAKE_CXX_STANDARD 14)
-  endif()
+# Enforce C++17 at least
+if(NOT DEFINED CMAKE_CXX_STANDARD OR CMAKE_CXX_STANDARD LESS 17)
```
waynemitchell, line 11: @victorapm , is this overriding the c++ standard set by the user? I'm trying to use cmake on lassen, which does not support c++17 and requires c++11 instead. I tried setting:
-DCMAKE_CXX_STANDARD=11
on my cmake configure line, but it's still using c++17 at compile time.

victorapm: Yes, this is wrong, fixed on the umpire-default PR. Thanks!

Sbozzolo:
Leaving a note here in case anyone ends up with a google/github search.

This commit fixes a second CUDA compatibility issue.

If you are seeing errors like:
```
/opt/gcc/include/c++/14.3.0/bits/alloc_traits.h(445): error: variable template "std::allocator_traits<_Alloc>::<error>" is not a valid class member template
   __enable_if_t<!__has_construct<_Alloc, _Tp, _Args...>, bool>
```

This is because you are using CUDA 12.8+ and GCC 14+. This combination requires C++17.

PR #1355, opened 2025-09-01
Remove error checking from hypre_gselim

rfalgout:


rfalgout:
Tux tests are clean.  Will this affect GPU?

victorapm:
It doesn't affect GPUs, but now I'm thinking it's better we remove `error` altogether here. Then, we could remove some hacky code like this one: https://github.com/hypre-space/hypre/blob/4dd96d0e83088890879612c58364f6c10756ed90/src/sstruct_ls/node_relax.c#L767

rfalgout:
> It doesn't affect GPUs, but now I'm thinking it's better we remove `error` altogether here. Then, we could remove some hacky code like this one:

I agree, since even if we do decide to throw error codes in hypre_gselim, we should just call the standard error functions anyway (we won't need the error argument).  Let's see what others think about removing the error reporting first, then if we all agree, I will make the additional changes.  Thanks!

liruipeng (approved):
Thanks!

rfalgout:
Note: In `hypre_BoomerAMGRelaxT`, we should really be returning the hypre error flag, but I am worried about side effects for the release.  So, this just returns zero at the moment.

victorapm (approved):

waynemitchell (approved):

PR #1358, opened 2025-09-03
[hypre 3.0]: Cmake and autotools builds

victorapm:
CMake and autotools install the same header files in the `include` folder, except for mup headers.

Future work: write a script for checking the above holds true


rfalgout (approved):
Looks good.  Thanks!

PR #1359, opened 2025-09-03
HYPRE 3.0 replace hypre_printf with hypre_error_w_msg in _pre functions

rfalgout:
This only affects the default case in the `_pre` functions. The error flag is not returned. See the todo doc for an explanation.


victorapm (approved):
Thanks, Rob!

PR #1362, opened 2025-09-04
Don't free data if we don't own it

lindsayad:
I ran into this in PETSc where we have hypre alias some other internal PETSc matrix's memory. When trying to do a cudaMemcpy from that PETSc matrix I got an invalid argument and found that the device pointer had been unregistered by this free


victorapm:
Thanks, Alex! I put some additional logic in PR #1363 to free the other fields as well, according to what is being done in our CSRMatrixDestroy

lindsayad:
Closing in favor of #1363

PR #1363, opened 2025-09-05
[hypre 3.0]: Moose Fixes

victorapm:
Reworked fixes reported by the Moose team

- Multiple CUDA archs: #1303 
- Freeing data at CSRMatrixMigrate: #1362
- IJMatrixGetValues fixes: #1366

cc @lindsayad


rfalgout (approved):

lindsayad (commented):
Thanks for taking a look at this @victorapm! I was thinking ... since post-migration we've allocated data at a new memory location which the previous data owner has no knowledge about, I think we should re-take owenership
```
hypre_CSRMatrixOwnsData(A) = 1;
```
What do you think?

victorapm:
Hi Alex, yes, I agree that's the right thing to do. Thanks for bringing it up! Done in [279f05d](https://github.com/hypre-space/hypre/pull/1363/commits/279f05df765b27766a9e5b26d6bff82951a0c747)

lindsayad (approved):

liruipeng (approved):
Looks good to me! Thanks

victorapm:
@waynemitchell could you take a look at this PR, particularly the latest commit [0b02d0a](https://github.com/hypre-space/hypre/pull/1363/commits/0b02d0a7c89b9782ff2c05fe79710a8145d91ba5)?

It has a fix on the device path of `HYPRE_IJMatrixGetValues`

Thanks!

src/IJ_mv/IJMatrix_parcsr_device.c
```diff
@@ -1094,6 +1094,92 @@ hypre_IJMatrixSetConstantValuesParCSRDevice( hypre_IJMatrix *matrix,
...
+         HYPRE_Int l_row = g_row - row_start_;
+         if (g_col >= col_start_ && g_col < col_end_)
+         {
+            HYPRE_Int        l_col   = g_col - col_start_;
+            const HYPRE_Int *p_begin = diag_j_ + diag_i_[l_row];
+            const HYPRE_Int *p_end   = diag_j_ + diag_i_[l_row + 1];
+            const HYPRE_Int *p_found = p_end;
+            if (p_begin != p_end)
+            {
+               /* Device assembly puts diag. first and sorts remaining columns in increasing order */
```
waynemitchell: @victorapm , @liruipeng , @rfalgout , more of a general question, but is it always true that our ordering for parcsr matrix rows is diagonal first, then remaining columns in ascending order on both CPU and GPU? I feel like I've run into issues with this at different times throughout the code, particularly on the GPU, where we may or may not have sorting requirements for things like spgemm (or the results of such routines may not be sorted how we expect).

victorapm: That's a good point. When matrices are assembled (via IJMatrixAssemble), that is always true (confirmed with Rui Peng)

lindsayad:
```suggestion
               /* Device assembly puts diag first and sorts remaining columns in increasing order */
```

waynemitchell (approved):

PR #1364, opened 2025-09-08
Machine lassen fix

waynemitchell:
Trying to get the full machine-lassen.sh test script to pass on lassen.


src/parcsr_mv/par_csr_matop.c
```diff
@@ -4161,7 +4161,6 @@ hypre_ParTMatmul( hypre_ParCSRMatrix  *A,
    if ( hypre_GetExecPolicy2(memory_location_A, memory_location_B) == HYPRE_EXEC_DEVICE )
    {
       hypre_CSRMatrixMoveDiagFirstDevice(hypre_ParCSRMatrixDiag(C));
-      hypre_SyncComputeStream();
```
liruipeng, line 4164: why removing this line?

waynemitchell: If you look in `hypre_CSRMatrixMoveDiagFirstDevice()`, there is a sync at the end of that function already.

liruipeng: Ah, OK. Good point. Thanks!

victorapm (approved):
Thanks, Wayne!

rfalgout (approved):
Thanks!

src/test/TEST_ij/solvers.sh
```diff
@@ -156,24 +156,52 @@ if [ "$OUTCOUNT" != "$RUNCOUNT" ]; then
...
- ${TNAME}.out.210\
- ${TNAME}.out.211\
- ${TNAME}.out.212\
- ${TNAME}.out.213\
- ${TNAME}.out.404\
- ${TNAME}.out.405\
-"
+HOST=`hostname`
+case $HOST in
+   lassen*)
```
waynemitchell: This doesn't actually work for excluding these tests from the lassen runs like I thought it would... the tests still get run because they are listed in solvers.jobs, and so the resulting error messages (floating point errors) get copied into the solvers.err file. What's the proper way to exclude these tests?

rfalgout: There is no easy way with a case statement or something similar.  You can break the solvers.jobs into two jobs files, maybe one called mgr.jobs, then create a file in the AUTOTEST directory called runtests-ij-lassen that lists all of the jobs in TEST_ij except for mgr.jobs (see the runtests-ij-noilu as an example).  In the machine-lassen.sh file, you would run the ij-lassen jobs instead of ij.  Let me know if you want to chat.

liruipeng (approved):
Looks good. Thanks @waynemitchell

src/test/TEST_ij/solvers.sh
rfalgout: Just to follow up on this.  We basically have the flexibility to control which runtest tests we run on a given machine (i.e., which foo.jobs / foo.sh pairs), but we don't have the flexibility to control which runs within a test to run.

waynemitchell: OK, thanks, I'll put something together to get this working.

PR #1366, opened 2025-09-10
A couple of fixes for hypre_IJMatrixGetValuesParCSRDevice

lindsayad:
- I believe the construction of row_indexes must use an inclusive scan
- hypre device storage seems to always put the diagonal first which you have to keep in mind before using lower_bound which assumes a sorted array

@victorapm I'm working on a PETSc wrapper of hypre's get values routine and these are things I had to change in order to get my tests to pass with a device matrix


victorapm:
Hi Alex,

Thanks a lot for this contribution! We’re currently focused on final bug fixes for our upcoming release next week, so reviewing this might take us a little longer than usual. Do you think this PR could wait until after the release, or is it important to get it in right away? If it’s critical, we’ll do our best to prioritize it.

Thanks again,
Victor

lindsayad:
I think it's fine to have it wait

victorapm:
Hi Alex, I moved this to #1363 for a few additional tweaks. We will try to include this in the release. Thank you again!

lindsayad:
great!

PR #1369, opened 2025-09-11
mixed precision conversion

ulrikeyang:
I have added some mixed precision conversion routines for matrices and vectors.

### DIFF
--- a/src/seq_mv/seq_mv_mp.c
+++ b/src/seq_mv/seq_mv_mp.c
@@ -243,7 +243,7 @@ hypre_SeqVectorConvert_mp (hypre_Vector *v,
             {
                case HYPRE_REAL_DOUBLE:
                {
-                  data_mp = (hypre_double *) hypre_CAlloc_dbl ((size_t)size, (size_t)sizeof(hypre_double), memory_location);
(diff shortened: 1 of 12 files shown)


src/seq_mv/seq_mv_mp.c
```diff
@@ -216,4 +216,240 @@ hypre_SeqVectorAxpy_mp( hypre_double alpha,
...
+         default:
+            hypre_error_w_msg_mp(HYPRE_ERROR_GENERIC, "Error: Undefined precision type!\n");
+      }
+      hypre_Free_dbl(data, memory_location);
+      hypre_CSRMatrixData(A) = data_mp;
+      hypre_CSRMatrixPrecision(A) = new_precision;
+   }
+   return hypre_error_flag;
+}
+
```
oseikuffuor1, line 454: This looks good Ulrike. I would recommend removing the _dbl from the CAlloc and Free routines, since the default should work fine (they don't depend on precision), or you could use the correct _>flt,dbl,ldbl> if you want to be explicit.

oseikuffuor1 (approved):
Looks good Ulrike. I left a minor comment for your review.

PR #1371, opened 2025-09-20
[hypre 3.0]: Fix MSVC build

victorapm:
- Remove VLA instances from the library
- Add `-Werror=vla` to strict checking test to prevent future VLA additions


rfalgout (approved):
This is good for now, but I'd like to revisit it and the idea of using a pointer and utility functions for allocating and freeing variable-length arrays so we don't need to have a max array macro.  Thanks!

victorapm:
Thanks, Rob! Yes, that would be a nicer solution

PR #1372, opened 2025-09-20
[hypre 3.0]: Fix PFMG tests with cte. coefs

victorapm:
Fix regressions on `TEST_struct/pfmgconstcoeffmp.sh` and `TEST_struct/vdpfmgRedBlackGS.sh`

### DIFF
--- a/src/struct_ls/pfmg_setup_rap7.c
+++ b/src/struct_ls/pfmg_setup_rap7.c
@@ -202,7 +202,6 @@ hypre_PFMGBuildCoarseOp7( hypre_StructMatrix *A,
 
    HYPRE_Real           *rap_cc, *rap_cw, *rap_ce, *rap_cs, *rap_cn;
    HYPRE_Real           *rap_cb, *rap_ca;
-   HYPRE_Real            center_int, center_bdy;
 
    HYPRE_Int             OffsetA;
    HYPRE_Int             OffsetP;
@@ -413,12 +412,14 @@ hypre_PFMGBuildCoarseOp7( hypre_StructMatrix *A,
 
       else if ( constant_coefficient == 1 )
       {
-         rap_cb[0] = rap_ca[0] = a_cb[0] * pa[0];
-
-         rap_cw[0] = rap_ce[0] = 2.0 * a_cw[0];
-         rap_cs[0] = rap_cn[0] = 2.0 * a_cs[0];
-
-         rap_cc[0] = a_cc[0] - 2.0 * ( a_cw[0] + a_cs[0] - rap_cb[0] );
+         hypre_LoopBegin(1, k)
+         {
+            rap_cb[0] = rap_ca[0] = a_cb[0] * pa[0];
+            rap_cw[0] = rap_ce[0] = 2.0 * a_cw[0];
+            rap_cs[0] = rap_cn[0] = 2.0 * a_cs[0];
+            rap_cc[0] = a_cc[0] - 2.0 * ( a_cw[0] + a_cs[0] - rap_cb[0] );
+         }
+         hypre_LoopEnd()
       }
 
       else if ( constant_coefficient == 2 )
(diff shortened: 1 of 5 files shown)


src/struct_ls/pfmg_setup_rap7.c
```diff
@@ -441,7 +442,7 @@ hypre_PFMGBuildCoarseOp7( hypre_StructMatrix *A,
                              A_dbox, fstart, stridef, iA,
                              RAP_dbox, cstart, stridec, iAc);
          {
-            rap_cc[iAc] = 2.0 * a_cc[iA] + center_int;
```
rfalgout, line 444: The reason for defining center_int and center_bdy was to prevent unecessary computations inside a grid-based loop. Compilers should be smart enough to figure that out, but it would be good to verify. I suppose it would be less clean to do the center_int_h thing again here. Maybe we stick with this and give a closer look later, after the release? Same comment for the other places where this occurs.

src/struct_ls/red_black_constantcoef_gs.c
```diff
@@ -190,17 +185,14 @@ hypre_RedBlackConstantCoefGS( void               *relax_vdata,
...
-                  AApd = 1.0 / Ap[Ai];
-
                   hypre_RedBlackLoopInit();
 
 #define DEVICE_VAR is_device_ptr(xp,bp)
                   hypre_RedBlackConstantcoefLoopBegin(ni, nj, nk, redblack,
                                                       bstart, bni, bnj, bi,
                                                       xstart, xni, xnj, xi);
                   {
-                     xp[xi] = bp[bi] * AApd;
```
rfalgout, line 203: Here we used multiplication because it was much faster than division at the time. I'm not sure this is a thing anymore.

rfalgout: Also, why Ap[0] and not Ap[Ai]?

src/struct_ls/red_black_constantcoef_gs.c
```diff
@@ -356,12 +333,12 @@ hypre_RedBlackConstantCoefGS( void               *relax_vdata,
                         {
                            xp[xi] =
                               (bp[bi] -
-                               App0 * xp[xi + xoff0] -
```
rfalgout, line 359: Why not App0[Ai]?

rfalgout (approved):
This looks good modulo my questions.  Let me know.  Thanks!

src/struct_ls/pfmg_setup_rap7.c
victorapm: Good point! We can revisit this. The issue here was the array indexing happening outside of the boxloop region. In GPU runs, the arrays used to form `center_int` (and the other variables) live on GPU memory, so we can access them only inside the BoxLoop

src/struct_ls/red_black_constantcoef_gs.c
victorapm: I think division should be fine here because we're limited by memory bandwidth anyway.

I double-checked and `Ap[0]` is correct because we are in the (all) constant coefficients case here.

src/struct_ls/red_black_constantcoef_gs.c
victorapm: In this branch we just have the diagonal coefficient as variable, while the others are constant. So, all the `App`s have to be accessed via `App[0]`

src/struct_ls/red_black_constantcoef_gs.c
rfalgout: Okay, I see now.  hypre_CCBoxIndexRank is always 0.  Thanks!

rfalgout: BTW, it does seem like there is still a significant difference in speed between floating point division and multiplication on modern hardware. But I agree that it probably won't make much difference to the bottom line for us. Note that I think the compiler is smart enough to only do the division once if you declare a variable inside the LoopBegin loop and set it equal to 1 over some loop-invariant value. So, it might be as easy as pulling the original code inside the loop. Just a thought. We can still worry about this later.

victorapm:
Hi Rob, I''ll merge this, but let's revisit the points you raised after the release. Thanks again!

PR #1374, opened 2025-09-20
[hypre 3.0]: Fix hypre headers

### LINKED ISSUE #898: Installed header files with name conflicts
When installing Hypre, some header files with fairly generic names are installed to the top level `INSTALL_PREFIX/include` directory which could lead to conflicts. For example `fortran.h` and `krylov.h`.

Would it be possible to rename these files to have some prefix as most of the other Hypre headers do, or to instead install them to a subdirectory at `INSTALL_DIR/include/hypre`? Or perhaps there's another preferred approach for avoiding these conflicts that could be taken. Thank you very much for your help!

victorapm:
This PR fixes our headers installation such that all private headers start with `_hypre_` and public ones start with `HYPRE_`.

If we decide to merge this before the hypre-3 release, we'll need to implement a few minor corrections on MFEM.

Closes #898


src/utilities/Makefile
```diff
@@ -123,14 +122,14 @@ all: libHYPRE_utilities${HYPRE_LIB_SUFFIX}
 	cp -fR $(srcdir)/_hypre_*.h $(HYPRE_BUILD_DIR)/include
 	cp -fR $(srcdir)/_hypre_utilities.hpp $(HYPRE_BUILD_DIR)/include
 	cp -fR $(srcdir)/_hypre_onedpl.hpp $(HYPRE_BUILD_DIR)/include
-	cp -fR $(srcdir)/fortran*.h $(HYPRE_BUILD_DIR)/include
+	cp -fR $(srcdir)/_hypre_fortran*.h $(HYPRE_BUILD_DIR)/include
```
rfalgout: Do we need this line at all?  Same question below.

rfalgout:
It looks like you are accidentally installing multiprecision.h in the cmake build.

rfalgout (approved):
This was a lot of work!  It looks great, and I definitely think we should include it in the hypre-3.0 release!  Thanks!

src/utilities/Makefile
victorapm: Thanks for catching. The earlier line comprises this case. Let me remove

victorapm:
Thanks, Rob! Just incorporated your comments and will merge this into hypre-3

PR #1377, opened 2025-09-22
[hypre 3.0]: Fix MGR issues

victorapm:
Fix MGR regressions on lassen/tioga

A few tests were using a BoomerAMG relaxation method that is not GPU-enabled. This PR updates the default AMG relaxation in these tests to a GPU-enabled one.

Also, `hypre_MGRSetupFrelaxVcycleData` was relying on UVM to work. This PR makes it general to the non-UVM case.

@oseikuffuor1 I would like to revisit this internal function in MGR. Now that we can specify the `A_FF` solver as an object outside MGR, it seems we don't need `hypre_MGRSetupFrelaxVcycleData` anymore

### DIFF
--- a/src/parcsr_ls/par_mgr.c
+++ b/src/parcsr_ls/par_mgr.c
@@ -912,6 +912,8 @@ hypre_MGRSetReservedCoarseNodes(void      *mgr_vdata,
 
 /*--------------------------------------------------------------------------
  * Set CF marker array
+ *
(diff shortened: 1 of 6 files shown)


oseikuffuor1 (approved):
@victorapm this looks good to me thanks. Yes we should revisit the vcycle relaxation routine after the release and remove it if it is not really needed anymore.

rfalgout (approved):
Looks good.  Thanks!

PR #1378, opened 2025-09-22
[hypre 3.0]: Update regression tests baselines

victorapm:
Update performance baselines and a few residual norm differences

PS: for the performance tests, we should probably update the results after running with umpire enabled


rfalgout (approved):
Looks good, especially PFMG.  Thanks!

PR #1379, opened 2025-09-23
[hypre 3.0]: Fix CMake's generated soversion

victorapm:
Change how `SOVERSION` is set to encode major and minor versions as a single integer (e.g., 2.33.0 → 233, 3.0.0 → 300). 

The current code in the `hypre-3.0` branch is translating `3.0.0` to `30`.

This ensures the library’s ABI version increases consistently with future releases.


rfalgout:
Regarding the translation, what about the third digit (patch).  What happens if we release hypre-3.0.1, for example.

Note that in `config/version.sh` we assume 2 digits max in each component (major, minor, patch):
```
hypre_number="$hypre_major*10000 + $hypre_minor*100 + $hypre_patch"
```
So, in `configure.in` the `HYPRE_NUMBER` is 30000 for `HYPRE_VERSION` 3.0.0. Not sure if this is relevant here.

rfalgout (approved):
Just had the one comment.  Thanks!

victorapm:
Good point about the patch... I'm assuming that patch releases are ABI-compatible, i.e., if we release `3.0.1` it should be ABI compatible with `3.0.0`, but not necessarily with `3.1.0`.

In other words, we would treat major and minor as not necessarily ABI compatible, but same major and minor with different patches would be ABI compatible (so no need to change SOVERSION for patch).

Let me know if this sounds correct, if not we will need to do something like `HYPRE_NUMBER` for `SOVERSION`

rfalgout:
I think that makes sense.  We probably should add a note to our Releases wiki about ensuring ABI compatibility for patch releases (otherwise, a minor release should be done instead).

victorapm:
Hi Rob, I just updated `INSTALL.md` with comments about ABI compatibility. Please, let me know if it looks good, thanks!

rfalgout:
> Hi Rob, I just updated `INSTALL.md` with comments about ABI compatibility. Please, let me know if it looks good, thanks!

I'm not sure I want to go as far as guaranteeing ABI compatibility until we discuss this more as a team. To do that, I think we would need to do something extra as part of the release process. I am okay with saying that our *intention* is for patch releases to be compatible within the same major.minor, but let's not promise too much on this quite yet.

victorapm:
Thanks, Rob! Just updated

rfalgout:
> Thanks, Rob! Just updated

Looks good.  Thanks!

PR #1380, opened 2025-09-23
Add CMake option to fetch and build Umpire

### LINKED ISSUE #1404: Finding CUDA with cmake
In https://github.com/hypre-space/hypre/blob/21845bffdf74f74370707f2d174b276454646774/src/config/cmake/HYPRE_SetupCUDAToolkit.cmake#L8-L32
`hypre` tries to set `CUDA_DIR` starting from other variables and heuristics. 

Later on, `hypre` uses CMake's `find_package` to find the CUDA toolkit. 
https://github.com/hypre-space/hypre/blob/21845bffdf74f74370707f2d174b276454646774/src/config/cmake/HYPRE_SetupCUDAToolkit.cmake#L69

I don't quite understand what is the purpose of `CUDA_DIR` in this second call, but what I am finding is that this current logic seems to defy part of the point of `find_package` (ie, _finding_ the package): if I don't pass a config that specifies CUDA_DIR, cmake fails with 
```
"CUDA_PATH or CUDA_HOME not set. Please set one of them to point to your CUDA installation."
```
so `find_package(CUDAToolkit)` is never called, even if `find_package(CUDAToolkit)` would have found my CUDA and set things up correctly.

This forces me to pass `CUDAToolkit_ROOT` to `hypre`, even when this was found with `find_package(CUDAToolkit)` somewhere else. 

I am guessing that what should happen is that `hypre` should not fail with a fatal error when `CUDA_DIR` cannot be set with the current heuristics, but should try to invoke `find_package(CUDAToolkit)` first.

victorapm:
The goal of this PR is to make it easier for users to build hypre with Umpire support.

 Details below:

- Added `HYPRE_BUILD_UMPIRE` option to auto-fetch and build Umpire via CMake's FetchContent (with blt, camp, fmt); installs into hypre’s prefix.
- Not added for the autotools build since AFAIK there isn't something like `FetchContent` to make this feature easier to implement.
- Normalize BLT’s cuda_runtime to CUDA::cudart across Umpire/camp targets.
- Version fix: sanitize Umpire’s config.hpp.in placeholders to avoid leading-zero issues.
- Better diagnostics: concise multi-line error message suggesting (1) HYPRE_BUILD_UMPIRE, (2) umpire_ROOT/umpire_DIR, (3) TPL_UMPIRE_*.
- Improve method for finding CUDA libraries. Closes #1404 
- Update documentation with instructions about `HYPRE_BUILD_UMPIRE`

Example:

```
cmake -DHYPRE_ENABLE_CUDA=ON -DHYPRE_BUILD_UMPIRE=ON -B build -S src
cmake --build build --parallel
```


rfalgout:
I would prefer to hold off on this one until after the release.

victorapm:
A common issue with user-provided Umpire builds is the absence of `C` interfaces in Umpire (turned off by default). These can be enabled via `-DUMPIRE_ENABLE_C=ON`.

https://github.com/hypre-space/hypre/pull/1380/commits/e285c092b12042e1e589f8c065ad8ee75312d59d improves the robustness of our build to check at config time whether Umpire's `C` interfaces are available. If not, it instructs users to enable them.

cc @helloworld922

waynemitchell:
Do folks have thoughts about having the auto-build of umpire happen by default? In the current version of the branch, if the user doesn't have an existing (properly configured) umpire install, then the configure command errors and tells the user to either let hypre build umpire or point hypre to an umpire install. What if instead the build system automatically enabled HYPRE_BUILD_UMPIRE whenever umpire is turned on and no existing umpire install is found? I can see arguments for either behavior, but I am maybe leaning towards the latter (fewer steps/less error prone)?

victorapm:
I think that's a good idea and improves user experience overall. If, for some reason, the auto build fails, then we could show an error message inviting folks to build Umpire externally and pass it to hypre

oseikuffuor1 (approved):
Thanks Victor!

Chiming in on comment by @waynemitchell, I also see arguments for either behavior. However, I am leaning towards the current behavior, which alerts the user to let hypre build UMPIRE or perhaps that they have entered an incorrect path to their UMPIRE install. The user can pass the appropriate path or request hypre to install UMPIRE if it indeed missing. If hypre requires some specific configuration of UMPIRE that is not standard, then perhaps installing it by default makes sense.

waynemitchell (approved):
I've tried this on a couple of systems now, and it seems to work great. I can't say I read carefully through all the cmake code here... I like this feature! Thanks, Victor!

Sbozzolo:
Hi! I am dropping by to say that I am finding the `fixup_umpire_cuda_runtime` function very useful in upgrading dependencies in [Palace](https://github.com/awslabs/palace). Thank you!

victorapm:
Thanks for letting us know, @Sbozzolo !

FYI all, that CMake function fixes a linker issue that happens when building umpire with CUDA and shared libs. The fix on BLT (upstream) seems to be in the works....

ulrikeyang (approved):
This looks like a lot of work, and I like all the messages included, but I can't really evaluate if there are any errors.

rfalgout (approved):
I did not take this for a test spin, but I like the idea and don't want to hold up the PR. My main concern is maintainability in the future because there is quite a bit of cmake code here focused just on supporting Umpire. Thanks!

victorapm:
Hi Rob, I understand the concern and agree. Some of this code will go away after the shared libs build for umpire is fixed upstream. I might also be able to simplify a few other things and will revisit everything after we get the upstream fix

liruipeng (approved):
Looks good to me!

PR #1381, opened 2025-09-23
 Add CMake check for AMDGPU architecture 

victorapm:
This changes explicitly avoids building hypre for CDNA and RDNA cards at the same time.

I'm marking this as "After 3.0"


rfalgout (approved):
Didn't try to understand this, but I'm sure it's good.  Thanks!

victorapm:
Context: CDNA uses a wavefront size of 64, while RDNA uses 32. Since we don't support multiple wavefront sizes in hypre, this check is just preventing problems ahead of time

PR #1382, opened 2025-09-23
Add machine matrix

victorapm:
Add test script for LC's matrix

TODO:
- [x] cmake-cuda-um-dbg
- [x] cmake-cuda-um-mixedint
- [x] cmake-cuda-um-shared
- [x] cmake-cuda-um-single
- [x] cmake-cuda-um-without-MPI
- [x] cmake-cuda-cpu 
- [x] cmake-cuda-bench
- [x] Autotools (same as CMake but without runs)
- [x] cmake-cuda-cpu baselines (Not needed since we rely on the `.saved` files used by tux)

Notes:
 - Internal code changes are needed to get the tests working when building hypre with GPU support on a machine without GPUs (e.g., dane) while setting the execution policy to host.
 - hypre's internal memory tracker was disabled temporarily due to crashes on matrix. To be investigated.
 - Residual tolerance for a couple of `TEST_ij/solvers.sh` runs was lowered so we can reuse the same saved files as tux when building with GPU support but setting the execution policy to host
 - Runs on a single node of matrix. Currently takes about 6h (to be improved)
 - Example job script for launching on matrix below:

```bash
#!/bin/bash
#SBATCH -t 10:00:00
#SBATCH -p pbatch
#SBATCH -J hypre
#SBATCH -o machine-matrix.out
#SBATCH -e machine-matrix.err
#SBATCH -N 1
#SBATCH -G 4
#SBATCH --exclusive

### Shell scripting
date; hostname; git branch --show-current
pwd=`pwd`
cd $(pwd)/AUTOTEST
echo -e "Current directory: $(pwd)"

### Launch parallel executable
./test.sh ./machine-matrix.sh ../src

date;
echo 'Done'
```


rfalgout:
Hi @victorapm .  I took a quick look through this, even though I'm technically not a reviewer. :)  It generally looks good to me.  The only question I have is if you've diff'ed these saved files with ones from another machine to make sure there is nothing out of the ordinary.  A few of the saved files had final residual numbers that were unexpected at first glance (I didn't dig into anything further).  Thanks!

victorapm:
Hi Rob, I was going to ask your review here after this is complete, so thanks for early input!

I haven't done that yet, but sounds a good idea to me. Could you point out a couple of examples where you saw the unexpected residual numbers? I'll take a closer look at those

src/test/TEST_sstruct/cycred.saved.matrix
```diff
@@ -0,0 +1,81 @@
...
+Final Relative Residual Norm = 0.000000e+00
+
+# Output file: cycred.out.1Dx.1
+Final Relative Residual Norm = 0.000000e+00
+
+# Output file: cycred.out.1Dx.2
+Final Relative Residual Norm = 0.000000e+00
+
+# Output file: cycred.out.3Dx.5
+Final Relative Residual Norm = 3.200000e+02
```
rfalgout, line 53: Here is one example.  The exact zero results in this file also stand out, though they could be just fine.

victorapm: In this case, the results on `matrix` (this file) match the baselines used in `tux` (cycred.saved). I agree these residual values look weird though...

victorapm: Rob, I checked the other `saved.matrix` files against the corresponding `saved.aurora` and they all look similar (a couple of iterations different in some cases and residual norms in the range of the requested error tolerance).

src/test/TEST_ij/solvers.saved
```diff
@@ -222,8 +222,8 @@ GMRES Iterations = 23
 Final GMRES Relative Residual Norm = 3.979845e-09
 
 # Output file: solvers.out.113
-GMRES Iterations = 25
-Final GMRES Relative Residual Norm = 8.744056e-09
+GMRES Iterations = 10
```
rfalgout, line 225: What was the reason for this drop?

victorapm: I changed the requested tolerance to `1e-3` so the residual norm computed on matrix and tux match to the precision printed here. This way, we can reuse the same `.saved` files from tux when running on matrix with host execution

rfalgout (approved):
Looks good to me.  Thanks!

src/test/TEST_ij/io.jobs
```diff
@@ -77,5 +77,5 @@ mpirun -np 4 ./ij -solver 2 -tol 1e-2 -printbin \
 mpirun -np 1 ./ij -printcsr > io.out.200
 mpirun -np 1 ./ij -fromonecsrfile csr.out.A.D.0 -rhsfromonefile csr.out.b.0 -x0fromonefile csr.out.x0.0 > io.out.201
 mpirun -np 3 ./ij -fromonecsrfile csr.out.A.D.0 -rhsfromonefile csr.out.b.0 -x0fromonefile csr.out.x0.0 > io.out.202
-mpirun -np 5 ./ij -fromonecsrfile csr.out.A.D.0 -rhsfromonefile csr.out.b.0 -x0fromonefile csr.out.x0.0 > io.out.203
-mpirun -np 7 ./ij -fromonecsrfile csr.out.A.D.0 -rhsfromonefile csr.out.b.0 -x0fromonefile csr.out.x0.0 > io.out.204
+mpirun -np 4 ./ij -fromonecsrfile csr.out.A.D.0 -rhsfromonefile csr.out.b.0 -x0fromonefile csr.out.x0.0 > io.out.203
```
liruipeng, line 80: What is the reason of changing np?

liruipeng (approved):

src/test/TEST_ij/io.jobs
victorapm: NP 5 and 7 gave me issues on matrix related to the GPUs scheduling. I will get back to this

PR #1386, opened 2025-09-27
Add `HYPRE_ENABLE_FORTRAN` option to CMake build

victorapm:
We talk about it in the documentation, but it was not implemented yet.

This option is mainly for enabling Fortran tests and examples. There's no code that gets compiled with Fortran and added to the library.

cc @helloworld922


rfalgout (approved):
We probably need to take another look at what we are doing to support fortran. These drivers aren't even tested, I think. Also, some of the test drivers are left over from an earlier project that was not completed.

victorapm:
Hi Rob, one option for now is to add `HYPRE_ENABLE_FORTRAN` to one of the tux tests, just to make sure the Fortran examples are compiled correctly. I agree we should discuss further testing, e.g., checking the output...

PR #1387, opened 2025-09-28
Fix issue with `TEST_ij/io`

victorapm:
Fix the following `TEST_ij/io` job when hypre is built with bigint support.

Not sure why `machine-tux.sh` didn't catch this before.

```
$ mpirun -np 2 ./ij -fromonecsrfile csr.out.A.D.0 -rhsfromonefile csr.out.b.0 -x0fromonefile csr.out.x0.0

Hypre init times:
Hypre init:
  wall clock time = 0.000002 seconds
  wall MFLOPS     = 0.000000
  cpu clock time  = 0.000002 seconds
  cpu MFLOPS      = 0.000000


Using HYPRE_DEVELOP_STRING: v3.0.0 (branch master)

  solver ID    = 0

  FromFile: csr.out.A.D.0
Generate Matrix:
Spatial Operator:
  wall clock time = 0.001078 seconds
  wall MFLOPS     = 0.000000
  cpu clock time  = 0.001078 seconds
  cpu MFLOPS      = 0.000000

  Number of vector components: 1
  RHS vector read from file csr.out.b.0

===================================================================================
=   BAD TERMINATION OF ONE OF YOUR APPLICATION PROCESSES
=   PID 1583363 RUNNING AT tux408
=   EXIT CODE: 9
=   CLEANING UP REMAINING PROCESSES
=   YOU CAN IGNORE THE BELOW CLEANUP MESSAGES
===================================================================================
YOUR APPLICATION TERMINATED WITH THE EXIT STRING: Segmentation fault (signal 11)
This typically refers to a problem with your application.
Please see the FAQ page for debugging suggestions
```

### DIFF
--- a/src/parcsr_mv/par_vector.c
+++ b/src/parcsr_mv/par_vector.c
@@ -53,6 +53,9 @@ hypre_ParVectorCreate( MPI_Comm      comm,
       partitioning[1] = partitioning_in[1];
    }
    local_size = (HYPRE_Int) (partitioning[1] - partitioning[0]);
+   hypre_assert(local_size >= 0);
+   hypre_assert(partitioning[0] >= 0);
+   hypre_assert(partitioning[1] >= 0);
 
    hypre_ParVectorAssumedPartition(vector) = NULL;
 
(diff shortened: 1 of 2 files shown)


rfalgout:
So, what was the bug here?  I see that you've taken a smarter approach to doing the broadcasts.  Is it something to do with the vec_starts array?

victorapm:
It was a subtle one... Before we were broadcasting the memory location as `HYPRE_MPI_INT`, which defaults to 64-bit integer in bigint builds. However, `HYPRE_MemoryLocation` is an `enum` and treated as 32-bit by the compiler. So the broadcast call was messing up with another memory address (which happened to be `vec_starts`)

rfalgout (approved):
Thanks!

PR #1393, opened 2025-10-01
Improve library build times

victorapm:
The goal of this PR is to make hypre's compilation faster by splitting large files into smaller ones.

Main changes:

- `pfmg_coarsen.c` was split into
  - `pfmg_cxyz_cc.c`
  - `pfmg_cxyz_vc.c`
- `struct_matvec_core.c` was split into
  - `struct_matvec_cc.c`
  - `struct_matvec_vc.c`
  - `struct_matvec_vcc.c`
- `struct_matmult_fuse.c` was split into
  - `struct_matmult_fuse_f.c`
  - `struct_matmult_fuse_c.c`
  - `struct_matmult_fuse_ff.c`
  - `struct_matmult_fuse_fc.c`
  - `struct_matmult_fuse_cc.c`
  - `struct_matmult_fuse_fcc.c`
  - `struct_matmult_fuse_ffc.c`
  - `struct_matmult_fuse_fff.c`

Profiling on a 16-cores machine (speedups on top of the bars are with respect to the corresponding Autotools build)

hypre-build-times


victorapm:
For reference, these are the numbers on LC's matrix for the CUDA build:

| Version       |    Autotools   |  CMake+Ninja |
|:------------|--------------:|---------------:|
| v2.33.0       |       142 s       |                 50 s  |
| v3.0.0         |       210 s       |                 70 s   |
| this branch |       177 s       |                 46 s |

A nice thing about this branch is that it provides even faster build times than the old v2.33.0 when using CMake+Ninja

@tzanio @v-dobrev FYI, this might be interesting for you. In summary, using CMake+Ninja and this branch improves hypre build times considerably

rfalgout:
Thanks @victorapm for all your work on this!  Based on the results, I'm not sure it's worth splitting up the files (my 2 cents).  It's really good to know that cmake and cmake+ninja are so much faster than autotools!  All the more reason to make it our default approach for development.

victorapm:
Hi Rob, that's fine. Even though I still think the improvements here are justifiable.

Should I close this PR?

rfalgout:
I think it's worth getting the opinion of others on the project, as I am just one person and they may feel differently. :)  I don't want to influence their opinions, though, so I'm not sure how best to go about that.  I think the cmake and cmake+ninja results are the most impactful, and it requires no code changes.  The improvements we gain from splitting 3 files into 13 seems less beneficial to me.  We gain 5 to 6 seconds on the CPU and 24-33 secs on GPUs (if I read this correctly).  Also, do we really need to split the files up this much?  Do we really need to split up all three of the files?

Please don't hesitate to get the opinions of the team.  I'll go with the consensus, no worries!

victorapm:
Sure, Rob! From my analysis we need to do the full split as implemented here, otherwise one the files would still slow down all the compilation

@waynemitchell @ulrikeyang @liruipeng @oseikuffuor1  Could you chime-in? Thank you!

rfalgout:
I just took a look at your original email to me, and it looks like matvec_core is the most problematic of the three.  If you split only that one up and kept the others as they were originally, do you get similar improvements overall?

victorapm:
That's a good point

That was the case for the CPU build, but unfortunately not for the CUDA build. Meaning that, to speedup the CUDA build, we really need to split the matmult kernels as they become the bottleneck there. So, the answer is that I don't get similar improvements overall after splitting only the matvec kernels into different files

waynemitchell:
I don't have a super strong opinion here, but I guess my vote would be to go ahead and split up these files as is done in this PR. I like the faster compile time, and I don't mind breaking up the huge files into smaller ones.

ulrikeyang (approved):
I am not really in favor of adding a lot more files, but compile times affect our users, and they don't see these files underneath. So, I am approving this.

liruipeng (approved):

oseikuffuor1 (approved):
Thanks Victor.

oseikuffuor1:
Just to add to the discussion, I am fine with splitting large files into smaller ones to improve readability and (in this case) benefit faster compile times. However, I also prefer fewer splits whenever possible. Not sure if there's an optimal way to split here to balance time with number of files, but given the amount of work already done to get this performance, I am happy with this as is.

src/struct_mv/_hypre_struct_mv.h
```diff
@@ -2412,6 +2415,143 @@ hypre_StructMatmultCompute_core( HYPRE_Int                  nterms,
                                  hypre_Index                Mdstart,
                                  hypre_Index                Mdstride );
 
+/* struct_matmult_fuse_fff.c */
+HYPRE_Int
+hypre_StructMatmultCompute_fuse_fff( HYPRE_Int            nprod,
```
rfalgout: I believe these functions will only ever be called from the file `_fuse.c` correct?  If so, I think it would be better to have a `_fuse.h` file that is explicitly included in `_fuse.c` and the other split-up files and *not* added to this general header file.

rfalgout: It would be nice to avoid having the multiprecision implementations for these as well.

src/struct_mv/struct_matmult.h
```diff
@@ -19,6 +19,9 @@
 #define HYPRE_MAX_MMTERMS 512
 #endif
 
+typedef HYPRE_Complex *hypre_3Cptrs[3];
```
rfalgout: Same comment here.  Can this be added to a local header that doesn't get included in the `_hypre_*.h` header files?

src/struct_mv/_hypre_struct_mv.h
victorapm: Thanks for catching, Rob! Moved the protos in [669042e](https://github.com/hypre-space/hypre/pull/1393/commits/669042ea74dfb089dbffcb5811519f16712a8172)

I'll take a look for your second point

src/struct_mv/struct_matmult.h
victorapm: Fixed in [7136e26](https://github.com/hypre-space/hypre/pull/1393/commits/7136e264994e542ac090a62285b393c9b7dce2bd)

src/struct_mv/_hypre_struct_mv.h
victorapm: Hi Rob, I'm not sure if it's possible to avoid having the multiprecision implementations.

I can't define them as `static` functions (and remove them from `mup.fixed`) because they need to be used in another translation unit. Maybe I'm missing something... Let me know!

rfalgout: I think if you keep them in mup.fixed but take the prototypes out of `_hypre_*.h`, then no wrapper code will be written (this is how it worked before you split the files up).  You will need to add the extra include lines in each of the split up C files, though.  Let me know if I am missing something here.

victorapm: Ah okay! Removed some additional mup generated code here: [5b12b2c](https://github.com/hypre-space/hypre/pull/1393/commits/5b12b2c1ab89ebccdd159d02c0206ee934c19329)

This should be all done now

rfalgout: In theory, we could do the same for the Matvec and Cxyz optimizations, and this raises an interesting question about our header file approach in general.  The internal headers are there for convenience and ease of maintenance because they minimize the number of include lines and potential changes in code (think of how many changes were necessary recently when you finished migrating our include file usage to use the `_hypre_` conventions), they improve consistency of the ordering of include lines, and minimize the number of header files we install at build time.  However, for functions that are intended to be highly specific (such as these optimization functions), it's nice to make them as close to "static" as possible and keep them out of general header files, etc.  We currently don't do that very well in the code, and it's not overly important anyway.  But, maybe it's a good idea to start doing it more often when it's clear that the functions are highly specific helper functions.  This could be a topic worth chatting about at one of our meetings.

Thanks for the changes!

rfalgout (approved):
Thanks!

src/struct_mv/_hypre_struct_mv.h
victorapm: I see the point, Rob, thanks! I agree that going forward we should be more careful of what we put in `protos.h`. Just to finish this work, I moved all the remaining core prototypes to dedicate headers so we get less auto-generated mup code [028da79](https://github.com/hypre-space/hypre/pull/1393/commits/028da79ccd0e836c8ad46c7850af7449c25c8885)

PR #1399, opened 2025-10-09
Preserve order of linking configuration arguments

lindsayad:
Since link line ordering is critical users (should) think about the order they pass them in to configure, so we should preserve the ordering they give us instead of reversing them. For me, this lead to an undefined reference

### DIFF
--- a/src/config/configure.in
+++ b/src/config/configure.in
@@ -1596,7 +1596,7 @@ AS_HELP_STRING([--with-umpire-lib=LIBS],
                [LIBS is space-separated linkable list (enclosed in quotes) of libraries
                 needed for UMPIRE. OK to use -L and -l flags in the list]),
 [for umpire_lib in $withval; do
-       HYPRE_UMPIRE_LIB="$umpire_lib $HYPRE_UMPIRE_LIB"
+       HYPRE_UMPIRE_LIB="$HYPRE_UMPIRE_LIB $umpire_lib"
  done;
  hypre_user_gave_umpire_lib=yes]
 )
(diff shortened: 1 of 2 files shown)


victorapm:
Hi Alex, thanks for this! I can do some tweaking and work on merging.

If you can share more details on the issue, that would be great

lindsayad:
From my `config.log`, this is what I passed to configure:
```
'--with-umpire-lib=-Wl,-rpath,/data/lindad/Umpire/installed/lib -L/data/lindad/Umpire/installed/lib -lumpire -lcamp -Wl,-rpath,/usr/local/cuda-12.9/lib64 -L/usr/local/cuda-12.9/lib64 -lcudart -lnvtx3interop -lcufft -lcublas -lcusparse -lcusolver -lcurand -L/usr/local/cuda-12.9/lib64/stubs -lcuda -lstdc++'
```
and this was the result in `HYPRE_UMPIRE_LIB`
```
HYPRE_UMPIRE_LIB='-lstdc++ -lcuda -L/usr/local/cuda-12.9/lib64/stubs -lcurand -lcusolver -lcusparse -lcublas -lcufft -lnvtx3interop -lcudart -L/usr/local/cuda-12.9/lib64 -Wl,-rpath,/usr/local/cuda-12.9/lib64 -lcamp -lumpire -L/data/lindad/Umpire/installed/lib -Wl,-rpath,/data/lindad/Umpire/installed/lib '
```
which ultimately led to
```
/usr/bin/ld: /data/lindad/Umpire/installed/lib/libumpire.a(ResourceAwarePool.cpp.o): in function `camp::resources::v1::Host camp::resources::v1::Resource::get<camp::resources::v1::Host>() const':
ResourceAwarePool.cpp:(.text._ZNK4camp9resources2v18Resource3getINS1_4HostEEET_v[_ZNK4camp9resources2v18Resource3getINS1_4HostEEET_v]+0x59): undefined reference to `camp::throw_re(char const*)'
collect2: error: ld returned 1 exit status
```
during the linking stage during `make`

PR #1400, opened 2025-10-10
Batched inverse fixes

oseikuffuor1:
This PR fixes a few issues in the GPU batched inverse implementation for MGR
o	Fix bugs in ROCSolver and ONEMKL implementations
o	Fix possible memory leaks
o	Fix bug in device function


victorapm (approved):
Thanks for the fix, Daniel! I just added another minor fix.

PR #1401, opened 2025-10-11
Fix MGR bugs

victorapm:
Fix a couple of issues in MGR

- Incorrect index used for accessing `l1_norms` array
- Divide-by-zero happening in operator complexity calculation


oseikuffuor1 (approved):
Thanks Victor!

PR #1403, opened 2025-10-11
Fix various warnings with `-Wconversion`

victorapm:
This pull request resolves all `-Wconversion` compiler warnings throughout the codebase.

This warning flags implicit type conversions that could lead to data loss, such as when a double is assigned to an int or a 64-bit integer is converted to a 32-bit one.  Fixing these warnings is crucial for correctness of our calculations and enhances portability, as explicitly handling conversions guarantees our code behaves predictably across different architectures where data type sizes can vary.

None of issues fixed here seemed critical, however it would be good to get the library clean for data conversion issues going forward for better long-term maintainability.


rfalgout:
Does this cause any compilation warnings when mixed precision is enabled?  Just wondering.  We do conversions in that code intentionally and so far it hasn't cause compiler warnings.  If it does, just let me know so we can discuss.  Thanks!

victorapm:
Hi Rob, I just pushed some trivial fixes for the mixed precision code; however, I still see close to 400 warnings involving `Wfloat-conversion` (implied by `Wconversion`), e.g.,

```
hypre-path/src/parcsr_ls/mup_pre.c:1142:72: warning: conversion from ‘hypre_long_double’ {aka ‘long double’} to ‘hypre_double’ {aka ‘double’} may change value [-Wfloat-conversion]
 1142 |          return HYPRE_BlockTridiagSetAMGStrengthThreshold_dbl( solver, thresh );
      |                                                                        ^~~~~~
```

So it seems we will need to do more work for mixed precision. If you want to double-check:

```
cd ${HYPRE_DIR}
cmake -DHYPRE_ENABLE_MIXED_PRECISION=ON -DCMAKE_C_FLAGS="-Wconversion" -B build -S src
cmake --build build --parallel
```

I'll leave this untouched for now so we can discuss it

victorapm:
Rob, another idea is to disable the implicit data type conversion warnings explicitly (via pragmas) on the auto-generated mup files. This way we will probably get a clean mixed precision build with `-Wconversion`

rfalgout:
Hi Victor.  I can fix most of these easily when I return by just changing some scripting.  My guess is the test drivers are also generating warnings, and for those files I think we should turn off `-Wconversion` (if that's doable).  Let's just leave this until I return tomorrow and I'll take a closer look (still traveling).

rfalgout:
Hi @victorapm .  I have some questions:
- If I add -Wconversion to the C compile line when using the autotools build, it's not clean even when I don't turn on mixed precision.  Is it supposed to be clean?
- When I try to compile with mixed precision using cmake, I get lots of link errors and I don't get the -Wconversion warning messages at all.

My understanding was that this PR fixes the -Wconversion warnings using the standard build, and that we were discussing what to do about mixed precision.  Did I misunderstand?

Thanks!

src/parcsr_ls/par_amg_solveT.c
```diff
@@ -34,7 +34,7 @@ hypre_BoomerAMGSolveT( void               *amg_vdata,
    HYPRE_Int      amg_print_level;
    HYPRE_Int      amg_logging;
    HYPRE_Real  *num_coeffs;
-   HYPRE_Int     *num_variables;
+   HYPRE_BigInt     *num_variables;
```
oseikuffuor1, line 37: Should this be HYPRE_Real like in par_amg_solveT.c above? I don't think it makes a difference, but just asking for consistency.

src/parcsr_ls/ads.c
```diff
@@ -1609,7 +1609,7 @@ HYPRE_Int hypre_ADSSolve(void *solver,
          r_norm = hypre_sqrt(hypre_ParVectorInnerProd(ads_data -> r0, ads_data -> r0));
          r0_norm = r_norm;
          b_norm = hypre_sqrt(hypre_ParVectorInnerProd(b, b));
-         if (b_norm)
+         if (b_norm != 0.0)
```
oseikuffuor1, line 1612: We should probably use HYPRE_REAL_EPSILON for comparisons against zero where the value is computed. -->  if (b_norm > HYPRE_REAL_EPSILON) ...

oseikuffuor1 (approved):
Hi Victor, thanks for this cleanup. This is a lot of work. I left a couple comments for your review. Once is simpler, the other might mean some further discussion, but that shouldn't impact moving this PR along.

src/parcsr_ls/ads.c
victorapm: That's a subtle one. I get the point, but `HYPRE_REAL_EPSILON` might be too large. If, for some reason, the linear system coefficients are all in the order of `1e-20`, that test would likely not work as expected. Maybe we could use `HYPRE_REAL_MIN`, but I guess in practice comparing to zero is fine in this situation (and that's what our sister libraries do)

src/parcsr_ls/par_amg_solveT.c
victorapm: That makes sense (for consistency), but I did this way just for minimizing the code changes (the other solution would require changes in three other places)... Still, let me know if the consistent approach is better and I'm happy to do it

src/parcsr_ls/ads.c
rfalgout: My opinion is we should keep it as `!= 0.0` so the code behaves as before.  If we decide this should be changed, let's do it in another PR so we can have some time to do sufficient testing, especially from users.

oseikuffuor1: Yes this is all fine with me @victorapm and @rfalgout. I agree a change may not be straightforward since  HYPRE_REAL_EPSILON is not necessarily the right choice everywhere and HYPRE_REAL_MIN might be more appropriate in other places (as @victorapm noted). We just need to be careful in cases where the if-statement expects the variable to be 0.0, but could be slightly off within roundoff, for example. Similarly, it would be a more robust way of comparing equality (for instance) between two floating point numbers. I think this is one of the reasons why we introduced these floating point limits in hypre anyway.

victorapm: Sounds great, Daniel! Let's revisit this for sure. I agree using `HYPRE_REAL_MIN` will be more appropriate than `!= 0.0` in some parts of the code...

rfalgout: One more comment here. I noticed in 'krylov/pcg.c' that you used '> 0.0' instead of '!= 0.0'. The latter should keep the code running as before, but '> 0.0' is probably okay there and here as well.  Should we be consistent? I am still mostly interested in not changing code behavior for now.

src/krylov/pcg.c
```diff
@@ -269,7 +269,7 @@ hypre_PCGSetup( void *pcg_vdata,
       (pcg_data -> r_old) = (*(pcg_functions->CreateVector))(b);
    }
 
-   if (rtol && recompute_residual_p && (!two_norm))
+   if (rtol > 0.0 && recompute_residual_p && (!two_norm))
```
rfalgout: Use '!= 0.0' for consistency with the other changes in this PR?

src/utilities/printf.c
```diff
@@ -129,12 +129,24 @@ free_format( char *newformat )
 HYPRE_Int
 hypre_ndigits( HYPRE_BigInt number )
```
rfalgout, line 130: Why not just have this function return a call to `hypre_ndigits_ull` with an argument cast?

rfalgout (approved):
Looks good.  Thanks!

src/utilities/printf.c
victorapm: Edit from previous message. Now I see it... I was thinking the other way around

src/krylov/pcg.c
victorapm: Done!

PR #1408, opened 2025-10-19
Update multiprecision scripts to use explicit casts for real types

rfalgout:
This PR is related to #1403

- Updates the code generation scripts to add explicit casts in the `mup_pre.c` code.
- Reorganizes the scripts to reduce known code replication and make script maintenance a bit easier.
- There are still `-Wconversion` warnings that need to be addressed (as we already know).  We should discuss first before doing anything further here.


victorapm (approved):
Thanks, Rob!

oseikuffuor1 (approved):
Thanks Rob! This looks good!

PR #1413, opened 2025-10-22
Improve MGR restriction

victorapm:
Generalizes the block column sum restriction in MGR for the case when the number of coarse points is larger than the number of fine points at a given level.

Example result from a multiphysics application:
iters_THM-Staircase


rfalgout:
Can you explain in words what you are doing algorithmically?

victorapm:
Sure, Rob! I'm dropping a few terms in the restriction operator. Particularly, those coupling with additional blocks in the coarse level system.

For example, let's consider a 3x3 block system in which the coarse level is 2x2 (A_11 is eliminated):
Screenshot 2025-11-02 at 9 42 19 PM

This approximation is important to remove unnecessary couplings in a pressure-thermal system for example (flow in porous media).

If we had more blocks, we could generalize this implementation even further and allow for choosing which terms in the restriction operator are zeroed out or computed. I would like to do this on another PR

rfalgout:
How are you deciding that W_31 should be zeroed out?  I couldn't see that by looking at the code.

victorapm:
The code is currently simply assuming that every time we have more C points than F points, then W_N1 will be zeroed out if N>=3.

This approach isn't general enough because we might want to keep or zero out different entries depending on the problem. But if this is not an issue, I would prefer working on the generalization in a separate PR so geos can already take advantage of these improvements

rfalgout:
I don't really follow the argument for this approach.  Couldn't it lead to worse behavior for some applications?  I guess I am wondering if it would be better to at least give users a simple switch to turn it all on or all off (off by default so it only affects user code that turns the switch on, e.g., GEOS, which you have control over).  A more general approach could be implemented later in another PR as you are suggesting.

victorapm:
Hi Rob, that's a good point. I can implement the switch as an interim option until we have the more general approach in place.

victorapm:
@oseikuffuor1 @rfalgout This is ready for review. I added a separate option to access the novel implementation discussed in this PR as an interim solution. Looking forward, we will need to implement a more general approach to select which submatrices in the restriction and prolongation operators should be zeroed out

rfalgout (approved):
Thanks!

oseikuffuor1 (approved):
Thanks Victor!

PR #1415, opened 2025-10-30
Add new options to MGR's F-relaxation solver

victorapm:
This PR adds new options to MGR's F-relaxation solver:
- Option 29: Sparse direct solver via SuperLU_Dist
- Option 32: ILU

In addition, this improves error handling in case of issues during F-relax setup/solve


rfalgout (approved):
Looks good to me.  Thanks!

src/parcsr_ls/par_mgr.h
```diff
@@ -241,7 +241,7 @@ typedef struct
 #define hypre_ParMGRDataFRelaxType(data)            ((data) -> Frelax_type)
 #define hypre_ParMGRDataFRelaxTypeI(data, i)        ((data) -> Frelax_type[i])
 #define hypre_ParMGRDataAFFsolver(data)             ((data) -> aff_solver)
-#define hypre_ParMGRDataAFFsolverI(data)            ((data) -> aff_solver[i])
+#define hypre_ParMGRDataAFFsolverI(data, i)         ((data) -> aff_solver[i])
```
oseikuffuor1, line 244: This looks like a bug before. Nice catch!

src/parcsr_ls/par_mgr.c
```diff
@@ -4060,7 +4060,13 @@ hypre_MGRDirectSolverSetup( void                *solver,
    HYPRE_UNUSED_VAR(f);
    HYPRE_UNUSED_VAR(u);
 
+#if defined(HYPRE_USING_DSUPERLU)
```
oseikuffuor1, line 4063: Was this changed to put the guards in here so you can add an error for the user? Otherwise why not avoid compiling it in the first place (as before)?

victorapm: Right, that was the main reason (easier access to user codes). Otherwise we would need to ifdef on multiple places. I think this solution is better

oseikuffuor1 (approved):
Looks good to me Victor. Just a minor comment for your review.

PR #1416, opened 2025-10-30
Sycl umpire

waynemitchell:
Add umpire support for the sycl backend.


src/utilities/memory.c
```diff
@@ -1725,11 +1725,17 @@ HYPRE_Int
...
    const char *pool_name = hypre_HandleUmpireDevicePoolName(handle);
 
+#if defined(HYPRE_USING_SYCL)
+   HYPRE_Int device_id;
+   hypre_GetDevice(&device_id);
+   hypre_sprintf(resource_name, "%s::%d", "DEVICE", device_id);
+#else
+   const hypre_int device_id = hypre_HandleDevice(handle);
    hypre_sprintf(resource_name, "%s::%d", "DEVICE", device_id);
+#endif
```
victorapm, line 1743: Minor reorg:

```suggestion
#if defined(HYPRE_USING_SYCL)
   HYPRE_Int device_id;
   hypre_GetDevice(&device_id);
#else
   const hypre_int device_id = hypre_HandleDevice(handle);
#endif
   hypre_sprintf(resource_name, "%s::%d", "DEVICE", device_id);
```

victorapm: Also, just curious... Shouldn't `hypre_HandleDevice(handle);` work for SYCL as well? Curious why we need to do something different for SYCL...

waynemitchell: The type of `hypre_HandleDevice(handle)` is different for sycl. It's an actual sycl device object rather than just an integer ID.

PR #1419, opened 2025-11-06
Fix HYPRE_Int -> HYPRE_BigInt

### LINKED ISSUE #1407: Potential undefined behavior in `hypre_GaussElimSetupDevice` and `hypre_GaussElimSetup`
These functions both compute the global system size as:

```c
HYPRE_Int  global_size = global_num_rows * global_num_rows;
```

Signed integer overflow is undefined behavior in both C and C++, so the check later on in `hypre_GaussElimSetupDevice` isn't a good way to handle it:

```c
if (global_size < 0)
{
    hypre_error_w_msg(HYPRE_ERROR_GENERIC, "Detected overflow!");
    return hypre_error_flag;
}
```

This can cause major issues for even quite small problems. For example say  `global_num_rows = 73270` and we compiled with mixed integer support. Computing the global size as a `HYPRE_BigInt` would give `5,368,492,900`, or about 2.5 times the signed 32-bit integer max. Assuming the compiler wraps around to negative integers on overflow (this is not at all guaranteed), this number actually wraps back around to a positive number! Specifically, around 1 billion. Thus, the above check misses the problem.

Important to note, `global_size` is never passed to a LAPACK routine; it is only used for memory allocation. As such, it should be a `HYPRE_BigInt` to prevent these potential problems.

As an aside, I recognize that these routines are typically only used in the coarse relaxation, so the system size shouldn't be that large. Regardless, better to be safe than sorry.

victorapm:
~~Adds `hypre_(Big)IntSafeMult` for multiplying two HYPRE_(Big)Ints safely, i.e., guaranteeing that overflow doesn't occur.  If multiplying the operands would lead to overflow, the function returns `HYPRE_(BIG_)INT_MAX` and lets the caller decide how to handle the error~~

Fix variable from HYPRE_Int to HYPRE_BigInt

Closes #1407

### DIFF
--- a/src/config/cmake/HYPRE_CMakeUtilities.cmake
+++ b/src/config/cmake/HYPRE_CMakeUtilities.cmake
@@ -778,6 +778,9 @@ function(maybe_build_umpire)
   set(ENABLE_HIP  ${HYPRE_ENABLE_HIP}  CACHE BOOL "Enable HIP in Umpire" FORCE)
   set(ENABLE_SYCL ${HYPRE_ENABLE_SYCL} CACHE BOOL "Enable SYCL in Umpire" FORCE)
 
+  # Rename Umpire's 'check' target to 'umpire_check' to avoid conflicts
+  set(BLT_CODE_CHECK_TARGET_NAME "umpire_check" CACHE STRING "Rename Umpire's check target" FORCE)
+
   # Ensure Umpire installs to the same prefix as hypre
   set(CMAKE_INSTALL_PREFIX "${CMAKE_INSTALL_PREFIX}" CACHE PATH "Install prefix" FORCE)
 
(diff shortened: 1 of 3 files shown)


rfalgout:
Did you have an issue that prompted this PR?  I would use these functions sparingly since it requires a bunch of logic and probably is useful in only rare instances.

victorapm:
Hi Rob, the issue was #1407. I agree we could probably take a simpler solution

rfalgout:
I think the right way to fix this potential issue is to use BigInt for `global_size`.  I am not against adding these functions for potential use elsewhere in the future, though.  Changing to BigInt here should dramatically reduce the possibility of overflow and make overflow to a positive value even less likely.

victorapm:
Thanks, Rob! Just implemented your suggested fix in [6c14fea](https://github.com/hypre-space/hypre/pull/1419/commits/6c14fea6bcd2eec805b9cc1c04d7af21c8d7e5cc)

rfalgout (approved):
Looks good.  If you want to add those routines back, that's fine.  Maybe they will come in handy some day.  Either way is okay with me.  Thanks!

Issue #1392, opened 2025-09-30
cudaErrorInvalidDevice: invalid device ordinal when running examples


moulin1024:
Hello,

I am trying to build hypre on our cluster with CMake with the following options:

```
cmake ../src -DBUILD_SHARED_LIBS=ON \
             -DHYPRE_ENABLE_PRINT_ERRORS=ON \
             -DHYPRE_ENABLE_CUDA=ON \
             -DCMAKE_CUDA_ARCHITECTURES='80' \
             -DHYPRE_ENABLE_UNIFIED_MEMORY=ON \
             -DHYPRE_BUILD_EXAMPLES=ON \
             -DHYPRE_BUILD_TESTS=ON \
             -DHYPRE_ENABLE_GPU_AWARE_MPI=ON \
             -DHYPRE_ENABLE_HYPRE_BLAS=OFF -DHYPRE_ENABLE_HYPRE_LAPACK=OFF \
             -DHYPRE_ENABLE_SUPERLU=ON -DCMAKE_PREFIX_PATH=/path/to/my/superlu
make -j8
```

When I try to run the ij test, it runs without any error

```
srun --ntasks=1 --gres=gpu:a100:1 --cpus-per-task=18 --mem=64G --time=01:00:00 ./build/test/ij
```

However, when I run the IJ example code (ex5),
```
srun --ntasks=1 --gres=gpu:a100:1 --cpus-per-task=18 --mem=64G --time=01:00:00 ./build/example/ex5
```

I get the following error:

```
terminate called after throwing an instance of 'thrust::THRUST_200500_800_NS::system::system_error'
  what():  after determining tmp storage requirements for exclusive_scan: cudaErrorInvalidDevice: invalid device ordinal
```

I also create  small sanity check program to further explore what went wrong:
```
// hypre_sanity_5x5.cu
#include <cstdio>
#include <cstdlib>
#include <vector>
#include <iostream>
#include <mpi.h>
#include <cuda_runtime.h>

#include "HYPRE.h"
#include "HYPRE_krylov.h"
#include "HYPRE_parcsr_ls.h"


static void my_cuda_state(const char* tag){
  int dev=-1, ndev=0; 
  cudaGetDevice(&dev); 
  cudaGetDeviceCount(&ndev);
  size_t f=0,t=0; 
  cudaMemGetInfo(&f,&t);
  printf("[%s] device=%d of %d, total=%.2f GiB, free=%.2f GiB\n",
[...]

PR #1420, opened 2025-11-06
Add logic for UVM prefetch

victorapm:
Run UVM prefetch only when the operating system and the NVIDIA card both allow it (and hypre is configured with unified memory support)

See https://github.com/hypre-space/hypre/issues/1392#issuecomment-3416588392


rfalgout (approved):
I don't understand the implications of doing this, but I'll leave that to you and Rui Peng to think about.  It looks good to me otherwise.  Thanks!

victorapm:
Thanks, Rob! For reference, I'm mimicking what umpire does: https://github.com/LLNL/Umpire/blob/30433c15454f019a338b541472690bf188433cda/src/umpire/op/CudaMemPrefetchOperation.cpp#L32-L40

victorapm:
@liruipeng Could you take a look at this PR?

liruipeng (approved):
Thanks @victorapm

PR #1425, opened 2025-11-17
Flexible AMG cycles

waynemitchell:
This is work from @dinesh-parthasarathy implementing flexible AMG cycles in which a user may define arbitrary order for visiting levels of the hierarchy as well as arbitrary relaxation methods and weights each time a level is visited. This capability may be combined with genetic programming to find optimized cycle structures that provide superior performance compared to traditional V-cycles.


src/parcsr_ls/HYPRE_parcsr_amg.c
```diff
@@ -980,6 +980,52 @@ HYPRE_BoomerAMGGetLogging( HYPRE_Solver solver,
    return ( hypre_BoomerAMGGetLogging( (void *) solver, logging ) );
 }
 
+/*----------------------------------------------------------------------------------------
+ *                               FLEXIBLE AMG CYCLING
+ *----------------------------------------------------------------------------------------*/
+
+HYPRE_Int
+HYPRE_BoomerAMGFlexibleSetCycleStruct( HYPRE_Solver solver,
```
waynemitchell: I think I would prefer the naming convention, `HYPRE_BoomerAMGSetFlexibleCycleStruct()`, etc. throughout this PR. Thoughts from others?

src/parcsr_ls/par_amg.c
```diff
@@ -2759,6 +2803,194 @@ hypre_BoomerAMGGetSmoothNumSweeps( void     *data,
...
+      return hypre_error_flag;
+   }
+
+   HYPRE_Int num_levels_flexible = 0;
+   HYPRE_Int i = 0;
+   HYPRE_Int lvl = 0;
+
+   while ( cycle_struct_flexible[i] == 0 || // same level
+           cycle_struct_flexible[i] == 1 || // cycle up
+           cycle_struct_flexible[i] == -1 ) // cycle down
```
waynemitchell, line 2830: If I understand this right, you are basically relying on the loop going off the end of the allocated cycle_struct_flexible array, which is not a safe way of doing things. You could either have the user explicitly set the length of the cycle structure or have something like the last entry in cycle_struct_flexible be a specified termination character.

src/parcsr_ls/amg_hybrid.c
```diff
@@ -1416,6 +1429,216 @@ hypre_AMGHybridSetLevelOuterWt( void   *AMGhybrid_vdata,
...
+    {
+       hypre_error_in_arg(2);
+       return hypre_error_flag;
+    }
+
+   if ((AMGhybrid_data -> cycle_struct_flexible))
+   {
+      hypre_TFree((AMGhybrid_data -> cycle_struct_flexible), HYPRE_MEMORY_HOST);
+   }
+   (AMGhybrid_data -> cycle_struct_flexible) = cycle_struct_flexible;
```
waynemitchell, line 1485: Need to free these in the destructor for AMGhybrid_data as well.

src/parcsr_ls/par_cycle.c
```diff
@@ -222,9 +241,28 @@ hypre_BoomerAMGCycle( void              *amg_vdata,
...
+               is_flexible[k] = 1;
+            else // recursive levels
+               is_flexible[k] = 0;
+         }
+         else  // the entire cycle is flexible
+         {
+            is_flexible[k] = 1;
+         }
+      }
+   }
```
waynemitchell: The additional is_flexible array seems unnecessary to me? Couldn't we just check `level < num_levels_flexible` in the main cycling loop?

src/parcsr_ls/par_cycle.c
```diff
@@ -64,6 +64,13 @@ hypre_BoomerAMGCycle( void              *amg_vdata,
    HYPRE_Int       block_mode;
    HYPRE_Int       cheby_order;
 
+   /* Flexible Cycle variables */
+   HYPRE_Int      *is_flexible, length_cycle_flexible, num_levels_flexible;
+   HYPRE_Int      *cycle_struct_flexible, *relax_types_flexible, *relax_orders_flexible;
+   HYPRE_Real     *outer_weights_flexible, *relax_weights_flexible, *cgc_scaling_factors_flexible;
+   HYPRE_Real     cgc_scaling_factor_tmp;
+   HYPRE_Int      index;
```
waynemitchell: Maybe name `index` as `flexible_index` or something in order to distinguish it.

src/test/ij.c
```diff
@@ -268,6 +268,11 @@ main( hypre_int argc,
    HYPRE_Int      CR_use_CG = 0;
    HYPRE_Int      P_max_elmts = 4;
    HYPRE_Int      cycle_type;
+   /* flexible cycling params*/
+   HYPRE_Int      flexible_cycle = 0, length_cycle_flexible = 0;
```
waynemitchell: Don't think we need the extra `flexible_cycle` flag.

src/test/ij.c
```diff
@@ -730,6 +735,178 @@ main( hypre_int argc,
          build_matrix_type      = -1;
          build_matrix_arg_index = arg_index;
       }
+      else if ( strcmp(argv[arg_index], "-flexamg") == 0 )
```
waynemitchell: I see that it is required to specify all the different flexible cycle parameter arrays together here if using the `-flexamg` option. Is there value in allowing the user to specify only a subset of the options?

src/parcsr_ls/par_cycle.c
```diff
@@ -378,6 +416,22 @@ hypre_BoomerAMGCycle( void              *amg_vdata,
...
+         if (num_levels_flexible > 0) 
+         {
+            relax_order = hypre_ParAMGDataRelaxOrder(amg_data);
+            if (is_flexible[level])
+            {
+               relax_type          = relax_types_flexible[index];
+               relax_order         = relax_orders_flexible[index];
+               relax_weight[level] = relax_weights_flexible[index];
+               omega[level]        = outer_weights_flexible[index];
+               num_sweep           = 1;
```
waynemitchell: There is an implicit requirement here that the user needs to set all of these parameter arrays if using flexible cycles (i.e. they cannot just set a subset). We should either make this requirement more explicit with some error checking, or relax it by allowing default values (e.g. if the user set `relax_type_flexible` but not `relax_order_flexible`, just use a default value for `relax_order`, etc.).

src/parcsr_ls/par_cycle.c
dinesh-parthasarathy: Agree! I have removed the additional is_flexible array (95590774d3970aaebe6db23a33d99b791190c135).

src/parcsr_ls/par_cycle.c
dinesh-parthasarathy: This has been renamed (95590774d3970aaebe6db23a33d99b791190c135).

src/parcsr_ls/par_cycle.c
dinesh-parthasarathy: I have updated such that the user can supply only a subset (95590774d3970aaebe6db23a33d99b791190c135).

src/parcsr_ls/par_amg.c
dinesh-parthasarathy: I wanted to keep the user inputs to a minimum, so I don't expect the length to be set explicitly. But the last entry of the list supplied would be a hint to terminate the loop (for example -2).

src/test/ij.c
dinesh-parthasarathy: I have made some modifications. The user must supply the values for `cycle_struct_flexible` to enable _flexible cycling_. The other arrays are optional only if you want to do something different at each step of the cycling. Else, the BoomerAMG parameters set using the other _non-flexible_ interfaces is simply used for _flexible cycling_ (01c7b7f79f76a910be977f995e87c29af89bd1e1).

src/test/ij.c
dinesh-parthasarathy: This has been removed (01c7b7f79f76a910be977f995e87c29af89bd1e1).

src/parcsr_ls/amg_hybrid.c
dinesh-parthasarathy: I believe the destructor for BoomerAMG called within `hypre_AMGHybridDestroy` should release memory allocated for flexible parameter arrays. 

`if (pcg_precond) { hypre_BoomerAMGDestroy(pcg_precond);`

src/parcsr_ls/par_cycle.c
```diff
@@ -378,6 +395,26 @@ hypre_BoomerAMGCycle( void              *amg_vdata,
...
+            if ((level < num_levels_flexible - 1) || (level == num_levels - 1))
+            {
+               if (relax_types_flexible)
+                  relax_type          = relax_types_flexible[index_flex];
+               if (relax_orders_flexible)
+                  relax_order         = relax_orders_flexible[index_flex];
+               if (relax_weights_flexible)
+                  relax_weight[level] = relax_weights_flexible[index_flex];
+               if (cgc_scaling_factors_flexible)
+                  omega[level]        = outer_weights_flexible[index_flex];
```
waynemitchell, line 420: Should be `if (outer_weights_flexible)`?

dinesh-parthasarathy: That's right. I pushed the fix to my fork f92495fb1bf4e8235d95b1a96edeb6e9b39d7e72, assuming it syncs to this branch?

src/parcsr_ls/HYPRE_parcsr_amg.c
waynemitchell: Went ahead and made this change.

src/parcsr_ls/par_amg.c
waynemitchell: Sounds good. I've documented this in the user-facing function description.

src/parcsr_ls/par_cycle.c
waynemitchell: I made the change on this branch (your fork won't sync here).

waynemitchell:
I have some small regression test changes to track down before this gets merged.

src/test/TEST_ij/solvers.jobs
```diff
@@ -105,6 +105,17 @@ mpirun -np 8 ./ij -n 40 40 40 -P 2 2 2 -nongalerk_tol 3 0.0 0.01 0.05 > solvers.
...
+#flexible cycles
+mpirun -np 2 ./ij -n 40 40 40 -solver 1 \
+                  -cycle_struct 0,0,-1,-1,-1,-1,1,1,1,1,-2 \
+                  -relax_types 0,0,14,-1,18,13,9,3,0,-1,14 \
+                  -relax_orders 1,1,0,0,0,1,0,0,0,0,0 \
+                  -outer_weights 1.2,0.9,1.0,0,1.0,1.0,1,0.95,0.8,0,1.0 \
+                  -relax_weights 0.95,1.85,1.0,0,1.0,1.0,1,1.8,1.2,0,1.0 \
+                  -cgc_scaling 0,0,1,1,1,1,1.35,0.7,1.3,1.1 > solvers.out.123
+mpirun -np 2 ./ij -n 40 40 40 -solver 20 \
+                  -cycle_struct -1,-1,-1,-1,1,0,0,0,-1,1,1,1,1,0,0,-2 > solvers.out.124
```
victorapm: Sorry for chiming in... These won't get parsed correctly because the IJ driver expects a `flexamg_` prefix

waynemitchell: Nice catch. Thanks, Victor. Feel free to add your comments/review as well!

src/parcsr_ls/amg_hybrid.c
waynemitchell: Fixed this.

src/test/ij.c
```diff
@@ -4498,6 +4730,7 @@ main( hypre_int argc,
       HYPRE_ParCSRHybridCreate(&amg_solver);
       HYPRE_ParCSRHybridSetTol(amg_solver, tol);
       HYPRE_ParCSRHybridSetAbsoluteTol(amg_solver, atol);
+      HYPRE_ParCSRHybridSetTwoNorm(amg_solver, two_norm);
```
waynemitchell, line 4733: Adding this line changes the setting that gets used here (default with no call to `Set` is 0, but `two_norm` defaults to 1 in this driver). This results in a small change in the regression tests. Adding the line seems sensible (we couldn't previously change this setting via the command line in the driver). So I guess I'll update the `.saved` files appropriately, if others think this is OK.

rfalgout: I agree.  This makes Hybrid PCG behave like PCG in this driver, so changing the saved files makes sense.  Thanks!

rfalgout (approved):
Looks good to me!  Thanks!

src/parcsr_ls/par_amg.c
```diff
@@ -654,6 +662,42 @@ hypre_BoomerAMGDestroy( void *data )
          hypre_IntArrayDestroy(hypre_ParAMGDataDofFunc(amg_data));
          hypre_ParAMGDataDofFunc(amg_data) = NULL;
       }
+
+      /* flexible cycles params */
+      if (hypre_ParAMGDataFlexibleCycleStruct(amg_data))
```
liruipeng, line 667: We typically don't need to write the `if` and `=NULL` in
```
if (ptr) {
   hypre_TFree(ptr);
   ptr = NULL;
}
```
since `hypre_TFree` and `hypre_Free` do those already internally. But it is probably not a big deal I guess.

src/test/TEST_ij/solvers.saved
```diff
@@ -32,7 +32,7 @@ Final Relative Residual Norm = 6.698760e-09
 Iterations = 11
 PCG_Iterations = 7
 DSCG_Iterations = 4
-Final Relative Residual Norm = 6.805037e-10
```
liruipeng, line 35: Just curious. Why does the PR change this result?

liruipeng (approved):
All looks good to me other than a few minor comments/questions. Thanks @waynemitchell @dinesh-parthasarathy  for the work!

src/test/TEST_ij/solvers.saved
waynemitchell: I think this is from the change discussed here:
https://github.com/hypre-space/hypre/pull/1425/changes#diff-6c7e332fcb45adb3e0b0f2da90ad6d6d63326db3661a53fe12d5f6aca436c0f9R4733

src/parcsr_ls/par_amg.c
waynemitchell: Yeah, I agree. There are a lot of places in the code, however, where we still explicitly have the `if` check and `=Null` set anyway... see the code directly above these additions, for example. I am inclined to leave this be for now, and maybe in the future we can try to go through the code and clean up this kind of thing in a more comprehensive way.

PR #1427, opened 2025-11-21
Fix for host exec with device build

tomstitt:
Only `hypre_SyncDevice` if the exec policy is device, otherwise we initialize the device context for a host-only run


victorapm:
Thanks, Tom! I'll do some cosmetics and reopen this in another PR

PR #1428, opened 2025-11-22
Fix for host execution with device build

victorapm:
Extends #1427 

Main goal is to run device synchronization only when GPUs are used.

cc: @tomstitt 

Question to @liruipeng: I see we have a `TODO` comment questioning whether the sync call is needed, do you any thoughts?


rfalgout (approved):
Looks good to me.  Thanks!

liruipeng (approved):
Looks good. Thanks!

PR #1431, opened 2025-11-27
Add option to control FPE trap

victorapm:
Autotools: `--enable-fpe-trap`
CMake: `-DHYPRE_ENABLE_FPE_TRAP=ON`

Floating-point exception traps are disabled by default and enabled when hypre is built in debug mode


rfalgout (approved):
Looks good.  Thanks!

PR #1433, opened 2025-12-03
Fix `-Wshadow` warnings

victorapm:
Cleans the code for the build: `cmake -B build -S src -DCMAKE_C_FLAGS=-Wshadow`

Resolving `-Wshadow` warnings is important to:

1. Preventing silent logic errors (such as using a local variable when we meant to use a global or outer-scope variable)
2. Improve readability: Removes ambiguity so developers know exactly which variable is being referenced at a glance.
3. Safeguard refactoring: Prevents bugs where moving code blocks into new scopes silently changes variable references.


src/IJ_mv/IJMatrix_parcsr.c
```diff
@@ -2451,7 +2451,6 @@ hypre_IJMatrixAssembleOffProcValsParCSR( hypre_IJMatrix       *matrix,
                                                    HYPRE_MEMORY_HOST);
             }
 
-            HYPRE_Int i;
```
rfalgout, line 2454: This needs to stay in, but with a new name like `li`.  It's inside an outer loop that is already using `i`.  In general, this simple renaming approach is the most reliable way to make these changes.  I'll try to look for other potential issues like this, but you might want to take a second look as well.

src/parcsr_ls/par_amg.c
```diff
@@ -614,7 +614,6 @@ hypre_BoomerAMGDestroy( void *data )
       }
       if (grid_relax_type)
       {
-         HYPRE_Int num_levels = hypre_ParAMGDataNumLevels(amg_data);
```
rfalgout, line 625: This one might need to stay in as well.  I would have to look at the logic here to be sure, but `num_levels` is initialized at the top of the routine so it's value may be needed outside of this `if` section.

src/IJ_mv/IJMatrix_parcsr.c
victorapm: Thanks for catching! This one was a real bug! Fixed

rfalgout:
Hi @victorapm .  I will wait until you take another look through your changes before looking through this some more.  We want to be careful not to potentially change the behavior of the code.  It is a good idea to test for this stuff, though.  Thanks!

src/parcsr_ls/par_amg.c
victorapm: I double checked this one and the current change is fine. We're in the destructor, so `num_levels` cannot change in that function. We can rely on the value initialized at the beginning

victorapm:
@rfalgout @oseikuffuor1 I updated this branch and reran the regression tests on tux. Everything is looking good. This is ready for review

src/parcsr_ls/par_mgr_solve.c
```diff
@@ -772,7 +772,7 @@ hypre_MGRCycle( void              *mgr_vdata,
                {
                   hypre_BoomerAMGRelax(A_array[fine_grid], F_array[fine_grid],
                                        NULL, level_smooth_type[fine_grid],
-                                       0, fp_one, fp_one, l1_norms,
+                                       0, fp_one, fp_one, l1_norms, //NULL,
```
rfalgout: Why is there //NULL here?

victorapm: This is from an old experiment. Removed!

src/struct_mv/box.h
```diff
@@ -120,13 +120,11 @@ hypre_max(0, (hypre_BoxIMaxD(box, d) - hypre_BoxIMinD(box, d) + 1))
   hypre_IndexD(index, d) <= hypre_BoxIMaxD(box, d) )
 #define hypre_BoxSpanIndex(box, index)\
 {\
-   HYPRE_Int d;\
-   for (d = 0; d < hypre_BoxNDim(box); d++)\
+   HYPRE_Int __d;\
```
rfalgout: I'm not sure I like this one.  There is no way to define a macro-local variable and be sure that it won't have the same name as another variable where it is used.  Is there a way to have the compiler flag ignore macros?  You are using the double underscore here to try to prevent it, which probably works.  Let me think a bit about this one.

victorapm: Thanks, Rob! I think alternative approaches here would be:
1. Use `_Pragma("GCC diagnostic push") _Pragma("GCC diagnostic ignored \"-Wshadow\"")` to skip the Wshadow check on that macro
2. Use a `static inline` function instead of a macro

rfalgout: Let's try the `_Pragma` approach here since it's C99 standard.  Maybe first check that it works in this one instance, then we can use it other places if needed (I would need to look at the PR details again).

victorapm: @rfalgout, added in [d3c9016](https://github.com/hypre-space/hypre/pull/1433/commits/d3c9016417092943e89e7f74815eed2e852906cb)

src/struct_mv/boxloop_host.h
```diff
@@ -61,17 +61,18 @@
 #define zypre_BoxLoop1Begin(ndim, loop_size,                                  \
                             dbox1, start1, stride1, i1)                       \
 {                                                                             \
-   HYPRE_Int i1;                                                              \
+   HYPRE_Int hypre_loop_i1;                                                   \
```
rfalgout: I would like to just put `_Pragma`s here as well and do some work on these loops in another PR.  The first `HYPRE_Int i1` can be declared as a single variable in `zypre_BoxLoopDeclare` best I can tell.  The second one can be left as is.  We really need to remove all the `zypre_` stuff as well.  These were intended to be temporary names but we never got around to fixing that properly.

src/test/zboxloop.c
```diff
@@ -32,7 +32,7 @@ main( hypre_int argc,
    HYPRE_Int         num_procs, myid;
    HYPRE_Int         dim;
    HYPRE_Int         rep, reps, fail, sum;
-   HYPRE_Int         size, xi1;
+   HYPRE_Int         size, boxloop0_idx;
```
rfalgout: We may not need these changes if we revert the `boxloop_host.h` stuff to use `_Pragma`.

src/utilities/gselim.h
```diff
@@ -8,57 +8,56 @@
...
-                  x[j] -= x[k] * A[j*n+k];             \
-               }                                       \
-            }                                          \
-         }                                             \
-      }                                                \
-      if (A[0] != 0.0) x[0] /= A[0];                   \
-   }                                                   \
+#define hypre_gselim(A,x,n)                                            \
+{                                                                      \
+   HYPRE_Int  __j, __k, __m;                                           \
```
rfalgout: I think it's safe to use `_Pragma` here as well.  When this macro is being used in the code, there is no confusion about variable names, and it's not possible to insert variables in the middle of the `{}` nesting below as we do in the boxloops.

src/utilities/mpistubs.c
```diff
@@ -8,28 +8,17 @@
 #include "_hypre_utilities.h"
 
 /******************************************************************************
- * This routine is the same in both the sequential and normal cases
- *
- * The 'comm' argument for MPI_Comm_f2c is MPI_Fint, which is always the size of
- * a Fortran integer and hence usually the size of hypre_int.
- ****************************************************************************/
+ * MPI stubs to generate serial codes without mpi
```
rfalgout, line 11: Was there a shadow variable issue here?

src/struct_mv/boxloop_host.h
victorapm: Addressed in https://github.com/hypre-space/hypre/pull/1433/commits/d251eb6002140d6abf586da4a9a991e20ac5409c

src/utilities/gselim.h
victorapm: Addressed in [2b6834f](https://github.com/hypre-space/hypre/pull/1433/commits/2b6834f1b31f4d28af03bc6339e71a164d1cea42)

src/utilities/mpistubs.c
victorapm: This is a bit out of scope

It's a fix for a an `-Winfinite-recursion` warning/error in sequential builds when HYPRE_HAVE_MPI_COMM_F2C is defined.

Before the change, `hypre_MPI_Comm_f2c()` lived above the #if defined(HYPRE_SEQUENTIAL) split and did this:
```
  #ifdef HYPRE_HAVE_MPI_COMM_F2C
     return (hypre_MPI_Comm) MPI_Comm_f2c(comm);
  #else
     return (hypre_MPI_Comm) (size_t)comm;
  #endif
```

But in sequential mode, `mpistubs.h` leads to

```
  #define MPI_Comm_f2c hypre_MPI_Comm_f2c
```

So that call expands into a call to itself:

```
  return (hypre_MPI_Comm) hypre_MPI_Comm_f2c(comm);
```

Some compilers may report that as `error: infinite recursion detected [-Werror=infinite-recursion]`

This branch moved `hypre_MPI_Comm_f2c()` inside the sequential/MPI path so we fix this warning issue

src/test/zboxloop.c
victorapm: That's true, thanks! Reverted in https://github.com/hypre-space/hypre/pull/1433/commits/c04a01860033310dbe703977c96dfe3e31345bf5

src/struct_mv/_hypre_struct_mv.h
```diff
@@ -3366,15 +3366,17 @@ hypre__J = hypre__thread;  i1 = i2 = 0; \
 #define zypre_BoxLoop1Begin(ndim, loop_size,                                  \
                             dbox1, start1, stride1, i1)                       \
 {                                                                             \
-   HYPRE_Int i1;                                                              \
    zypre_BoxLoopDeclare();                                                    \
    zypre_BoxLoopDeclareK(1);                                                  \
    zypre_BoxLoopInit(ndim, loop_size);                                        \
-   zypre_BoxLoopInitK(1, dbox1, start1, stride1, i1);                         \
+   zypre_BoxLoopInitK(1, dbox1, start1, stride1, hypre__ik);                  \
```
rfalgout: The `hypre__ik` argument shouldn't be needed at all.

victorapm: Addressed in https://github.com/hypre-space/hypre/pull/1433/commits/fa6d498c98821d742ddca4ac471d57c47979d47a

src/struct_mv/boxloop_host.h
```diff
@@ -61,15 +56,21 @@
...
                             dbox1, start1, stride1, i1)                       \
 {                                                                             \
+   HYPRE_DIAGNOSTIC_PUSH                                                      \
+   HYPRE_DIAGNOSTIC_IGNORE_WSHADOW                                            \
    HYPRE_Int i1;                                                              \
+   HYPRE_DIAGNOSTIC_POP                                                       \
    zypre_BoxLoopDeclare();                                                    \
    zypre_BoxLoopDeclareK(1);                                                  \
    zypre_BoxLoopInit(ndim, loop_size);                                        \
    zypre_BoxLoopInitK(1, dbox1, start1, stride1, i1);                         \
```
rfalgout: What I was trying to say in the last comment is that the `InitK` macro doesn't need to have that last argument at all (it can be a 4-argument macro).  It can just use `hypre__ik` since you added it to the `Declare` macro.  It doesn't look to me like that variable is needed anywhere else but `InitK`.  Maybe I am missing something.

victorapm: Sorry, Rob, I had missed your point. Yes, looking twice at this, it does make sense to remove that last variable. Done in [efb944d](https://github.com/hypre-space/hypre/pull/1433/commits/efb944df56ff58ac3dde64712cd2c6e88c1911e6)

rfalgout (approved):
This looks good to me now.  Let's be sure to run an extensive set of regression tests again to be extra careful.  Thanks!

PR #1440, opened 2025-12-18
Casts to avoid overflow of 32-bit integers before casting to uint64

rietmann-nv:
I came across this issue when running larger solves on GPUs. The error was in `par_krylov_func.c`
```
array_data = hypre_CTAlloc(HYPRE_Complex, (n * size * num_vectors), memory_location);
```
The expression `n * size * num_vectors` overflowed, leading to undefined behavior and crashes. I found a number of other places where this could possibly happen and fixed them as well, but they are less likely because they don't have the extra `n` parameter.


liruipeng:
@rietmann-nv Thanks for reporting the bug. We had another PR for this issue https://github.com/hypre-space/hypre/pull/1421 . We will merge yours to that PR, and then to `master`.

liruipeng (approved):

PR #1450, opened 2026-01-02
Update SuperLU interface + MGR

victorapm:
The main goal of this PR is to conform the interface to SuperLU_Dist with the general solver structure in hypre. Particularly:

- Add `hypre_SLUDistCreate` for creating the solver object
- Change the `hypre_DSLUData` struct members to pointers
- Remove `print_level` from `hypre_SLUDistSetup` and create a dedicate `hypre_SLUDistSetPrintLevel`
- Update the setup/solve `hypre_SLUDist` function signatures to take A, b, and x
- Update SuperLU_Dist and `aff_solver` usage in MGR to reflect the above changes

### DIFF
--- a/AUTOTEST/machine-tux.sh
+++ b/AUTOTEST/machine-tux.sh
@@ -104,7 +104,7 @@ co="--enable-maxdim=4 --enable-debug"
 ./test.sh basic.sh $src_dir -co: $co -mo: $mo -eo: -maxdim
 ./renametest.sh basic $output_dir/basic--enable-maxdim=4
 
-co="--enable-complex --enable-maxdim=4 --enable-debug"
+co="--enable-complex --enable-maxdim=4 --enable-debug --with-extra-CFLAGS='-Wno-incompatible-pointer-types'"
 ./test.sh basic.sh $src_dir -co: $co -mo: $mo -eo: -complex
 # ignore complex compiler output for now
 rm -fr basic.dir/make.???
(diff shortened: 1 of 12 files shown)


rfalgout (approved):
Looks good to me.  Don't forget the mixed precision updates.

oseikuffuor1 (approved):
Looks good. Thanks Victor! I am also noting the needed mixed-precision updates as @rfalgout mentioned.

victorapm:
Thanks, all! `mup_check` and `mup_code` came out with no changes. Looking at mup.fixed in parcsr_ls, it seems the SLUDist routines have never been added (probably because they are defined only when using superlu_dist and we haven't testes superlu_dist + mixed precision yet). Should this be done in a separate PR?

rfalgout:
> Thanks, all! `mup_check` and `mup_code` came out with no changes. Looking at mup.fixed in parcsr_ls, it seems the SLUDist routines have never been added (probably because they are defined only when using superlu_dist and we haven't testes superlu_dist + mixed precision yet). Should this be done in a separate PR?

I forgot about that.  Daniel's original `mup_check` added a couple of other configuration options that probably turned this on (I can't remember).  I removed the extra options for some reason that I also can't remember. :)  I think it introduced another set of issues that I wanted to save for later.  I guess now is the time to look into it again.  For this PR, don't worry about it.  Thanks!

oseikuffuor1:
> Thanks, all! `mup_check` and `mup_code` came out with no changes. Looking at mup.fixed in parcsr_ls, it seems the SLUDist routines have never been added (probably because they are defined only when using superlu_dist and we haven't testes superlu_dist + mixed precision yet). Should this be done in a separate PR?

Thanks Victor. I agree with Rob's comment. Don't worry about it for this PR. We can address it later or perhaps in the mixed-kyrlov PR.

victorapm:
Thank you Daniel and Rob, I'm leaving the mixed precision work for another PR

PR #1426, opened 2025-11-19
Add initial CI capabilities

victorapm:
This PR contains the initial steps toward implementing CI in hypre.

1. Added `.github/workflows/ci.yml` that runs on pushes to `master` and on pull requests
2. Configured a Debug CMake build for CI
3. Added `check` target to CI to quickly test the compiled library.
4. Added a `require-label` job that fails pull requests lacking the `Run CI` label; developers must add the label to trigger the build job and unblock merging.

PR #1451, opened 2026-01-06
[CI]: Add code checks

victorapm:
Adds basic code checks to CI involving:

- Code style
- Mixed precision code check (`mup_check` script)
- License headers
- MPI function usage (enforce `hypre_MPI_`)
- basic data types usage (enforce `HYPRE_Int`, `HYPRE_Real`,...)
- case insensitive filenames

These tests are also done in `machine-tux.sh`. They are here in CI to help catching issues earlier.

CI will fail an print the contents of the `.err` file if any of these checks return such a file.

Depends on #1426


victorapm:
Hi @rfalgout @liruipeng @oseikuffuor1 @waynemitchell , this PR adds basic code checks to CI. It should go after #1426 is merged in. Would you be able to take a look at both? Thank you!

.github/workflows/ci.yml
```diff
@@ -57,3 +57,158 @@ jobs:
...
+          else
+            echo "astyle version: $INSTALLED_VERSION"
+          fi
+
+          # Run astyle-apply.sh to check and apply formatting
+          # This will modify files in place if formatting is needed
+          ./config/astyle-apply.sh .
+
+          # Check if any source files were modified (formatting changes)
+          if ! git diff --exit-code -- '*.c' '*.h' '*.cc' '*.cpp' '*.cxx' '*.C' '*.hpp' '*.hxx' '*.H'; then
```
rfalgout: Why are all these file extensions needed here?

victorapm: Some of them refer to files in `fei-mv`, but I just realized that we don't run the code style in that folder, so I can simplify this list. Let me do it. Thanks for bringing this up!

.github/workflows/ci.yml
```diff
@@ -57,3 +57,158 @@ jobs:
...
+            echo "::error::Code formatting check failed. Some source files need formatting."
+            echo "::error::Run './config/astyle-apply.sh .' from the src directory to fix formatting issues."
+            echo "::error::Then commit the formatted files."
+            echo "::error::Modified files:"
+            git diff --name-only -- '*.c' '*.h' '*.cc' '*.cpp' '*.cxx' '*.C' '*.hpp' '*.hxx' '*.H'
+            exit 1
+          fi
+
+          # Check if any header files were generated/modified
+          if ! git diff --exit-code -- '*_hypre*.h'; then
```
rfalgout: I think I answered my own question.  Looks like astyle-apply.sh also runs `headers` and you are trying to distinguish between the two.  I'm not sure we need to do that.  In either case, we should simply run `astyle-apply.sh`.  Being specific here ties the yaml code too tightly to the script itself - more maintenance worries.

victorapm: Sounds a good plan to me, I'll update!

.github/workflows/ci.yml
```diff
@@ -57,3 +57,148 @@ jobs:
...
+            cat check-headers.err
+            exit 1
+          fi
+          echo "Headers check passed"
+
+      - name: Check case-insensitive filenames
+        working-directory: AUTOTEST
+        run: |
+          SRC_DIR="${{ github.workspace }}/src"
+          ./test.sh check-case.sh "$SRC_DIR/.."
```
rfalgout, line 188: This just made me wonder why we didn't catch the case issue recently reported by Paul on the ALE3D team.  The current script only catches this within subdirectories.  I'll look into potentially changing that (not sure yet if we should).

rfalgout (approved):
This looks good to me, and it makes sense to always do these checks when pushing code.  Thanks!

.github/workflows/ci.yml
victorapm: That's an interesting point... I can take a look if we decide to change anything. Let me know!

.github/workflows/ci.yml
```diff
@@ -57,3 +57,148 @@ jobs:
...
+      - name: Check code formatting
+        working-directory: src
+        run: |
+          # Configure git to recognize the workspace
+          git config --global --add safe.directory ${{ github.workspace }}
+
+          # Check astyle version
+          ASTYLE_VERSION="Artistic Style Version 3.1"
+          INSTALLED_VERSION=$(astyle --version)
+          if [ "$INSTALLED_VERSION" != "$ASTYLE_VERSION" ]; then
```
waynemitchell: Why not specify the astyle version when installing above?

victorapm: Sounds a good idea, thanks! Done in [9f97fdc](https://github.com/hypre-space/hypre/pull/1451/commits/9f97fdcd14ffb91373b93865793be6ac1c4b29d5)

oseikuffuor1 (approved):
Thanks Victor!

PR #1452, opened 2026-01-08
Update and add to sycl/Aurora regression tests

waynemitchell:
This PR updates the regression tests that we run on Aurora and also adds sstruct tests.


src/test/TEST_sstruct/neumann.saved.aurora
```diff
@@ -0,0 +1,48 @@
+# Output file: neumann.out.0
+Iterations = 6
+Final Relative Residual Norm = 2.366245e-07
+
+# Output file: neumann.out.1
+Iterations = 100
+Final Relative Residual Norm = 1.495032e+90
```
waynemitchell, line 7: I'm getting bad results for a couple of the neumann regression jobs. The reason is that this is a singular problem, and the coarsest level system is a zero matrix. There is a check for this in pfmg setup:
https://github.com/hypre-space/hypre/blob/master/src/struct_ls/pfmg_setup.c#L363
This in turn relies on `hypre_StructMatrixZeroDiagonal()`, which checks for exact equality to 0.0:
https://github.com/hypre-space/hypre/blob/master/src/struct_mv/struct_matop.c#L87
If I run on my laptop, the coarsest-level matrix is exactly zero and is correctly caught by the check in pfmg setup (and everything converges fine, etc). If I run on Aurora, however, the coarsest-level matrix has close-to-zero entry on the diagonal (on the order of 1e-15), and the check does not identify it as a zero matrix. Then during the solve, we end up doing jacobi relaxation on the coarsest level, which makes things blow up.

What's the best fix here...? Should we change the `hypre_StructMatrixZeroDiagonal()` to use a tolerance instead of checking exact equality to zero? Then we have the issue of how to set the tolerance in a robust way...

victorapm: Hmm, that's tricky... How about we use a check with `HYPRE_REAL_MIN`? I think we do that already in other parts of the code

victorapm (approved):

src/test/TEST_sstruct/neumann.saved.aurora
waynemitchell: `HYPRE_REAL_MIN` is too small in this case.

rfalgout: I think one thing we could do is compare the diagonal entries on the coarse level to the corresponding ones on the previous level (or any other level, actually).  So, something like this: `if diagcoarse < theta*diagcompare` then count `diagcoarse` as "zero", where `diagcoarse` is a diagonal entry on the coarsest grid, `diagcompare` is from another level, and `theta` is set to something like 1e-12.  I'm not sure this would be so easy to implement, though.  You would need to pass something else into ZeroDiagonal() to serve as a reference for the comparison.  Probably a coarsest-grid-sized vector of values that holds the diagonal entries from one of the other matrices.

Alternatively, if we have matrix row sums lying around from previous levels, if a diagonal entry on the coarsest level is of the same order as the corresponding row sum, then we can count the entry as "zero".

In general, I think we need something to compare to to get the scaling right.  To account for coefficient jump problems, we can't use just one comparison value for the entire matrix.  Anyway, that's the best I can offer.  Let me know if you have other ideas.

rfalgout: By the way, I don't think we need to fix this for this PR.  It would require broader testing, so you can just leave the big numbers in the saved files here.

rfalgout (approved):
Looks good.  Thanks!

src/test/TEST_sstruct/neumann.saved.aurora
waynemitchell: OK, yes, I think your proposals here make sense, Rob. I'll leave this out of scope for this PR as you suggest.

PR #1448, opened 2025-12-29
Add ParCSR overlap capabilities

victorapm:
This PR implements the core infrastructure for computing overlapping subdomains in ParCSR matrices. This is the first step towards implementing overlapping Schwarz preconditioners in hypre.

Main changes are given below:

* `par_csr_overlap.h`: Defines the `hypre_OverlapData` structure and associated accessor macros to store extended subdomain information.
* `par_csr_overlap.c`: Implements the overlap computation logic, including:
  * `hypre_OverlapDataCreate` and `hypre_OverlapDataDestroy`.
  * `hypre_ParCSRMatrixComputeOverlap`: Determines extended row indices for a given overlap order.
  * `hypre_ParCSRMatrixGetExtendedRows`: Handles MPI communication to retrieve external matrix rows.
  * `hypre_ParCSRMatrixBuildExtendedMatrix`: Constructs the local extended matrix.
* `src/parcsr_mv/CMakeLists.txt`: Added new source files to the build system.
* `src/parcsr_mv/protos.h`: Added prototypes for the new functions.
* `src/parcsr_mv/par_csr_matrix.c`: Minor cleanup in error handling logic.

TODOs

- [x] Test MPI neighbors of neighbors (Tests 3, 6, 7 in `test_csr_overlap.c`)
- [x] Variable naming
- [x] Initial timings

PR #1453, opened 2026-01-09
Unit tests for ParCSRMatrix overlap code

victorapm:
- Refactor `test_csr_overlap.c` and reduce its size
- Add `benchmark-matmat` mode to `test_csr_overlap.c` to serve as a reference for ParCSRMatrix overlap calculations
- Add `TEST_unit` folder in `src/test` for unit testing
- Add unit test job to `machine-tux408.sh` regression test

Depends on #1448


rfalgout (approved):
Looks good.  I don't understand the need for the cmake targets, but that's okay.  Thanks!

victorapm:
Thanks, Rob! Just removed an old `ctest` entry that is not needed. The other two remaining are for 1. running the unit tests when invoking `ctest` from the build directory and 2. a convenience target that allows running unit tests via `make unit_tests`

PR #1454, opened 2026-01-09
Fix incorrect handling of rows in `hypre_ParCSRMatrixGenerateFFFC`

victorapm:
Fixes a bug in `hypre_ParCSRMatrixGenerateFFFCHost` where the code assumed that when `S == A`, the diagonal is always the first entry in each row. For rows without a diagonal, this skipped the first off-diagonal entry, causing missing entries in `A_CF` and `A_FC`.


rfalgout (approved):
Looks good to me.  Thanks!

PR #1455, opened 2026-01-13
Minor update to AUTOTEST/check-case.sh for portability

rfalgout:


rfalgout:
Hi @victorapm .  This just fixes a portability issue on my mac.  I don't think we should try to generalize the script to check case outside of the same directory.  There are too many issues with that.

victorapm (approved):
Sounds good, thanks Rob!

PR #1462, opened 2026-01-23
Fix an autotest egrep issue on macos

rfalgout:
The use of '*' in the `configure.filters` file was causing an error message on macos.  The addition of these lines was originally intended to fix tests on Lassen, which is no longer around.  Not sure now where to test it.  Let me know if you have ideas @waynemitchell .


victorapm:
Hi Rob, we should test this on `matrix`. I can give it a try (only the job without umpire)

victorapm:
This branch worked fine on matrix, i.e., it filtered correctly the warning message

rfalgout:
BTW, I will not merge this until after the release is done.

victorapm (approved):

PR #1464, opened 2026-01-24
MinGW-w64: fix undefined ffs() in src/utilities/_hypre_utilities.h

brechtsanders:
Fix undefined ffs() in src/utilities/_hypre_utilities.h when building with MinGW-w64.

### DIFF
--- a/src/utilities/_hypre_utilities.h
+++ b/src/utilities/_hypre_utilities.h
@@ -2678,6 +2678,8 @@ first_lsb_bit_indx( hypre_uint x )
          x >>= 1;
       }
    }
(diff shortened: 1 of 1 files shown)


liruipeng (approved):

liruipeng (changes requested):
see Rob's comment on where to put these line and running header. Thanks @brechtsanders 

> This change needs to be put in ‘hopscotch_hash.h’ first, then run ‘headers’.  Thanks!
>  
> -Rob

victorapm:
I'm happy to add this change to #1463, which contains a few other fixes, if that makes sense

PR #1469, opened 2026-02-03
Fix option tables in SSAMG docs

victorapm:
Latex syntax in tables was not being parsed correctly by the hypre docs scripts. This PR updates the table to a simpler format

### DIFF
--- a/src/sstruct_ls/HYPRE_sstruct_ls.h
+++ b/src/sstruct_ls/HYPRE_sstruct_ls.h
@@ -316,10 +316,8 @@ HYPRE_SStructSSAMGSetRelChange(HYPRE_SStructSolver solver,
  *
  * Current operators set by {\tt rap\_type} are:
  *
- * \begin{tabular}{l@{ -- }l}
- * 0 & Galerkin (default) \\
- * 1 & non-Galerkin 5-pt or 7-pt stencils \\
- * \end{tabular}
+ *    - 0 : Galerkin (default)
+ *    - 1 : non-Galerkin 5-pt or 7-pt stencils
  *
  * Both operators are constructed algebraically. The non-Galerkin option
  * maintains a 5-pt stencil in 2D and a 7-pt stencil in 3D on all grid levels.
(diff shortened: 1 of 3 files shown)


src/sstruct_ls/HYPRE_sstruct_ls.h
```diff
@@ -363,12 +359,10 @@ HYPRE_SStructSSAMGSetInterpType(HYPRE_SStructSolver solver,
...
  * Current relaxation methods set by {\tt relax\_type} are:
  *
- * \begin{tabular}{l@{ -- }l}
- * 0  & Jacobi \\
- * 1  & Weighted Jacobi (default) \\
- * 2  & L1-Jacobi \\
- * 10 & Red/Black Gauss-Seidel (symmetric: RB pre-relaxation, BR post-relaxation) \\
- * \end{tabular}
+ *    - 0  : Jacobi
+ *    - 1  : Weighted Jacobi (default)
```
waynemitchell, line 363: Looking at ssamg.c, it seems the default relax type is 0.

https://github.com/hypre-space/hypre/blob/master/src/sstruct_ls/ssamg.c#L32

Should we change the default or list 0 as the default in the docs here? Not sure what the best default relaxation really is.

src/test/sstruct.c
```diff
@@ -2312,13 +2312,19 @@ PrintUsage( char *progname,
...
+      hypre_printf("  -relax <r>         : (Sys)PFMG - relaxation type\n");
       hypre_printf("                        0 - Jacobi\n");
       hypre_printf("                        1 - Weighted Jacobi (default)\n");
       hypre_printf("                        2 - R/B Gauss-Seidel\n");
       hypre_printf("                        3 - R/B Gauss-Seidel (nonsymmetric)\n");
       hypre_printf("\n");
-      hypre_printf("                       ParCSR - relaxation type\n");
+      hypre_printf("                       SSAMG - relaxation type\n");
+      hypre_printf("                        0 - Jacobi\n");
+      hypre_printf("                        1 - Weighted Jacobi (default)\n");
```
waynemitchell, line 2323: Same question on the default here.

src/sstruct_ls/HYPRE_sstruct_ls.h
victorapm: Hi Wayne, we should use weighted Jacobi as default since it's more robust

waynemitchell: OK, I changed the default ssamg relax_type to 1.

waynemitchell:
@victorapm , I'm good to merge this if you are.

waynemitchell (approved):

PR #1472, opened 2026-02-06
Update memory location check for ParCSRMatrix

victorapm:
Replace check on abstract memory location with the actual memory location

Fixes issues with printing SStructMatrix to file and solving with SSAMG after reading it.


rfalgout (approved):

waynemitchell (approved):
Thanks for finding and fixing this, Victor!

PR #1474, opened 2026-02-11
Improve build robustness with external BLAS/LAPACK

### LINKED ISSUE #1467: Hypre LOBPCG solver fails with 64-bit integers
I have noticed that the LOBPCG eigensolver fails when Hypre is built with 64-bit integers, but works with 32-bit integers. I have verified this by running [MFEM's example 11](https://github.com/mfem/mfem/blob/master/examples/ex11p.cpp), which implements a simple eigenvalue problem with a diffusion integrator. When MFEM is built against a 32-bit Hypre, the example runs normally and returns the correct result. If it is built against a 64-bit Hypre, I get the following output:

```
Number of unknowns: 1361

Solving generalized eigenvalue problem with preconditioning

block size 5

No constraints

Error in LOBPCG:
Bad initial vectors: orthonormalization failed
DPOTRF INFO = 129944235540480

Eigenvalue lambda                    nan
Eigenvalue lambda                    nan
Eigenvalue lambda                    nan
Eigenvalue lambda                    nan
Eigenvalue lambda                    nan
Residual   0.00000000000000e+00
Residual   0.00000000000000e+00
Residual   0.00000000000000e+00
Residual   0.00000000000000e+00
Residual   0.00000000000000e+00

0 iterations
```

Is this issue known? And if so, is there a fix for it other than using 32-bit integer Hypre?

victorapm:
When users request:

1. An external (32-bit) BLAS/LAPACK distribution in hypre (e.g., spack does it by default)
2. 64-bit integers support via the BigInt build option (`HYPRE_ENABLE_BIGINT`) 

hypre will build successfully, but will lead to runtime errors such as the one reported in https://github.com/hypre-space/hypre/issues/1467#issue-3876485233 

There are a couple of ways of handling this issue:

1. Use `HYPRE_ENABLE_MIXEDINT`, so hypre will pass 32-bit integer pointers to BLAS/LAPACK as we rely on `HYPRE_Int`
2. Link to an external BLAS/LAPACK distribution that is 64-bit enabled
3. Do not rely on external BLAS/LAPACK (instead rely on our internal one, which is usually the case).

Still, the scenario that was originally described (BigInt + external 32-bit BLAS/LAPACK) could be better handled in hypre, which brings to the purpose of this PR: 

- if users request "BigInt + external 32-bit BLAS/LAPACK", this branch will error out during build time giving a nice error msg to users (only via our primary cmake build):
```
CMake Error at config/cmake/HYPRE_CMakeUtilities.cmake:742 (message):
  Incompatible external LAPACK integer ABI for HYPRE_ENABLE_BIGINT=ON.

  The configured external LAPACK appears to use LP64 integers (32-bit), but
  BIGINT requires an ILP64-compatible BLAS/LAPACK integer interface.

  Use one of:

    1) Provide ILP64 BLAS/LAPACK libraries
    2) Set -DHYPRE_ENABLE_MIXEDINT=ON (recommended for external LP64 BLAS/LAPACK)
    3) Use internal BLAS/LAPACK: -DHYPRE_ENABLE_HYPRE_BLAS=ON -DHYPRE_ENABLE_HYPRE_LAPACK=ON

Call Stack (most recent call first):
  CMakeLists.txt:483 (check_bigint_external_lapack_abi)
```

A separate PR (on spack-packages) will add this logic to the spack build as well.

Closes #1467

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -217,6 +217,7 @@ set_hypre_option(SKIP TPL_LAPACK_LIBRARIES              "Absolute paths to LAPAC
 set_hypre_option(SKIP TPL_FEI_INCLUDE_DIRS              "Absolute paths to FEI include directories [Optional]." "")
 
 # Extra flags
+set_hypre_option(SKIP HYPRE_ALLOW_BIGINT_WITH_LP64_BLAS "Allow potentially unsafe BIGINT with external LP64 BLAS/LAPACK" OFF)
 set(HYPRE_WITH_EXTRA_CFLAGS "" CACHE STRING             "Define extra C compile flags")
 set(HYPRE_WITH_EXTRA_CXXFLAGS "" CACHE STRING           "Define extra CXX compile flags")
 set(HYPRE_SYCL_TARGET "" CACHE STRING                   "Target SYCL architecture, e.g. 'spir64_gen'.")
@@ -479,6 +480,7 @@ setup_tpl(magma)
 setup_tpl(caliper)
 setup_tpl_or_internal(blas)
 setup_tpl_or_internal(lapack)
+check_bigint_external_lapack_abi()
 
 # Some TPLs need C++ to be enabled (GPU already enables C++ by default)
 if(HYPRE_NEEDS_CXX AND NOT HYPRE_USING_GPU)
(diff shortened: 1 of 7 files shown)


rfalgout (approved):
Looks good, Victor.  My only concern with all the checking we are starting to do is that it will take some effort to maintain the check scripts over time.  But we can cross that bridge if we ever get to it.  Thanks!

PR #1254, opened 2025-03-29
Add support to tagged residual norms

victorapm:
This PR introduces several changes focusing on adding support for tagged vectors and improving the GMRES solver's functionality. The most important changes include the addition of new functions for setting tags on vectors, modifications to the GMRES solver to handle tagged vectors, and updates to the documentation to reflect these changes.

Example usage:

When solving a linear elasticity problem, now we can ask for the residual norms of individual DOF variables, e.g., displacements. This is useful for multiphysics problems for understanding how each variable type converges.

```
$ mpirun -np 4 test/ij -fromfile test/TEST_ij/data/beam_tet_dof2475_np4/A.IJ -fromrhsfile test/TEST_ij/data/beam_tet_dof2475_np4/b.IJ -funcsinterleaved -nf 3 -solver 3 -k 500 -iout 3
=============================================
Hypre init times:
=============================================
Hypre init:
  wall clock time = 0.000001 seconds
  wall MFLOPS     = 0.000000
  cpu clock time  = 0.000001 seconds
  cpu MFLOPS      = 0.000000


Using HYPRE_DEVELOP_STRING: v2.32.0-41-g022e6c3b7 (branch tagged-res; not the develop branch)

Running with these driver parameters:
  solver ID    = 3

=============================================
Generate Matrix:
=============================================
Spatial Operator:
  wall clock time = 0.004846 seconds
  wall MFLOPS     = 0.000000
  cpu clock time  = 0.004847 seconds
  cpu MFLOPS      = 0.000000

  Number of vector components: 1
  RHS vector has unit coefficients
  Initial guess is 0
  Calling BuildFuncsInterleaved with num_functions = 3
=============================================
IJ Vector Setup:
=============================================
RHS and Initial Guess:
  wall clock time = 0.000062 seconds
  wall MFLOPS     = 0.000000
  cpu clock time  = 0.000062 seconds
  cpu MFLOPS      = 0.000000

Solver: AMG-GMRES
HYPRE_GMRESGetPrecond got good precond


 Num MPI tasks = 4

 Num OpenMP threads = 1


BoomerAMG SETUP PARAMETERS:

 Max levels = 25
 Num levels = 6

 Strength Threshold = 0.250000
 Interpolation Truncation Factor = 0.000000
 Maximum Row Sum Threshold for Dependency Weakening = 1.000000

 Coarsening Type = HMIS 
 measures are determined locally


 No global partition option chosen.

 Interpolation = extended+i interpolation
 Number of functions = 3
[...]

PR #1476, opened 2026-02-14
Add Krylov residual print interface

victorapm:
This branch modularizes Krylov residual-print formatting and reduces the amount of redundant code

It extends the work started in #1254 so we have multi-tag residual reporting in FlexGMRES (and possibly other solvers in the future).


src/parcsr_ls/HYPRE_parcsr_flexgmres.c
```diff
@@ -34,7 +34,7 @@ HYPRE_ParCSRFlexGMRESCreate( MPI_Comm comm, HYPRE_Solver *solver )
          hypre_ParKrylovMatvecCreate,
          hypre_ParKrylovMatvec,
          hypre_ParKrylovMatvecDestroy,
-         hypre_ParKrylovInnerProd,
+         (hypre_KrylovPtrToInnerProd) hypre_ParKrylovInnerProdTagged,
```
rfalgout: Why do we need the cast for just the one function?  Should we be prototyping this Create function differently instead?

victorapm: Thanks, Rob! That sounds a better solution. Let me work on it

rfalgout (approved):
This looks good modulo the one question.  Thanks!

src/parcsr_ls/HYPRE_parcsr_flexgmres.c
victorapm: Fixed in [df9caa1](https://github.com/hypre-space/hypre/pull/1476/commits/df9caa14ee106fb9ac7caf22e97005f0cca10187)

PR #1485, opened 2026-03-04
Reuse GMRES for operators of different sizes

nmnobre:
Hi,

This PR is really an issue with a suggested (incomplete) change in case we decide to take it.

The issue:
If a downstream app sets up a GMRES solver, conducts the solve with a given matrix A, but then resizes A (think, e.g. AMR) and decides to reuse the same solver, even if `hypre_GMRESSetup` is called again, the internal vectors used by the solver, e.g. basis vectors `p` and residual `r`, aren't resized and hypre eventually seg faults.

I see two options:
1) I can see for PCG solvers, `hypre_PCGSetup` reallocates its internal vectors unconditionally. The suggestion here does the same, and in case we decide to do this, I'll expand the change to the other GMRES variants. I'm not sure this is the most efficient though, and I'd like to understand why we decided to do this for PCG but not GMRES up until now.
2) The downstream app is responsible for reconstructing the GMRES solver if A changes its size, which is fine, though the design will (continue to) be inconsistent with what we do with PCG.

Cheers,
-Nuno

### DIFF
--- a/src/krylov/gmres.c
+++ b/src/krylov/gmres.c
@@ -220,42 +220,55 @@ hypre_GMRESSetup( void *gmres_vdata,
     * ecompute phases of matvec and the preconditioner.
     *--------------------------------------------------*/
 
-   if ((gmres_data -> p) == NULL)
+   if ( (gmres_data -> p) != NULL )
    {
-      (gmres_data -> p) = (void**)(*(gmres_functions->CreateVectorArray))(k_dim + 1, x);
+      for (HYPRE_Int i = 0; i < (gmres_data -> k_dim + 1); i++)
+      {
+         if ( (gmres_data -> p)[i] != NULL )
+         {
+            (*(gmres_functions->DestroyVector))( (gmres_data -> p) [i]);
+         }
+      }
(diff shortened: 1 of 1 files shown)


victorapm:
Hi Nuno,

thank you once again for your contribution and careful investigation inside the hypre library

We had a team discussion about this topic. This pull request is valid and fully correct. However, if we look at the bigger picture, which is supporting:

1.  `KrylovSetup` on system A,
2. `KrylovSolve`on system A,
3. `KrylovSetup` on system B.

We realized this would require a reasonable amount of extra work from our end (propagating this behavior to all krylov solvers, adding regression tests, etc...).

If we look at current alternatives already implemented in the library, this situation could be handled by simply adding a `KrylovDestroy` call between steps 2 and 3. From a computational cost perspective, this approach and the originally intended one should be very similar.

Therefore, for now, we prefer to keep the solvers lightweight and rely on the user to reconstruct them (call Destroy) when the problem dimensions change. We will, however, keep this issue in mind for future refactors where we might aim for better consistency across all Krylov methods.

We appreciate the effort you put into this!

PR #1489, opened 2026-03-17
sstruct driver fixes

waynemitchell:
This fixes a couple of issues with the sstruct driver.

First, when reading sstruct vectors from file, the underlying `hypre_StructVectorReadData()` routine was re-initializing the struct vectors, causing the data not to be in the correct memory location for the sstruct vector (this screws up things like matvec down the line, since the parcsr portion of the matvec will point to the wrong memory location for the vector data).

Second, I changed the right-hand side construction such that it can do 0, 1, random right-hand sides while also reading a matrix from file (previously reading a matrix from file REQUIRED reading the rhs vector from file as well). I added the random rhs option while I was at it just for fun.

### DIFF
--- a/src/struct_mv/struct_vector.c
+++ b/src/struct_mv/struct_vector.c
@@ -1601,7 +1601,6 @@ hypre_StructVectorReadData( FILE               *file,
    }
 
    /* Set vector values */
-   hypre_StructVectorInitialize(vector, 1);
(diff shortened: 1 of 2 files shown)


rfalgout (approved):

victorapm (approved):
Thanks, Wayne!

PR #1492, opened 2026-03-18
Fix Device/Kernel functions names

liruipeng:
Fixing the naming of `hypre_GPUKernelFoo` and `hypre_FooDevice` functions.


victorapm (approved):
Looks great, thanks for all the work!

We still have functions that don't follow the code style in the wiki, e.g., `hypre_qsort0` (and alike), `hypre_index_of_minimum`, `hypre_matinv`, ... 

We could fix it in another PR and add some sort of script to check for these things (happy to volunteer for this work if that makes sense)

@rfalgout

waynemitchell (approved):
Thanks for doing all this, Rui Peng!

waynemitchell:
@liruipeng , I found another, similar naming convention issue in `utilities/functors.h`:
https://github.com/hypre-space/hypre/blob/master/src/utilities/functors.h
We have `hypreFunctor_...`. Don't know if you want to change these as well here or leave it for later.

waynemitchell:
There are also some `hypreSycl_*` names here:
https://github.com/hypre-space/hypre/blob/master/src/utilities/_hypre_onedpl.hpp
But I will be addressing those in an upcoming PR, so don't change those.

liruipeng:
> @liruipeng , I found another, similar naming convention issue in `utilities/functors.h`: https://github.com/hypre-space/hypre/blob/master/src/utilities/functors.h We have `hypreFunctor_...`. Don't know if you want to change these as well here or leave it for later.

Fixed those names, Thanks!

liruipeng:
> There are also some `hypreSycl_*` names here: https://github.com/hypre-space/hypre/blob/master/src/utilities/_hypre_onedpl.hpp But I will be addressing those in an upcoming PR, so don't change those.

I was trying to fix them but it seems that @waynemitchell already did that. Thanks!

rfalgout (approved):
This looks good to me, as long as all the testing goes through okay (I know I wasn't requested to review, but I decided to give it a look).  Thanks!

src/struct_mv/mup_fixed_gpu.c
```diff
@@ -0,0 +1,31 @@
+
```
oseikuffuor1, line 1: @liruipeng Makefile needs updating to include these new mup_*_gpu.c functions. See [here](https://github.com/hypre-space/hypre/blob/a6671459b740a50f084afe63dba357d73d285a20/src/IJ_mv/Makefile#L51-L60) for an example. Similarly for CMakeLists update.

src/struct_mv/_hypre_struct_mv_mup.hpp
```diff
@@ -0,0 +1,60 @@
+
```
oseikuffuor1, line 1: @liruipeng headers script also needs updating to include the new _hypre_*_mup.hpp file. See [here](https://github.com/hypre-space/hypre/blob/hypreGPUKernel/src/IJ_mv/headers) for an example (grep for hypre_MP_BUILD). You'll have to re-run headers again after that.  In general both generated .h and .hpp headers will have the same additions. Note the order and location of this block is important.

oseikuffuor1 (approved):
Rui, this looks good to me, apart from the comments I left for you. The struct_mv folder needs a few more updates for multiprecision. Let me know if you need help with that or want me to go ahead and make the changes for you.

src/struct_mv/mup_fixed_gpu.c
liruipeng: @oseikuffuor1  Thanks, can you take another look?

src/struct_mv/_hypre_struct_mv_mup.hpp
liruipeng: @oseikuffuor1  Thanks, can you take another look?

src/struct_mv/_hypre_struct_mv.h
```diff
@@ -3744,6 +3744,7 @@ hypre__J = hypre__thread;  i1 = i2 = 0; \
 #ifndef hypre_MP_BUILD
 #include "_hypre_struct_mv_mup_undef.h"
 #include "_hypre_struct_mv_mup.h"
+#include "_hypre_IJ_mv_mup.hpp"
```
oseikuffuor1: Name should be _hypre_struct_mv_mup.hpp. Similar changes below

src/struct_mv/_hypre_struct_mv.hpp
```diff
@@ -7,6 +7,10 @@
 
 #include "_hypre_utilities.hpp"
 
+#ifdef HYPRE_MIXED_PRECISION
+#include "_hypre_IJ_mv_mup_def.h"
```
oseikuffuor1: See above comment - IJ should be struct.

src/struct_mv/_hypre_struct_mv.hpp
```diff
@@ -1634,5 +1638,17 @@ else                                                            \
 }
 #endif
 
+#ifdef HYPRE_MIXED_PRECISION
+/* The following is for user compiles and the order is important.  The first
+ * header ensures that we do not change prototype names in user files or in the
+ * second header file.  The second header contains all the prototypes needed by
+ * users for mixed precision. */
+#ifndef hypre_MP_BUILD
```
oseikuffuor1, line 1646: same comment

src/struct_mv/headers
```diff
@@ -81,6 +81,7 @@ cat >> $INTERNAL_HEADER <<@
 #ifndef hypre_MP_BUILD
 #include "_hypre_struct_mv_mup_undef.h"
 #include "_hypre_struct_mv_mup.h"
```
oseikuffuor1, line 83: same comments here too

src/struct_mv/_hypre_struct_mv.h
liruipeng: fixed. Thanks

src/struct_mv/_hypre_struct_mv.hpp
liruipeng: fixed.

PR #1503, opened 2026-04-14
Fix code annotations

victorapm:
- Fix `HYPRE_ANNOTATE` region mismatches
- Add python script to scan for issues
- Add CI job for checking code annotations. See [failing example in CI](https://github.com/hypre-space/hypre/actions/runs/24380604503/job/71203102069?pr=1503#step:4:1)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -240,9 +240,13 @@ jobs:
           - mpi
           - memory
           - case
+          - annotation
     steps:
       - name: Checkout sources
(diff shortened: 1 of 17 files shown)


rfalgout (approved):
Looks good, Victor.  My only concern is the complexity of the python script.  But since it only affects the check, I think we can run with it and see how things go.  Thanks!

victorapm:
Thanks, Rob! Sure, I'll take care of the script if there are any issues in the future

PR #1504, opened 2026-04-14
Add internal support to solver reuse

victorapm:
This PR adds a new flag in the base solver indicating whether setup has been called or not.

The main driver for this feature is to allow reuse of specific parts of MGR (F-relaxation, G-relaxation...) during a sequence of linear solves


rfalgout (approved):
This all looks good to me.  I did not look carefully through the logic of the MGR changes, but I get the overall gist of what's going on.  Thanks!

src/test/ij.c
```diff
@@ -9672,59 +9715,77 @@ main( hypre_int argc,
...
-      {
-         HYPRE_BoomerAMGSetMaxIter(amg_solver, max_iter);
-         HYPRE_BoomerAMGSetPrintLevel(amg_solver, 3);
-      }
-      else
-      {
-         HYPRE_BoomerAMGSetMaxIter(amg_solver, precon_cycles);
-         HYPRE_BoomerAMGSetTol(amg_solver, 0.0);
-         HYPRE_BoomerAMGSetPrintLevel(amg_solver, 1);
+         /* set the MGR coarse solver. Comment out to use default CG solver in MGR */
```
oseikuffuor1: We can remove this comment now right?

victorapm: Hi Daniel, I think the comment is still valid. Maybe I'm missing your point

oseikuffuor1: The part that says comment out to use default CG may not be necessary anymore since this call is embedded in the if-statement !use_default_cg above. Or am I missing something?

victorapm: I see now (it's been sometime I wrote this :). 

Agreed with you. I'll remove

src/parcsr_ls/par_mgr_setup.c
```diff
@@ -1303,11 +1102,13 @@ hypre_MGRSetup( void               *mgr_vdata,
...
 
                   /* Setup F-solver */
+                  /* hypre_ParPrintf(comm, */
+                  /*                 "hypre_MGRSetup: setting up F-solver at level %d\n", lev); */
                   fgrid_solver_setup(aff_solver[lev],
                                      A_ff_array[lev],
                                      F_fine_array[lev + 1],
                                      U_fine_array[lev + 1]);
-                  (mgr_data -> fsolver_mode) = 1;
+                  aff_solver_type[lev] = HYPRE_MGR_SOLVER_TYPE_USER;
```
oseikuffuor1: What does it mean for a user-provided solver type? Is the plan to support any arbitrary solver type by the user? something that is not supported in hypre?

oseikuffuor1 (approved):
@victorapm, this looks good to me overall and I agree with the structure you are trying to follow. I have a couple general comments:
1. These are a lot of changes and I suppose it will mean updating the reference manual for some of the options right?
2. Reusable and persistent suggest the same thing to me, but they are being used for two different sets of operations. Could you use a more descriptive name for the those two functions?

victorapm:
Thanks, Daniel!

1. Yes, we need to update the docs. I'm planning to do this on another PR with a major rewrite on the MGR docs (it's missing many things currently)
2. Thanks for the suggestion, I'll do it on this PR

src/parcsr_ls/par_mgr_setup.c
victorapm: 1. `HYPRE_MGR_SOLVER_TYPE_USER` means the user creates a solver object and pass it to MGR, e.g., BoomerAMG, ILU, etc...
2. This supports `hypre_BoomerAMG`, `hypre_ILU`, `hypre_FSAI`, and `hypre_MGRDirectSolver`
3. All options need to be supported by hypre, e.g., no external solver here

PR #1505, opened 2026-04-16
Update semi-struct code to use a single set of base boxes

rfalgout:
This PR changes the semi-structured code to use a single set of base boxes throughout.  Since the data space for vectors and matrices now aligns with these base boxes, there is no need to "rebase" to different base grids as before.  This simplifies the logic in Matvec in particular.

Lots of files have been touched, but most of it is due to the trivial macro name changes.  The best place to start is with the header files `struct_grid.h`, `struct_vector.h`, and `struct_matrix.h` because the comments at the top now have diagrams to help visualize what's going on internally.  After that, the main files to pay attention are:
```
struct_grid.c
coarsen.c
struct_vector.c
struct_matrix.c
utilities.c
communication_info.c
struct_communication.c
struct_matvec.c
```

Some additional details:

- Changed structures and accessor macros for StructGrid, StructVector, and StructMatrix
- Removed data mapping from StructVector because the data space and grid boxes now live on the same index space.
- Removed the `stride` argument from CreateCommInfo and CreateComputeInfo
- Made some name changes inside the `CommInfo` structure (for sanity)
- Updated the communication routines to use base boxnums passed in through the `BoxArrayArray` IDs of `send_boxes` and `recv_boxes` in the `CommInfo` structure.  This allows correct access into the data, since the data space now lines up with the base boxes.  See the changes in the `CreateCommInfo` routine.
- Removed VectorRebase from the code entirely
- PFMG can now use red/black GS with `v_memory_mode = 2` - before it always changed it to `0`

TODO:
- [x] Update the code for mixed precision
- [x] Change a few more comment blocks (e.g., Matvec and Matmult)


src/struct_mv/struct_vector.c
```diff
@@ -184,27 +92,21 @@ hypre_StructVectorDestroy( hypre_StructVector *vector )
 /*--------------------------------------------------------------------------
  * Set vector stride, nboxes, and boxnums.
  * If stride == NULL, set default values.
+ *
+ * RDF BASE: Need this only to set boxnums for now
  *--------------------------------------------------------------------------*/
 
 HYPRE_Int
 hypre_StructVectorSetStride( hypre_StructVector *vector,
                              hypre_IndexRef      stride )
```
victorapm: In this function, the input `stride` is never passed in to `hypre_StructGridComputeBoxnums` . Should we remove that input argument or fix the implementation?

rfalgout: I am going to remove this function altogether, possibly before merging this PR (see 'RDF BASE' comment at the top).  I will remove the `vec_boxnums` component of the vector at the same time.

rfalgout: I just removed this function.

victorapm: Thanks, Rob! I'm still going over everything :)

src/struct_mv/struct_copy.c
```diff
@@ -97,66 +97,3 @@ hypre_StructCopy( hypre_StructVector *x,
...
- *
- * A BoxArrayArray of boxes are needed- for each box of x, only an array
- * of subboxes (i.e., a boxarray for each box of x) are copied.
- *
- * RDF TODO: Remove this routine and the SStruct routine that calls it.  Need to
- * remove the FAC solver as well (it's the only thing that uses it).
- *--------------------------------------------------------------------------*/
-
-HYPRE_Int
-hypre_StructPartialCopy( hypre_StructVector  *x,
```
victorapm, line 111: Should we also remove this function's declaration in protos.h and regenerate the private headers?

rfalgout: Good catch!  Done.

src/struct_ls/pfmg_setup.c
```diff
@@ -74,22 +74,13 @@ hypre_PFMGSetup( void               *pfmg_vdata,
...
    hypre_sprintf(region_name, "%s", "PFMG-Init");
    HYPRE_ANNOTATE_REGION_BEGIN("%s", region_name);
    hypre_GpuProfilingPushRange(region_name);
    hypre_MemoryPrintUsage(comm, hypre_HandleLogLevel(hypre_handle()), "PFMG setup begin", 0);
 
-   /* RDF: For now, set memory mode to 0 if using R/B GS relaxation */
-   if (relax_type > 1)
-   {
-      v_memory_mode = 0;
-   }
```
waynemitchell, line 91: Out of curiosity, what was the initial reason for setting memory mode this way for R/B GS?

rfalgout: It didn't work otherwise, because the R/B smoother couldn't deal with a rebased vector.  Since Jacobi builds on Matvec, it was okay.  Now that we don't rebase, everything is good.

src/struct_mv/HYPRE_struct_vector.c
```diff
@@ -37,21 +37,6 @@ HYPRE_StructVectorDestroy( HYPRE_StructVector struct_vector )
...
-/*--------------------------------------------------------------------------
- *--------------------------------------------------------------------------*/
-
-/* RDF: Need a good user interface for setting the grid. */
-
-/*--------------------------------------------------------------------------
- *--------------------------------------------------------------------------*/
-
-HYPRE_Int
-HYPRE_StructVectorSetStride(HYPRE_StructVector vector,
```
waynemitchell, line 49: This function is still declared in `HYPRE_struct_mv.h`.

rfalgout: Good catch!  Fixed.  Thanks!

src/struct_mv/struct_matrix.h
```diff
@@ -150,24 +172,60 @@ typedef struct hypre_StructMatrix_struct
...
+#define hypre_StructMatrixRanBox(matrix, i) \
+hypre_StructMatrixBox(matrix, hypre_StructMatrixRanBoxnum(matrix, i))
+#define hypre_StructMatrixRanDataBox(matrix, i) \
+hypre_StructMatrixBoxDataBox(matrix, hypre_StructMatrixRanBoxnum(matrix, i))
+#define hypre_StructMatrixRanData(matrix, i, s) \
+hypre_StructMatrixBoxData(matrix, hypre_StructMatrixRanBoxnum(matrix, i), s)
+#define hypre_StructMatrixRanDataValue(matrix, i, s, index) \
+hypre_StructMatrixBoxDataValue(matrix, hypre_StructMatrixRanBoxnum(matrix, i), s, index)
+#define hypre_StructMatrixDomBaseBoxnum(matrix, i) \
+hypre_StructMatrixBaseBoxBoxnum(matrix, hypre_StructMatrixDomBoxnum(matrix, i))
```
victorapm:
```suggestion
hypre_StructMatrixBaseBoxnum(matrix, hypre_StructMatrixDomBoxnum(matrix, i))
```

src/struct_mv/coarsen.c
```diff
@@ -102,33 +102,63 @@ hypre_StructMapCoarseToFine( hypre_Index cindex,
...
+                             hypre_Index     stride1,
+                             hypre_IndexRef  origin2,  // = comp_origin on input
+                             hypre_Index     stride2,  // = comp_stride on input
+                             HYPRE_Int       ndim )
+{
+   HYPRE_Int  d;
+
+   for (d = 0; d < ndim; d++)
+   {
+      origin2[d] = (origin2[d] - origin1[d]) / stride1[d];
```
victorapm, line 163: Just an idea: we could add a `hypre_assert` here just to make sure that both numbers are divisable. This could help with future debugging if we ever need

victorapm (approved):
Thanks for all the work, Rob!

src/struct_mv/struct_matrix.h
rfalgout: Good catch!

src/struct_mv/coarsen.c
rfalgout: I just added a couple of asserts.  Let me know if it looks okay to you.  Thanks!

victorapm: Thanks, Rob!

waynemitchell (approved):
Thanks for getting all this done, Rob!

rfalgout:
@victorapm @waynemitchell - FYI, I am just going to run some tux tests on this before merging.  It should be all good.

rfalgout:
Hi @victorapm and @waynemitchell .  I had a couple of issues in the tux tests that needed to be resolved:

1. I discovered that the `cmake.filters` file was stripping out errors (all errors?) that should have been reported.  I made some changes to it, but it may affect how this runs on gpu systems where we put out warning messages about using umpire.  The filters files are tricky.  You can't just put anything in them because some characters are treated specially.  @victorapm Would you be able to check this on one of the gpu systems to see if it still filters the umpire stuff out correctly?
2. There was a small residual difference in a couple of SSAMG runs.  The runs are for the `marinak` problem on a single processor.  Namely the following two in `TEST_sstruct/solvers.jobs`:
```
mpirun -np 1  ./sstruct -in sstruct.in.marinak -P 1 1 1 -solver 4 > solvers.out.36
mpirun -np 1  ./sstruct -in sstruct.in.marinak -P 1 1 1 -solver 14 > solvers.out.135
```
Previously, the residual for each of these runs matched the 8-processor runs that follow them in the jobs file (see the lines for output files `37` and `136`).  The differences are small and normally might be attributable to the BoomerAMG part of SSAMG, but I don't think any of that part of the code changed.  I'm going to think about it a bit more, but this may be some subtle thing that would be too difficult to find and probably not an issue.  Note that if I use `-solver 1` (PFMG split) then the final residuals match between the 1-proc and 8-proc runs, but if I use `-solver 5` (BoomerAMG) they don't.

rfalgout:
Hi @waynemitchell and @victorapm .  With the two saved-file changes mentioned above, both the tux and tux-valgrind tests run cleanly except for one cmake build that Victor is taking a look at (the cmake issue is already a problem in the master branch but isn't getting reported correctly).  I am leaning towards merging this today even if we don't find answers to these two questions beforehand.  What do you think?

Though, maybe we should run a quick GPU test somewhere.

In the meantime, I will continue to think about and investigate the saved file question.  Let me know.  Thanks!

rfalgout:
Hi @victorapm and @waynemitchell .  Just another update.
- I've run this on tioga and it passes except for the umpire cmake filtering I mentioned before and the mixed-precision shared cmake test that Victor is also looking into.  Note that I had to make a change to the `AUTOTEST/cmake.sh` script so it would use the make options correctly.  These cmake changes are also necessary on the `master` branch.
- I haven't been able to figure out why the single-process marinak runs have residual differences.  I'm sure it's not due to AMG as a coarse solver, though, because the issue remains when using weighted Jacobi.  There must be some subtle difference relative to the previous code, but it only shows up on these two runs. :(
- I forgot to mention that the benchmark runs on tioga actually don't pass, but it's good news because the setup for the struct solvers is now faster. :) We will have to change the saved file at some point, but I want to wait until the dust settles.

waynemitchell:
> Hi @waynemitchell and @victorapm . With the two saved-file changes mentioned above, both the tux and tux-valgrind tests run cleanly except for one cmake build that Victor is taking a look at (the cmake issue is already a problem in the master branch but isn't getting reported correctly). I am leaning towards merging this today even if we don't find answers to these two questions beforehand. What do you think?
> 
> Though, maybe we should run a quick GPU test somewhere.
> 
> In the meantime, I will continue to think about and investigate the saved file question. Let me know. Thanks!

I'm fine if you want to merge this in its current state. Thanks for all the updates on how the testing is going!

rfalgout:
Hi @victorapm and @waynemitchell .  I found the SSAMG issue with the marinak problem and fixed it.  It was indeed a subtle bug that has to do with the fact that we compute the data size of the SStructVector separately from the computation of the data space in the StructVector and rely on them being computed in the same way.  This allows us to use one vector for both the Struct and ParCSR pieces of the SStructMatvec.  I had changed something in the StructVector data space computation that put them out of sync, but only in very special situations, hence only two regression tests being a little off.  The fix was simple (see commit 4eb51c7).

Note that one of the things Wayne and I are planning to look into is to remove this ghost-layer computation in SStruct and just do copies instead.  That will allow us to use vector resizing in SStruct as well.

I am rerunning regression tests, but I'm pretty confident I'll be merging this in today.  Thanks!

PR #1507, opened 2026-04-16
Update the error test to ignore the line numbers

rfalgout:
In the error test `TEST_error/error.sh`, replace the line numbers with `###` to minimize the need to update the saved file.


victorapm (approved):
Thanks, Rob!

PR #1511, opened 2026-04-20
Update struct driver help string

waynemitchell:
The current help string for the struct driver has the wrong default solver listed: SMG instead of PFMG (the actual default).


rfalgout (approved):
Thanks!

PR #1502, opened 2026-04-13
Remove DPCT

waynemitchell:
This removes the dependency on dpct, which will not be available in upcoming oneAPI releases. In addition, I did some name changes from `hypreSycl_*` to `hypre_*Sycl`. Important changes relating to working around dpct are in `_hypre_onedpl.hpp` and `device_utils.h`.

PR #1512, opened 2026-04-20
Add umpire + SYCL support

victorapm:
Add Umpire support to hypre's SYCL backend

In addition, this reduces the overhead of pool allocations by saving the address of the pool inside hypre's structs rather than relying on repetitive calls to `umpire_rm_get_allocator_by_name` 

#1502 should be merged into master first


rfalgout:
It looks like this needs to be brought up to date with master?  Maybe of these changes look the same as the previous PR, is that right?

victorapm:
Doing it now :)

victorapm:
@rfalgout this is ready now

rfalgout (approved):
Looks good to me.  Thanks!

src/config/cmake/HYPRE_SetupSYCLToolkit.cmake
```diff
@@ -109,3 +143,26 @@ endif()
...
+  set(_new "#if defined(UMPIRE_ENABLE_SYCL)\n    *SHCXX_rv = SH_this->makeAllocator<umpire::strategy::QuickPool>(\n        SHCXX_name, *SHCXX_allocator, initial_size, block, 256UL);\n#else\n    *SHCXX_rv = SH_this->makeAllocator<umpire::strategy::QuickPool>(\n        SHCXX_name, *SHCXX_allocator, initial_size, block);\n#endif\n    SHC_rv->addr = SHCXX_rv;\n    SHC_rv->idtor = 1;\n    return SHC_rv;\n    // splicer end class.ResourceManager.method.make_allocator_quick_pool")
+  string(REPLACE "${_old}" "${_new}" _content "${_content}")
+  if(NOT _content MATCHES "256UL")
+    message(WARNING "Failed to patch QuickPool alignment in ${_wrap_rm} — file format may have changed. Umpire allocations may not meet oneMKL's 256-byte alignment requirement.")
+    return()
+  endif()
+  file(WRITE "${_wrap_rm}" "${_content}")
+  message(STATUS "Patched Umpire wrapResourceManager.cpp: QuickPool SYCL pool alignment set to 256 bytes")
+endfunction()
```
waynemitchell, line 168: Is this related to the PR I see you opened with Umpire?
https://github.com/llnl/Umpire/pull/1070
Will we be able to remove this patch after the fix on the Umpire side?

victorapm: Hi Wayne, yes it's related. The plan is to remove it after the umpire update is merged

src/config/cmake/HYPRE_SetupSYCLToolkit.cmake
```diff
@@ -52,20 +52,54 @@ endif()
...
+    $<$<COMPILE_LANGUAGE:CXX>:${HYPRE_ONEDPL_COMPILE_OPTIONS}>)
+endif()
+
+get_target_property(HYPRE_ONEDPL_LINK_LIBS oneDPL INTERFACE_LINK_LIBRARIES)
+if(HYPRE_ONEDPL_LINK_LIBS)
+  foreach(HYPRE_ONEDPL_LINK_ITEM IN LISTS HYPRE_ONEDPL_LINK_LIBS)
+    target_link_libraries(${PROJECT_NAME} PUBLIC $<LINK_ONLY:${HYPRE_ONEDPL_LINK_ITEM}>)
+  endforeach()
+endif()
+
```
waynemitchell, line 102: What does all of this do/why is it needed?

victorapm: Simply put is to improve robustness of the build. 

The old CMake hard-coded a guessed oneDPL layout, while this new version uses the actual CMake target exported by `find_package`, which is more reliable (I had a few issues on another machine - not aurora).

The various `target` calls here are to propagate oneDPL's required compile definitions/options (the older approach never copied those, so if oneDPL says onsumers need particular defines or options, hypre would ignore them, which could lead to suble issues downstream depending on the installed oneDPL version)

waynemitchell (approved):
Looks good. Just had a couple questions mostly just for my own understanding. Thanks for getting this working!

victorapm:
Thanks, Wayne! I'll hold this off until the mixed precision and @rfalgout's struct PR are merged into master

PR #1515, opened 2026-04-24
Add GPU+Mixed precision support to CMake build

victorapm:
This PR makes the mixed-precision build work consistently across

1. plain CPU CMake
2. CUDA CMake and autotools
3. HIP CMake
4. SYCL CMake and autotools
5. CUDA/HIP CMake with umpire
6. CUDA/HIP autotools with umpire

The base branch `gpu-mixed-krylov` has the core mixed-precision pieces in place (and most of the work). This branch solves a few remaining issues:

1. some mixed-precision object libraries were built with the wrong config or wrong language
2. GPU `.c` sources that should have been compiled by CUDA/HIP were sometimes compiled as plain C
3. some backend-specific GPU mixed-precision wrappers, e.g., ending with Cusparse, were not being generated correctly
4. some public unsuffixed APIs still had no mixed-precision runtime dispatcher
5. Umpire code was precision independent but being compiled in places where MP would emit it more than once
7. the autotools MP archive was too broad and could pull in stale or wrong objects


src/seq_mv/mup.fixed_gpu
```diff
@@ -1,6 +1,9 @@
 hypreDevice_CSRSpAdd
 hypreDevice_CSRSpGemm
 hypreDevice_CSRSpGemmBinnedGetBlockNumDim
+hypreDevice_CSRSpGemmCusparse
+hypreDevice_CSRSpGemmCusparseGenericAPI|defined(HYPRE_USING_CUSPARSE) && CUSPARSE_VERSION >= CUSPARSE_NEWAPI_VERSION
```
rfalgout: I'm pretty sure we were trying to avoid dealing with include guards in the code generation for the original PR #1423 .  I think the idea was to move the guards into the functions so that the APIs were always present.  We had a discussion about this, if I recall.  Although this is implements the alternative approach, I think we need to discuss it again unfortunately.

src/blas/mup.functions
```diff
@@ -0,0 +1,2 @@
+hypre_ddot
```
rfalgout: It doesn't make sense to have `ddot` in multiple precisions.  Why did you need to do this?  There should be an alternative fix that will also work with externally linked blas (which would only expect double precision).

victorapm: Thanks, Rob, you're right. This probably slipped in when I was trying to fix the drivers build. Let me revert this

oseikuffuor1:
Thanks @victorapm , I will have to take a closer look and perhaps cherry-pick some of these into my branch. Were the linker issues you saw specific to CMake? The UMPIRE issue makes sense, since I did not have UMPIRE support fully covered yet. The rest, as @rfalgout noted, are issues that we may have discussed already and taken an alternate path. I will ping you with any questions soon.

src/seq_mv/mup.fixed_gpu
victorapm: I can revert this alternative approach, I don't have a strong argument for preferring it rather than the other approach. I will just mention here that we should probably rename `hypreDevice_CSRSpGemmCusparse` to `hypreDevice_CSRSpGemmVendor` so it becomes backend agnostic

rfalgout: I don't have a problem with renaming the function.  I would prefer to keep the mup.* files simple as before without the extra include guard annotations.  Thanks!

rfalgout:
> Thanks @victorapm , I will have to take a closer look and perhaps cherry-pick some of these into my branch.

Which branch are you talking about, out of curiosity?

src/blas/mup.functions
victorapm: Fixed in [df28450](https://github.com/hypre-space/hypre/pull/1515/commits/df2845016c64ef35f28dc55a5ea0365275ac7198)

src/blas/CMakeLists.txt
```diff
@@ -37,9 +37,15 @@ set(REGULAR_SRCS
   xerbla.c
 )
 
+set(MUP_SRCS
```
rfalgout, line 40: Should this also appear in the lapack CMakeLists.txt file?

oseikuffuor1:
> > Thanks @victorapm , I will have to take a closer look and perhaps cherry-pick some of these into my branch.
> 
> Which branch are you talking about, out of curiosity?

Oh I meant the gpu-mixed-krylov branch

src/seq_mv/mup.fixed_gpu
victorapm: Rob, this should be handled by https://github.com/hypre-space/hypre/pull/1515/commits/30ffcf76938cc9f569f917173aee14bfae80613c Thanks!

src/blas/CMakeLists.txt
victorapm: Hi Rob, yes, I think I did that in the first commit in this branch https://github.com/hypre-space/hypre/pull/1515/commits/e15569d92bf3716a043e76319a2158f8a327de5f

rfalgout:
> > > Thanks @victorapm , I will have to take a closer look and perhaps cherry-pick some of these into my branch.
> > 
> > 
> > Which branch are you talking about, out of curiosity?
> 
> Oh I meant the gpu-mixed-krylov branch

Okay, thanks!  In that case, I would say that it would be better not to cherry pick some of it, but try to get a clean merge directly into it if possible.

src/distributed_matrix/distributed_matrix_mp.c
```diff
@@ -0,0 +1,136 @@
+/******************************************************************************
```
rfalgout: What's going on with this file?  It looks like it implements some of the things our code generation is designed for.

src/IJ_mv/_hypre_IJ_mv.h
```diff
@@ -581,6 +581,9 @@ HYPRE_Int hypre_IJVectorUpdateValuesDevice( hypre_IJVector *vector, HYPRE_Int nu
 #ifndef hypre_MP_BUILD
 #include "_hypre_IJ_mv_mup_undef.h"
 #include "_hypre_IJ_mv_mup.h"
+#if defined(HYPRE_USING_GPU) && defined(__cplusplus)
```
rfalgout: These include guards are already inside the header file.  Do we need them here?

src/lib/Makefile
```diff
@@ -14,11 +14,11 @@ FEMLIFILES = ${HYPRE_FEI_FEMLI_FILES}
 # multiprecision build
 ifeq (${MP_BUILD}, 1) 
 
-IJMVFILES           = ${HYPRE_SRC_TOP_DIR}/IJ_mv/*.o
+IJMVFILES           = ${HYPRE_SRC_TOP_DIR}/IJ_mv/mup*.o
```
rfalgout, line 17: Can you explain why this change is needed?  Same question for the other changes.

src/utilities/device_utils.h
```diff
@@ -859,21 +859,97 @@ struct hypre_DeviceData
 
 hypre_DeviceData*     hypre_DeviceDataCreate();
 void                  hypre_DeviceDataDestroy(hypre_DeviceData* data);
+hypre_DeviceStream    hypre_DeviceDataStream(hypre_DeviceData *data, HYPRE_Int i);
```
rfalgout, line 862: Can you explain why these changes were needed?

rfalgout:
This looks pretty good now, @victorapm !  Just a few questions I'd like to understand before approving.  Let me know.  Thanks!

src/blas/CMakeLists.txt
rfalgout: Oh, I probably was looking at a specific commit when I made this comment, but thought I was looking at all of the code changes.  Thanks!

src/lib/Makefile
victorapm: I did this to improve robustness of the mixed precision build. The old "*.o" is too broad and could pick stale objects from a previous build. On the other hand `mup*.o` is more specific and should span all the objects we need (unless I'm missing something here, let me know!)

src/utilities/device_utils.h
victorapm: I moved some helper functions here to `static inline` so they don't emitted multiple times from the precision-specific objects. Let me know if I'm missing something!

src/distributed_matrix/distributed_matrix_mp.c
victorapm: Sorry Rob, I got confused with this one. We actually don't need this. Removed in [b0c1d03](https://github.com/hypre-space/hypre/pull/1515/commits/b0c1d0398cfb1a30c341111a393370587ed25b3d)

src/IJ_mv/_hypre_IJ_mv.h
victorapm: Thanks for catching! Removed in [b0c1d03](https://github.com/hypre-space/hypre/pull/1515/commits/b0c1d0398cfb1a30c341111a393370587ed25b3d)

src/lib/Makefile
rfalgout: Okay.  I don't quite understand this yet, but let me think a bit more about it.  Thanks!

src/utilities/device_utils.h
rfalgout: Okay, sounds good.  Thanks!

src/lib/Makefile
rfalgout: I understand now.  This shouldn't be a problem if there aren't old `.o` files hanging around from a non-mixed-precision build (I think), but I think it's a good idea.  Thanks again!

rfalgout (approved):
Thanks, @victorapm !

src/lib/Makefile
victorapm: Hi Rob, great, yes! This shouldn't be an issue if we are starting from a fresh build. Just to make things a little more robust...

rfalgout:
Hi @oseikuffuor1 .  Just wanted to keep this on your radar since it affects your PR #1423 which I would like to get merged before I merge my PR #1505 .  Let me know.  Thanks!

oseikuffuor1:
> Hi @oseikuffuor1 . Just wanted to keep this on your radar since it affects your PR #1423 which I would like to get merged before I merge my PR #1505 . Let me know. Thanks!

Yes I'm working on it. I hope to have it merged into #1423 soon.

src/IJ_mv/headers
```diff
@@ -64,6 +64,7 @@ cat >> $INTERNAL_HEADER <<@
 #ifndef hypre_MP_BUILD
 #include "_hypre_IJ_mv_mup_undef.h"
 #include "_hypre_IJ_mv_mup.h"
+#include "_hypre_IJ_mv_mup.hpp"
```
oseikuffuor1, line 67: @victorapm I'm a bit confused why these lines are being added again, since they were already in gpu-mixed-krylov branch?

src/utilities/device_utils.c
```diff
@@ -1907,14 +1907,14 @@ template<typename T>
...
-#if 0
-   HYPRE_THRUST_CALL( fill_n, d_x, n, v);
-#else
    if (n <= 0)
    {
       return hypre_error_flag;
    }
 
+#if defined(HYPRE_USING_HIP)
+   HYPRE_THRUST_CALL(fill_n, d_x, n, v);
```
oseikuffuor1, line 1916: Was this a bug before? Or is explicitly calling THRUST here more efficient for HIP?

src/seq_mv/csr_matrix_cuda_utils.c
oseikuffuor1, line 6: Is this file still needed?

src/IJ_mv/headers
victorapm: Thanks Daniel, I see the point. Do we really need to create a new `_hypre_IJ_mv.hpp` ? This is a one line change that gets away with it

src/utilities/device_utils.c
victorapm: This is a bit off-topic, sorry! It's not a bug, but something that I did for efficiency

src/seq_mv/csr_spgemm_device_rocsparse.c
```diff
@@ -56,8 +59,16 @@ hypreDevice_CSRSpGemmRocsparse(HYPRE_Int           m,
    /* RL: for matrices with long rows, it seemed that the sorting is still needed */
    /* VPM: Adding sorting back since it is necessary for correctness in a few cases */
 #if 1
-   hypre_SortCSRRocsparse(m, k, nnzA, descrA, d_ia, d_ja_sorted, d_a_sorted);
-   hypre_SortCSRRocsparse(k, n, nnzB, descrB, d_ib, d_jb_sorted, d_b_sorted);
+   hypre_CSRMatrixSortDevice(A);
```
oseikuffuor1: This call seems to do something slightly different. Do we need the actual matrix A sorted here as well? It doesn't seem like it was before.

src/seq_mv/csr_matrix_cuda_utils.c
victorapm: Good catch, removed in https://github.com/hypre-space/hypre/pull/1515/commits/724ce12293b7fbeaf13ad3194c3f20a657231666

src/IJ_mv/headers
oseikuffuor1: Unfortunately we have to do that to allow easy code generation with the script (without having to specialize it for each case). It is only true for folders with GPU functions. 

I was only concerned because it showed it had been added in this PR, when the parent branch should already have that. I was worried you may have started from an un-updated branch. But the line is correct, and I see you also run the headers script (which I forgot to do in the parent branch, so this is also good).

oseikuffuor1 (approved):
@victorapm this looks really good, thanks a lot. I left a few minor comments for your review. Otherwise, please merge into gpu-mixed-krylov and let me know if you approve the PR. I will do one final check once everything is merged then I can get it into master.

src/seq_mv/csr_spgemm_device_rocsparse.c
victorapm: Thanks for catching, Daniel. Can you take another look at this?

oseikuffuor1: @victorapm Thanks, this looks good to me. Only suggestion would be to use "use_sorted_storage" or something more descriptive for the flag. I am fine with what you have if you prefer to be more terse.

PR #1516, opened 2026-04-24
Update GPU .saved files

waynemitchell:
I had not update the .saved files for aurora or matrix after the recent addition/changes to the TEST_ij solvers jobs.

### DIFF
--- a/src/test/TEST_ij/solvers.saved.matrix
+++ b/src/test/TEST_ij/solvers.saved.matrix
@@ -32,7 +32,7 @@ Final Relative Residual Norm = 6.698760e-09
 Iterations = 11
 PCG_Iterations = 7
 DSCG_Iterations = 4
-Final Relative Residual Norm = 2.361078e-09
+Final Relative Residual Norm = 2.646554e-09
 
 # Output file: solvers.out.10
 Iterations = 10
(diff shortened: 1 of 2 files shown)


src/test/TEST_ij/solvers.saved.matrix
```diff
@@ -222,7 +222,7 @@ GMRES Iterations = 38
 Final GMRES Relative Residual Norm = 7.222662e-09
 
 # Output file: solvers.out.113
-GMRES Iterations = 10
+GMRES Iterations = 11
```
waynemitchell, line 225: This and the change on line 253 below both seem super weird to me? This is what I got when I ran the solvers jobs on matrix. Note that the final reported residual norm is EXACTLY the same, but the number of iterations is off by one both here and at line 253. Seems quite odd. Don't know why this would change and also why it would change in such a specific and odd way.

waynemitchell:
By the way, do we regularly run the `machine-matrix.sh` autotest? I don't see it as part of the autotest repo. I'm curious when the machine-matrix regressions were last checked (especially regarding the odd iteration count changes that I flagged above).

rfalgout:
> By the way, do we regularly run the `machine-matrix.sh` autotest? I don't see it as part of the autotest repo.

No, but I should add it to the mix.  I'll make a note.

src/test/TEST_ij/solvers.saved.matrix
victorapm: This is possibly my fault when I created that saved file... I may have written the iteration number manually wrongly. I wouldn't worry too much here

victorapm (approved):

rfalgout (approved):

victorapm:
BTW: same applies to `machine-tioga.sh` (regarding autotest)

PR #1517, opened 2026-04-27
Fix mixed precision SYCL build

victorapm:
Extends #1515 with SYCL support


rfalgout (approved):
Looks good to me.  Thanks!

PR #1526, opened 2026-04-30
disambiguates path to avoid unnecessary recompiles

andrewcorrigan:
I just added hypre to as a TPL to another code, and noticed that it kept rebuilding.  This fix disambiguates a path between the build and source folder to avoid unnecessary recompiles.  The AI agent offered the detailed explanation below.

Written with the assistance of AskSage:

if(NOT EXISTS "relative") resolves to cmake's working directory (the build dir) under CMP0132 OLD, while file(WRITE "relative") has always resolved to CMAKE_CURRENT_SOURCE_DIR. Since cmake_minimum_required(VERSION 3.21) locks CMP0132 to OLD, the existence check always fails, and the files get rewritten every configure, which updates their timestamps and triggers recompilation.

Using ${CMAKE_CURRENT_SOURCE_DIR}/ in the existence check is the minimal correct fix — it's consistent with what file(WRITE) already does, and doesn't depend on any policy version.


victorapm:
Andrew, thank you for proposing a fix. By coincidence, today I worked on a related topic and ended up applying the same kind of fix as you suggested on #1527, therefore this PR became obsolete. I'm happy to include you as co-author in the other PR for fair attribution. Please, let me know!

andrewcorrigan:
No need to include me as co author.   I'm glad it's resolved either way.

PR #1528, opened 2026-05-03
Fix SSAMG coarsened grid boundary handling

BenWibking:
## Summary

Fix two SSAMG issues triggered by coarsened semi-structured grids with pruned boxes.

- Initialize SSAMG `interp_type` to `-1`, matching the public API documentation that describes structured interpolation as the default.
- Make `hypre_SStructMatrixHaloToUMatrix` look up part-boundary box IDs in the current structured grid ID list instead of assuming every part-boundary ID is present and ordered contiguously. Boundary entries whose source boxes were pruned by coarsening are skipped.

## Root Cause

Bug 1: The SSAMG default interpolation type was documented as `-1`, but the solver data object initialized it to `0`, enabling the unstructured interpolation path by default.

Bug 2: In the structured interpolation path, RAP setup calls `hypre_SStructMatrixHaloToUMatrix`; after coarsening/pruning, the current `StructGrid` can contain IDs like `0,2,4,...` while the part-boundary metadata still contains IDs such as `1`. The old linear scan walked past the end of the grid boxes in that case.

## Validation

Validated in an AMReX SStruct SSAMG reproducer against this HYPRE build:

- `make -j8` in `hypre/src`
- `make -B -j8 main3d.gnu.DEBUG.MPI.ex`
- `./main3d.gnu.DEBUG.MPI.ex inputs-mlhypre-ssamg-works`
- `./main3d.gnu.DEBUG.MPI.ex inputs-mlhypre-ssamg-works max_grid_size=64`
- `./main3d.gnu.DEBUG.MPI.ex inputs-mlhypre-ssamg-works max_grid_size=8`

The original reproducer now completes with 27 SSAMG iterations and relative residual `5.649857648e-11`.

### DIFF
--- a/src/sstruct_ls/ssamg.c
+++ b/src/sstruct_ls/ssamg.c
@@ -30,7 +30,7 @@ hypre_SSAMGCreate( hypre_MPI_Comm comm )
    (ssamg_data -> zero_guess)       = 0;
    (ssamg_data -> max_levels)       = 0;
    (ssamg_data -> relax_type)       = 1;
-   (ssamg_data -> interp_type)      = 0;
+   (ssamg_data -> interp_type)      = -1;
    (ssamg_data -> skip_relax)       = 0;
    (ssamg_data -> usr_relax_weight) = 1.0;
    (ssamg_data -> usr_set_rweight)  = 0;
--- a/src/sstruct_mv/sstruct_matrix.c
+++ b/src/sstruct_mv/sstruct_matrix.c
@@ -2973,6 +2973,7 @@ hypre_SStructMatrixHaloToUMatrix( hypre_SStructMatrix   *A,
    hypre_BoxArray        *convert_boxa;
    hypre_BoxArray        *pbnd_boxa;
    hypre_BoxArray        *grid_boxes;
+   HYPRE_Int             *grid_box_ids;
    hypre_Box             *box;
    hypre_Box             *grow_box;
    hypre_Box             *grid_box;
@@ -2984,7 +2985,6 @@ hypre_SStructMatrixHaloToUMatrix( hypre_SStructMatrix   *A,
    HYPRE_Int              num_boxes;
    HYPRE_Int              pbnd_boxaa_size;
    HYPRE_Int              convert_boxaa_size;
-   HYPRE_Int              grid_box_id;
    HYPRE_Int              convert_box_id;
 
    HYPRE_ANNOTATE_FUNC_BEGIN;
@@ -3008,6 +3008,7 @@ hypre_SStructMatrixHaloToUMatrix( hypre_SStructMatrix   *A,
          sgrid = hypre_SStructPGridSGrid(pgrid, var);
          num_boxes  = hypre_StructGridNumBoxes(sgrid);
          grid_boxes = hypre_StructGridBoxes(sgrid);
+         grid_box_ids = hypre_BoxArrayIDs(grid_boxes);
 
          /* Exit this loop if there are no grid boxes */
          if (!num_boxes)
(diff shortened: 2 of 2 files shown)


victorapm:
Thanks for proposing a fix, Ben. Would you be able to share your input file`inputs-mlhypre-ssamg-works`?

BenWibking:
> Thanks for proposing a fix, Ben. Would you be able to share your input file`inputs-mlhypre-ssamg-works`? 

This is the input file:

```
max_level = 1
ref_ratio = 2
n_cell = 64
max_grid_size = 32

amrex.fpe_trap_invalid = 1
amrex.fpe_trap_overflow = 1
amrex.fpe_trap_zero = 1

use_mlhypre = 1
use_hypre_ssamg = 1
```

BenWibking:
Here is a self-contained reproducer that uses AMReX: 
[ssamg_reproducer.tar.gz](https://github.com/user-attachments/files/27326033/ssamg_reproducer.tar.gz)

BenWibking:
Codex claims to have found two additional bugs from debugging the code when interp_type is set to 0 (where it still segfaults, even with this patch):
```
  The concrete crash path is:

  1. hypre_SSAMGCreateInterpOp marks the center interpolation entry as constant for P:
     /Users/benwibking/nelim-fld/hypre/src/sstruct_ls/ssamg_interp.c:109 and /Users/benwibking/nelim-fld/hypre/src/
     sstruct_ls/ssamg_interp.c:140.
  2. ssamg_uinterp later calls:
     /Users/benwibking/nelim-fld/hypre/src/sstruct_ls/ssamg_uinterp.c:572

     hypre_SStructMatrixCompressUToS(P, 0);

  3. But CompressUToS explicitly says constant stencil entries are not handled:
     /Users/benwibking/nelim-fld/hypre/src/sstruct_mv/sstruct_matrix.c:2156

     > what if there are constant stencil entries?... avoid this case
  4. LLDB confirmed that on the bad level, before CompressUToS, P_s center constant is 1; immediately after
     CompressUToS, it is 0.

  That zeroed interpolation center then produces A_l[11] with a zero structured diagonal at part 0, box id 0. Jacobi
  then divides by that zero diagonal.

  There is also a separate indexing bug in the CF-marker construction in ssamg_uinterp: it assumes per-box odd/even
  marking over ghost-grown boxes matches the projected coarse grid, which fails when level 10 has 1x1 boxes and level
  11 drops whole boxes. That explains the earlier invalid-column / truncate crash.

  So the current diagnosis is:

  - Math: semi-coarsening plus unstructured boundary interpolation is plausible.
  - Implementation: broken for this mesh layout.
  - Main bugs: CF-marker/coarse-column mismatch, plus CompressUToS(P, 0) clobbering a constant P_s center coefficient.

  The likely fix is to make P entries that can be overwritten by U-to-S compression non-constant, especially the center
  entry, and fill the center coefficient over the actual structured boxes. Separately, the CF split should be derived
  from whether each fine row maps to an actual coarse-domain grid point, not from per-box odd/even marking.
```

rfalgout:
Hi @BenWibking .  Thanks for reporting this and offering solutions, too!  FYI, the interp_type 0 mode is experimental right now and not something you should try to use.  Even when it does work, it is extremely slow.  We are planning to revamp much of this soon, so some of the problems you are finding will go away because of a new implementation.

The interp_type = -1 issues are something we need to look at, so we will do that.  Thanks again!

BenWibking:
> Hi @BenWibking . Thanks for reporting this and offering solutions, too! FYI, the interp_type 0 mode is experimental right now and not something you should try to use. Even when it does work, it is extremely slow. We are planning to revamp much of this soon, so some of the problems you are finding will go away because of a new implementation.
> 
> The interp_type = -1 issues are something we need to look at, so we will do that. Thanks again!

Ok, thanks for the warning about that interp_type.

I went ahead and asked the tool to audit the broader SSAMG code for bugs and created issues for those. Hopefully they are useful. I'm afraid I don't understand the implementation well enough to be certain they are bugs, but they look plausible to me.

PR #1542, opened 2026-05-14
Build systems refactoring

victorapm:
This is a draft

Adds selective compilation for certain source files according to the build options passed to hypre.

For now:

1. `_device` files are compiled only for CUDA/HIP/SYCL builds

@rfalgout


src/lib/Makefile
```diff
@@ -5,6 +5,8 @@
 
 include ../config/Makefile.config
 
+wildcard_if_any = $(wildcard $(1))
```
rfalgout, line 10: What's this about?

victorapm: Rob, the wildcard is needed because `.obj` files are now optional. In this branch, when CUCC is empty, the subdirectory Makefiles no longer build the `.obj` files.

Without `$(wildcard ...)`, make keeps an unmatched pattern like `parcsr_mv/*.obj` literally. Then `src/lib/Makefile` tries to use `parcsr_mv/*.ob`j as an actual file when building libHYPRE, and the non-GPU build can fail because that file does not exist.

So the intent here is: include *`.obj` files if they exist; otherwise include nothing.

I also noticed we may need to apply the same treatment to the remaining optional mixed-precision `.obj` entries, such as `mup*.obj`, `utilities_mp_device.obj`, and `umpire.obj`, since those can also be not present when CUCC is empty.

src/sstruct_mv/Makefile
```diff
@@ -48,12 +48,13 @@ FILES =\
...
  sstruct_matop.c\
  sstruct_scale.c\
- sstruct_stencil.c
-
-CUFILES =\
+ sstruct_stencil.c\
  sstruct_matrix.c\
  sstruct_vector.c
 
+CUFILES =\
```
rfalgout: I am confused about the Makefile structure changes you are making, especially here.  If I am reading this correctly, `CUOBJS`, `CUOBJS_single`, etc. are always empty.  So, why do you need and `ifeq`-`else` for setting `OBJS`?

rfalgout: Related to this, why were the files `sstruct_matrix.c` and `sstruct_vector.c` listed in `CUFILES` to begin with?

victorapm: Hi Rob, I understand the confusion here. I think I can improve this, let me take a stab

PR #1543, opened 2026-05-16
compilation fixes for compiling HYPRE with CUDA support using Clang

andrewcorrigan:
This PR makes it possible to compile HYPRE with CUDA support enabled, using Clang as the CUDA compiler, rather than NVCC.

It fixes two errors:

1. The compiler flag `--extended-lambda` is NVCC-specific.  There's no equivalent flag with Clang, which has better support for lambdas, even without this flag.  I added a guard to only specify this flag with NVCC.
2. Clang's CUDA wrapper includes CUDA toolkit code that includes memory.h, which resolves to hypre's internal include instead of the standard memory.h.  This led to the error message shown below. Since this is just an internal include (not installed), I renamed it to hypre_memory.h. 

```
[  0%] Building CUDA object CMakeFiles/HYPRE.dir/struct_mv/struct_matvec.c.o
In file included from <built-in>:1:
In file included from /u/acorriga/install/llvm-libstdcxx-22.1.0/lib/clang/22/include/__clang_cuda_runtime_wrapper.h:493:
In file included from /usr/local/install/cuda-12.4/include/curand_mtgp32_kernel.h:108:
/u/acorriga/hypre/src/utilities/memory.h:104:15: error: unknown type name 'HYPRE_MAYBE_UNUSED_FUNC'
  104 | static inline HYPRE_MAYBE_UNUSED_FUNC hypre_MemoryLocation
```


victorapm:
@andrewcorrigan thanks for the contribution! Just curious, could you share your experience on using clang in the place of nvcc for CUDA builds? What is the minimum clang version that would be required? Thanks!

andrewcorrigan:
Regarding using clang instead of nvcc for CUDA builds, I'm a newcomer to HYPRE, so I can't share any experience with that yet.  In general, I've been using Clang instead of NVCC since it became available around 10 years ago, possibly before it was mainlined (when it was "gpucc").   I used to use NVCC for my code (circa 2010-2014), but compilation speed became unworkable: 2+ hour builds with nvcc vs. 2 minutes using clang/gcc for CPU builds.  Also, the lag in language feature support with nvcc was unbearable: C++11 support only became available with nvcc around 2014/2015.    It's gotten better with language support since then, but lambda support is still completely inadequate with NVCC, due to limitations such as,

```
error: The enclosing parent function for an extended __host__ __device__ lambda must not have deduced return type
error: __host__ __device__ extended lambdas cannot be generic lambdas
error: init-captures are not allowed for extended __host__ __device__ lambdas
```

With clang ,lambdas are supported without any CUDA-specific limitations and compilation speed is good too.

Regarding the clang version required, I wouldn't recommend going this far back, but according to https://llvm.org/docs/CompileCudaWithLLVM.html it was introduced in Clang 3.9.

At this point, given CMake's CUDA language support, it's completely seamless to switch between the two.

victorapm:
Merged in `cuda-clang` for internal testing, thanks @andrewcorrigan !

PR #1545, opened 2026-05-16
Extend CUDA builds with clang support

victorapm:
From #1543 

The main goal of this PR is to allow building hypre with CUDA support via clang, rather than the default `nvcc`.
In addition, secondary changes improve the CMake build robustness (there were warnings when using cmake >3.30)

It seems there's a slightly speedup in build times when using `clang` in the place of `nvcc`. See build times on 112 cores (2x Intel(R) Xeon(R) Platinum 8480+ CPUs) below:

## `nvcc` as CUDA compiler
```
$ CMD="cmake --fresh -DHYPRE_BUILD_UMPIRE=ON -DHYPRE_ENABLE_CUDA=ON -DCMAKE_CUDA_ARCHITECTURES=90 -B build -S src -G Ninja && cmake --build build -j 112"; time $(eval "${CMD}")

real	1m0.237s
user	28m16.646s
sys	    2m41.664s
```

## `clang-19` as CUDA compiler
```
$ CMD="cmake --fresh -DCMAKE_CUDA_COMPILER=clang++ -DHYPRE_BUILD_UMPIRE=ON -DHYPRE_ENABLE_CUDA=ON -DCMAKE_CUDA_ARCHITECTURES=90 -B build -S src -G Ninja && cmake --build build -j 112"; time $(eval "${CMD}")

real	0m46.590s
user	16m2.668s
sys	    1m8.102s
```

cc @andrewcorrigan


src/utilities/headers
```diff
@@ -47,7 +47,7 @@ cat printf.h                   >> $INTERNAL_HEADER
 cat error.h                    >> $INTERNAL_HEADER
 cat mpistubs.h                 >> $INTERNAL_HEADER
 cat smp.h                      >> $INTERNAL_HEADER
-cat memory.h                   >> $INTERNAL_HEADER
+cat hypre_memory.h             >> $INTERNAL_HEADER
```
rfalgout: Why is this rename needed?  Is there a conflict with a standard header file or something?

victorapm: Hi Rob, yes. According to @andrewcorrigan in his original PR:

Clang's CUDA wrapper includes CUDA toolkit code that includes memory.h, which resolves to hypre's internal include instead of the standard memory.h. This led to the error message shown below. Since this is just an internal include (not installed), I renamed it to hypre_memory.h.

```
[  0%] Building CUDA object CMakeFiles/HYPRE.dir/struct_mv/struct_matvec.c.o
In file included from <built-in>:1:
In file included from /u/acorriga/install/llvm-libstdcxx-22.1.0/lib/clang/22/include/__clang_cuda_runtime_wrapper.h:493:
In file included from /usr/local/install/cuda-12.4/include/curand_mtgp32_kernel.h:108:
/u/acorriga/hypre/src/utilities/memory.h:104:15: error: unknown type name 'HYPRE_MAYBE_UNUSED_FUNC'
  104 | static inline HYPRE_MAYBE_UNUSED_FUNC hypre_MemoryLocation
```

rfalgout: Okay, but I would prefer a different renaming.  All of the memory related files in `utilities` are of the form `memory*` and there are no other files that start with `hypre`.  I though about a few alternative ways to deal with this like renaming all `memory*` files to `mem*` or `memutils*` or `memtools*`, but `mem.h` is also a reserved header for some compilers (apparently) and I don't particularly like the other two.  So, I think your approach is okay, but I would use the name `memory_header.h` instead and add a comment just before the `cat` line in `headers` about the fact that this name was chosen carefully to avoid conflicts with reserved header file names.  We don't want to reintroduce this problem in the future.

rfalgout (approved):
Looks good to me modulo my comments.  I did not try to understand the cmake logic, nor did I take it for a test spin.

src/utilities/headers
victorapm: Thanks, Rob! Fixed in [a9d0e47](https://github.com/hypre-space/hypre/pull/1545/commits/a9d0e478e3140705034c2e48e488568edc89c95d)

PR #1546, opened 2026-05-18
Update tioga tests

victorapm:
- Update baselines used in the struct benchmarks. Setup is faster since Rob's improvements
- Fix mixed precision issues. Some helpers were not being properly compiled for the different precisions, which would cause memory-related crashes in the GPU + mixed precision tests


src/utilities/_hypre_utilities.hpp
```diff
@@ -2622,6 +2622,19 @@ hypre_HYPREIntToCusparseIndexType()
 #if defined(HYPRE_USING_CUDA) || defined(HYPRE_USING_HIP)
 #if !defined(HYPRE_USING_RAJA) && !defined(HYPRE_USING_KOKKOS)
 
+/* These helpers depend on HYPRE_Real's size, so keep their C++/device symbols
```
rfalgout: I don't understand why this is needed, when other struct types do not require these defines (e.g., hypre_PFMGData).  What's special here?  Are these used in a mixed precision context such as the struct for the mixed precision BoomerAMG solver?

victorapm: H Rob, I think what is special here is that these are C++/GPU header-defined helpers, not ordinary C solver data structs. The structs `HYPRE_Real2/4/6` have layouts that depend on `HYPRE_Real`, and their names participate in C++/device symbol mangling through constructors/operators, `ReduceSum<T>`, and `OneBlockReduceKernel<T>`.

In a mixed-precision GPU build, this header is compiled in both the single and double precision object builds. Without suffixing, both builds can give symbols for things like `HYPRE_Real2::operator+=` or `OneBlockReduceKernel<HYPRE_Real2>` with the same C++ type name, even though the layout is `float,float` in one build and `double,double` in the other. This would lead to memory issues

I think `hypre_PFMGData` is different in the sense that it is a C data struct/typedef and does not itself produce C++/device symbols. The functions operating on solver data are what get renamed by the existing mixed-precision machinery. So, we are good there.

That's my understanding of the mixed precision build... There might be a better fix to the one I'm proposing...

rfalgout: I assume this was intended to fix the 'cmake-hip-mup-shared' errors that we got?  If so, why do we not get the same errors for 'basic-hip-mup'?  I think I need a discussion on this one as well.

victorapm: That's correct, those changes were intended to fix `cmake-hip-mup-shared`. That was puzzling to me as well and I did some research on this.... My understanding is that it has to do with static vs shared build. But the reason is not fully clear to me to be honest...

victorapm: @rfalgout @oseikuffuor1 , should we revisit this?

victorapm:
@oseikuffuor1 @rfalgout I reverted the mup changes in this branch (will be addressed in Daniel's branch), so it's only updating the baselines

rfalgout (approved):
Thanks, @victorapm !

PR #1547, opened 2026-05-19
#include <thrust/tuple.h>

andrewcorrigan:
`#include <thrust/tuple.h>`

before, this was included transitively from other thrust headers, in more recent versions it must be included explicitly

### DIFF
--- a/src/utilities/_hypre_utilities.hpp
+++ b/src/utilities/_hypre_utilities.hpp
@@ -454,6 +454,7 @@ using hypre_DeviceItem = void*;
 #include <thrust/remove.h>
 #include <thrust/version.h>
 #include <thrust/pair.h>
+#include <thrust/tuple.h>
 
 /* VPM: this is needed to support cuda 10. not_fn is the correct replacement going forward. */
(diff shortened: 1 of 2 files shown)


victorapm (approved):
Thanks, @andrewcorrigan! Out of curiosity, which CUDA version surfaced this issue to you?

andrewcorrigan:
It wasn't the CUDA version that triggered the errors since I'm still stuck on an old CUDA version.  Unfortunately, I'm not sure when this change was introduced to Thrust.

My code bundles an up-to-date version of CCCL/Thrust, which is currently a month or two behind their main branch.

As soon as I pointed HYPRE to that copy instead of whatever is bundled with CUDA 12.4, that triggered the missing include errors.

rfalgout (approved):

nmnobre:
> Out of curiosity, which CUDA version surfaced this issue to you?

This change is included in CCCL 3.3.0 and above, which affects CUDA 13.3 which bundles CCCL 3.3.3.
Thanks @andrewcorrigan. :)

PR #1552, opened 2026-05-21
Prepare ParCSR overlap extraction for Schwarz

victorapm:
This PR improves our current overlap extraction capabilities, more specifically:

  - Builds the extended row set for overlap order delta.
  - Creates the communication package for fetching external overlap rows.
  - Fetches external row data collectively across MPI ranks.
  - Extracts a local CSR matrix over the extended subdomain.
  - Adds host-memory guards so GPU/device matrices fail explicitly.
  - Fixes collective behavior so ranks with no local external rows still participate when other ranks need communication.
  - Replaces repeated linear/binary searches with unordered maps in key places.

### DIFF
--- a/src/parcsr_mv/_hypre_parcsr_mv.h
+++ b/src/parcsr_mv/_hypre_parcsr_mv.h
@@ -567,6 +567,7 @@ typedef struct hypre_OverlapData_struct
    HYPRE_Int            num_extended_rows;     /* Total rows in extended domain */
    HYPRE_Int            num_overlap_rows;      /* External rows (from overlap) */
    HYPRE_BigInt        *extended_row_indices;  /* Global indices of all extended rows */
+
    HYPRE_Int           *row_is_owned;          /* 1 if row is owned, 0 if external */
 
    /* Communication package for fetching overlap data */
--- a/src/parcsr_mv/par_csr_overlap.c
+++ b/src/parcsr_mv/par_csr_overlap.c
@@ -541,7 +541,7 @@ hypre_ParCSRMatrixComputeOverlap(hypre_ParCSRMatrix  *A,
       }
 
       hypre_MPI_Allreduce(&ext_count, &global_ext_count, 1, HYPRE_MPI_INT,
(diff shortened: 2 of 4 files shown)


rfalgout:
Hi @victorapm .  Can you explain in words how the function `hypre_ParCSRMatrixComputeOverlap` has changed?  What was the basic logic before versus now?  Thanks!

victorapm:
Sure, Rob!

Before `hypre_ParCSRMatrixComputeOverlap` did this:
1. Start the overlap set with locally owned rows.
2. For each overlap level: Scan local rows already in the set. Then,  add off-processor column neighbors as potential new overlap rows. Then, for levels greater than 0, fetch external row sparsity and add neighbors of those external rows. Then, sort/deduplicate the new rows. Lastly, union them into the current extended row set.
3. Build row_is_owned.
4. If this rank had external rows, build an overlap communication package. Duplicate checks were done with binary search against current_set, then a linear scan through the temporary potential_external list.

Now, the basic overlap expansion is still the same: start from local rows, grow by graph distance, fetch external row patterns for higher overlap levels, then produce the final extended row set. The important changes are:
1. It now rejects device-memory matrices early, because this overlap construction path requires host data.
2. It makes sure the matrix has an assumed partition before building extension communication packages.
3. It uses an unordered map to track potential_external, replacing repeated linear scans for duplicate detection.
4. Communication is now guarded by global collective checks making sure we don't hit any deadlocks. This fixes a main risk in the old version: some ranks could skip collective communication while others entered it, especially when only some ranks had overlap rows. The new version makes participation consistent across MPI ranks.

So, in short: the mathematical idea did not really change. The function still computes a graph-distance overlap around local rows. The change is mostly that the implementation is now safer for MPI collectives, handles ranks with zero local requests correctly, creates required partition metadata for the next PR on Schwarz and uses a faster method for tracking duplicates.

rfalgout:
Thanks for the explanation, @victorapm !

rfalgout (approved):
Looks good to me.  Thanks!

PR #1553, opened 2026-05-21
Add internal overlapping Schwarz implementation

victorapm:
Depends on #1552 

This PR adds the internal solver implementation for the new overlapping Schwarz variants.

It builds on the ParCSR overlap extraction utilities from the earlier work: setup uses those utilities to construct each rank's
extended subdomain matrix, then configures a local solver on that subdomain.

A detailed list of changes is given below:
- Adds RAS and classical additive Schwarz formulations for the overlapping implementation.
- Supports ILU(k), ILUT, AMG, and SuperLU_dist as local subdomain solvers.
- Extends `hypre_SchwarzData` append-only so the existing ABI prefix is preserved.
- Implements setup, solve, destroy, local solver ownership, work vectors, residual/statistics bookkeeping, and setter validation.
- Adds host-memory checks since the current implementation is not ported to GPUs yet
- Keeps the existing (legacy) domain-based Schwarz variants on their original code path.

On-going work that will be added to future PRs:
- Benchmarking and performance results
- Unit tests for new Schwarz capabilities
- IJ regression tests


src/parcsr_ls/par_schwarz.c
```diff
@@ -7,12 +7,1182 @@
...
+#include "dsuperlu.h"
+
+/*--------------------------------------------------------------------------
+ * Wrapper for hypre_SLUDistSolve to match HYPRE_PtrToSolverFcn signature.
+ * Keep a local wrapper so the Schwarz code can store the same function
+ * pointer type for all local solvers.
+ *--------------------------------------------------------------------------*/
+
+static HYPRE_Int
+hypre_SchwarzSLUDistSolveWrapper(void               *solver,
```
rfalgout: This has the same signature as `hypre_SLUDistSolve` so why is it needed?

victorapm: I'll fix this. It's from a few months ago when I started writing it when the SLUDist API look differently. We shouldn't need this wrapper anymore. Thanks!

victorapm: Addressed in [d009395](https://github.com/hypre-space/hypre/pull/1553/commits/d009395441f8b50b18043bf5758eacc5d5ed2b1c)

rfalgout (approved):
Looks good.  Thanks!

PR #1564, opened 2026-06-13
Improve MGR performance on GPUs

victorapm:
This PR adds a centralized `hypre_BlockDiagInvDevice` in MGR for computing the full inverse of block diagonal matrices on GPUs. This operation is needed in a few MGR options.

### DIFF
--- a/src/parcsr_ls/ads.c
+++ b/src/parcsr_ls/ads.c
@@ -592,18 +592,11 @@ HYPRE_Int hypre_ADSComputePi(hypre_ParCSRMatrix *A,
          HYPRE_BigInt global_num_rows = hypre_ParCSRMatrixGlobalNumRows(F2V);
          HYPRE_BigInt global_num_cols = 3 * hypre_ParCSRMatrixGlobalNumCols(F2V);
          HYPRE_BigInt *row_starts = hypre_ParCSRMatrixRowStarts(F2V);
-         HYPRE_BigInt *col_starts;
-         HYPRE_Int col_starts_size;
          HYPRE_Int num_cols_offd = 3 * hypre_CSRMatrixNumCols(hypre_ParCSRMatrixOffd(F2V));
          HYPRE_Int num_nonzeros_diag = 3 * hypre_CSRMatrixNumNonzeros(hypre_ParCSRMatrixDiag(F2V));
          HYPRE_Int num_nonzeros_offd = 3 * hypre_CSRMatrixNumNonzeros(hypre_ParCSRMatrixOffd(F2V));
          HYPRE_BigInt *col_starts_F2V = hypre_ParCSRMatrixColStarts(F2V);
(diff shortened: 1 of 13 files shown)


rfalgout (approved):
Looks good.  I assume you've got some mup code to write as well?

victorapm:
Thanks, Rob! Thanks for reminding. This wasn't flagged by `./mup_code`, but as we know already `./mup_code` is not checking the GPU build (there's another PR for that...). Let me add the missing `mup` prototypes here

victorapm:
Rob, I added `mup` code for the new prototype added in this PR, however `./mup_check` misinterpreted the new `mup` code and it flagged an issue (see https://github.com/hypre-space/hypre/actions/runs/27485128010/job/81239836436). It looks like we will need to add GPU support to `mup_check` (PR #1524) so it will understand new prototypes appearing only in device builds

oseikuffuor1:
> Rob, I added `mup` code for the new prototype added in this PR, however `./mup_check` misinterpreted the new `mup` code and it flagged an issue (see https://github.com/hypre-space/hypre/actions/runs/27485128010/job/81239836436). It looks like we will need to add GPU support to `mup_check` (PR #1524) so it will understand new prototypes appearing only in device builds

@victorapm there's actually some partial support for mup_check for the GPU case. I will take a look at PR#1524, but you can trigger this by adding "-gpu" tag to the call to "mup_check_dir.sh" in mup_check.sh. This will catch the GPU checks.

victorapm:
Thanks, Daniel! I'm using that on #1524  (via the check_type variable). I just updated that branch wrt master

src/parcsr_ls/ads.c
```diff
@@ -592,18 +592,11 @@ HYPRE_Int hypre_ADSComputePi(hypre_ParCSRMatrix *A,
...
          HYPRE_Int num_nonzeros_diag = 3 * hypre_CSRMatrixNumNonzeros(hypre_ParCSRMatrixDiag(F2V));
          HYPRE_Int num_nonzeros_offd = 3 * hypre_CSRMatrixNumNonzeros(hypre_ParCSRMatrixOffd(F2V));
          HYPRE_BigInt *col_starts_F2V = hypre_ParCSRMatrixColStarts(F2V);
-         col_starts_size = 2;
-         col_starts = hypre_TAlloc(HYPRE_BigInt, col_starts_size, HYPRE_MEMORY_HOST);
-         for (i = 0; i < col_starts_size; i++)
-         {
-            col_starts[i] = 3 * col_starts_F2V[i];
-         }
+         HYPRE_BigInt col_starts[2] = {3 * col_starts_F2V[0], 3 * col_starts_F2V[1]};
```
victorapm, line 599: cc @tzanio We had a memory leak on ADS. This PR fixes the issue

PR #1568, opened 2026-06-27
Add support to user-provided Schur complement in MGR

victorapm:
This PR adds `HYPRE_MGRSetCoarseGridMatrixAtLevel`: an optional API letting an application supply its own coarse-grid operator for a given MGR reduction level, activated by setting that level's `coarse_grid_method` to 6.

This allow application codes to provide a spectrally-equivalent, cheaper coarse approximation to a Schur complement (e.g., scaled pressure-mass Schur complement for saddle-point systems). The supplied matrix is used as-is and it is borrowed, not owned by MGR.


rfalgout (approved):
Looks good.  Thanks!

src/parcsr_ls/par_mgr.c
```diff
@@ -46,6 +46,7 @@ hypre_MGRCreate(void)
    (mgr_data -> R_array) = NULL;
    (mgr_data -> RT_array) = NULL;
    (mgr_data -> RAP) = NULL;
+   (mgr_data -> user_coarse_grid_matrix) = NULL;
```
oseikuffuor1: Should the name be user_coarse_grid_matrix_array to be consistent with the rest of the array of operators?

victorapm: Good point, I can change it

victorapm: Done https://github.com/hypre-space/hypre/pull/1568/commits/d2c70e56ed6e48e122c9e0b53f93f0c0a932fcbf

src/parcsr_ls/par_mgr.c
```diff
@@ -327,7 +328,39 @@ hypre_MGRCleanupBuildData( void      *mgr_vdata,
...
 {
    hypre_ParMGRData *mgr_data = (hypre_ParMGRData*) mgr_vdata;
-   HYPRE_Int i;
+   HYPRE_Int i, lvl;
+
+   /* Detach application-provided coarse-grid (Schur) matrices from the hierarchy
+      before cleaning it up. They are borrowed -- the application owns and frees them --
+      but they are aliased into A_array / RAP. Clearing those slots lets the generic
+      destroy loops below skip them through their existing NULL checks, so no
+      per-matrix ownership test is needed. */
```
oseikuffuor1: Could this code-block for detaching the user data be simplified?
* Do we need the outer loop over max_num_coarse_levels? It seems to me the inner block if (mgr_data->A_rray) is what is necessary since outer loop lvl and inner loop i should be the same within num_coarse_levels. So the comparison check for user_mat in the inner loop should suffice. 
* The RAP check is repeated over max_num_coarse_levels, but I suppose we know what the RAP (coarsest) level is at this time, right?
* General question: I suppose we don't expect this user_coarse_grid capability to be set on level 0, right? Should we check for that in the hypre_MGRSetCoarseGridMatrixAtLevel function and skip i=0 here?

src/parcsr_ls/par_mgr.c
```diff
@@ -3056,6 +3092,46 @@ hypre_MGRSetCoarseGridMethod( void      *mgr_vdata,
...
+ * for a given MGR reduction level, used in place of the Galerkin RAP product when
+ * that level's coarse_grid_method is 6. MGR uses the matrix in place: it never
+ * frees the matrix and never alters its entries. If the matrix has no
+ * communication package, MGR creates one during setup (reclaimed when the
+ * application destroys the matrix). Its row/column partitioning must match the
+ * compressed coarse (C-point) space that hypre_MGRCoarseParms produces for that
+ * level.
+ *--------------------------------------------------------------------------*/
+
+HYPRE_Int
```
oseikuffuor1, line 3103: Do we allow level == 0 here? See previous comment

oseikuffuor1:
@victorapm I left a few comments for your review. I suppose you are yet to run the mup_code script right? I don't see the mup codes in the PR yet.

src/parcsr_ls/par_mgr.c
victorapm: Thanks for the questions, Daniel!

1. The outer loop is over reduction levels in user_Ac_array, while the matching entry in the hierarchy is `A_array[level + 1]` (set in `hypre_MGRBuildCoarseOperator`). So `lvl` and `i` are not the same; a direct `lvl == i` check would miss the alias. The nested pointer comparison is a bit defensive, but it avoids relying on that off-by-one in cleanup. I can simplify it to clear `A_array[lvl + 1]` when it equals `user_Ac_array[lvl]` if you prefer.
2. Agreed on RAP: it is only set for the coarsest reduction (`level + 1 == num_coarse_levels`), so the RAP nulling can be a single check rather than repeated inside the loop. I’ll clean that up.
3. Level 0 is intentionally allowed and is the main use case. `SetCoarseGridMatrixAtLevel(..., 0, S)` with `coarse_grid_method[0] = 6` supplies the operator after the first reduction (stored as A_array[1]); it does not replace the fine matrix A_array[0]. Rejecting level 0 would break the typical one-level Schur-complement workflow.

src/parcsr_ls/par_mgr.c
victorapm: Level 0 is intentionally allowed and is the main use case. SetCoarseGridMatrixAtLevel(..., 0, S) with coarse_grid_method[0] = 6 supplies the operator after the first reduction (stored as A_array[1]); it does not replace the fine matrix A_array[0]. Rejecting level 0 would break the typical one-level Schur-complement workflow.

victorapm:
Thanks Daniel! Updated mup code on https://github.com/hypre-space/hypre/pull/1568/commits/a089362870ad3cafc213233fa89e382d0eba6fa9. I generally leave it as the last item in the PRs :)

src/parcsr_ls/par_mgr.c
oseikuffuor1: Yes, this makes sense. Sorry for the confusion.

src/parcsr_ls/par_mgr.c
oseikuffuor1: Ok thanks. For 1. I was mostly concerned about performance, since it is looping over max_num_coarse_levels and checking num_coarse_level times internally. Maybe "break" after you match user_mat in the inner loop, or simplify to a single loop like you mentioned above. lvl = i+1 is correct. 

3. makes sense. Thanks!

victorapm: Thanks Daniel! Agreed on simplifying that block for clarity (I don't think though it's a performance issue because `max_num_coarse_levels` is small). 

Done in https://github.com/hypre-space/hypre/pull/1568/commits/d8bccd60dd94cf5d1e63eeb6a986423d72bf7337

oseikuffuor1 (approved):
Thanks Victor!

PR #1577, opened 2026-07-06
Memory bug fix for SStructGraphAssemble

rfalgout:
This PR fixes a memory bug in `HYPRE_SStructGraphAssemble` in cases where `GraphAddEntries` sets graph entries within the same part.

The PR also adds a regression test to `TEST_sstruct/periodic.jobs` to check this scenario.


rfalgout:
Hi @waynemitchell .  Use "hide whitespace" when looking at the code.  I basically just added an outer if block around the problematic code.  Also, note that the regression does actually fail (on the tux box you need to turn on valgrind to see the issue) with the old code.  Let me know what you think, then we can point Arjun to the branch.

waynemitchell (approved):
This seems fine for now as a quick fix. Thanks, Rob!

PR #1579, opened 2026-07-13
Fix HMIS issue with device execution policy

### LINKED ISSUE #1573: CUDA illegal memory access in BoomerAMG with HMIS + ext+i on H100 (intermittent, via PETSc PCHYPRE)
We are seeing an intermittent CUDA crash in HYPRE BoomerAMG when using HMIS coarsening with ext+i interpolation on NVIDIA H100. The run is launched through PETSc PCHYPRE, but the failure appears inside HYPRE.

Actual behavior:
- Solver intermittently fails with an illegal memory access in hypreDevice_extendWtoP.
- The same case can run successfully in other attempts, so it is not a deterministic immediate failure.
- PMIS is stable on the same setup, while HMIS shows this instability.
- Unified Memory variants are sometimes slower and sometimes faster than non-UM runs.

Expected behavior:
- We expect HMIS + ext+i to be stable on GPU for repeated runs of the same case.

Environment:
- GPU: NVIDIA H100
- CUDA: 12.6.85 (CUDA Toolkit 12.6, nvcc reports release 12.6, V12.6.85)
- MPI: NVIDIA HPC-X Open MPI v4.1.7rc1 (GPU-aware enabled in PETSc configure)
- HYPRE: hypre 3.0.0
- PETSc: 3.24.4
- Application: Fortran CFD code, pressure solve through PETSc + BoomerAMG

PETSc/HYPRE options used:
- ksp_type preonly
- pc_type hypre
- pc_hypre_type boomeramg
- pc_hypre_boomeramg_coarsen_type hmis
- pc_hypre_boomeramg_interp_type ext+i

Questions:
- Is this a known issue for HMIS + ext+i on GPU?
- Are there recommended safer settings for H100 (for example interpolation/coarsening combinations)?
- If helpful, should we test a specific branch/patch to narrow this down?

victorapm:
Closes #1573

This PR fixes a device HMIS coarsening bug that could leave some CF markers undecided and later cause interpolation setup to abort.

### DIFF
--- a/src/parcsr_ls/par_coarsen_device.c
+++ b/src/parcsr_ls/par_coarsen_device.c
@@ -371,14 +371,6 @@ hypreGPUKernel_PMISCoarseningInit(hypre_DeviceItem &item,
    if (CF_marker_i == 0 && measure_diag[i] < 1.0)
    {
       CF_marker_i = F_PT;
-   }
-
-   /* Keep points selected by the initial HMIS pass out of subsequent
-    * independent sets.  These points are not included in graph_diag, so a
-    * nonzero measure would allow a neighboring graph point to reset their
-    * marker to zero without adding them back to the graph. */
-   if (CF_marker_i != 0)
-   {
       measure_diag[i] = 0.0;
    }
 
--- a/src/parcsr_ls/par_coarsen_device.c
+++ b/src/parcsr_ls/par_coarsen_device.c
@@ -364,6 +364,14 @@ hypreGPUKernel_PMISCoarseningInit(hypre_DeviceItem &item,
    if (CF_marker_i == 0 && measure_diag[i] < 1.0)
    {
       CF_marker_i = F_PT;
+   }
+
+   /* Keep points selected by the initial HMIS pass out of subsequent
+    * independent sets.  These points are not included in graph_diag, so a
+    * nonzero measure would allow a neighboring graph point to reset their
+    * marker to zero without adding them back to the graph. */
+   if (CF_marker_i != 0)
+   {
       measure_diag[i] = 0.0;
    }
 
--- a/src/parcsr_ls/par_coarsen_device.c
+++ b/src/parcsr_ls/par_coarsen_device.c
@@ -343,13 +343,6 @@ hypreGPUKernel_PMISCoarseningInit(hypre_DeviceItem &item,
       {
          measure_diag[i] = 0.0;
       }
-
-      /* Retained HMIS C-points are excluded from graph_diag.  Clear their
-       * measures so subsequent independent sets cannot reset their markers. */
-      if (CF_marker_i > 0)
-      {
-         measure_diag[i] = 0.0;
-      }
    }
    else
    {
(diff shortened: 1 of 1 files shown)


src/parcsr_ls/par_coarsen_device.c
```diff
@@ -364,6 +364,14 @@ hypreGPUKernel_PMISCoarseningInit(hypre_DeviceItem &item,
    if (CF_marker_i == 0 && measure_diag[i] < 1.0)
    {
       CF_marker_i = F_PT;
+   }
+
+   /* Keep points selected by the initial HMIS pass out of subsequent
+    * independent sets.  These points are not included in graph_diag, so a
+    * nonzero measure would allow a neighboring graph point to reset their
+    * marker to zero without adding them back to the graph. */
+   if (CF_marker_i != 0)
```
rfalgout: I am not confident about this change.  In the logic above, when CF_marker_i = SF_PT (special F-pt), the measure_array[i] flag is not set to zero.  This is also what happens in the CPU code.  With this change, it would be set to zero.

victorapm: Thanks for catching, Rob, that makes sense to me. I fixed the conditional so it only applies to retained HMIS C-points

rfalgout: I still don't get what's going on here.  The bug is for HMIS, though the code change is in a PMIS function.  So, presumably this is an issue after HMIS has already selected some coarse points and PMIS is filling in points along process boundaries.  But why is 'measure_diag' so important in this scenario?  The CF_marker is supposed to be the array that determines when the coarsening algorithm is done.  The new code sets measure_diag to zero if CF_marker is set, but why isn't the latter enough to begin with?

Sorry for all the questions, but it feels like the bug is really somewhere else in the code, in a place where measure_diag is being checked to determine if a point has been marked instead of checking CF_marker.  But this is confusing code (always has been), so maybe I'm just missing something.  Thanks!

victorapm: Hi Rob, no worries. The fix is in the PMIS implementation because HMIS performs an initial Ruge coarsening and then invokes PMIS with `CF_init = 1`, passing the existing C/F markers into it:
```
  hypre_BoomerAMGCoarsenRuge(..., &CF_marker);
  hypre_BoomerAMGCoarsenPMIS(S, A, 1, debug_flag, CF_marker_ptr);
```

I agree the proposed fix here (involving `measure_diag`) doesn't feel very natural. I'll revisit this later and try to come up with a better fix. @liruipeng feel free to chime in if you have ideas. Thanks!

rfalgout: Quick question.  Where in the original code is the nonzero measure_diag causing an error?  This is supposed to just be a measure of how good a point is as a potential C-point choice.  But if the point has already been decided (C-pt or F-pt), then measure_diag has no meaning.  Since the only thing this fix does is change measure_diag, that implies it is being used incorrectly somewhere.

victorapm: Hi Rob, here's the issue: https://github.com/hypre-space/hypre/blob/01b4be15b1f3a3bbd3a642efff7e02c94510c934/src/parcsr_ls/par_indepset_device.c#L59-L74

According to the logic above: If point J has nonzero measure, it might get reclassified under HMIS. I'm evaluating if the fix can be implemented somewhere in that code block

victorapm: @rfalgout I pushed a new fix involving `graph_diag` mainly. In `hypre_PMISCoarseningInitDevice`, the initial `graph_diag` now uses a filter depending on `CF_init`. When it's one, it keeps all points with `CF_marker >= 0`, so retained C-points stay in the graph like in the host code and get their measures cleared by `hypre_PMISCoarseningUpdateCFDevice`

src/parcsr_ls/par_coarsen_device.c
```diff
@@ -428,21 +428,47 @@ hypre_PMISCoarseningInitDevice( hypre_ParCSRMatrix  *S,               /* in */
...
+      new_end = HYPRE_THRUST_CALL( remove_copy_if,
+                                   thrust::make_counting_iterator(0),
+                                   thrust::make_counting_iterator(num_rows_diag),
+                                   CF_marker_diag,
+                                   graph_diag,
+                                   is_negative<HYPRE_Int>());
+   }
+   else
+   {
+      new_end = HYPRE_THRUST_CALL( remove_copy_if,
```
rfalgout, line 465: Am I understanding correctly that this only copies values with CF_marker_diag = 0?  If so, do you know which coarsening algorithm needs to overwrite previously-selected C-points?  I guess I am wondering if the if/else is really needed (or can we just use your new remove_copy_if), and I'm also trying to understand what's going on.  Thanks!

PR #1581, opened 2026-07-15
Address code quality issues

victorapm:
Address issues pointed out by GitHub's CodeQL

### DIFF
--- a/src/examples/ex4.c
+++ b/src/examples/ex4.c
@@ -516,7 +516,7 @@ int main (int argc, char *argv[])
                                                       to the offsets */
          double *values;
 
-         values = (double*) calloc(5 * (size_t) n * (size_t) n, sizeof(double));
+         values = (double*) calloc(5 * (n * n), sizeof(double));
 
          /* The order is left-to-right, bottom-to-top */
          for (k = 0, j = 0; j < n; j++)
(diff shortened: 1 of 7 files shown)


src/examples/ex4.c
```diff
@@ -516,7 +516,7 @@ int main (int argc, char *argv[])
                                                       to the offsets */
          double *values;
 
-         values = (double*) calloc(5 * (n * n), sizeof(double));
+         values = (double*) calloc(5 * (size_t) n * (size_t) n, sizeof(double));
```
rfalgout: I would prefer to not include this casting in the examples for readability (here and in all the other cases as well).  It is safe to upcast from 'int' to 'long long', and safe to upcast from 'int' to 'size_t' except when the 'int' holds a negative value.  The latter would be an exceptional situation caused by either a bug or an integer value larger than we want to support here anyway.

rfalgout (approved):
I am okay with the ci.yml change and the CMake change, but would prefer to not make add the casting to the examples.

src/examples/ex4.c
victorapm: Thanks, Rob! Addressed in [d028759](https://github.com/hypre-space/hypre/pull/1581/commits/d028759283c64737863f8a3e56bba3514d6c951f)

PR #1584, opened 2026-07-17
Add CUDA/HIP device graph utilities

victorapm:
Extends #1576 by @multitalentloes 

Adds a backend-independent device graph abstraction for CUDA and HIP

  - Introduces `hypre_DeviceGraphData` to store the graph, executable graph, and capture state.
  - Adds `hypre_DeviceGraph*` macros for capture, instantiation, launch, and destruction.
  - Regenerates `_hypre_utilities.hpp`.

GPU graphs reduce launch overhead when the same sequence of kernels is executed repeatedly. These will be used by the device level-set ILU implementation


rfalgout (approved):
Looks good to me.  Thanks!

PR #1587, opened 2026-07-20
Fix A*A for struct and sstruct

waynemitchell:
Some bug fixes for multiplying A*A (that is, two square matrices in a row) for both struct and sstruct. Also adds regression tests.


src/struct_mv/struct_grid.c
```diff
@@ -278,6 +279,14 @@ hypre_StructGridAssemble( hypre_StructGrid *grid )
 
    void                *entry_info = NULL;
 
+   /* don't re-assemble */
+   if (hypre_StructGridIsAssembled(grid))
```
rfalgout, line 283: Why does this throw an error flag?  Is this code just for debugging purposes?

src/struct_mv/struct_matmult.c
```diff
@@ -1035,6 +1035,14 @@ hypre_StructMatmultInitialize( hypre_StructMatmultData  *mmdata,
          }
       }
    }
+   if (cdata_space == NULL)
+   {
+      cdata_space = hypre_BoxArrayClone(fdata_space);
+   }
+   if (fdata_space == NULL)
```
rfalgout: This second `if` shouldn't be needed.  I believe we are always guaranteed `fdata_space` non-NULL (except in the case when `nmatrices = 0` but that's not a real case).

src/struct_mv/struct_grid.c
waynemitchell: I was mirroring behavior in `hypre_BoxManAssemble()`:
https://github.com/hypre-space/hypre/blob/master/src/struct_mv/box_manager.c#L1158

The assert below shouldn't be there... that was probably something I put in when debugging. Let me know if you think an error flag should now be thrown.

src/struct_mv/struct_matmult.c
waynemitchell: I couldn't really see this from just looking briefly at the code... `fdata_space` is set to `NULL` above in 995. Then there's that loop over the matrices. Is guaranteed that `mtypes[m]` will be 0 for at least one matrix?

I guess looking back further in the routine, `mtypes` is initialized to all zeros, then only set to 1 if you find a coarser stride or something... So at least one of the matrices has `mtype[m] == 0`. I think I get it now. Let me know if I've misinterpreted anything. I'll remove the unnecessary code here.

rfalgout:
Hi @waynemitchell .  There is something amiss in the structmat tests.  The `runtest.sh TEST_structmat/structmat.sh` run isn't clean due to something about missing files.  I did run one A*A test to confirm it works now, though.

src/struct_mv/struct_matmult.c
rfalgout: Yup, that's correct.

src/struct_mv/struct_grid.c
rfalgout: Interesting.  In both cases, it seems like the flag is there only to ensure that re-assemble is never called.  If we allow assemble to be called more than once, then we shouldn't throw an error.  However, in both of these cases, we are assuming that re-assemble is being called with no changes to the structure in between assemble calls.  That's probably an okay assumption to make for what we need here.  The problem with keeping the error flag is that many of our users check the error flag, so we only want to throw errors when we actually consider them to be errors.

rfalgout (approved):
This looks good to me once you get the regression tests running.  Thanks!

waynemitchell:
> Hi @waynemitchell . There is something amiss in the structmat tests. The `runtest.sh TEST_structmat/structmat.sh` run isn't clean due to something about missing files. I did run one A*A test to confirm it works now, though.

Hmm... I'm not able to reproduce this. Looks like it is running clean for me.

rfalgout:
> > Hi @waynemitchell . There is something amiss in the structmat tests. The `runtest.sh TEST_structmat/structmat.sh` run isn't clean due to something about missing files. I did run one A*A test to confirm it works now, though.
> 
> Hmm... I'm not able to reproduce this. Looks like it is running clean for me.

Maybe it's an issue on my mac.  Don't let me hold you up.  I'll look at it later.

rfalgout:
Seems to be working now.  Maybe a run failed randomly the first time I checked.  Thanks!

PR #1589, opened 2026-07-22
Fix communication exchanges for multicomponent vectors

### LINKED ISSUE #1586: ParCSR communication metadata becomes invalid after a 2-to-1 multivector component transition
When the same hypre_ParCSRCommPkg is reused with the sequence

1 component → 2 components → 1 component

the final entries of send_map_starts and recv_vec_starts become **zero** after returning to one component.

The suspected source is:

hypre_ParCSRCommPkgUpdateVecStarts()

where the extents are updated using:

send_map_starts[i] *= num_components_in / num_components;
recv_vec_starts[i] *= num_components_in / num_components;

For a 2 → 1 transition, integer division evaluates 1 / 2 as zero, causing the communication extents to be cleared.

The expected behavior is that a 1 → 2 → 1 transition restores the original scalar communication metadata.

A possible correction is:

send_map_starts[i] =
    (send_map_starts[i] / num_components) * num_components_in;

recv_vec_starts[i] =
    (recv_vec_starts[i] / num_components) * num_components_in;

The reconstruction of send_map_elmts may also need to use the original scalar mapping.

Environment
HYPRE: v3.0.0-24-gfca98afcd
Open MPI: 4.1.2
Compiler: g++ 11.4.0
Platform: Linux x86-64
HYPRE_USING_CUDA: enabled
HYPRE_USING_GPU: enabled
GPU-aware MPI: disabled
Persistent communication: disabled
HYPRE_Int: 4 bytes
HYPRE_BigInt: 4 bytes
HYPRE_Real: 8 bytes

The issue occurs with multiple MPI ranks and is not observable with one rank because no inter-rank communication is required.

victorapm:
Closes #1586 

When `hypre_ParCSRCommPkg` is reused for multi-component vectors with a varying number of components such as `1, 2, 1`, the arrays `send_map_starts` and `recv_vec_starts` were zeroed out incorrectly. This happened because the extents were rescaled by the integer ratio `num_components_in / num_components`, which truncates to 0 when transitioning down (e.g. 1/2). With zeroed extents, matvec would pack/receive the wrong data and produce incorrect results.

The fix in this PR computes the scalar element count once and uses it for the allocation size and all `send_map_elmts` reconstruction loop bounds, and rescales the extents via `send_map_starts[i] = (send_map_starts[i] / num_components) * num_components_in` to avoid getting a zero result

### DIFF
--- a/src/parcsr_mv/par_csr_communication.c
+++ b/src/parcsr_mv/par_csr_communication.c
@@ -1089,18 +1089,24 @@ hypre_ParCSRCommPkgUpdateVecStarts( hypre_ParCSRCommPkg *comm_pkg,
 
    HYPRE_Int    *send_map_elmts_new;
 
+   HYPRE_Int     num_send_elmts;
+
    HYPRE_Int     i, j;
 
    hypre_assert(num_components > 0);
 
    if (num_components_in != num_components)
    {
+      /* Number of scalar send elements. This is computed from the current
+         (not-yet-rescaled) extents, which hold num_send_elmts * num_components. */
+      num_send_elmts = send_map_starts[num_sends] / num_components;
+
       /* Update number of components in the communication package */
       hypre_ParCSRCommPkgNumComponents(comm_pkg) = num_components_in;
 
       /* Allocate send_maps_elmts */
       send_map_elmts_new = hypre_CTAlloc(HYPRE_Int,
-                                         send_map_starts[num_sends] * num_components_in,
+                                         num_send_elmts * num_components_in,
                                          HYPRE_MEMORY_HOST);
 
       /* Update send_maps_elmts */
@@ -1108,7 +1114,7 @@ hypre_ParCSRCommPkgUpdateVecStarts( hypre_ParCSRCommPkg *comm_pkg,
       {
          if (num_components == 1)
          {
-            for (i = 0; i < send_map_starts[num_sends]; i++)
+            for (i = 0; i < num_send_elmts; i++)
             {
                for (j = 0; j < num_components_in; j++)
                {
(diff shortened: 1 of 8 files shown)


src/parcsr_mv/par_csr_communication.c
```diff
@@ -1089,26 +1089,32 @@ hypre_ParCSRCommPkgUpdateVecStarts( hypre_ParCSRCommPkg *comm_pkg,
...
+         (not-yet-rescaled) extents, which hold num_send_elmts * num_components. */
+      num_send_elmts = send_map_starts[num_sends] / num_components;
+
       /* Update number of components in the communication package */
       hypre_ParCSRCommPkgNumComponents(comm_pkg) = num_components_in;
 
       /* Allocate send_maps_elmts */
       send_map_elmts_new = hypre_CTAlloc(HYPRE_Int,
-                                         send_map_starts[num_sends] * num_components_in,
+                                         num_send_elmts * num_components_in,
```
rfalgout, line 1109: The size of this alloc is very different now.  In the discussion, you talk about the rescaling below truncating to zero, but it seems like there was more going wrong here.  Is that correct?

victorapm: Right, sorry, I forgot to mention this in the description. In this code path, `send_map_starts` is already scaled (`num_send_elmts * num_components`). So, now I'm recovering `num_send_elmts` and recomputing the right allocation size via `num_send_elmts * num_components_in`. When the new value is smaller, we shouldn't need to reallocate `send_map_elmts`, but this fix is reallocating always. Let me know if you prefer the "smarter" fix

rfalgout: Sounds good.  Just wanted to verify.

rfalgout (approved):

PR #1590, opened 2026-07-22
Fix re-asseble of struct grid under sstruct matmult

waynemitchell:


src/sstruct_mv/sstruct_grid.c
```diff
@@ -319,6 +319,7 @@ hypre_SStructPGridAssemble( hypre_SStructPGrid  *pgrid )
 
    cell_sgrid = hypre_SStructPGridCellSGrid(pgrid);
    HYPRE_StructGridSetPeriodic(cell_sgrid, periodic);
+   hypre_SStructPGridCellSGridDone(pgrid) = hypre_StructGridIsAssembled(cell_sgrid);
```
waynemitchell, line 322: This removes the re-assemble issue. Not sure whether this is the best fix. There are a few comments from you, @rfalgout, in the code regarding the CellSGrid. See here and the following comments around when grids are assembled or not:
https://github.com/hypre-space/hypre/blob/master/src/sstruct_mv/sstruct_matmult.c#L318

rfalgout: If the regression tests run, then I think this fix is sufficient.  PMatmult only works for the cell-centered case currently.  I think many of these comments have to do with generalizing it at a later date.

rfalgout (approved):

PR #1591, opened 2026-07-24
Additional support for mixed-precision on GPUs

oseikuffuor1:
This PR add additional support for mixed-precision on GPUs. In particular, tests like ij_mp.c and struct_mp.c have been refactored to support GPU runs. Regression tests have also been added to test some mixed-precision solver functionality. Other mixed-precision functions to support mixed-precision solver development have also been added.

Regression tests:
- [x] tux (Rob)
- [x] tux valgrind (Rob)
- [x] tioga (Daniel)
- [ ] matrix (Victor)
- [x] aurora (Victor)


AUTOTEST/machine-tux.sh
```diff
@@ -139,6 +139,7 @@ co="--enable-bigint --enable-debug"
 ./renametest.sh basic $output_dir/basic--enable-bigint
 
 co="--enable-debug --enable-mixed-precision"
+ro="-mixed-precision -rt -save ${save}"
```
rfalgout: We want this to continue to run the normal tests defined with `ro`.  So, this should be changed to
```
RO="$ro -mixed-precision -rt -save ${save}"
```
then use `$RO` on the next `test.sh` line.  I can make the change if you agree.

src/struct_ls/CMakeLists.txt
```diff
@@ -90,9 +90,11 @@ set(REGULAR_SRCS
 
 # Mixed precision sources
 set(MUP_SRCS
+  HYPRE_struct_ls_mp.c
   mup_fixed.c
   mup_functions.c
   mup_pre.c
+  HYPRE_struct_ls_mp.c
```
rfalgout: Looks like this file appears twice

src/test/Makefile
```diff
@@ -80,9 +80,7 @@ HYPRE_UNIT_TESTS =\
 HYPRE_MP_DRIVERS =\
  ij_mp.c\
  test_mp.c\
```
rfalgout, line 82: Do we want to keep `test_mp.c`?

src/test/ij_mp.c
```diff
@@ -320,1609 +840,3460 @@ int main (int argc, char *argv[])
...
+               hypre_printf("Solver: DOUBLE PRECISION DS-PCG\n");
+               HYPRE_PCGSetPrecond_pre(solver_precision, pcg_solver,
+                                       (HYPRE_PtrToSolverFcn) HYPRE_ParCSRDiagScale_dbl,
+                                       (HYPRE_PtrToSolverFcn) HYPRE_ParCSRDiagScaleSetup_dbl,
+                                       pcg_precond);
+            }
+            else
+            {
+               hypre_printf("Solver: LONG DOUBLE PRECISION DS-PCG\n");
+               HYPRE_PCGSetPrecond_pre(solver_precision, pcg_solver,
```
victorapm: Shouldn't the `HYPRE_PCGSetPrecond_pre` calls happen outside the `if (myid == 0)` block?

Same pattern occurs in other blocks below

src/struct_mv/struct_mv_mp.c
```diff
@@ -54,96 +52,28 @@ hypre_StructVectorCopy_mp( hypre_StructVector *x,
    /* Implicit conversion to generic data type (void pointer) */
    xp = hypre_StructVectorData(x);
    yp = hypre_StructVectorData(y);
+   /* copy data */
+   hypre_RealArrayCopy_mp(hypre_StructVectorPrecision (x), xp, hypre_StructVectorMemoryLocation(y),
```
victorapm:
```suggestion
   hypre_RealArrayCopy_mp(hypre_StructVectorPrecision (x), xp, hypre_StructVectorMemoryLocation(x),
```

src/struct_mv/struct_mv_mp.c
```diff
@@ -54,96 +52,28 @@ hypre_StructVectorCopy_mp( hypre_StructVector *x,
...
+hypre_StructVectorClone_mp( hypre_StructVector *x, HYPRE_Precision new_precision )
+{
+   hypre_StructVector   *y;
+
+   if (hypre_StructVectorPrecision(A) == new_precision)
+   {
+      return hypre_ParCSRMatrixClone_pre( hypre_ParCSRMatrixPrecision(A), A, 1 );
+   }
+}
+*/
```
victorapm: wondering why this code block is commented out

src/test/struct_mp.c
```diff
@@ -1770,17 +1724,42 @@ main( hypre_int argc,
...
+      {
+         HYPRE_StructSMGDestroy_pre(precond_precision, precond);
+      }
+      else if (solver_id == 1 || solver_id == 2 || solver_id == 3 ||
+               solver_id == 11 || solver_id == 12 || solver_id == 13 ||
+               solver_id == 21 || solver_id == 22 || solver_id == 23 ||
+               solver_id == 31 || solver_id == 32 || solver_id == 33)
+      {
+         HYPRE_StructPFMGDestroy_pre(precond_precision, precond);
+      }
```
victorapm, line 1743: Looks like the largest `solver_id` available in this driver is 24. So we shouldn't check numbers larger than that like 30, 31, ...

src/test/ij_mp.c
```diff
@@ -320,1609 +840,3460 @@ int main (int argc, char *argv[])
...
+         hypre_printf("       20=DiagScale-FlexGMRES\n");
+         hypre_printf("       21=AMG-FlexGMRES\n");
+         hypre_printf("       30=DiagScale-BiCGSTAB\n");
+         hypre_printf("       31=AMG-BiCGSTAB\n");
+         hypre_printf("\n");
+         hypre_printf("  -solver_precision <p>     : solver precision\n");
+         hypre_printf("                       Precision for outer solver (Krylov solver) \n");
+         hypre_printf("                        0 = flt, 1 = dbl, 2 = long_dbl\n");
+         hypre_printf("\n");
+         hypre_printf("  -precond_precision <p>     : precond precision\n");
```
victorapm:
```suggestion
         hypre_printf("  -pc_precision <p>          : precond precision\n");
```

src/test/ij_mp.c
oseikuffuor1: Not good. Thanks for catching this Victor!

src/test/Makefile
oseikuffuor1: I wanted to keep it temporarily for now, but I might archive it somewhere else.

AUTOTEST/machine-tux.sh
oseikuffuor1: @rfalgout yes that is correct. I will fix that if you haven't already.

src/mup_check
```diff
@@ -27,15 +27,34 @@ fi
...
+            )
             check_type="GPU"
             ;;
         --hip)
-            opts="$opts --without-umpire --with-hip"
+            opts=(
+                $opts 
+                --without-umpire 
+                --with-hip 
+                --with-MPI-include=${MPICH_DIR}/include
```
rfalgout, line 42: Do we generally need to specify the MPI stuff when compiling with hip?  Seems annoying to have to do that.

victorapm: Yes! Not for the CMake build, but currently yes for the autotools build because it isn't "smart enough"

src/test/Makefile
rfalgout: I mentioned it only because it does not also appear in the cmake build, so I thought maybe this was an oversight.  Whatever you choose to do here is okay with me.  Thanks!

src/test/struct_mp.c
oseikuffuor1: Yes thanks, I missed that.

src/mup_check
oseikuffuor1: @rfalgout I believe so. We do that in our regression tests as well. @victorapm correct me if I'm wrong.

AUTOTEST/machine-tux.sh
oseikuffuor1: resolved in [5b21e00](https://github.com/hypre-space/hypre/pull/1591/commits/5b21e00e9cd5746b658bece2d0da9df63b45edb5)

src/struct_ls/CMakeLists.txt
oseikuffuor1: resolved in [5b21e00](https://github.com/hypre-space/hypre/pull/1591/commits/5b21e00e9cd5746b658bece2d0da9df63b45edb5)

src/test/ij_mp.c
oseikuffuor1: Resolved in [a4fc222](https://github.com/hypre-space/hypre/pull/1591/commits/a4fc222d443e959898470c201e36ca541b067374)

src/test/struct_mp.c
oseikuffuor1: resolved in [a4fc222](https://github.com/hypre-space/hypre/pull/1591/commits/a4fc222d443e959898470c201e36ca541b067374)

src/test/ij_mp.c
oseikuffuor1: resolved in [a4fc222](https://github.com/hypre-space/hypre/pull/1591/commits/a4fc222d443e959898470c201e36ca541b067374)

src/struct_mv/struct_mv_mp.c
oseikuffuor1: This was actually a placeholder for me to do later, but I had some questions, so I have decided to defer it until we discuss it first. I have cleaned that up.

src/struct_mv/struct_mv_mp.c
oseikuffuor1: resolved in [2cd0c73](https://github.com/hypre-space/hypre/pull/1591/commits/2cd0c736596ba25845fd34b0c7a0ae68f91e4a50)

src/utilities/device_reducer.h
```diff
@@ -293,9 +312,12 @@ struct ReduceSum
    {
       if (hypre_HandleReduceBuffer(hypre_handle()) == NULL)
       {
-         /* allocate for the max size for reducing HYPRE_Real6 type */
+         /* Allocate for the max size for reducing HYPRE_Real6 type. This buffer is
+          * held by the (precision-agnostic) handle and shared by all precisions, so
+          * size it with the largest real type instead of the build's HYPRE_Real. */
          hypre_HandleReduceBuffer(hypre_handle()) =
-            hypre_TAlloc(HYPRE_Real6, HYPRE_MAX_NTHREADS_BLOCK, HYPRE_MEMORY_DEVICE);
+            hypre_TAlloc(char, 6 * sizeof(hypre_long_double) * HYPRE_MAX_NTHREADS_BLOCK,
```
rfalgout:
```suggestion
            hypre_TAlloc(hypre_long_double, 6 * HYPRE_MAX_NTHREADS_BLOCK,
```

rfalgout: Alternatively, would the following work?
```
hypre_TAlloc(hypre_RealPack6<hypre_long_double>, HYPRE_MAX_NTHREADS_BLOCK, ...
```
This is more faithful to the "typed alloc" macro we use, assuming there isn't something weird about using templated types here.

src/utilities/device_reducer.h
```diff
@@ -15,44 +15,56 @@
 
 template<typename T> void OneBlockReduce(T *d_arr, HYPRE_Int N, T *h_out);
```
rfalgout: Not related to this PR, but the function `OneBlockReduce` doesn't appear to be used anywhere in hypre.

src/utilities/device_reducer.h
victorapm: Thanks Rob, that's a better fix. Added in [e27a644](https://github.com/hypre-space/hypre/pull/1591/commits/e27a6444fcfd1d17fbff4af11a4c0f31b1f804c7)

src/utilities/device_reducer.h
victorapm: Agreed. Removed in [e27a644](https://github.com/hypre-space/hypre/pull/1591/commits/e27a6444fcfd1d17fbff4af11a4c0f31b1f804c7)

src/utilities/device_reducer.h
```diff
@@ -15,44 +15,56 @@
...
 
-struct HYPRE_Real2
+/* Packed real types used by the reducers below.
+ *
+ * These are templated on the scalar type instead of referring to HYPRE_Real
```
rfalgout, line 18: I still don't understand why there is a difference between static and shared compiles here.  But because these C++ structs have overloaded functions associated with them, I think this templated approach is the right approach anyway, so this looks good.  Thanks!

Are we confident about the discussion in this comment block?

victorapm: Hi Rob, I think the comment block describes relatively well the problem and how we fixed it. Is there anything that you believe could be improved in the comment block?

rfalgout: The first part of the comment about structs having the same name in each of the builds is true for all the structs in hypre, yet they don't cause problems with either static or shared builds.  Presumably, the structs are immediately incorporated into the object code at compile time, not link time, so the names don't matter.  For some reason, this is not handled quite the same way in C++, probably because of the overloaded operators, where it makes sense to mimic what happens with non-static functions.  But even considering that, there is a difference between static and shared linkage for some reason.  I guess I'm not sure we fully understand what's going on here.

We can keep the comment as is.  I just wanted to ask what you thought.

rfalgout (approved):
This looks good, @oseikuffuor1 !  I am running the full set of tux tests now, but I ran a mixed precision test that was clean, so I think it well be okay.  I'll let you know.  Thanks!

rfalgout:
FYI, both tux and tux-valgrind are clean.

victorapm (approved):
Thanks for all the work, Daniel!

`machine-matrix`is running. I added a list of tests in the PR description

oseikuffuor1:
Thanks Rob and Victor for taking a look at this PR. @rfalgout I have fixed a bug in PCG for when the true residual is recomputed here [3683008](https://github.com/hypre-space/hypre/pull/1591/commits/3683008f0621d3f31965350596a7ba077dd6c0b1). Could you take a quick look and let me know if you like the fix. I have a print statement to alert users. Let me know if you's like to rephrase that in any way.

rfalgout:
> Thanks Rob and Victor for taking a look at this PR. @rfalgout I have fixed a bug in PCG for when the true residual is recomputed here [3683008](https://github.com/hypre-space/hypre/pull/1591/commits/3683008f0621d3f31965350596a7ba077dd6c0b1). Could you take a quick look and let me know if you like the fix. I have a print statement to alert users. Let me know if you's like to rephrase that in any way.

Hi @oseikuffuor1 .  Can you explain the bug?  What was actually going wrong?  Also, does this still run through our current regression tests okay?  Thanks!

oseikuffuor1:
> > Thanks Rob and Victor for taking a look at this PR. @rfalgout I have fixed a bug in PCG for when the true residual is recomputed here [3683008](https://github.com/hypre-space/hypre/pull/1591/commits/3683008f0621d3f31965350596a7ba077dd6c0b1). Could you take a quick look and let me know if you like the fix. I have a print statement to alert users. Let me know if you's like to rephrase that in any way.
> 
> Hi @oseikuffuor1 . Can you explain the bug? What was actually going wrong? Also, does this still run through our current regression tests okay? Thanks!

Ah sorry for not being clear. The bug is an inconsistency in the PCG search direction and beta and gamma updates when the recursive residual drifts from the true residual and so iterations restart from the true residual. This shows up when the option to recompute residual is set. 
* In the case where we use the 2-norm to check for convergence [line 821](https://github.com/hypre-space/hypre/commit/3683008f0621d3f31965350596a7ba077dd6c0b1#diff-18e5b29f310002c0bf4342a0f35a791b461e413446e15c01298ecdc49810b02dR821) we not compute the preconditioned residual (from the new residual: s = C*r). If the convergence check fails in [line 834](https://github.com/hypre-space/hypre/commit/3683008f0621d3f31965350596a7ba077dd6c0b1#diff-18e5b29f310002c0bf4342a0f35a791b461e413446e15c01298ecdc49810b02dL834), then the standard p = s + beta*p where s is inconsistent (since it hasn't been updated from the new true residual) -- see branch [here](https://github.com/hypre-space/hypre/commit/3683008f0621d3f31965350596a7ba077dd6c0b1#diff-18e5b29f310002c0bf4342a0f35a791b461e413446e15c01298ecdc49810b02dL974-L984). Additionally, it should actually be restarting PCG from the true residual, so taking the second path p = s should be the correct action, which is not what was happening.

* In the non 2-norm case, the preconditioned residual is computed correctly, but the search update branch also selects p = s + beta*p (using the old direction built from the old r), instead of p = s;

The commit fixes these issues to ensure consistency in the restart. Hopefully this is clear to you.

src/krylov/pcg.c
```diff
@@ -831,7 +831,42 @@ hypre_PCGSolve( void *pcg_vdata,
...
+            if (two_norm)
+            {
+               /* compute preconditioned resdidual - s = C*r */
+               (*(pcg_functions->ClearVector))(s);
+               precond(precond_data, precond_Mat, r, s);
+               /* iprod = gamma = <r,s> */
+               gamma = (*(pcg_functions->InnerProd))(r, s);
+            }
+            /* set flag to restart pcg from true residual */
+            recompute_true_residual = 1;
```
rfalgout, line 868: This flag is already set - that's how we ended up in this else block.

oseikuffuor1: It is actually not set. It is only set if the user chooses to periodically compute the residual [here](https://github.com/hypre-space/hypre/blob/0beb8b47a8bee4011f7b74685fb147231b90d352/src/krylov/pcg.c#L581). I set it in the convergence check to also ensure that the convergence triggered restart is done correctly. Upon restart, it goes back to being unset until the periodic condition is met or the residual convergence check fails again.

rfalgout: Sorry about that!  I was looking at the diff summary page which puts a big gap between the `if` block I was looking at and this piece of code.  I should have been more careful.

I will give this a deeper look.  One big complaint I have about our Krylov solvers, and especially PCG, is that they are really difficult to navigate.  The original PCG code was simple and easy to follow, but it has grown into a beast.  I want to be sure I understand how this change fits in with everything else.  Stay tuned.  Thanks!

oseikuffuor1: Thanks Rob. I also agree it is much more complex than the standard PCG :-). Doing this relatively simple fix took me sometime to make sure all the logic remained consistent and nothing else broke. I see some code refactoring in our future :-). Let me know if you have other questions.

rfalgout: Whew!!  Really complex code!!  Must have taken you a while to unravel it.  Looks correct.  Thanks!

rfalgout:
Hi @oseikuffuor1 .  I will run the tux tests again, just to be sure nothing changed due to the PCG mods.

oseikuffuor1:
> Hi @oseikuffuor1 . I will run the tux tests again, just to be sure nothing changed due to the PCG mods.

Thanks Rob! Let me know if anything needs updating. Feel free to push any updates you see, particularly from machine-tux.

rfalgout:
> > Hi @oseikuffuor1 . I will run the tux tests again, just to be sure nothing changed due to the PCG mods.
> 
> Thanks Rob! Let me know if anything needs updating. Feel free to push any updates you see, particularly from machine-tux.

The tux tests are clean.  Looks like there is an astyle issue, but otherwise this looks good.  Thanks again!

oseikuffuor1:
> > Hi @oseikuffuor1 . I will run the tux tests again, just to be sure nothing changed due to the PCG mods.
> 
> Thanks Rob! Let me know if anything needs updating. Feel free to push any updates you see, particularly from machine-tux.

Thanks Rob. I will fix the astyle issues before merging.

PR #1592, opened 2026-07-24
Sstruct matrix set array values

waynemitchell:
Adds routines to set/add/get arrays of values for (s)struct matrices. This provides addition flexibility for setting values at arbitrary indexes compared to the SetBoxValues routines and aims to provide greater performance (on the GPU) than setting things one at a time via the SetValues routines.


src/struct_mv/struct_matrix.c
```diff
@@ -1610,6 +1611,339 @@ hypre_StructMatrixSetBoxValues( hypre_StructMatrix *matrix,
...
+                           j,
+                           nvalues,
+                           indexes,
+                           stencil_indices,
+                           values,
+                           mat_box_data,
+                           done_arr );
+      }
+   }
+   if (add_to_ghost)
```
victorapm: Need to recompute `mat_box_data` in this loop

src/struct_mv/struct_matrix.c
```diff
@@ -1610,6 +1611,339 @@ hypre_StructMatrixSetBoxValues( hypre_StructMatrix *matrix,
...
+
+   const dim3 bDim = hypre_GetDefaultDeviceBlockDimension();
+   const dim3 gDim = hypre_GetDefaultDeviceGridDimension(nvalues, "thread", bDim);
+
+   for (i = 0; i < hypre_StructMatrixRanNBoxes(matrix); i++)
+   {
+      grid_box = hypre_StructMatrixBox(matrix, i);
+      data_box = hypre_StructMatrixBoxDataBox(matrix, i);
+
+      for (j = 0; j < stencil_size; j++)
```
victorapm, line 1697: Check for constant and symmetric stencil entries inside this loop

src/sstruct_mv/sstruct_matrix.c
```diff
@@ -2146,6 +2688,308 @@ hypre_SStructMatrixSetInterPartValues( HYPRE_SStructMatrix  matrix,
...
+      {
+         struct_stencil_indices_h[i] = smap[tentries_h[i]];
+      }
+      hypre_TMemcpy(struct_stencil_indices, struct_stencil_indices_h, HYPRE_Int, ntvalues,
+                    HYPRE_MEMORY_DEVICE, HYPRE_MEMORY_HOST);
+      hypre_TFree(tentries_h, HYPRE_MEMORY_HOST);
+      hypre_TFree(struct_stencil_indices_h, HYPRE_MEMORY_HOST);
+
+      /* zero out values in PMatrix (possibly in ghost) */
+      hypre_StructMatrixSetArrayValuesDevice(smatrix, ntvalues, tindexes, struct_stencil_indices, NULL,
```
victorapm, line 3039: This zeroes the PMatrix using an `smatrix` that was last assigned by the box-union loop at line 2877
(`hypre_SStructPMatrixSMatrix(pmatrix, var, vars[i])` with `i == stencil_size-1`), not the smatrix belonging to the entries being zeroed.

src/sstruct_mv/sstruct_matrix.c
```diff
@@ -1941,6 +2184,303 @@ hypre_SStructMatrixSetBoxValues( HYPRE_SStructMatrix  matrix,
...
+
+/*--------------------------------------------------------------------------
+ * (action > 0): add-to values
+ * (action = 0): set values
+ * (action < 0): get values
+ * (action =-2): get values and zero out
+ *--------------------------------------------------------------------------*/
+
+HYPRE_Int
+hypre_SStructMatrixSetArrayValuesDevice( HYPRE_SStructMatrix  matrix,
```
victorapm, line 2458: If I understand correctly, `hypre_SStructMatrixSetArrayValuesDevice` splits the caller's `values` into fresh
`Svalues`/`Uvalues` device buffers, fills them via `hypre_StructMatrixGetArrayValuesDevice` / `hypre_SStructUMatrixSetArrayValuesDevice`, and returns without scattering them back into `values`. Should we do the scatter back?

src/sstruct_mv/sstruct_matrix.c
```diff
@@ -1233,6 +1234,246 @@ hypre_SStructUMatrixSetValues( hypre_SStructMatrix *matrix,
...
 
+#if defined(HYPRE_USING_GPU)
+/*--------------------------------------------------------------------------
+ * (action > 0): add-to values
+ * (action = 0): set values
+ * (action < 0): get values
+ *--------------------------------------------------------------------------*/
+
+HYPRE_Int
+hypre_SStructUMatrixSetArrayValuesDevice( hypre_SStructMatrix *matrix,
```
victorapm, line 1246: `hypre_SStructUMatrixSetArrayValuesDevice` sorts `rows` with `cols`/`values` zipped and compacts `values` with `remove_if`. For `action < 0` (Get) the values that `HYPRE_IJMatrixGetValues` writes back are in row-sorted, compacted order, which no longer corresponds to the caller's `indexes`/`entries` ordering. Looks like the public API returns a permuted array with no way to map it back. It also permutes the caller's `values` buffer in place for Set/AddTo

src/sstruct_mv/sstruct_matrix.c
```diff
@@ -2146,6 +2688,308 @@ hypre_SStructMatrixSetInterPartValues( HYPRE_SStructMatrix  matrix,
...
+                  /* ibox1 should be a single index */
+                  for (d = 0; d < ndim; d++)
+                  {
+                     hypre_assert(hypre_BoxIMinD(ibox1, d) == hypre_BoxIMaxD(ibox1, d));
+                  }
+                  tlocations_h[ntvalues++] = i;
+
+                  /* Update list of part boundaries */
+                  /* WM: todo - I think we want to move away from using pbnd boxaa, but for now this is needed for ssamg */
+                  hypre_ForBoxI(j, grid_boxes)
```
victorapm: The device inter-part routine appends to `pbnd_boxaa` and runs the `hypre_UnionBoxes` pass unconditionally; the CPU version does this only under `if (action >= 0)` (line 2619).

src/sstruct_mv/sstruct_matrix.c
```diff
@@ -2146,6 +2688,308 @@ hypre_SStructMatrixSetInterPartValues( HYPRE_SStructMatrix  matrix,
...
+   ibox0     = hypre_BoxCreate(ndim);
+   ibox1     = hypre_BoxCreate(ndim);
+   ibox2     = hypre_BoxCreate(ndim);
+   tobox     = hypre_BoxCreate(ndim);
+   frbox     = hypre_BoxCreate(ndim);
+
+   /* Copy indexes and entries to the host */
+   /* WM: todo - rethink these checks on memory location... I only expect this
+    * function to be called with inputs and the matrix in device memory? */
+   tlocations_h = hypre_TAlloc(HYPRE_Int, nvalues, HYPRE_MEMORY_HOST);
```
victorapm, line 2875: Not sure if `nvalues` is enough space here given what happens in line 2849. Maybe it is, just wanted to double check

src/struct_mv/struct_matrix.c
```diff
@@ -1610,6 +1611,339 @@ hypre_StructMatrixSetBoxValues( hypre_StructMatrix *matrix,
...
+                                           HYPRE_Complex     *values,
+                                           HYPRE_Complex     *mat_box_data )
+{
+   HYPRE_Int d;
+   HYPRE_Int done = 0;
+   HYPRE_Int ndim = grid_box.ndim;
+   HYPRE_Int my_index[3];
+   HYPRE_Int  i = hypre_gpu_get_grid_thread_id<1, 1>(item);
+   /* Note: can pass values = NULL to set to 0.0 */
+   HYPRE_Complex val = values ? values[i] : 0.0;
```
victorapm: I think we need to guard this read `values[i]` for threads in the final thread block

src/sstruct_mv/sstruct_matrix.c
```diff
@@ -1233,6 +1234,246 @@ hypre_SStructUMatrixSetValues( hypre_SStructMatrix *matrix,
...
+                                                  &(cols_h[nvalues_set++]), matrix_type);
+         }
+      }
+      else
+      {
+         /* non-stencil entries */
+         entry -= size;
+         hypre_SStructGraphGetUVEntryRank(graph, part, var, index, &Uverank);
+
+         /* WM: todo - is this good error handling? What do we want the code to do if we are getting bad entries? */
```
victorapm: Maybe we could give some descriptive error message about what went wrong in this code path. We are also missing a few `hypre_TFree` calls here

src/sstruct_mv/sstruct_matrix.c
```diff
@@ -1233,6 +1234,246 @@ hypre_SStructUMatrixSetValues( hypre_SStructMatrix *matrix,
...
+   /* WM: todo */
+#else
+
+   /* WM: note that the values array passed to this function is modified here... I think that's fine. */
+   HYPRE_THRUST_CALL( remove_if,
+                      values,
+                      values + nvalues,
+                      skip_val,
+                      thrust::identity<HYPRE_Int>() );
+   ncols = hypre_TAlloc(HYPRE_Int, nvalues_set, memory_location);
```
victorapm: Looks like this has no matching `hypre_TFree`

src/sstruct_mv/sstruct_matrix.c
```diff
@@ -2146,6 +2688,308 @@ hypre_SStructMatrixSetInterPartValues( HYPRE_SStructMatrix  matrix,
...
+                         tvalues,
+                         tvalues + ntvalues,
+                         tlocations,
+                         values );
+#endif
+
+   } /* end if action */
+
+   /* Free memory */
+   hypre_BoxDestroy(box);
```
victorapm, line 3067:
```suggestion
   hypre_BoxDestroy(set_box);
   hypre_BoxDestroy(box);
```

src/sstruct_mv/_hypre_sstruct_mv.h
```diff
@@ -1058,9 +1061,15 @@ HYPRE_Int hypre_SStructMatrixSetValues ( HYPRE_SStructMatrix matrix, HYPRE_Int p
 HYPRE_Int hypre_SStructMatrixSetBoxValues( HYPRE_SStructMatrix  matrix, HYPRE_Int part,
                                            hypre_Box *set_box, HYPRE_Int var, HYPRE_Int nentries, HYPRE_Int *entries, hypre_Box *value_box,
                                            HYPRE_Complex *values, HYPRE_Int action );
+HYPRE_Int hypre_SStructMatrixSetArrayValuesDevice( HYPRE_SStructMatrix  matrix, HYPRE_Int part,
```
rfalgout, line 1071: It looks like this prototype appears twice.

src/struct_mv/struct_matrix.c
```diff
@@ -1610,6 +1611,339 @@ hypre_StructMatrixSetBoxValues( hypre_StructMatrix *matrix,
...
 
+#if defined(HYPRE_USING_GPU)
+
+/*--------------------------------------------------------------------------
+ *--------------------------------------------------------------------------*/
+
+template <bool clear_ghost>
+__global__ void
+hypreGPUKernel_StructMatrixSetArrayValues( hypre_DeviceItem  &item,
+                                           hypre_Box          grid_box,
```
rfalgout, line 1622: We usually pass boxes as `hypre_Box *`.  It looks like you call this routine below (line 1698) by passing a pointer, too.  There are other differences in the API as well, so I'm not sure what's going on.  Maybe I'm just misreading this.

src/struct_mv/struct_matrix.c
```diff
@@ -1610,6 +1611,339 @@ hypre_StructMatrixSetBoxValues( hypre_StructMatrix *matrix,
...
+                                           HYPRE_Int          set_stencil_index,
+                                           HYPRE_Int          nvalues,
+                                           HYPRE_Int         *indexes,
+                                           HYPRE_Int         *stencil_indices,
+                                           HYPRE_Complex     *values,
+                                           HYPRE_Complex     *mat_box_data )
+{
+   HYPRE_Int d;
+   HYPRE_Int done = 0;
+   HYPRE_Int ndim = grid_box.ndim;
```
rfalgout, line 1633: This may be old fashioned at this point, but we use accessor macros pretty extensively in hypre mainly to make it easy to use `grep` to find all places in the code where data in a given struct is accessed.  Exceptions are in some of the drivers, but they are self contained.  I'd like to have a consistent policy one way or the other.

src/struct_mv/struct_matrix.c
```diff
@@ -1610,6 +1611,339 @@ hypre_StructMatrixSetBoxValues( hypre_StructMatrix *matrix,
...
+      grid_box = hypre_StructMatrixBox(matrix, i);
+      data_box = hypre_StructMatrixBoxDataBox(matrix, i);
+
+      for (j = 0; j < stencil_size; j++)
+      {
+         mat_box_data = hypre_StructMatrixBaseData(matrix, hypre_StructMatrixBaseBoxnum(matrix, i), j);
+
+         if (clear_ghost)
+         {
+            HYPRE_GPU_LAUNCH( (hypreGPUKernel_StructMatrixSetArrayValues<1>),
```
rfalgout: This doesn't appear to match the function prototype.

rfalgout: Scratch this.  I was misreading.

src/struct_mv/struct_matrix.c
rfalgout: Scratch the API mismatch comment.  Is there a reason to pass the box and not a reference to it?  Just wondering.

src/sstruct_mv/sstruct_matrix.c
```diff
@@ -2023,6 +2563,8 @@ hypre_SStructMatrixSetInterPartValues( HYPRE_SStructMatrix  matrix,
          hypre_BoxManEntryGetExtents(toentries[toi],
                                      hypre_BoxIMin(tobox),
                                      hypre_BoxIMax(tobox));
+         /* Why does this intersection work? If the tobox lives in another part, why would their intersection be meaningful?
```
rfalgout: Recall that we put "to boxes" from SetNeighborBox and SetNeighborPart into the box manager, and the indexes are all relative to the "from part".  So your guess is correct. :)

rfalgout (approved):
This looks good to me so far, modulo doing a full set of tests.  Thanks!

src/sstruct_mv/sstruct_matrix.c
```diff
@@ -2146,6 +2688,308 @@ hypre_SStructMatrixSetInterPartValues( HYPRE_SStructMatrix  matrix,
...
+   for (i = 0; i < hypre_SStructStencilSize(stencil); i++)
+   {
+      smatrix = hypre_SStructPMatrixSMatrix(pmatrix, var, vars[i]);
+      sgrid = hypre_StructMatrixGrid(smatrix);
+      grid_boxes = hypre_StructGridBoxes(sgrid);
+      hypre_ForBoxI(j, grid_boxes)
+      {
+         box_id = hypre_StructGridID(sgrid, j);
+         pbnd_boxa = hypre_BoxArrayArrayBoxArray(pbnd_boxaa, box_id);
+         hypre_UnionBoxes(pbnd_boxa);
```
waynemitchell: OK, actually my strategy for the pbnd boxaa here does not work. One issue is that when I union boxes above, the result is an array of non-overlapping boxes, but the boxes in pbnd boxaa may (and sometimes should) overlap in general. Also, it's just generally not possible to recover the correct pbnd boxaa here depending on what arrays of indexes the user chooses to pass in at one time.

The pbnd boxaa stuff is just for ssamg and is also an implementation strategy that we want to eventually move away from, correct? I think my solution for now will be to simply not test the SetArrayValues routines together with ssamg. A better solution in the mid term might be to put the generation of the pbnd boxaa in its own separate routine that is called somewhere other than SetInterPartValues routine. For example, I don't thing pnbd boxaa will be setup correctly if the user calls SetValues to construct the matrix instead of SetBoxValues. Perhaps the pbnd boxaa should be constructed during matrix assembly? At any rate, I'd leave this kind of thing to a future PR.

victorapm: That sounds a good plan, Wayne. Agreed, thanks!

src/sstruct_mv/_hypre_sstruct_mv.h
waynemitchell: Good catch. Should have been `...UMatrix...`. I was also missing a prototype for `hypre_SStructMatrixSplitArrayEntriesDevice()`. Fixed now.

src/struct_mv/struct_matrix.c
waynemitchell: Yes, the box data structure is getting passed by value to the GPU for the kernel execution. You can do this with simple structs (things just get copied to the device similar to passing just a single int/double, etc.). If I passed a pointer to a `hypre_Box`, the kernel would then be trying to access things on the host memory when dereferencing. Does that make sense?

src/struct_mv/struct_matrix.c
waynemitchell: Related to your question above, I am using the `hypre_Box` struct directly here, instead of a pointer to it.

src/struct_mv/struct_matrix.c
rfalgout: Yes.  I wondered if you needed to copy the actual struct because of the whole CPU/GPU memory thing.  Sounds like you do.  Thanks!

src/struct_mv/struct_matrix.c
waynemitchell: Good catch, thanks!

src/struct_mv/struct_matrix.c
waynemitchell: I actually have a question looking at the code in `SetBoxValues` for the constant case. Does the implementation in `SetBoxValues` require unified memory for setting constant values? It looks like we are directly dereferencing a pointer to the data (which may live in device memory). Again, thanks for catching this, I'll fix it in the new routines.

src/sstruct_mv/sstruct_matrix.c
waynemitchell: Ended up removing the pbnd boxaa stuff.

src/struct_mv/struct_matrix.c
waynemitchell: I also see that `hypre_StructMatrixSetValues()` just throws an error if it ends up trying to set a constant value. I'll probably just take that approach for `SetArrayValues`, that is, throw an error and expect the user not to try to set constant values with this function.

src/sstruct_mv/sstruct_matrix.c
waynemitchell: Nice catch. This actually brings up a larger issue with the implementation here... Right now, all connections between different variables are put in the parcsr matrix (rather than struct), so really the only `smatrix` that we end up dealing with is `hypre_SStructPMatrixSMatrix(pmatrix, var, var)`. In general, in the future, I guess we might expect the user to pass an array of matrix values here that includes connections from `var` to multiple other variables and we would have to call `hypre_StructMatrixSetArrayValuesDevice()` multiple different struct matrices. That will require some additional sorting and separating of the passed indexes/values, I guess. This will also impact the implementation in `hypre_SStructMatrixSetArrayValuesDevice()`, where you can see that I have a comment in the code about assuming we only consider the struct matrix for `var` to `var` connections. 

I think for now, I'll change the implementation such that it is explicitly setting `smatrix = hypre_SStructPMatrixSMatrix(pmatrix, var, var)` and leave a note about our assumption that inter-variable connections get placed in parcsr. I think our long-term plan for supporting inter-variable connections in struct involves a more significant refactor throughout the code (and probably the Engwer trick stuff that Rob has planned).

src/sstruct_mv/sstruct_matrix.c
waynemitchell: Ah, yes, I have not tested the `Get` routines here... I do need to do some more work to get the up and running and tested. I will need to do a scatter back as you suggest here and also a similar scatter back in the `UMatrix` portion as you point out below.

In general, it seems that we don't really test the sstruct get values routines in our regression tests? Unless I'm just not seeing where we do that.

src/sstruct_mv/sstruct_matrix.c
waynemitchell: Should be OK, I think. The number of values that are inter-part values that we need to move from struct to parcsr will be less than or equal to the number of values passed into this function. So `ntvalues` should always be less than or equal to `nvalues`.

src/struct_mv/struct_matrix.c
waynemitchell: Yup, thanks.

src/sstruct_mv/sstruct_matrix.c
waynemitchell: Fixed

src/sstruct_mv/sstruct_matrix.c
waynemitchell: Fixed

src/sstruct_mv/sstruct_matrix.c
waynemitchell: OK, I added something a bit more descriptive.

src/sstruct_mv/sstruct_matrix.c
waynemitchell: Get routines should be fixed now.

src/sstruct_mv/sstruct_matrix.c
waynemitchell: Get routines should be fixed now.

src/sstruct_mv/sstruct_matrix.c
waynemitchell: Oh, I think I'm actually not quite right about this... It seems that we DO allow inter-variable couplings to be stored in struct format provided the variables are of the same type. However, there appear to be (kind of sneaky) limitations in the code in the SetValues and SetBoxValues routines that the user should only be passing in connections to one variable type at a time (see my added comments in the code). Just to make things easier on myself (and in the interest of getting this PR done) I will impose the same limitation on the SetArrayValues routines for now.

src/struct_mv/mup.fixed_gpu
```diff
@@ -1,2 +1,11 @@
 hypre_BoxRanksToIndexesDevice
 hypre_GPUKernelBoxRanksToIndexes
+hypre_GPUKernelStructMatrixGetArrayValues
+hypre_GPUKernelStructVectorAddToArrayValues
+hypre_GPUKernelStructVectorGetArrayValues
```
waynemitchell, line 5: I followed the `mup_check` script for this, but I was a bit confused as to why only some of the kernel functions need to be added here. For example, the script made me add `hypre_GPUKernelStructMatrixGetArrayValues`, but not the corresponding `Set` or `AddTo` functions (similarly for the `Vector` kernels, the `Set` kernel is left out). Curious what the reason for this is.

oseikuffuor1: @waynemitchell first can you confirm you are doing mup_check --sycl (or --hip or --cuda) for the appropriate machine? That should turn on "all" GPU supported functionality so that they get added as well, if missing. Are these corresponding Set functions new functions as well? If they existed before, then they might already be in the mup.fixed list depending on how they are guarded in the code (for example, by HYPRE_USING_GPU)

waynemitchell: Yes, I was doing `mup_check --hip` on tioga. The `Set` functions are also newly added and don't exist in any `mup.*` files. In fact, if I do add `hypre_GPUKernelStructMatrixSetArrayValues` to the `mup.fixed_gup` file, the `mup_check --hip` script complains that I should remove it.

oseikuffuor1: Ok good. I took a quick look and it looks like both functions are within the GPU header guards, so they should both be detected. I will take a closer look and see if I spot anything.

oseikuffuor1: @waynemitchell I took a closer look at your branch. The set function is templated and hence tagged as a weak (W) symbol. The mup lists only needs the strong/ standard global functions (T). So the script will ignore the weak symbols and not request their inclusion in the list. Multiple weak symbols with the same name will only have one instance in the library so I was concerned about the HYPRE_Complex argument in your function. However, with your templated function, the C++ compiler mangles the template and precision names appropriately, so all the weak symbols are distinctly present, so there are no issues there. Hopefully this helps address your concerns .

waynemitchell: Oh, interesting. I think I get what you are saying. Sounds like the bottom line is that templated functions need not be included, so the script is behaving as expected and the mixed precision build should work fine as is. Thanks for taking a look, Daniel!

PR #1599, opened 2026-08-02
 Fix crashes and hangs when a rank owns no rows

victorapm:
Three related failures appearing when a parallel decomposition of block linear systems with `dofmap` leaves one or more ranks with empty local blocks. 
 
1. Integer division by zero in hypre_ParCSRMatrixCreateFromParVector (crash)

```
  HYPRE_Int num_cols = (HYPRE_Int) (col_starts[1] - col_starts[0]);
  HYPRE_Int blk_dim  = hypre_max(1, num_rows / num_cols);   /* num_cols == 0 -> SIGFPE */
```

`blk_dim` is a global property of the block structure but was computed from local row and column counts. Fixed by computing it from `global_num_rows / global_num_cols`, which is identical on every rank.

2. Mismatched collective in `hypre_ParVectorInnerProdTagged` (hang)

The routine ends in a reduction whose length depends on the tag count `(num_tags + 1)`, but the branch choosing the tagged vs untagged path was taken from rank-local state. Like the previous case, we need the global value. Fixed by agreeing the tag count across the communicator first (one MPI_Allreduce of a single int)

3. Uninitialized output pointer in `hypre_ParCSRMatrixBlockColSum` (crash)

Both "not divisible by the block dimension" error paths returned without assigning `*B_ptr`, so the caller reads an uninitialized pointer and segfaults. Fixed by setting `*B_ptr = NULL` on every error path and checking the create result.

### DIFF
--- a/src/parcsr_mv/par_vector.c
+++ b/src/parcsr_mv/par_vector.c
@@ -667,51 +667,51 @@ hypre_ParVectorInnerProdTagged( hypre_ParVector  *x,
       iprod = *iprod_ptr;
    }
 
-   /* The tag count must agree across ranks: this routine performs collective
-    * reductions whose size depends on it, and a rank owning no rows carries no tags.
-    * Take the global maximum so every rank follows the same branch below. */
+   /* This routine ends in a collective reduction whose length depends on the tag
+    * count, so every rank must take the same branch below. A rank that owns no rows
+    * carries no tags (tags == NULL), and would otherwise fall back to the untagged
+    * inner product - reducing a single value - while ranks holding tags reduce
+    * num_tags + 1 values. That mismatched collective deadlocks. Decide globally:
+    * if any rank holds a tagged vector, all ranks run the tagged path, and ranks
+    * without local tags simply contribute zeros. */
    {
-      HYPRE_Int global_num_tags = num_tags;
+      HYPRE_Int local_num_tags  = (tags && num_tags > 1) ? num_tags : 0;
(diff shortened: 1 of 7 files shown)


src/parcsr_mv/par_vector.c
```diff
@@ -667,22 +667,51 @@ hypre_ParVectorInnerProdTagged( hypre_ParVector  *x,
...
+    * if any rank holds a tagged vector, all ranks run the tagged path, and ranks
+    * without local tags simply contribute zeros. */
    {
-      iprod[0] = hypre_ParVectorInnerProd(x, y);
+      HYPRE_Int local_num_tags  = (tags && num_tags > 1) ? num_tags : 0;
+      HYPRE_Int global_num_tags = 0;
 
-      *num_tags_ptr = 1;
-      *iprod_ptr = iprod;
+      hypre_MPI_Allreduce(&local_num_tags, &global_num_tags, 1, HYPRE_MPI_INT,
```
rfalgout: The only downside I see here is the extra Allreduce (for performance reasons).  Can this information be computed ahead of time in VectorAssemble or something?  Otherwise, this all looks good.

rfalgout (approved):

src/parcsr_mv/par_vector.c
victorapm: Thanks Rob! That's a better fix. Implemented in https://github.com/hypre-space/hypre/pull/1599/commits/0768d5b04507c3c3de66765484b7be7ea7cdc336

PR #1605, opened 2026-08-09
Fix valgrind issue

victorapm:
`hypre_CSRMatrixData(G)` is already allocated on the `CSRMatrixInitialize` call that happens above


oseikuffuor1 (approved):
Thanks Victor!

PR #1607, opened 2026-08-18
Fix mup treatment of `box_device`

victorapm:
Fixes regression issues on `machine-tioga`

### DIFF
--- a/src/struct_mv/Makefile
+++ b/src/struct_mv/Makefile
@@ -57,7 +57,6 @@ FILES =\
  struct_stmatrix.c
 
 CUFILES =\
- box_device.c\
  struct_matvec.c\
  struct_matvec_cc.c\
  struct_matvec_icc.c\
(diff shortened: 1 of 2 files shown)


src/struct_mv/Makefile
```diff
@@ -94,6 +93,15 @@ MP_FILES = \
  HYPRE_struct_mv_mp.c\
  struct_mv_mp.c
 
+# box_device.c is precision-independent; compile it once in mixed precision
+MP_CUFILES =
+
+ifeq (${MP_BUILD}, 1)
```
rfalgout, line 99: We only use this if/else construct in one other place as far as I know, in `utilities`.  So, what is special about this "device" file that is different from all the others in hypre?

rfalgout: I think we need to introduce a `mup.fixed_gpu` file here instead.

oseikuffuor1: @rfalgout this is correct. There was a new function added in a recent merge that included a GPU-specific function that warranted this file to be created. Incidentally @liruipeng and I have addressed this in his PR (https://github.com/hypre-space/hypre/pull/1492/changes/fceb81c725e15db215e8b8bb03e2c14cd43b92ea)

oseikuffuor1: After adding mup.fixed_gpu, MP_CUFILES should no longer to empty, presumably.

victorapm: Thanks, all! Should I close this branch in favor of #1492 ? Looks like it's ready to merge

oseikuffuor1: @victorapm if you agree that branch addresses this concern, then we can close this. I think it is waiting on my review so I will take a look too.

rfalgout: I just reviewed the other PR, so I am happy to use that one instead (assuming it's almost ready).

victorapm: Closing this one, thanks all!

PR #1610, opened 2026-08-26
Fix memory leak in MGR

victorapm:
Add a missing destroy call


victorapm:
Hi @rfalgout, I'm merging this branch since the change is trivial. Looks like we need to improve our regression tests coverage involving MGR. I'll be doing this on a separate PR. Thanks!

PR #1612, opened 2026-08-27
Use static declaration to avoid weak definition 

oseikuffuor1:
Use static declaration for hypre_CurandUniform_core to specify internal linkage and avoid weak symbol definitions. 

Without static, the linker could coalesce the two weak core<float> definitions, causing both call paths to use whichever (hypre_handle’s) implementation won. With internal linkage, there is no coalescing, so each wrapper uses the correct build-specific handle, and error handler.


src/utilities/device_utils.c
```diff
@@ -1745,7 +1745,7 @@ hypre_DeviceDataCurandGenerator(hypre_DeviceData *data)
  *--------------------------------------------------------------------*/
 
 template <typename T>
-HYPRE_Int
+static HYPRE_Int
```
rfalgout, line 1748: This change makes sense to me, but what about the other two on lines 1681 and 1805?

rfalgout (approved):
Looks good modulo my one question.  Thanks!

src/utilities/device_utils.c
oseikuffuor1: Thanks @rfalgout. I fixed those in [here](https://github.com/hypre-space/hypre/pull/1612/commits/f1d76aaf01311724f5a8b8fe0a1628153b80f8f0)

PR #1613, opened 2026-08-29
Add hypre_CSRMatrixFnormDevice

victorapm:
Add the device implementation for computing Frobenius norm of CSR matrices


rfalgout (approved):
Looks good to me.  Thanks!

liruipeng:
essentially, it returns $\Vert F \Vert^2$ instead of $\Vert F \Vert$. Same issue in the CPU function.

victorapm:
Hi Rui Peng, note we return `hypre_sqrt`. Thanks for checking!

liruipeng:
> Hi Rui Peng, note we return `hypre_sqrt`. Thanks for checking!

yes never mind. I missed that.

PR #1615, opened 2026-09-03
Struct MatPrec preconditioner and matrix support routines

rfalgout:
This PR does the following:
- Adds a Struct matrix preconditioner called `MatPrec` that currently only builds a matrix preconditioner for weighted Jacobi.  There is a placeholder for building a Chebyshev preconditioner, and it can be set up to take a user provided preconditioning matrix as well (these latter two will be implemented later).
- Add several Struct matrix routines such as: `MatrixAddMat`, `MatrixPoly`, `MatrixGetDiagMat`, and `MatrixDiagonal`.
- Adds regression testing via the `sstruct` driver that compares regular Jacobi preconditioning to MatPrec preconditioning.
- Some of the StMatrix code in StructMatmult was reorganized so it could be used in the polynomial code as well.

This solver can be extended at a later date to work with SStruct and ParCSR matrices.

TODO: Add the mixed precision code.


rfalgout:
@waynemitchell .  Just wanted to bring your attention to this in case you are interested in having a look.

src/struct_mv/struct_matop.c
```diff
@@ -992,3 +992,433 @@ hypre_StructMatrixScale( hypre_StructMatrix *A,
...
+   hypre_StructMatrix   *D;
+   HYPRE_Complex        *Ddata;
+   hypre_StructStencil  *stencil;
+   hypre_Index           offset;
+
+   hypre_SetIndex(offset, 0);
+   HYPRE_StructStencilCreate(hypre_StructGridNDim(grid), 1, &stencil);
+   HYPRE_StructStencilSetEntry(stencil, 0, offset);
+   HYPRE_StructMatrixCreate(hypre_StructGridComm(grid), grid, stencil, &D);
+   HYPRE_StructMatrixInitialize(D);
```
victorapm, line 1354: We should use `_v2` here and pass in the same memory location as `A`

src/struct_mv/struct_matop.c
```diff
@@ -992,3 +992,433 @@ hypre_StructMatrixScale( hypre_StructMatrix *A,
...
+      Ddata = hypre_StructMatrixConstData(D, 0);
+
+      if (type == 0)
+      {
+         Ddata[0] = weight * Adata[0];
+      }
+      else if (type == 1)
+      {
+         Ddata[0] = weight / Adata[0];
+      }
```
victorapm, line 1369: This is potentially dereferencing device memory and therefore it's unsafe.

src/struct_mv/struct_matop.c
rfalgout: What is `_v2`?

src/struct_mv/struct_matop.c
rfalgout: Thanks!  I have not worried about the GPU side yet, but will get to it.

src/struct_mv/struct_matop.c
rfalgout: I see that there is `HYPRE_IJMatrixInitialize_v2`.  Are you suggesting we implement something similar here?

rfalgout: Everything in Struct seems to use the memory location in the `hypre_handle`.  I think we can continue doing that, right?

victorapm: Sorry, Rob, Now I recall there isn't a `HYPRE_StructMatrixInitialize_v2` that takes the memory location as input.

Instead we rely on `hypre_StructSetMemoryLocation`. But we are missing the public API, so we would need to add it.

We should rely on the memory location of the objects (matrix or vector) instead of `hypre_handle` to be able to do backend switch during runtime correctly, otherwise we might hit some corner cases and memory issues. I'm trying to recall a case that I debugged in the past related to this...

src/struct_mv/struct_matop.c
victorapm: Sure, I can take care of the device implementation here and potentially other parts of this branch. Just let me know if that's fine

rfalgout: Let me take a stab at it first.  I think it's only `AddMat` and `GetDiagMat` in this file that need special treatment in the constant case.  There is some additional optimization that could be done to make the GPU code faster here, but I want to wait on that because I have a thought (vague) about how we might approach BoxLoop kernel optimizations in general.  I'll let you know if I have questions.

PR #1566, opened 2026-06-17
Update rocSPARSE SpMV APIs usage in hypre

victorapm:
This PR updates our rocSPARSE SpMV usage by using the modern APIs in rocSPARSE

We are also adding the (global) option of switching the SpMV algorithm type in use.

TODO:
- [x] Mixed precision code
- [x] Performance tests
- [x] Stable set of default options

cc @YvanMokwinski @amontoison @kliegeois @kswirydo

PR #1618, opened 2026-09-09
ij.c:  partition matrix market among procs

mikethebos:
The ij executable cannot read or partition matrices in Matrix Market format.  This PR uses the existing `HYPRE_IJMatrixReadMM` function to read a Matrix Market matrix.

### DIFF
--- a/src/test/ij.c
+++ b/src/test/ij.c
@@ -741,7 +741,13 @@ main( hypre_int argc,
 
    while ( (arg_index < argc) && (!print_usage) )
    {
-      if ( strcmp(argv[arg_index], "-frombinfile") == 0 )
+      if ( strcmp(argv[arg_index], "-fromMMfile") == 0 )
+      {
+         arg_index++;
+         build_matrix_type      = -3;
+         build_matrix_arg_index = arg_index;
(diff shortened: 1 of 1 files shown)


mikethebos:
I quickly discovered that `HYPRE_IJMatrixReadMM` does not distribute the matrix.  

In my latest commit (this commit was AI assisted by Claude, add labels as needed), I implemented `BuildParFromOneMMFile` in `ij.c` to handle the partitioning.

If you would like me to implement this in `hypre_IJMatrixRead` or perhaps a ParCSR function instead, I can attempt to do so.

mikethebos:
I just noticed that PR #1566 also adds MM reading to ij.  However, I believe it falls into the same trap I found - `HYPRE_IJMatrixReadMM` does not partition the matrix among the ranks (see `hypre_IJMatrixRead`).

With my latest commit, I have confirmed that running ij with `print_matrix_info` turned on on Tuolumne does correctly partition all of the several MM matrices I tried from SuiteSparse.  PCG solves applicable SPD systems and matrix info prints correct number of global rows/columns and global nonzeros.

PR #1619, opened 2026-09-10
Make hypre re-entrant by giving its process-wide state thread-local storage

danielepanozzo:
## What

hypre keeps nine pieces of state in process-wide globals — the error record, the handle, the library state machine, the timing table, the random seed, the `printf` scratch buffer, and Euclid's four. Two threads cannot drive hypre independently today; they collide on all of it.

This marks those globals `HYPRE_THREAD_LOCAL`, a macro defined once in `HYPRE_utilities.h`. 15 files, +126 / −63, most of it the same one-word marking repeated.

## It is a no-op for OpenMP builds, by construction

`HYPRE_THREAD_LOCAL` expands to **nothing** when `HYPRE_USING_OPENMP` is defined.

That is deliberate. Six sites raise an error from inside an OpenMP parallel region:

```
parcsr_ls/par_fsai_setup.c:601,857
IJ_mv/IJMatrix_parcsr.c:3233,3602,3872,4216
```

With per-thread error records, an error raised there by a worker thread would be invisible to `HYPRE_GetError()` on the master thread. Rather than trade one bug for another, an OpenMP build gets exactly the storage it has always had.

For what it is worth, those six sites are unsynchronised read-modify-writes on the shared record as things stand — `IJMatrix_parcsr.c` carries a disabled `#pragma omp atomic` next to one of them with the comment *"error_flag is currently not used anywhere"*. This PR neither fixes that nor makes it worse. Happy to open a separate issue if it is of interest.

## User-visible impact

None for existing builds.

- **OpenMP builds** are bit-identical, and the storage class is unchanged — `nm` puts `hypre__global_error` in `__DATA,__data` exactly as before.
- **Non-OpenMP builds** get per-thread storage. A single-threaded program cannot tell the difference; a multi-threaded one gets independent hypre state per thread, which is the point.

No API change, no new build option, nothing to configure.

## Testing

**OpenMP builds are unchanged.** `test/ij -n 30 30 30 -solver 1`, Open MPI 5.0.10 + libomp, against clean `master`:

| configuration | clean master | this PR |
|---|---|---|
| `-np 1`, `OMP_NUM_THREADS=4` | 9 iters, 1.838656e-09 | 9 iters, 1.838656e-09 |
| `-np 2`, `OMP_NUM_THREADS=4` | 10 iters, 1.279282e-09 | 10 iters, 1.279282e-09 |
| `-np 4`, `OMP_NUM_THREADS=2` | 10 iters, 3.475134e-09 | 10 iters, 3.475134e-09 |

**Re-entrancy works.** BoomerAMG-preconditioned PCG on a 40³ 7-point Laplacian, driven from several threads of one process:

| threads | iters | final relative residual |
|---|---|---|
| 1 | 5 | 6.784053e-11 |
| 2 | 6 | 1.040472e-11 |
| 4 | 6 | 2.458001e-11 |
| 8 | 7 | 2.972584e-11 |

Also builds clean with `HYPRE_ENABLE_MPI=OFF`. The C++ spelling of the macro is needed because nvcc compiles the GPU sources as C++, where `_Thread_local` does not exist.

## Motivation

A library that embeds hypre cannot ask the application linking it to relaunch under `mpirun`. If hypre is re-entrant, its ranks can instead be threads of the calling process, keeping the normal domain decomposition — which is the fast path; the OpenMP backend's setup phase is largely sequential and caps out far lower.

Downstream we do that by building hypre against an MPI whose ranks are threads ([nano-mpi](https://github.com/danielepanozzo/nano-mpi)). That needs **no** hypre change beyond this one: nano-mpi installs a header named `mpi.h`, so hypre's ordinary `find_package(MPI)` finds it and `mpistubs.c` narrows `HYPRE_Int` to `int` exactly as it does for Open MPI.

Nothing here depends on that, though. The change is re-entrancy, and it stands on its own.

## Notes

- Euclid's headers do not include `HYPRE_utilities.h`, so they carry the same macro guard locally.
- Per CONTRIBUTING.md this exceeds 100 lines, and I am glad to move the discussion to an issue first if you would rather — the line count is mostly the same marking repeated across Euclid's two headers.

## AI disclosure

Per the repository's AI policy: **this change was written with AI assistance** (Claude Code). I have reviewed and understand the change, and I take responsibility for its correctness, licensing and attribution. The `AI-assisted` label applies — I do not have permission to set labels on this repository, so please add it, or tell me and I will note it here instead.

The testing above was run by hand and the numbers are real measurements, not generated text.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/distributed_ls/Euclid/_hypre_Euclid.h
+++ b/src/distributed_ls/Euclid/_hypre_Euclid.h
@@ -355,6 +355,22 @@ you need to write EUCLID_GET_ROW() functions: see src/getRow.c
 #include <string.h>
 #include <math.h>
 #include <limits.h>
+
+/* thread-local storage: see HYPRE_utilities.h */
+#if !defined(HYPRE_THREAD_LOCAL)
+#if defined(__cplusplus)
+#define HYPRE_THREAD_LOCAL thread_local
+#elif defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L)
+#define HYPRE_THREAD_LOCAL _Thread_local
+#elif defined(__GNUC__) || defined(__clang__)
+#define HYPRE_THREAD_LOCAL __thread
+#elif defined(_MSC_VER)
+#define HYPRE_THREAD_LOCAL __declspec(thread)
+#else
+#define HYPRE_THREAD_LOCAL
+#endif
+#endif
+
 #include <stdarg.h>
 
 #define REAL_DH HYPRE_Real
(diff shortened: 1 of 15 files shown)


victorapm:
Thank you for the PR @danielepanozzo. It will be reviewed by the hypre team

